PR #6, opened 2014-12-29
call register() once instead of having every documentation node type call it

yole:


orangy:
Do we have all descriptors/nodes registered? I remember previously some were not registered. But then I added parameters, so I'm unsure.

PR #9, opened 2014-12-30
Support annotation classes and annotations in model and rendered output

yole:


src/Formats/StructuredFormatService.kt
```diff
@@ -1,6 +1,7 @@
 package org.jetbrains.dokka
 
 import java.util.LinkedHashMap
+import com.intellij.openapi.util.text.StringUtil
```
orangy: It's from Kotlin compiler, so it will only be available in Kotlin's module, when we will split the project.

src/Formats/StructuredFormatService.kt
```diff
@@ -200,14 +221,17 @@ public abstract class StructuredFormatService(val locationService: LocationServi
...
+                        DocumentationNode.Kind.AnnotationClass,
                         DocumentationNode.Kind.Constructor,
                         DocumentationNode.Kind.Property,
                         DocumentationNode.Kind.Package,
                         DocumentationNode.Kind.Function,
                         DocumentationNode.Kind.PropertyAccessor,
                         DocumentationNode.Kind.ClassObjectProperty,
-                        DocumentationNode.Kind.ClassObjectFunction
+                        DocumentationNode.Kind.ClassObjectFunction,
+                        DocumentationNode.Kind.ExternalClass
```
orangy: Please vote: https://youtrack.jetbrains.com/issue/KT-3762

src/Kotlin/DocumentationBuilder.kt
```diff
@@ -6,9 +6,21 @@ import org.jetbrains.jet.lang.types.*
 import org.jetbrains.jet.lang.types.lang.*
 import org.jetbrains.jet.lang.resolve.name.*
 import org.jetbrains.jet.lang.resolve.lazy.*
+import org.jetbrains.jet.lang.descriptors.annotations.Annotated
+import org.jetbrains.jet.lang.descriptors.annotations.AnnotationDescriptor
+import org.jetbrains.jet.lang.resolve.DescriptorUtils
+import org.jetbrains.jet.lang.resolve.constants.CompileTimeConstant
+import com.intellij.openapi.util.text.StringUtil
```
orangy: Kotlin compiler dependency.

src/Kotlin/DocumentationBuilder.kt
```diff
@@ -104,6 +125,16 @@ class DocumentationBuilder(val session: ResolveSession, val options: Documentati
             node.appendProjection(typeArgument)
     }
 
+    fun DocumentationNode.appendAnnotations(annotated: Annotated) {
+        annotated.getAnnotations().forEach {
+            val annotationNode = it.build()
+            if (annotationNode != null) {
+                append(annotationNode,
+                        if (annotationNode.name == "deprecated") DocumentationReference.Kind.Deprecation else DocumentationReference.Kind.Annotation)
```
orangy: I would rather make it pluggable to support "adornments" on items. We might want to support serializable, final/sealed, or open types, etc.

src/Kotlin/DocumentationBuilder.kt
```diff
@@ -128,7 +175,11 @@ class DocumentationBuilder(val session: ResolveSession, val options: Documentati
         for ((packageName, declarations) in descriptors) {
             println("  package $packageName: ${declarations.count()} nodes")
```
orangy: Shouldn't println here be fixed to adhere to different count of nodes generated below?

orangy: Also we probably need to employ sl4j or something instead of dumb println.

src/Kotlin/KotlinLanguageService.kt
```diff
@@ -1,15 +1,6 @@
...
 
-import org.jetbrains.dokka.symbol
-import org.jetbrains.dokka.text
-import org.jetbrains.dokka.identifier
-import org.jetbrains.dokka.link
-import org.jetbrains.dokka.keyword
-import org.jetbrains.dokka.LanguageService
-import org.jetbrains.dokka.DocumentationNode
-import org.jetbrains.dokka.ContentNode
 import org.jetbrains.dokka
```
orangy: shouldn't it be removed as well? it's same package

src/Kotlin/KotlinLanguageService.kt
```diff
@@ -171,24 +164,46 @@ class KotlinLanguageService : LanguageService {
         }
     }
 
+    private fun ContentNode.renderAnnotationsForNode(node: DocumentationNode) {
+        node.annotations.forEach {
+            renderAnnotation(it)
+        }
+    }
+
+    private fun ContentNode.renderAnnotation(node: DocumentationNode) {
```
orangy: We may try to put annotation(s) in square brackets for easier reading. There also could be cases when they are mandatory, not sure.

orangy:
On "deprecated" support, we wanted to have an option to include/exclude deprecated members from the output. Another option is to put them in a separate section on a site, or in a separate group in the list of members, etc. Need to experiment.

src/Kotlin/DocumentationBuilder.kt
yole: I'd rather not over-generalize too early. We can refactor this later when the requirements are more clear.

src/Kotlin/DocumentationBuilder.kt
yole: Changed "nodes" to "declarations" so that the number remains correct. We can refactor to start using a proper logging framework later. (And no slf4j please - I'd rather use JDK logging.)

src/Kotlin/KotlinLanguageService.kt
yole: done

src/Formats/StructuredFormatService.kt
yole: done

src/Formats/StructuredFormatService.kt
yole: done

src/Kotlin/KotlinLanguageService.kt
yole: I've looked at the generated docs now, and I think that they're easier to read without square brackets. I also plan to make it possible to exclude certain annotations (e.g. inlineOptions) from the output.

yole:
I do plan to make it possible to exclude deprecated members later.

src/Kotlin/KotlinLanguageService.kt
orangy: inline options should ideally be translated to some human-readable note/text in the docs. It's important to understand what's being inlined in performance-critical code.

orangy: Also see https://devnet.jetbrains.com/message/5532560#5532560, I think it's pretty good idea for convention.

PR #16, opened 2015-01-14
generate source links in the documentation

yole:


src/main.kt
```diff
@@ -14,6 +13,10 @@ class DokkaArguments {
     ValueDescription("<path>")
     public var src: String = ""
 
+    Argument(value = "srcLink", description = "Mapping between a source directory and a Web site for browsing the code")
+    ValueDescription("<path>=<url>[#lineSuffix]")
+    public var srcLink: String = ""
+
```
orangy: Can't there be multiple mappings? E.g. when project consists of multiple source repositories checked out into single folder structure.

src/main.kt
```diff
@@ -40,6 +43,15 @@ class DokkaArguments {
 
 }
 
+class SourceLinkDefinition(val path: String, val url: String, val lineSuffix: String?)
+
+private fun parseSourceLinkDefinition(srcLink: String): SourceLinkDefinition {
+    val urlAndLine = srcLink.substringAfter("=")
```
orangy: this should work here:
val (path, urlAndLine) = srcLink.split("=")

src/main.kt
yole: The underlying logic supports multiple mappings, but I'm not such a big fan of trying to represent this in a command line parameter.

orangy: Just use ":" between mappings :) Like in classpath.

src/main.kt
yole: Done

src/main.kt
yole: : is valid in URLs

PR #20, opened 2015-01-16
support in-page anchors in locations

yole:


yole:
Reimplemented differently.

PR #29, opened 2015-07-13
Gradle plugin

cy6erGn0m:
Initial gradle plugin implementation


cy6erGn0m:
@yole actually there is the only JDK 1.8: all other are outdated. However I think you right: we shouldn't force exact JDK version here as we can simply let IDEA to choose any

PR #31, opened 2015-07-24
Kotlin upgrade + source position in documentation node

cy6erGn0m:


cy6erGn0m:
@yole extension function has been reworked to have ConstantValue as receiver instead of Any?. But I see no good alternatives to null values except json with no corresponding fields if no information available. Json here looks like some kind over-engineering

PR #38, opened 2015-11-23
Enable browsing of the output HTML with CSS

mikehearn:


yole:
I'll merge this, but this will need further refactoring: the CSS file should really be copied only when the HTML format is used, because other formats don't use this CSS file.

PR #40, opened 2016-01-03
Allow multiple dokka tasks for one project, up-to-date check

ingokegel:
For example, if multiple formats should be generated, adding more tasks like

    task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
        outputFormat = 'javadoc'
        outputDirectory = "$buildDir/javadoc"
    }

is now possible.

Also,  dokka gradle tasks now know when they are up-to-date and don't have to generate anything.


yole:
Thanks a lot for your contribution!

PR #48, opened 2016-01-05
@Suppress("NOT_DOCUMENTED"), fix for linking to targets with underscores, gradle plugin improvements

ingokegel:
@Suppress("NOT_DOCUMENTED"): Sometimes you have classes that you do not want to document without getting swamped by warnings.

Btw, the recent revision of kotlin-for-upsource.jar does not work with kotlin beta 4. Is beta 5 expected soon?

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,10 +132,17 @@ buildscript {
 apply plugin: 'org.jetbrains.dokka'
 ```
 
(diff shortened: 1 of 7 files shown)


yole:
Thanks a lot for your contributions!

PR #64, opened 2016-03-20
Android support and custom sources

### LINKED ISSUE #61: Dokka can't find sourceSets defined for android modules
For android development, sourceSets are defined within the android block of the build.gradle file, so the only way to get dokka to find them for now is to duplicate the sourceSets outside of the android block.

porokoro:
This adds an initial support for Android library projects.

Also you are able to set the source directories directly from the Gradle plugin now, for the case that you have a special configuration or you have plugins that specify the sources in another way.

I separated the Android support in its own plugin so that users which have a plain Kotlin library don't need to pull the Android dependencies, which they don't need obviously.

Fixes #61. 

**Please note**: I modified the project in Gradle only (build system wise) and didn't change anything in the IntelliJ IDEA configuration. In my opinion the whole _.idea_ folder should be removed because the configuration gets overridden every time you sync with Gradle anyway.
But I have seen there are some special configs for generating Kotlin's docs and if you tell me what needs to be changed in the IntelliJ config I will change it accordingly.


porokoro:
I updated the .idea settings now as well. It looked like a lot of changes at first but this is due to the fact that the structure changed a bit with IntelliJ IDEA 2016.1. For example the `.iml` files seem to be located inside the `.idea` folder now.

Issue #71, opened 2016-05-18
A module file included has extra LF inserted in bullet points


apatrida:
I include a module file using the `includes` parameter to gradle plugin, the source has lists like:

```
In addition there are extension methods on the `Vertx` class that provide promise based versions of deployment, shutdown and close.

* `Vertx.promiseDeployVerticle(verticle: Verticle)`
* `Vertx.promiseDeployVerticle(verticle: Verticle, options: DeploymentOptions)`
* ...
```

But in the resulting doc file, when this is inserted into it, extra LF are inserted so it becomes:

```
In addition there are extension methods on the `Vertx` class that provide promise based versions of deployment, shutdown and close.

* 
`Vertx.promiseDeployVerticle(verticle: Verticle)`

* 
`Vertx.promiseDeployVerticle(verticle: Verticle, options: DeploymentOptions)`

* 
...
```

Which can break the display

Issue #72, opened 2016-05-18
dokka markdown generation doesn't generate table header correctly.


apatrida:
For the list of packages in a module, dokka markdown generates:

```
| [uy.klutter.vertx](uy.klutter.vertx/index.md) |  |
| [uy.klutter.vertx.json](uy.klutter.vertx.json/index.md) |  |
```

and that needs a header for the table or most markdown processors will not generate a correct table, it should be:

```
|package name|whatever|
|---|---|
| [uy.klutter.vertx](uy.klutter.vertx/index.md) |  |
| [uy.klutter.vertx.json](uy.klutter.vertx.json/index.md) |  |
```

see https://help.github.com/articles/organizing-information-with-tables/

apatrida:
this makes package list, and all types useless in markdown for those (which is most) viewers that do not process tables without the header.

apatrida:
and the index of a package.  it turns out almost every page is affected by this.

yole:
Resolved by c0064d2b43ec30237d0d39bf5a9aaa4ce5e14744

karvar:
I am still running into this issue. I am using `mkdoc` and it does not render tables without a header. dokka is omitting the table headers.

fgsalomon:
I'm also having this issue with 0.9.18.

kamilok1965:
Do you use the `GMF` markdown format in dokka?

fgsalomon:
I tried with just `markdown` and `jekyll` as output format. With `GMF` works ok. Sorry for the inconvenience.

kamilok1965:
No problem, happy to help

PR #83, opened 2016-06-25
many markdown fixes

apatrida:
Don't escape within code blocks because HTML entities are not processed by markdown
Fence a triple backtick code block with 4 invisible spaces per line in case the code block contains backticks that might confuse the fencing
Check for backticks in inline code snippets and fence with more backticks than those present (which is the standard way to escape)
Change LF to <br/> or empty depending on the container (i.e. list items cannot have LF but rather can have <p> or <br/>, and tables as well in markdown cannot have LF but can have the <p> and <br/>
Fix bug where list items copied from includes add a line after the * and before the text (markdown parser generates a paragraph which is NOT intended to be a line break in the middle of a list item
Fix tables, they require a header row and the divider |---|---| style thing

should fix:  #71, #72 and others not yet reported

### DIFF
--- a/core/src/main/kotlin/Formats/HtmlFormatService.kt
+++ b/core/src/main/kotlin/Formats/HtmlFormatService.kt
@@ -27,6 +27,10 @@ open class HtmlFormatService @Inject constructor(@Named("folders") locationServi
         return "<span class=\"identifier\"$id>${formatText(text)}</span>"
     }
 
+    override fun appendCode(to: StringBuilder, bodyAsText: ()->String) {
+        to.append("<code>${bodyAsText()}</code>")
+    }
+
     override fun appendBlockCode(to: StringBuilder, language: String, bodyAsLines: ()->List<String>) {
         to.append("<pre><code>")
         to.append(bodyAsLines().joinToString("\n"))
@@ -91,10 +95,6 @@ open class HtmlFormatService @Inject constructor(@Named("folders") locationServi
 
     override fun formatStrikethrough(text: String): String {
         return "<s>${text}</s>"
-    }
-
-    override fun formatCode(code: String): String {
-        return "<code>${code}</code>"
     }
 
     override fun formatUnorderedList(text: String): String = "<ul>${text}</ul>"
--- a/core/src/main/kotlin/Formats/HtmlFormatService.kt
+++ b/core/src/main/kotlin/Formats/HtmlFormatService.kt
@@ -27,9 +27,9 @@ open class HtmlFormatService @Inject constructor(@Named("folders") locationServi
         return "<span class=\"identifier\"$id>${formatText(text)}</span>"
     }
 
-    override fun appendBlockCode(to: StringBuilder, lines: List<String>, language: String) {
+    override fun appendBlockCode(to: StringBuilder, language: String, bodyAsLines: ()->List<String>) {
         to.append("<pre><code>")
-        to.append(lines.joinToString("\n"))
+        to.append(bodyAsLines().joinToString("\n"))
         to.append("</code></pre>")
     }
 
@@ -45,20 +45,18 @@ open class HtmlFormatService @Inject constructor(@Named("folders") locationServi
         to.appendln("$text<br/>")
     }
 
+    override fun appendList(to: StringBuilder, body: () -> Unit) {
+        body()
+    }
+
     override fun appendAnchor(to: StringBuilder, anchor: String) {
         to.appendln("<a name=\"${anchor.htmlEscape()}\"></a>")
     }
 
-    override fun appendTable(to: StringBuilder, body: () -> Unit) {
+    override fun appendTable(to: StringBuilder, columnCount: Int, body: () -> Unit) {
         to.appendln("<table>")
         body()
         to.appendln("</table>")
-    }
-
-    override fun appendTableHeader(to: StringBuilder, body: () -> Unit) {
-        to.appendln("<thead>")
-        body()
-        to.appendln("</thead>")
     }
 
     override fun appendTableBody(to: StringBuilder, body: () -> Unit) {
--- a/core/src/main/kotlin/Formats/MarkdownFormatService.kt
+++ b/core/src/main/kotlin/Formats/MarkdownFormatService.kt
@@ -1,6 +1,7 @@
 package org.jetbrains.dokka
 
 import com.google.inject.Inject
+import java.util.*
 
 
 open class MarkdownFormatService(locationService: LocationService,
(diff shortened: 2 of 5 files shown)


apatrida:
This is somewhat hacky, because the generator API needs to be redone to do this better.

I might take a look at that in the future.  I think it might be best to generate more of a MarkDown ContentNode tree for everything, then use the normal rendering of that as HTML for that output, and for markdown render it from the tree.  Currently it is a mix of tree visiting and ad hoc writes to the output.

apatrida:
I still need to update the unit tests, but if you have comments before then let me know.  All markdown unit tests fail.

apatrida:
Ok, looks like the last remaining main issue (for those I have checked) is to change the signatures not to be in backtick blocks, but `<code>` blocks instead so that the links will work.

apatrida:
a build of this is in my bintray for core and gradle plugin:
https://bintray.com/jaysonminard/kohesive/dokka/view

apatrida:
test docs generated using this version:

https://github.com/kohesive/klutter/blob/master/docs/markdown/klutter-vertx3-jdk8/uy.klutter.vertx/index.md

https://github.com/kohesive/klutter/blob/master/docs/markdown/klutter-core-jdk6/index.md

yole:
This is unfortunately very hard to review in this form. Would it be possible to split this into multiple pull requests which change one thing at a time? I'll leave some comments for this PR according to what I can see in this version of the code.

core/src/main/kotlin/Formats/HtmlFormatService.kt
```diff
@@ -45,22 +49,20 @@ open class HtmlFormatService @Inject constructor(@Named("folders") locationServi
...
     }
 
-    override fun appendTable(to: StringBuilder, body: () -> Unit) {
+    override fun appendTable(to: StringBuilder, columnCount: Int, body: () -> Unit) {
         to.appendln("<table>")
         body()
         to.appendln("</table>")
     }
 
-    override fun appendTableHeader(to: StringBuilder, body: () -> Unit) {
```
yole: Why did you remove this method?

core/src/main/kotlin/Formats/StructuredFormatService.kt
```diff
@@ -55,6 +56,29 @@ abstract class StructuredFormatService(locationService: LocationService,
         return StringBuilder().apply { formatText(location, content, this, listKind) }.toString()
     }
 
+    fun ContentNode.ignoreParagraphAtTop(): ContentNode {
```
yole: This logic is already handled by the buildInlineContentTo() function; doesn't need to be duplicated.

core/src/main/kotlin/Formats/MarkdownFormatService.kt
```diff
@@ -44,44 +118,57 @@ open class MarkdownFormatService(locationService: LocationService,
...
+    override fun appendBlockCode(to: StringBuilder, language: String, bodyAsLines: ()->List<String>) {
+        to.append(allowedLf().requiredWhitespace)
+
+        val codeTags = allowedLf().inlineCode
+        if (codeTags.start.startsWith('`')) {
+            changeLf(LfPriority.Code) {
+                // this does not adjust line feeds, a code block in a list or table will just break formatting
+                to.appendln(if (language.isEmpty()) "```" else "``` $language")
+                // use 4 zero width spaces which act as a code fence escaping any use of ``` within the code block
+                to.appendln(bodyAsLines().map { "​\u200B\u200B\u200B\u200B$it" }.joinToString("\n"))
```
yole: Why exactly is this needed? And are you sure this is going to be processed as you expect by all Markdown renderers?

core/src/main/kotlin/Formats/HtmlFormatService.kt
apatrida: It wasn't used and to add it causes a problem for Markdown I didn't want to solve unless this is every brought back to be used.  Then ONLY if it is really needed, do the hard work to solve the problem.

core/src/main/kotlin/Formats/MarkdownFormatService.kt
apatrida: Few markdown renderers support headerless tables, and common mark the parser you are using doesn't either.  It is currently being discussed how headerless tables would be in markdown and there is no consensus.  The table format that was there before was the rare unsupported case of having no headers and no `---|---|---` row.

core/src/main/kotlin/Formats/StructuredFormatService.kt
apatrida: I'll look at that, previously everything was getting extra paragraphs that shouldn't causing line breaks which broke the markdown and made it invalid.

core/src/main/kotlin/Formats/MarkdownFormatService.kt
yole: I mean, why are these four zero-width spaces needed?

apatrida:
In general most of the changes work together, the rendering model is really hard for markdown.  It was built for HTML then tried to adapt to Markdown.  Would be better to build a complete tree using the Markdown AST no matter what renderer you are using, then render from there.  In the HTML case the default markdown-to-html rendering could be used instead of a custom renderer.

core/src/main/kotlin/Formats/MarkdownFormatService.kt
apatrida: Ah code fencing.  It is because if you use backticks inside of code it can break the code block.  

You can create a code block alternatively using 4 white-space characters.  

So by using 4 zero space white space characters it ensures no matter what fencing is used, that backticks within the code block are ignored.

It is basically escaping everything each code block line.  

It is a work around until the "add more back ticks per fence" method is supported by more markdown renderers.

apatrida: (assuming you don't have a parser that doesn't treat all whitespace the right way, then this would do nothing for that parser)

yole:
Rewriting Dokka to use the Markdown AST for everything isn't going to happen, but it's definitely possible to adapt the content block structure so that it would work better with Markdown, so if you can point to specific places which cause problems, please do.

apatrida:
Maybe the rendering methods could change then instead to be workable for both.  I had to force-fit the HTML rendering style pattern to work with Markdown.

Another option is to use html tags in the markdown, but it isn't clear which renderers support which basic tags, and if a person might have them turned off.  

So these changes try to use the standard markdown language unless it would not work at all in some cases then use html tags which might be stripped (which is fine, it does best attempt).  

The current model also wanted to escape things before building strings and was escaping for HTML and not for markdown, so that was a problem too.  

Basically this commit represents the basic "make everything work right" at least from the renderers I checked including Github's.  (likely the most common use of this is Github yes?)

yole:
I've been looking at this some more, and I think that it would be much more valuable to have testdata on which the current formatter is producing incorrect results, rather than the changes to the formatter itself. I think that the LfPriority mechanism you've implemented is a little too complicated, and can be easier handled by separating MarkdownTextBuilder (which knows the current state of the output Markdown) from MarkdownFormatService. However, in the current state, where you broke all tests and didn't provide updated testdata, I have no way to ensure that such a refactoring would solve the problems that motivated you to make these changes.

yole:
Also it looks like most Markdown renderers support headerless tables, and only GFM requires a header row (and renders the current headers without text in a fairly ugly way). I think it would be cleaner to provide the text for table headers to the rendering service and to provide a separate GFM output format, which would render the headers as supported by GFM.

yole:
I decided to approach the problem differently and did a fairly massive rewrite that produces a much cleaner output (fewer redundant line breaks, for one) with a simpler implementation. This should fix all the issues that you're addressing with this pull request, aside from multiple backticks inside a code block (you're welcome to file a separate issue about this). I'd appreciate it if you could test the master version on your library and let me know if there are any issues with formatting. Please use "-format gfm" if you plan to publish the content on GitHub Pages.

In any case, thanks for your contribution!

Issue #94, opened 2016-07-29
In the Google browser display will be garbled words in Chinese


dreamkidd:
In the Google browser display will be garbled words in Chinese
![2016-07-29_18:59:52.jpg](http://o9vq8w7xd.bkt.clouddn.com/2016-07-29_18:59:52.jpg)
but use Safari is normal 
I think should add the `<meta charste="utf-8">` when generate the html file  to prevent this 
thanks

Issue #95, opened 2016-08-09
Dokka appears to delete apostrophes from the output


mikehearn:
I have a KDoc comment with the word "couldn't" in it, and it's rendered as "couldnt".

simonorono:
Same happens to me. This comment:

~~~Kotlin
    /**
     * Retrieves this pool's corresponding quark for [str].
     *
     * @param str The string
     * @return An integer corresponding to a quark in the pool
     */
~~~

Turns into:
`Retrieves this pools corresponding quark for str.`

nexbit:
Same here..
The only workaround I've found is to use `&apos;` html entity in my comments.. Horrible to see, but it works, I'll refactor them when this issue will be resolved!

PR #101, opened 2016-11-01
Various fixes

semoro:
Fix for GH #94


semoro:
Fix for GH #95 , KT-13300 : dokka eats apostrophes

Issue #115, opened 2016-11-16
Multiple keywords of KDoc Syntax are ignored for javadoc output format


lkjh654:
There is no output generated for `@param <name>`, `@return` in the final javadoc.

1.  checkout `https://github.com/JetBrains/kotlin-examples/tree/master/gradle/dokka-gradle-example`
2. modify source code to add  `@param <name>`, `@return` keywords according to the documentation at `https://kotlinlang.org/docs/reference/kotlin-doc.html`
3. generate documentation using `dokkaJavadoc` command.

There is no output generated for the keywords, they simply seem to be ignored. The proper output is generated for `html` format.

PR #119, opened 2016-11-28
Fix for #115  `@param` and `@return` tags is missing in javadoc output

semoro:


core/src/main/kotlin/javadoc/docbase.kt
```diff
@@ -416,11 +442,15 @@ open class ClassDocumentationNodeAdapter(module: ModuleNodeAdapter, val classNod
             .map { ClassDocumentationNodeAdapter(module, it) }
             .toTypedArray()
 
-    override fun typeParamTags(): Array<out ParamTag> = (classNode.details(NodeKind.TypeParameter).filter { it.content.summary !is ContentEmpty || it.content.description !is ContentEmpty || it.content.sections.isNotEmpty() }.map {
-        ParamTagAdapter(module, this, it.name, true, it.content.children)
-    } + classNode.content.sections.filter { it.subjectName in typeParameters().map { it.simpleTypeName() } }.map {
-        ParamTagAdapter(module, this, it.subjectName ?: "?", true, it.children)
-    }).toTypedArray()
+    override fun typeParamTags(): Array<out ParamTag> =
+            (classNode.details(NodeKind.TypeParameter)
```
yole: I think this should be extracted into a method to reduce dupllication

core/src/main/kotlin/javadoc/docbase.kt
```diff
@@ -365,6 +380,17 @@ class MethodAdapter(module: ModuleNodeAdapter, node: DocumentationNode) : Docume
     override fun isDefault(): Boolean = false
 
     override fun returnType(): Type = TypeAdapter(module, node.detail(NodeKind.Type))
+
+    override fun tags(tagname: String?) = super.tags(tagname)
+
+    override fun tags(): Array<out Tag> {
+        val tags = super.tags().toMutableList()
+        node.content.findSectionByTag(ContentTags.Return)?.let {
```
yole: @throws should be supported too

PR #123, opened 2016-12-02
Now dokka-fatjar complete isolated from poisonous Gradle environment

semoro:


runners/gradle-plugin/src/main/kotlin/main.kt
```diff
@@ -83,17 +120,38 @@ open class DokkaTask : DefaultTask() {
...
+                outputFormat,
+                false,
+                false,
+                false,
+                false,
+                6,
+                true,
+                linkMappings.map {
+                    val path = project.file(it.dir).absolutePath
+                    return@map "$path=${it.url}${it.suffix}"
```
yole: "return@map" is redundant here

runners/gradle-plugin/src/main/kotlin/main.kt
```diff
@@ -8,22 +8,38 @@ import org.gradle.api.file.FileCollection
...
+import java.net.URLClassLoader
 import java.util.*
+import java.util.function.Predicate
 
 open class DokkaPlugin : Plugin<Project> {
+
+    val properties = Properties()
+
+    init {
+        properties.load(javaClass.getResourceAsStream("/META-INF/gradle-plugins/org.jetbrains.dokka.properties"))
```
yole: Why "plugins"?

integration/src/main/kotlin/org/jetbrains/dokka/ProxyUtils.kt
```diff
@@ -0,0 +1,76 @@
...
+ * Warning! Hard reflection magic used here.
+ *
+ * Creates [java.lang.reflect.Proxy] with pass through invocation algorithm,
+ * to create access proxy for [delegate] from [targetClassLoader] to [delegateClassLoader].
+ *
+ * Every object type contained in method calls will be translated to proxy, if [filter.test] will success for it's class
+ *
+ */
+@Suppress("UNCHECKED_CAST")
+fun <T> automagicTypedProxy(targetClassLoader: ClassLoader, targetType: Class<T>,
```
yole: `<reified T>` and remove explicit targetType parameter?

Issue #85, opened 2016-07-07
Missing class: DocErrorReporter


norswap:
On the command line, running `java -jar dokka-fatjar.jar src -java` requires the class `com.sun.javadoc.DocErrorReporter` (and maybe others).

The fix, include `tools.jar` on the classpath, which makes the correct command (note `-jar` is incompatible with `-cp`):

```
java -cp $JAVA_HOME/lib/tools.jar:dokka-fatjar.jar org.jetbrains.dokka.MainKt src -java
```

This should probably be documented in the README.

This relates to #58, which you might wish to revisit that one as well, because on my machine (OS X El Capitan with jdk 1.8.0_92 (also confirmed on 1.8.0_51 -- I'm suspecting jdk 8 in general)), `classes.jar` doesn't exist, and `tools.jar` is not `$JAVA_HOME/../lib/tools.jar` but `$JAVA_HOME/lib/tools.jar` (with `JAVA_HOME = /usr/libexec/java_home`).

AlmasB:
Can confirm the issue is present with dokka-maven-plugin 0.9.10 with goal set to `javadocJar` (windows 10x64, jdk 1.8.0_112).

Issue #58, opened 2016-02-26
maven plugin 0.9.7 results in ClassNotFoundException: com/sun/javadoc/DocErrorReporter


AlmasB:
Steps to reproduce: change outputFormat to javadoc in maven plugin 0.9.7.

Then
```maven
 mvn pre-site
```
            <plugin>
                <groupId>org.jetbrains.dokka</groupId>
                <artifactId>dokka-maven-plugin</artifactId>
                <version>0.9.7</version>
                <executions>
                    <execution>
                        <phase>pre-site</phase>
                        <goals>
                            <goal>dokka</goal>
                        </goals>
                        <configuration>
                            <outputFormat>javadoc</outputFormat>
                        </configuration>
                    </execution>
                </executions>
            </plugin>

Manual addition of $JAVA_HOME/lib/tools.jar to classpath fixes it, so I'm assuming the dependency was not in fatjar for some reason.

cy6erGn0m:
I used to think tools.jar is in the classpath for jdk.. Perhaps we can add system dependency as well to avoid issue

cy6erGn0m:
Perhaps we can do http://stackoverflow.com/a/3080695

AlmasB:
Yeah, should do the trick

yole:
Fixed in 62cd24119a46770a533013c262a324d59110a966

PR #125, opened 2016-12-07
tools.jar

semoro:
Fix for #58 Regression, and for #85

### DIFF
--- a/runners/cli/src/main/kotlin/cli/main.kt
+++ b/runners/cli/src/main/kotlin/cli/main.kt
@@ -102,22 +102,32 @@ object MainKt {
     }
 
     fun createClassLoaderWithTools(): ClassLoader {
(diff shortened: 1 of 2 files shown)


runners/cli/src/main/kotlin/cli/main.kt
```diff
@@ -45,41 +46,90 @@ class DokkaArguments {
...
+                documentationOptions)
+
+        generator.generate()
+        DokkaConsoleLogger.report()
+    }
+
+    fun findToolsJar(): File {
+        val javaHome = System.getProperty("java.home")
+        val default = File(javaHome, "../lib/tools.jar")
+        val mac = File(javaHome, "../Classes/classes.jar")
```
yole: Is this actually needed? I think this is the same jar where all the rest of JRE classes are located, so if it's not on the classpath, we won't get to this point.

PR #149, opened 2017-03-30
Add a line between table rows

deanveloper:
Table rows are kinda hard to read as there is nothing separating table rows. This is a nice simple solution which separates table rows by a grey line. Here's what it looks like: https://kbukkit.deanveloper.com/kotlin/-k-bukkit/index.html


MarcinAman:
Hi,
since this PR was created dokka has been rewritten. In the new HTML format table rows have spacing between

PR #152, opened 2017-04-28
Fix missing whitespace around links and tokens in Markdown formatted text

alex2069:
When a link or non-text token (e.g. comma) is inside an emphasis, strong, or header node, the whitespace surrounding the link (or following the non-text token) is dropped.
This PR fixes this and adds tests.

Of note - a previous test `markdownInLinks` changed as a result of this.
Was there a reason the spaces surrounding "kas" were being dropped? This seemed like a related bug, thus the expected markdownInLinks.html file was adjusted.


semoro:
Thanks for your contribution! :slightly_smiling_face:

PR #154, opened 2017-05-02
Fix dokka-android plugin with Android Gradle 2.4.0 build tools

alex2069:
The 2.4.0 build tools delay variant dependency analysis until needed,
thus we do the same; delay processing variant configurations until
Dokka task execution.

Changes also tested on Android Gradle 2.2.0 and 2.3.0.


semoro:
Merged manually. 

PR #162, opened 2017-05-22
Sort package-list output

alex2069:
If left unsorted, then package list can change each time even though no packages were changed.


semoro:
Thank you for the contribution a lot! :slightly_smiling_face:

PR #175, opened 2017-06-21
Fix link in the last breadcrumbs item

nd:


semoro:
Generated links seems incorrect, 
image

Maybe `href = '#top'` would be better?

nd:
Sorry for delay, for some reason github notification were marked as spam. The link was indeed incorrect. As far as I understand not every page has the `#top` anchor. I've submitted a different fix which adds a link to the file itself using relative path so link should be correct now. Please take a look.

semoro:

PR #176, opened 2017-06-21
Put samples in a section

nd:


semoro:
Not sure that we should put samples in section, cause you may want to write some description around a sample. I think it's better to remove grouping in section in IDE

nd:
I wonder how one can write a description for sample? The `@sample` doesn't allow that. I've ended up writing comments inside the sample function. Is there a better way?

MarcinAman:
Hi, 
dokka has been rewritten and now samples have its own section in the documentation. May want to check out latest version: 1.4.10 👍

PR #192, opened 2017-07-12
Add global configuration "includeNonPublic"

### LINKED ISSUE #181: includeNonPublic as top level configuration
It'd be useful if "includeNonPublic" could be made a top level configuration option outside packageOptions. 

Default on top-level should be false for obvious reasons, but we have a use-case of creating documentation for non-public functions for complete projects. When packageOptions is properly working (see #178) I _assume_ I could solve it with a very globally setup prefix, too - but I'm not sure if having to provide packageOptions for global setups is the best way to do this.

HaasJona:
Fixes #181

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,9 @@ dokka {
     // When set to default, caches stored in $USER_HOME/.cache/dokka
     cacheRoot = 'default' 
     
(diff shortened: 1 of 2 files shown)


semoro:
Thank you!

Issue #221, opened 2017-10-05
External documentation links to nested classes are incorrect for Android documentation.


adesentenac:
When dokka generates links to subclasses, or interfaces in classes, from external documentation (for example for Android), instead of outputting a . for the last separator, it outputs a /

i.e. instead of creating a link to https://developer.android.com/reference/android/view/GestureDetector.SimpleOnGestureListener.html it generates a link to https://developer.android.com/reference/android/view/GestureDetector/SimpleOnGestureListener.html
(note that the correct case is GestureDetector.Simple... and the incorrect is GestureDetector/Simple...)

Idem for interfaces, instead of: https://developer.android.com/reference/android/view/View.OnClickListener.html it generates: https://developer.android.com/reference/android/view/View/OnClickListener.html

The same issue appears when using custom external documentation (in my case to Exoplayer classes).

adesentenac:
I've created a pull request to fix this issue, and also handling external documentations which uses dash instead of parenthesis and commas as separators between parameters, like: https://google.github.io/ExoPlayer/doc/reference/com/google/android/exoplayer2/Player.EventListener.html#onPlayerStateChanged-boolean-int-

Let me know if you see anything wrong with my fix.

kamildoleglo:
AFAIK this has been fixed in the `1.4.10.2` release,  please feel free to reopen if the issue persists

PR #223, opened 2017-10-11
#221 - Updates external documentation link configuration

adesentenac:
Fix for #221 
Updates external documentation link generation configuration to support keeping the dot as separator between a class and its inner class, as well as using the dash for as a parameter separator, since both are used in Google API documentations.

Adds `useDotAsSubclassSeparator` to keep `.` as separator between a class and its subclass. Fixes links to Android documentation (e.g. https://developer.android.com/reference/android/view/View.OnClickListener.html)

Adds `useDashAsParameterSeparator` to replaces parenthesis and commas as parameters separator, to support some Google API documentations (like http://google.github.io/guice/api-docs/latest/javadoc/com/google/inject/grapher/Edge.html#copy-com.google.inject.grapher.NodeId-com.google.inject.grapher.NodeId- or https://google.github.io/ExoPlayer/doc/reference/com/google/android/exoplayer2/Player.EventListener.html#onLoadingChanged-boolean-) 

I have tested the gradle plugins. The maven plugin and ant task should also work, since I didn't see anything specific in the code, but are untested.
The command line is not supported, since I don't know how we would want to add the additional parameters to the command line (adding ^ separators might not be really clear since both new parameters are booleans).

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,12 @@ dokka {
         
         // If package-list file located in non-standard location
         // packageListUrl = new URL("file:///home/user/localdocs/package-list") 
+
+        // If the external documentation uses the dot as separator between a class and its inner class instead of a slash
+        // useDotAsSubclassSeparator = true
+
+        // If the external documentation uses the dash as separator between parameters
+        // useDashAsParameterSeparator = true
     }
     
     // Allows to customize documentation generation options on a per-package basis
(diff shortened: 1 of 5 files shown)


xit0c:
I'm facing the same issue this PR is trying to resolve. Any update about it?

tynn:
Your parameter suggests that you're handling subclasses, but you're actually handling nested classes

kamildoleglo:
Dokka has been rewritten since the creation of this PR, this issue has been fixed with multiple  ExternalLocationProviders ☺️

PR #234, opened 2017-10-31
Suppress all generated files (Java and Android projects)

### LINKED ISSUE #197: Suppressed files not honored when generating Javadoc (DokkaAndroidTask)
When generating Javadocs, the `collectSuppressedFiles(sourceRoots: List<SourceRoot>)` override is not honored. Therefore, Javadocs show generated files in the documentation. I came across this when generating docs for an Android library (DokkaAndroidTask).

The default documentation does honor the suppression, so the Kotlin docs do not have the generated files. The Kotlin documentation builder does a check for the suppressed files (see [DocumentationBuilder](https://github.com/Kotlin/dokka/blob/ec2f9aef111cd44eb7fb438c5fd03cf32be9eced/core/src/main/kotlin/Kotlin/DocumentationBuilder.kt#L862)).

I do not see a similar check in [JavaPsiDocumentationBuilder](https://github.com/Kotlin/dokka/blob/ec2f9aef111cd44eb7fb438c5fd03cf32be9eced/core/src/main/kotlin/Java/JavaPsiDocumentationBuilder.kt) or [KotlinAsJavaDocumentationBuilder](https://github.com/Kotlin/dokka/blob/ec2f9aef111cd44eb7fb438c5fd03cf32be9eced/core/src/main/kotlin/Kotlin/KotlinAsJavaDocumentationBuilder.kt)

codeprogression:
Documentation for generated files (in the `generated/source` path of the project build directory) were being suppressed for Android projects (`R`, `BuildConfig`). However, generated files were included in Java projects. This caused all generated classes in Java projects (such as Dagger annotation processing) to be documented which is not expected.

Moved the body of `collectSuppressedFiles()` function to `main.kt` from `androidMain.kt`.

Fixes #197.

### DIFF
--- a/runners/android-gradle-plugin/src/main/kotlin/mainAndroid.kt
+++ b/runners/android-gradle-plugin/src/main/kotlin/mainAndroid.kt
@@ -22,11 +22,6 @@ open class DokkaAndroidTask : DokkaTask() {
 
     @Input var noAndroidSdkLink: Boolean = false
 
-    override fun collectSuppressedFiles(sourceRoots: List<SourceRoot>): List<String> {
-        val generatedSubpath = "${project.buildDir}/generated/source".replace("/", File.separator)
-        return sourceRoots.filter { generatedSubpath in it.path }.flatMap { File(it.path).walk().toList() }.map { it.absolutePath }
-    }
-
     init {
         project.afterEvaluate {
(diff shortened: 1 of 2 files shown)


codeprogression:
I suppose I misdiagnosed the problem initially. After digging in, I realized what was happening. The issue for Android projects was fixed. (It didn’t seem to work in the published version of 0.9.15, but it was working if built from master.)

semoro:
Closing as the problem already fixed

PR #249, opened 2017-12-08
Update plugin-publish-plugin

timyates:
Hiya, Tim from Gradle here.

There was a [bug in versions prior to 0.9.7](https://discuss.gradle.org/t/plugin-authors-please-use-the-latest-plugin-publish-plugin-others-may-result-in-a-broken-upload/23573) of the gradle `plugin-publish-plugin`, where artifacts would sometimes  silently not be pushed into the repo.

Your plugin has not been affected by the bug yet, but you should still upgrade.

This won't fix issue 146 (I'm not sure what has caused the pom differences seen in that issue), but it should avoid any problems publishing future versions.


semoro:
Merged into dev manually

PR #255, opened 2017-12-25
MPP fixes

elizarov:


semoro:
It is not worth merging this PR, as it does not provides complete nor correct MPP support, rather then number of hacks

PR #271, opened 2018-02-13
Document that kotlin-android plugin is required even for non-kotlin p…

mathjeff:
…rojects

I'm not actually confident that this is in fact required or the most proper way to do this; hopefully reviewers can tell me whether this is expected.

I made a small project at https://github.com/mathjeff/dokka-tests/blob/master/kotlin-android-required/README.txt demonstrating missing documentation until the kotlin-android plugin is added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ buildscript {
 }
 
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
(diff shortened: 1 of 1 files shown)


semoro:
Hi! Looks like there is a problem with automatic classpath configuration (Which is works best if Kotlin plugin used)
It’s better to file an issue, rather than documenting that `kotlin-android` required for Dokka to work

mathjeff:
Excellent; thanks

PR #286, opened 2018-03-10
fix doc linkMapping, requires absolute path

robstoll:
Might be that the implementation is wrong (https://github.com/Kotlin/dokka/blob/54c3c87acfb31afc22afc5f20229384f755b677f/core/src/main/kotlin/Model/SourceLinks.kt#L16) but it requires an absolute path and not a relative path. Also `/` or `\` plays a role and thus it is best to create a file and get the absolute path to it

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ dokka {
     // Repeat for multiple mappings
     linkMapping {
         // Source directory
-        dir = "src/main/kotlin"
+        dir = file("${project.projectDir}/src/main/kotlin").absolutePath
          
         // URL showing where the source code can be accessed through the web browser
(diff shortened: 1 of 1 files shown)


robstoll:
Totally agree (if it is wrong, I don't know), but as long as it isn't fixed we should document it, don't you think so?

semoro:
It's incorrect, and shouldn't be documented.
But should be reported as an issue

robstoll:
Good, I'll close this pull request then, and open an issue. Shall I provide a fix for dokka/core/src/main/kotlin/Model/SourceLinks.kt or do you prefer a fix in the dokka plugin?

semoro:
Better to fix at `dokka/core/src/main/kotlin/Model/SourceLinks.kt`, but check that absolute paths also works.

PR #313, opened 2018-05-29
Resolve method equals(Object) in javadoc links

atoulme:


atoulme:
Also tests for integer and boolean. Works OK for those two.

core/src/main/kotlin/Model/DocumentationReference.kt
```diff
@@ -41,7 +41,14 @@ class NodeReferenceGraph() {
     val references = arrayListOf<PendingDocumentationReference>()
 
     fun register(signature: String, node: DocumentationNode) {
-        nodeMap.put(signature, node)
+        val javaSignature = signature.replace("kotlin.Any", "java.lang.Object").
```
semoro: This shouldn't be done on `DocumentationReference` side, it should be changed on somehow in JavaDocParser, to make it emit proper signatures for both AsJava and AsKotlin mode.

KrystianUjma:
@atoulme would you tell me what do you want to achieve with your PR? I tested it with javadoc format + java class and it generated wrong function signature.

atoulme:
It's been a very long while. When I generated javadoc with dokka, it would not resolve the java parameters.

Issue #213, opened 2017-09-01
 Exception while loading package-list from ExternalDocumentationLinkImpl


ChansEbm:
> Guice provision errors:
  
  1) Error injecting constructor, java.lang.RuntimeException: Exception while loading package-list from ExternalDocumentationLinkImpl(url=https://developer.android.com/reference/, packageListUrl=https://developer.android.com/reference/package-list)
    at org.jetbrains.dokka.ExternalDocumentationLinkResolver.<init>(ExternalDocumentationLinkResolver.kt:25)
    at org.jetbrains.dokka.ExternalDocumentationLinkResolver.class(ExternalDocumentationLinkResolver.kt:25)
    while locating org.jetbrains.dokka.ExternalDocumentationLinkResolver
      for parameter 4 at org.jetbrains.dokka.DeclarationLinkResolver.<init>(DeclarationLinkResolver.kt:9)
    while locating org.jetbrains.dokka.DeclarationLinkResolver
      for parameter 0 at org.jetbrains.dokka.PackageDocs.<init>(PackageDocs.kt:13)
    at org.jetbrains.dokka.PackageDocs.class(PackageDocs.kt:12)
    while locating org.jetbrains.dokka.PackageDocs
  
  1 error

This issus shown when i use ./gradlew dokka
how can i fix this.

semoro:
Try again, or maybe you are behind the corporate proxy? Then you should manually download `https://developer.android.com/reference/package-list`, place it somewhere and use packageListUrl parameter to configure its location.
Also please post full stack-trace including `caused by`

allo86:
Hi. I am getting this error now with 0.9.15 version. This is the full stack-trace:

```
* What went wrong:
Execution failed for task ':ui:dokka'.
> Guice provision errors:
  
  1) Error injecting constructor, java.lang.RuntimeException: Exception while loading package-list from ExternalDocumentationLinkImpl(url=http://docs.oracle.com/javase/6/docs/api/, packageListUrl=http://docs.oracle.com/javase/6/docs/api/package-list)
    at org.jetbrains.dokka.ExternalDocumentationLinkResolver.<init>(ExternalDocumentationLinkResolver.kt:25)
    at org.jetbrains.dokka.ExternalDocumentationLinkResolver.class(ExternalDocumentationLinkResolver.kt:25)
    while locating org.jetbrains.dokka.ExternalDocumentationLinkResolver
[...]

PR #314, opened 2018-05-31
Proxy support with basic authentication has been added using system properties

jacombo:
Dokka is great plugin unfortunatelly can't be used in corporates behind proxy therefore proxy support with basic authentication has been added using system properties below

-Dhttp.proxyHost=host.domain.com
-Dhttp.proxyPort=8080
-Dhttps.proxyHost=host.domain.com
-Dhttps.proxyPort=8080
-DproxySet=true
-DproxyUser=user
-DproxyPassword=password

if proxySet is true and proxyUser with proxyPassword is set then basic proxy authentication is added

### DIFF
--- a/core/src/main/kotlin/Kotlin/ExternalDocumentationLinkResolver.kt
+++ b/core/src/main/kotlin/Kotlin/ExternalDocumentationLinkResolver.kt
@@ -46,6 +46,16 @@ class ExternalDocumentationLinkResolver @Inject constructor(
         connection.connectTimeout = timeout
         connection.readTimeout = timeout
 
+        // Add proxy basic authentication when proxyUser, proxyPassword is set and proxySet is true
+        val systemSettings = System.getProperties()
+        if (systemSettings.getValue("proxySet").toString() == "true"
+                && systemSettings.getValue("proxyUser").toString().isNotEmpty()
+                && systemSettings.getValue("proxyPassword").toString().isNotEmpty()) {
+            val encoder = sun.misc.BASE64Encoder()
+            val encodedUserPwd = encoder.encode((systemSettings.getValue("proxyUser").toString() + ":" + systemSettings.getValue("proxyPassword").toString()).toByteArray())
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/Kotlin/ExternalDocumentationLinkResolver.kt
```diff
@@ -46,6 +46,16 @@ class ExternalDocumentationLinkResolver @Inject constructor(
         connection.connectTimeout = timeout
         connection.readTimeout = timeout
 
+        // Add proxy basic authentication when proxyUser, proxyPassword is set and proxySet is true
+        val systemSettings = System.getProperties()
+        if (systemSettings.getValue("proxySet").toString() == "true"
```
semoro: As far, as I debugged, Gradle already support basic authentification, so no need for such code in Gradle runner plugin. 
Isn't it better to use `java.net.Authenticator`? It looks like more straightforward approach.

semoro: And why not to use `http.proxyUser` and so on?

semoro:
Dokka already supports proxy authentication according to my comment in https://github.com/Kotlin/dokka/issues/213#issuecomment-393955468

core/src/main/kotlin/Kotlin/ExternalDocumentationLinkResolver.kt
jacombo: could you debug and verify also  Maven + proxy + basic authentication ?

I will try also  will try use "systemProp.jdk.http.auth.tunneling.disabledSchemes" too and let you know

we are no using Graddle at the moment so we need to fix Maven first

semoro:
Checked, doesn't work properly, looks like Dokka for Maven should use Maven transport API http://maven.apache.org/wagon/

semoro:
So, anyway, changes is required

kamildoleglo:
Dokka has been rewritten since the creation of this PR,  please open a new PR if the issue still persists ☺️

Issue #188, opened 2017-07-10
To add an ability to disable link to java documentation


neverov-test:
It would be great to disable link to java documentation, similar to ```noStdlibLink``` for Kotlin stdlib. Without that docs generation fails with the error like this when network is not available:

```
[ERROR] Failed to execute goal org.jetbrains.dokka:dokka-maven-plugin:0.9.15:dokka (default-cli) on project configs-dsl-kotlin: Execution default-cli of goal org.jetbrains.dokka:dokka-maven-plugin:0.9.15:dokka failed: Guice provision errors:
[ERROR]
[ERROR] 1) Error injecting constructor, java.lang.RuntimeException: Exception while loading package-list from ExternalDocumentationLinkImpl(url=http://docs.oracle.com/javase/6/docs/api/, packageListUrl=http://docs.oracle.com/javase/6/docs/api/package-list)
[ERROR] at org.jetbrains.dokka.ExternalDocumentationLinkResolver.<init>(ExternalDocumentationLinkResolver.kt:25)
[ERROR] at org.jetbrains.dokka.ExternalDocumentationLinkResolver.class(ExternalDocumentationLinkResolver.kt:25)
[ERROR] while locating org.jetbrains.dokka.ExternalDocumentationLinkResolver
[ERROR] for parameter 4 at org.jetbrains.dokka.DeclarationLinkResolver.<init>(DeclarationLinkResolver.kt:9)
[ERROR] while locating org.jetbrains.dokka.DeclarationLinkResolver
[ERROR] for parameter 0 at org.jetbrains.dokka.PackageDocs.<init>(PackageDocs.kt:13)
[ERROR] at org.jetbrains.dokka.PackageDocs.class(PackageDocs.kt:12)
[ERROR] while locating org.jetbrains.dokka.PackageDocs
```

semoro:
See `cacheRoot` option of Dokka, it can cache package-lists and prevent you from failing when there is no network.

nd:
As far as I understand, ExternalDocumentationLinkResolver always opens a connection to the package list url (when the cached entry exists, it checks the modified date) and will fail when there is no network.

semoro:
It should show a warning, not fail when checking for modified date

semoro:
There is a `try catch` https://github.com/Kotlin/dokka/blob/master/core/src/main/kotlin/Kotlin/ExternalDocumentationLinkResolver.kt#L63

neverov-test:
[...]

PR #318, opened 2018-06-08
Add options to include links to stdlib.

stephenbyrne-mfj:
https://github.com/Kotlin/dokka/issues/188
Do not connect to the internet for stdlib links by default.
Removed "noStdlibLink" option (which only controlled kotlin stdlib, not
java, and was enabled by default).
Added "kotlinStdlibLink" and "javaStdlibLink" options, which are
disabled by default.


semoro:
Sorry, but we can't accept this PR because of:
 * Option to disable JDK links already added, but not yet merged into dev (8e9e768d51d83c2ae1641a0c2d9ec4614c109117)
 * It changes the reasonable default
 * Disabling default external links will result in a drop in overall documentation quality, as no-one will override the default
 * It doesn't provide a deprecation cycle. This would result in unexpected behavior change for users and will be treated as a regression

stephenbyrne-mfj:
I would argue that requiring a connection to oracle.com at run time is **not** a reasonable default, but rest of the points stand.

semoro:
Providing option to disable such behavior seems fine to resolve this issue. Don't you think so?

stephenbyrne-mfj:
Yes, when 8e9e768 is released, I will be content.

PR #325, opened 2018-06-21
use canonicalPath instead of absolutePath for srcLink

### LINKED ISSUE #289: linkMapping does not work with relative path (requires absolute path)
See #286 for the discussion

robstoll:
This way a user can define "./" instead of an absolute path to the
root of the project dir (or a user can use ../ etc.). Thus:
- use canonicalPath in:
  - SourceLinkDefinitionImpl::parseSourceLinkDefinition
  - and DocumentationNode.appendSourceLink => here because if the
    config is deserialized we bypass parseSourceLinkDefinition

Moreover:
- make sure the comparison works for unix and windows paths
- fixes #289

### DIFF
--- a/core/src/main/kotlin/Model/SourceLinks.kt
+++ b/core/src/main/kotlin/Model/SourceLinks.kt
@@ -10,25 +10,47 @@ import java.io.File
 
 fun DocumentationNode.appendSourceLink(psi: PsiElement?, sourceLinks: List<SourceLinkDefinition>) {
     val path = psi?.containingFile?.virtualFile?.path ?: return
+    val canonicalPath = File(path).canonicalPath
 
     val target = if (psi is PsiNameIdentifierOwner) psi.nameIdentifier else psi
-    val absPath = File(path).absolutePath
-    val linkDef = sourceLinks.firstOrNull { absPath.startsWith(it.path) }
-    if (linkDef != null) {
-        var url = linkDef.url + path.substring(linkDef.path.length)
-        if (linkDef.lineSuffix != null) {
+    val pair = determineSourceLinkDefinition(canonicalPath, sourceLinks)
+    if (pair != null) {
+        val (sourceLinkDefinition, sourceLinkCanonicalPath) = pair
+        var url = determineUrl(canonicalPath, sourceLinkDefinition, sourceLinkCanonicalPath)
+        if (sourceLinkDefinition.lineSuffix != null) {
             val line = target?.lineNumber()
             if (line != null) {
-                url += linkDef.lineSuffix + line.toString()
+                url += sourceLinkDefinition.lineSuffix + line.toString()
             }
         }
-        append(DocumentationNode(url, Content.Empty, NodeKind.SourceUrl),
-                RefKind.Detail);
+        append(DocumentationNode(url, Content.Empty, NodeKind.SourceUrl), RefKind.Detail)
     }
 
     if (target != null) {
         append(DocumentationNode(target.sourcePosition(), Content.Empty, NodeKind.SourcePosition), RefKind.Detail)
     }
+}
+
+private fun determineSourceLinkDefinition(
+    canonicalPath: String,
+    sourceLinks: List<SourceLinkDefinition>
+): Pair<SourceLinkDefinition, String>? {
+    return sourceLinks
+        .asSequence()
+        .map { it to File(it.path).canonicalPath }
+        .firstOrNull { (_, sourceLinkCanonicalPath) ->
+            canonicalPath.startsWith(sourceLinkCanonicalPath)
+        }
+}
+
+private fun determineUrl(
+    canonicalPath: String,
+    sourceLinkDefinition: SourceLinkDefinition,
+    sourceLinkCanonicalPath: String
+): String {
+    val relativePath = canonicalPath.substring(sourceLinkCanonicalPath.length)
+    val relativeUrl = relativePath.replace('\\', '/').removePrefix("/")
+    return "${sourceLinkDefinition.url.removeSuffix("/")}/$relativeUrl"
 }
 
 private fun PsiElement.sourcePosition(): String {
--- a/README.md
+++ b/README.md
@@ -100,11 +100,11 @@ dokka {
     // If provided, Dokka generates "source" links for each declaration.
     // Repeat for multiple mappings
     linkMapping {
-        // Source directory
-        dir = "src/main/kotlin"
+        // Directory relative to the root of the project (where you execute gradle respectively). 
+        dir = "src/main/kotlin" // or simply "./"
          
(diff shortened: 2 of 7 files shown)


core/src/main/kotlin/Kotlin/DeclarationLinkResolver.kt
```diff
@@ -6,8 +6,6 @@ import org.jetbrains.kotlin.descriptors.CallableMemberDescriptor
 import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
 import org.jetbrains.kotlin.descriptors.TypeAliasDescriptor
 import org.jetbrains.kotlin.idea.kdoc.resolveKDocLink
-import org.jetbrains.kotlin.resolve.descriptorUtil.isEffectivelyPrivateApi
```
semoro: This change isn't relevant

core/src/test/kotlin/format/HtmlFormatTest.kt
```diff
@@ -3,7 +3,6 @@ package org.jetbrains.dokka.tests
 import org.jetbrains.dokka.*
 import org.jetbrains.kotlin.cli.jvm.config.JavaSourceRoot
 import org.jetbrains.kotlin.config.KotlinSourceRoot
-import org.junit.Before
```
semoro: This change isn't relevant

core/src/test/kotlin/format/KotlinWebSiteHtmlFormatTest.kt
```diff
@@ -1,7 +1,6 @@
 package org.jetbrains.dokka.tests
 
 import org.jetbrains.dokka.*
-import org.junit.Before
```
semoro: This change isn't relevant

core/src/test/kotlin/format/MarkdownFormatTest.kt
```diff
@@ -1,7 +1,6 @@
 package org.jetbrains.dokka.tests
 
 import org.jetbrains.dokka.*
-import org.junit.Before
```
semoro: This change isn't relevant

core/src/test/kotlin/issues/IssuesTest.kt
```diff
@@ -2,7 +2,6 @@ package issues
 
 import org.jetbrains.dokka.DocumentationNode
 import org.jetbrains.dokka.NodeKind
-import org.jetbrains.dokka.tests.toTestString
```
semoro: This change isn't relevant

core/src/test/kotlin/markdown/ParserTest.kt
```diff
@@ -5,7 +5,8 @@ import org.jetbrains.dokka.toTestString
 import org.jetbrains.dokka.parseMarkdown
 import org.junit.Ignore
 
-@Ignore public class ParserTest {
+@Ignore
```
semoro: This change isn't relevant

core/src/test/kotlin/model/CommentTest.kt
```diff
@@ -4,7 +4,7 @@ import org.junit.Test
 import org.junit.Assert.*
 import org.jetbrains.dokka.*
 
-public class CommentTest {
```
semoro: This change isn't relevant

core/src/test/kotlin/model/FunctionTest.kt
```diff
@@ -5,7 +5,6 @@ import org.jetbrains.dokka.NodeKind
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Test
-import kotlin.test.assertNotNull
```
semoro: This change isn't relevant

core/src/test/kotlin/model/PackageTest.kt
```diff
@@ -7,7 +7,7 @@ import org.jetbrains.kotlin.config.KotlinSourceRoot
 import org.junit.Assert.*
 import org.junit.Test
 
-public class PackageTest {
+class PackageTest {
```
semoro: This change isn't relevant

core/src/main/kotlin/Model/SourceLinks.kt
```diff
@@ -6,31 +6,61 @@ import com.intellij.psi.PsiNameIdentifierOwner
...
     if (target != null) {
         append(DocumentationNode(target.sourcePosition(), Content.Empty, NodeKind.SourcePosition), RefKind.Detail)
     }
 }
 
+private fun determineSourceLinkDefinition(
+    path: String,
+    sourceLinks: List<SourceLinkDefinition>
+): Pair<SourceLinkDefinition, String>? {
+    val absPath = File(path).absolutePath
```
semoro: Maybe better to use canonical path here also? Is it changes something?

core/src/main/kotlin/Model/SourceLinks.kt
```diff
@@ -6,31 +6,61 @@ import com.intellij.psi.PsiNameIdentifierOwner
...
+            absPath.startsWith(sourceLinkAbsPath)
+        }
+}
+
+private fun determineUrl(
+    path: String,
+    sourceLinkDefinition: SourceLinkDefinition,
+    sourceLinkAbsPath: String
+): String {
+    val relativePath = path.substring(sourceLinkAbsPath.length)
```
semoro: There should be better methods to determine relative path, like `File.relativeTo`

core/src/main/kotlin/Kotlin/DeclarationLinkResolver.kt
robstoll: fair enough, I'll create another pull request with the cleanup changes

core/src/main/kotlin/Model/SourceLinks.kt
robstoll: I'll simplify it further. Moreover I will no longer catch an IOException. It is better if the user sees the error message than beeing surprised that source linking does not work

robstoll: It would change something if `psi?.containingFile?.virtualFile?.path` returns paths containing `.`or `..`. I'll change it just to be sure

core/src/main/kotlin/Model/SourceLinks.kt
robstoll: A bit overkill here IMO. We already checked that both paths have the same root. We don't need to do this again with `relativeTo`

BenDenger:
do you already know when this will be merged?

robstoll:
I fixed all open change request some time ago, any other requests?

semoro:
Accepted, will be available in 0.9.18

Thank you for contribution 👍 
Sorry for the long wait 🙁

PR #357, opened 2018-09-15
Fixes #196 - make gradle plugin compatible with Kotlin-DSL by providing hooks according to Gradles API in addition to Groovy specific ones

### LINKED ISSUE #196: Gradle plugin does not play well with gradle-kotlin-dsl
The methods on DokkaTask all use Groovy Closures, which cannot be implemented in a convenient way from the gradle-kotlin-dsl.

If you use the the `Action<Subject>` interface provided by gradle, it works with all languages, plus it saves the work of configuring the closure delegate each time, as gradle will handle it for you.

```
fun externalDocumentationLink(action: Action<in DokkaConfiguration.ExternalDocumentationLink.Builder>)  {
    val builder = ...
    action.execute(builder)
    ...
}
```

(Providing additional methods with Kotlin lambda functions would make it even more convenient though)
(Also the gradle-kotlin-dsl helpers `delegateClosureOf()` do generate `Closure<Unit>` and are also not compatible in kotlin)

ramonwirsch:
added Action-based variants of configuration methods using closures:
linkMapping()
sourceRoot()
packageOptions()
externalDocumentationLink()

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/main.kt
+++ b/runners/gradle-plugin/src/main/kotlin/main.kt
@@ -1,10 +1,7 @@
 package org.jetbrains.dokka.gradle
 
 import groovy.lang.Closure
-import org.gradle.api.DefaultTask
-import org.gradle.api.Plugin
-import org.gradle.api.Project
-import org.gradle.api.Task
+import org.gradle.api.*
 import org.gradle.api.file.FileCollection
 import org.gradle.api.plugins.JavaBasePlugin
 import org.gradle.api.plugins.JavaPluginConvention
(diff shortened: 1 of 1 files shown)


svenjacobs:
Any news on this? :) I struggle with the `packageOptions` syntax in a Gradle Kotlin script.

svenjacobs:
In case anyone is having the same issue: The `delegateClosureOf()` extension functions helps wrapping a Gradle `Closure` in Kotlin DSL.

kamildoleglo:
Dokka has been rewritten since the creation of this PR and now supports Kotlin DSL as a first-class citizen

PR #395, opened 2018-12-12
389 modifiers sorted in wrong order

KrystianUjma:


core/src/main/kotlin/Kotlin/DocumentationBuilder.kt
```diff
@@ -332,7 +333,11 @@ class DocumentationBuilder
 
     fun DocumentationNode.appendModifiers(descriptor: DeclarationDescriptor) {
         val psi = (descriptor as DeclarationDescriptorWithSource).source.getPsi() as? KtModifierListOwner ?: return
-        KtTokens.MODIFIER_KEYWORDS_ARRAY.filter { it !in knownModifiers }.forEach {
+        KtTokens.MODIFIER_KEYWORDS_ARRAY.filter {
+            it !in knownModifiers
+        }.sortedWith (
```
semoro: Maybe `sortedBy`?

core/testdata/format/suspendInlineFunction.kt
```diff
@@ -0,0 +1,6 @@
+/**
+ * returns 1
+ */
+suspend inline fun foo(): Int {
```
semoro: Maybe it worth adding a test for different than canonical order?

KrystianUjma: done

PR #401, opened 2019-01-10
328 nullable left off function type parameter with default value

KrystianUjma:


core/src/main/kotlin/Kotlin/KotlinLanguageService.kt
```diff
@@ -166,8 +166,17 @@ class KotlinLanguageService : CommonLanguageService() {
             keyword("dynamic")
             return
         }
+
+        val nullabilityModifier = node.details(NodeKind.NullabilityModifier).singleOrNull()
```
semoro: `detailOrNull` might be better here

core/src/main/kotlin/Kotlin/KotlinLanguageService.kt
```diff
@@ -166,8 +166,17 @@ class KotlinLanguageService : CommonLanguageService() {
...
+
         if (node.isFunctionalType()) {
-            renderFunctionalType(node, renderMode)
+            nullabilityModifier ?.apply {
```
semoro: `?.let` is better, as a receiver not changed

core/src/main/kotlin/Kotlin/KotlinLanguageService.kt
KrystianUjma: done

core/src/main/kotlin/Kotlin/KotlinLanguageService.kt
KrystianUjma: agree but, in case of `?.let` if `symbol(it.name)` returns null for some reason, then `renderFunctionalType(node, renderMode)` will get executed. I will use `?.also` in this case.

core/src/main/kotlin/Kotlin/KotlinLanguageService.kt
```diff
@@ -167,14 +167,14 @@ class KotlinLanguageService : CommonLanguageService() {
             return
         }
 
-        val nullabilityModifier = node.details(NodeKind.NullabilityModifier).singleOrNull()
+        val nullabilityModifier = node.detailOrNull(NodeKind.NullabilityModifier)
 
         if (node.isFunctionalType()) {
-            nullabilityModifier ?.apply {
+            nullabilityModifier ?.also {
```
semoro: You can also use plain old `if`, nullabilityModifier will automatically cast to not-null type after check

KrystianUjma: done

Issue #403, opened 2019-01-17
(gradle dokka task output) add support to remove md extension in links for Gitlab Wiki formating


comdiv:
Now links in `gfm` forma are looks like `[text](dir/page.md)` it's not compatible with Gitlab Wiki while it uses    `[text](dir/page)` format.

Suggested 
1. add option to remove extension from links, for example `extension: String` 
2. add new output format  `glw` (GitLab Wiki) that will be `gfm` clone with setting `extension=""` by default

PR #404, opened 2019-01-17
GitLab wiki format added #403

comdiv:
Gitlab has markdown very close to GitHUB, but if it's used for Gitlab WIKI it has differene

1. Link should not have `.md` extensions `index.md -> index`
2. Link should be explicitly relative `index -> ./index`

So it require to rewrite href in links.

Decided to add 2 formats - `glm` (Gitlab Markdown) with option to be "wiki" and `glmw` for (GitLab Wiki) that is extension for `glm` with wiki option ON


MarcinAman:
Hi,
since then dokka has been rewritten. If you would like to see this feature in new dokka please adjust the code and make another PR.

Right now dokka uses a system of plugins. In order to create another plugin please refer to developer guide: https://kotlin.github.io/dokka/1.4.10/developer_guide/introduction/
In this plugin you would need to create a `LocationProvider` that would handle a logic related to resolving links and its extensions.

PR #407, opened 2019-01-18
Do not begin filenames with minus (-)

dschuermann:
Filenames are converted to lowercase, e.g., by replacing "A" to "-a". So "LongClassExampleFactory" will be "-long-class-example-factory". IMHO It should be "long-class-example-factory" without the minus (-) at the first character.

This pull request fixes this.

### DIFF
--- a/core/src/main/kotlin/Locations/Location.kt
+++ b/core/src/main/kotlin/Locations/Location.kt
@@ -45,7 +45,8 @@ fun relativePathToNode(node: DocumentationNode) = relativePathToNode(node.path.m
 
 fun identifierToFilename(path: String): String {
     val escaped = path.replace('<', '-').replace('>', '-')
-    val lowercase = escaped.replace("[A-Z]".toRegex()) { matchResult -> "-" + matchResult.value.toLowerCase() }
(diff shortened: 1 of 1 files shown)


yole:
How are you going to differentiate the filenames for declarations called "List" and "list"?

dschuermann:
Dang, I haven't thought that Kotlin/Java is case sensitive, you are right.

For now I will keep this patch in my upcoming dokka extension for building docs integrated in Hugo: https://github.com/cotechde/dokka-hugo-extension . Hugo "normalizes" certain URLs and in this process also removes the minus prefix resulting in broken links in dokka docs.

Could this be made configurable? IMHO URLs with a minus prefix are ugly in general. Or what about keeping the camel case as javadoc does?

yole:
Keeping the camel case won't work because we want Dokka to function on case-insensitive filesystems. Making this configurable could be possible, but we don't want to have a configuration option that generates broken docs in some environments. If you have suggestions for a less ugly escaping mechanism, please let us know.

hamoid:
Would it be possible to write a plugin to define our own naming rules?

yole:
I'm no longer involved in Dokka development, please don't @ me.

PR #417, opened 2019-01-29
300 constructor tag does not work

KrystianUjma:


core/src/main/kotlin/Kotlin/DescriptorDocumentationParser.kt
```diff
@@ -62,7 +64,10 @@ class DescriptorDocumentationParser
                 ?.resolveToDescriptorIfAny()
                 ?: descriptor
 
-        var kdocText = kdoc.getContent()
+        var kdocText = if (isDefaultNoArgConstructor) {
+             ((DescriptorToSourceUtils.descriptorToDeclaration(descriptor)?.navigationElement as? KtElement) as KtDeclaration).docComment?.findSectionByTag(KDocKnownTag.CONSTRUCTOR)?.getContent() ?: kdoc.getContent()
```
semoro: Please, extract this to a separate function. 
Also, maybe it makes sense to fix findKDoc function also

PR #424, opened 2019-02-14
Change annotation name to qualified name.

fvoidCN:
Annotation name is the simple name. It wouldn't distinguish the annotation Which has used the same name. So I think it should use the qualified name.


kamildoleglo:
Dokka has been rewritten since the creation of this PR, you can create a plugin to fix this if the issue still persists ☺️ 
If you wanted to do so, here's a short introduction for plugin development: https://kotlin.github.io/dokka/1.4.10.2/developer_guide/introduction/

Issue #319, opened 2018-06-09
Can't find node by signature not implemented for deserialized class KtFile with psi: null


arturbosch:
When building the [detekt](https://github.com/arturbosch/detekt) project with `gradle build` we are getting dokka warnings like:

```
Can't find node by signature not implemented for deserialized class KtFile with psi: null, referenced at io.gitlab.arturbosch.detekt.api.RuleSet$accept(org.jetbrains.kotlin.psi.KtFile, kotlin.collections.Set((kotlin.String))) (RuleSet.kt:22)
Can't find node by signature not implemented for deserialized class KtFile with psi: null, referenced at io.gitlab.arturbosch.detekt.api.RuleSet$accept(org.jetbrains.kotlin.psi.KtFile, kotlin.collections.Set((kotlin.String))) (RuleSet.kt:22)
null:-1:-1: Tag @see cannot be used in inline documentation.  It can only be used in the following types of documentation: overview, package, class/interface, constructor, field, method.
null:-1:-1: Tag @see cannot be used in inline documentation.  It can only be used in the following types of documentation: overview, package, class/interface, constructor, field, method.
Can't find node by signature not implemented for deserialized class KtFile with psi: null, referenced at io.gitlab.arturbosch.detekt.api.RuleSet$accept(org.jetbrains.kotlin.psi.KtFile, kotlin.collections.Set((kotlin.String))) (RuleSet.kt:22)
```

The mentioned `RuleSet` class looks like this:
```kotlin
package io.gitlab.arturbosch.detekt.api

import org.jetbrains.kotlin.psi.KtFile

/**
 * A rule set is a collection of rules and must be defined within a rule set provider implementation.
 *
 * @author Artur Bosch
 */
class RuleSet(val id: String, val rules: List<BaseRule>) {

	init {
		validateIdentifier(id)
	}

	/**
	 * Visits given file with all rules of this rule set, returning a list
	 * of all code smell findings.
	 */
	fun accept(file: KtFile): List<Finding> = rules.flatMap { it.visitFile(file); it.findings }

	/**
	 * Visits given file with all non-filtered rules of this rule set.
	 * If a rule is a [MultiRule] the filters are passed to it via a setter
	 * and later used to filter sub rules of the [MultiRule].
	 *
	 * A list of findings is returned for given [KtFile]
	 */
	fun accept(file: KtFile, ruleFilters: Set<String>): List<Finding> =
			rules.filterNot { it.id in ruleFilters }
[...]

Issue #284, opened 2018-03-09
Can't find node by signature, defined in other project


mathjeff:
Full reproduction case at https://github.com/mathjeff/dokka-tests/tree/master/import-from-other-project-not-found

Essentially there are two projects.

project JefflibA contains class com.jeffclass.JeffClass which defines a static string field, STATIC_FIELD
project JefflibB contains class com.jeffclass2.JeffClass2 which references com.jeffclass.JeffClass#STATIC_FIELD

However, dokka doesn't resolve the reference and instead prints "Can't find node by signature com.jeffclass.JeffClass$STATIC_FIELD"

The reference is found when coming from within the same project, however.

kamilok1965:
This bug willl be fixed in the upcoming release, 0.9.18.
You will need to add to your project jefflibB `build.gradle` file something like this: 
```
 evaluationDependsOn(":jefflibA")
 afterEvaluate {
   dokka {
     classpath = configurations.compile
     externalDocumentationLink {
       packageListUrl = new URL("file://${project(":jefflibA").buildDir}/dokka/jefflib-a/package-list")
       url = new URL("file://${project(":jefflibA").buildDir}/dokka/jefflib-a/")
     }
   }
 }
```

Issue #278, opened 2018-03-01
Can't find node by signature not implemented for value-parameter 


madorb:
I have the following code:

```
/**
 * Return a errors as a Set. If [errors] is empty an empty set is returned.
 * @param errors An array of errors
 */
fun createErrors(vararg errors: Error): Set<Error> = if (errors.isEmpty()) {
    emptySet()
} else {
    HashSet(errors.asList())
}
```

I've defined a `gradle` task to generate javadoc:

```
task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
        outputFormat = 'javadoc'
        outputDirectory = "$buildDir/dokkaJavadoc"
}
```
running task `gradle dokkaJavadoc` i get the following error:

```
Can't find node by signature not implemented for value-parameter vararg errors: com.foo.exception.error.Error defined in com.foo.exception.error.createErrors[ValueParameterDescriptorImpl@6fe7f16c] with psi: VALUE_PARAMETER, referenced at com.foo.exception.error$createErrors(kotlin.Array((com.foo.exception.error.Error))) (ErrorUtil.kt:64)
```

skalarproduktraum:
I'm seeing the same issue. @semoro: Is there some workaround or something? What's the cause of this problem?

KrystianUjma:
PR created [here](https://github.com/Kotlin/dokka/pull/431).

Issue #269, opened 2018-02-05
Can't find node by signature spam


ZacSweers:
This is a duplicate of #184 but I was asked to make a new issue.

Here is a repro project on 0.9.16-EAP-3 - https://github.com/uber/AutoDispose

and an example of the output data

```
> Task :android:autodispose-android-archcomponents-kotlin:dokka 
warning: duplicate source root: /Users/pandanomic/dev/java/contributing/AutoDispose/android/autodispose-android-archcomponents-kotlin/src/main/kotlin
Can't find node by signature com.uber.autodispose.LifecycleScopeProvider, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner) (autodisposeAndroidLifecycle.kt:28)
Can't find node by signature android.arch.lifecycle.LifecycleOwner, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner) (autodisposeAndroidLifecycle.kt:28)
Can't find node by signature com.uber.autodispose.LifecycleScopeProvider, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner) (autodisposeAndroidLifecycle.kt:28)
Can't find node by signature android.arch.lifecycle.LifecycleOwner, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner) (autodisposeAndroidLifecycle.kt:28)
Can't find node by signature com.uber.autodispose.LifecycleScopeProvider, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner, android.arch.lifecycle.Lifecycle.Event) (autodisposeAndroidLifecycle.kt:34)
Can't find node by signature android.arch.lifecycle.LifecycleOwner, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner, android.arch.lifecycle.Lifecycle.Event) (autodisposeAndroidLifecycle.kt:34)
Can't find node by signature com.uber.autodispose.LifecycleScopeProvider, referenced at com.uber.autodispose.android.lifecycle$scope(android.arch.lifecycle.LifecycleOwner, io.reactivex.functions.Function((android.arch.lifecycle.Lifecycle.Event, ))) (autodisposeAndroidLifecycle.kt:43)
[...]

PR #427, opened 2019-02-15
support for gradle multiproject

Kordyjan:
There are lots of problems (eg. #284 #269 #278 #319) resulting from the fact that dokka gradle task is only aware of the project that it was defined in and not its dependencies.

In this pull request I'm making dokka aware of dependencies between projects. Now generating documentation depends on generating documentation for dependencies. In the same manner documentations of dependencies are treated as inputs to project documentation generation. This allows us to have a realiable relative links between documentations of different subprojects.

Sadly this solution is not working for javadoc output format, because resolver for javadoc [is not finished](https://github.com/Kotlin/dokka/blob/dev-pre/core/src/main/kotlin/Kotlin/ExternalDocumentationLinkResolver.kt#L230), but as soon as more cases are added to this resolver the links to javadoc will start to work properly.


pyricau:
Any chance this will make it into a future release?

RaymondChan1:
I would definitely love to see this implemented. Good luck!

RingerJK:
are you folks gonna merge it?

kamildoleglo:
This has been fixed in the new Dokka

Issue #316, opened 2018-06-04
Dokka fatJar resolution in Gradle plugin should use all resolution results


diendanyoi54:
I'm receiving this error when trying to build my project with Spring Auto Restdocs amongst a decent amount of other Spring tools.  Here's a link to the issue I created in the Auto Restdocs repository.  https://github.com/ScaCap/spring-auto-restdocs/issues/235.  Google only returns one semi-related item which resulted in a fix to the code.

semoro:
Could you please provide a full stack trace? By running Gradle with `--stacktrace` option

diendanyoi54:
Is this what you were looking for?
```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':dokka'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:103)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:73)
	at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
	at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:59)
	at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:101)
	at org.gradle.api.internal.tasks.execution.FinalizeInputFilePropertiesTaskExecuter.execute(FinalizeInputFilePropertiesTaskExecuter.java:44)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:91)
	at org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter.execute(ResolveTaskArtifactStateTaskExecuter.java:62)
[...]

PR #432, opened 2019-02-25
316 dokka fatjar repack

KrystianUjma:


KrystianUjma:
#316

runners/gradle-plugin/src/main/kotlin/main.kt
```diff
@@ -216,24 +218,21 @@ open class DokkaTask : DefaultTask() {
...
     }
 
-    fun tryResolveFatJar(project: Project): File {
+    fun tryResolveFatJar(project: Project): Set<File> {
         return try {
-            val dependency = project.buildscript.dependencies.create(dokkaFatJar)
-            val configuration = project.buildscript.configurations.detachedConfiguration(dependency)
-            configuration.description = "Dokka main jar"
-            configuration.resolve().first()
+            dokkaRuntime!!.resolve()
```
semoro: Looks like if `dokkaRuntime` wasn't initialized, this code will fail

runners/gradle-plugin/src/main/kotlin/main.kt
```diff
@@ -298,6 +297,7 @@ open class DokkaTask : DefaultTask() {
 
     @TaskAction
     fun generate() {
+        project.configurations.getByName("dokkaRuntime").defaultDependencies{ dependencies -> dependencies.add(project.dependencies.create(dokkaFatJar)) }
```
semoro: Looks like you should initialize property `dokkaRuntime` in case it is null and apply `defaultDependencies` to it

PR #466, opened 2019-04-16
fix fatjar argument parsing

vincentlauvlwj:
Using the fatjar with a `-jdkVersion 8` argument will cause an exception now: 

````raw
Exception in thread "main" java.lang.IllegalArgumentException: Could not find constructor in class int that takes a string
        at com.sampullara.cli.Args.setProperty(Args.java:351)
        at com.sampullara.cli.Args.processProperty(Args.java:163)
        at com.sampullara.cli.Args.parse(Args.java:55)
        at org.jetbrains.dokka.MainKt.main(main.kt:172)
Caused by: java.lang.NoSuchMethodException: int.<init>(java.lang.String)
        at java.lang.Class.getConstructor0(Class.java:3082)
        at java.lang.Class.getDeclaredConstructor(Class.java:2178)
        at com.sampullara.cli.Args.createValue(Args.java:408)
        at com.sampullara.cli.Args.getValue(Args.java:401)
        at com.sampullara.cli.Args.setProperty(Args.java:346)
        ... 3 more
````

So I changed the types of all the properties in DokkaArguments to String to fix this bug.

### DIFF
--- a/runners/cli/src/main/kotlin/cli/main.kt
+++ b/runners/cli/src/main/kotlin/cli/main.kt
@@ -36,10 +36,10 @@ class DokkaArguments {
     var classpath: String = ""
 
     @set:Argument(value = "nodeprecated", description = "Exclude deprecated members from documentation")
-    var nodeprecated: Boolean = false
+    var nodeprecated: String = "false"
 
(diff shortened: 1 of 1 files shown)


MarcinAman:
Hi, 
dokka has been rewritten and now it uses kotlinx.cli to parse configuration and those changes are no longer needed

PR #485, opened 2019-07-03
Include fallback monospace font in style.css

whbboyd:
Fallback to a monospace font on systems without any of the listed concrete
fonts. (This includes a very large number of Linux systems.) Otherwise they
will fall back to the system default, which is almost certainly proportional.


kamildoleglo:
Please rebase

PR #499, opened 2019-09-15
Don't create filenames that conflict with Windows reserved filenames.

drmoose:
A project that has classes/members whose names match any of the [windows reserved filenames](https://superuser.com/a/613335) can cause Dokka to omit documentation for that member, sometimes with cryptic log messages such as:

```
java.io.IOException: The device is not connected
```

This PR attempts to address that issue by blacklisting the Windows-reserved filenames in the `identifierToFilename` function.

A minimalist test project exists [here](https://github.com/drmoose/dokka-windows-example). The latest release tries to create a file `prn.html` for the example class, which fails on Windows. With this PR, that file is called `--prn--.html`, which is allowed.

```kotlin
/** An example class to trigger the filename bug */
data class Example(
    /** gps satellite ID number */
    val prn: Int)
```

### DIFF
--- a/core/src/main/kotlin/Locations/Location.kt
+++ b/core/src/main/kotlin/Locations/Location.kt
@@ -43,10 +43,12 @@ fun relativePathToNode(qualifiedName: List<String>, hasMembers: Boolean): String
 
 fun relativePathToNode(node: DocumentationNode) = relativePathToNode(node.path.map { it.name }, node.members.any())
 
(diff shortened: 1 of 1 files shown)


MarcinAman:
Hi, i think same changes are [already merged into master](https://github.com/Kotlin/dokka/blob/abe5113df0aeece35fc959e6b3b15ee125fe101c/plugins/base/src/main/kotlin/resolvers/local/DokkaLocationProvider.kt#L108), so i'll close this PR.

Issue #503, opened 2019-09-22
Display actual class names instead of `<ERROR CLASS>`


cjbrooks12:
When the project classpath is not provided to Dokka, references to classes in other libraries are displayed as `<ERROR CLASS>`. For projects using the Gradle plugins to run Dokka, this is fine since they are able to properly provide the classpath to Dokka, but for running Dokka on the CLI directly from its fat jar, this is not fine. 

I don't fully understand the internals of how Dokka parses the Kotlin files, but it seems like Dokka should be able to simply output the String value of that class reference instead of completely erroring out, since that String is just a part of the source code and can be deduced just by reading that single file, without needing to look anything up from the classpath. Since this is documentation, even an unlinked String for a type name is better than `<ERROR CLASS>`.

cjbrooks12:
I think I've got a fix for this. I'll open a PR shortly for review

semoro:
It is not possible in case of implicit function/property type
```
fun returnsUnresolvedString() = ""
```

cjbrooks12:
Yeah, that's what I had found as well, but I was able to fix it for the majority of cases. 

However, after a bit more digging, I found that I was able to get the entire expression as a String directly from the PSI tree. Maybe we could just render that whole raw expression directly into the docs? It's not ideal, as none of its types will be linked, but it will still be better than simply display an error message. I could update my PR with this new finding.

semoro:
It is not clear, why you even want to run Dokka with incorrect configuration and get something?

cjbrooks12:
I'm building a [documentation site generator](https://github.com/JavaEden/Orchid) that has functionality to wrap Dokka and embed that data in a custom theme (think Jekyll, but with plugins for code documentation). To do this, I am running Dokka as a fatjar on the command-line with a custom formatter that renders the Dokka models as JSON, which can then be used by Orchid to render that data in theme templates.

The reason I am trying to fix this and get it displayed properly is that I am not running Dokka from Gradle, and thus am unable to provide the complete configuration Dokka expects. In fact, a big feature of Orchid is that it is not tied directly to any build system, and so could be used for things like one repo cloning several others and documenting them all from one place. The most common use-case for using Orchid is to set it up in a single `docs` subproject and point it to your other modules in a repo, rather than having Orchid set up for each individual module. This makes it really nice for aggregating documentation from many different projects into a single website with minimal configuration, but it also means that it cannot easily access the classpath of those other modules it is documenting.

[...]

PR #509, opened 2019-10-05
#503: type references not on classpath displayed as ERROR CLASS

cjbrooks12:
Fixes issue #503: type references not on classpath displayed as ERROR CLASS.

Missing classes used for the bounds of generic methods and for inferred method return types do not seem possible to resolve.


cjbrooks12:
I've just rebased this branch off the current master (0.10.0) and also added support for inlining the AST text of expressions that could not be resolved normally (such as implicit function/property types)

kamildoleglo:
Dokka has been rewritten since the creation of this PR, please create a new one if the issue persists ☺️

PR #527, opened 2019-11-15
Resolved problems with android multiflavors documentation shadowing

BarkingBad:
The problem was with generating documentation for multiflavored android project. 
From now it will generate documentation for all stated flavors through dokka task's parameter.

Note it will actually squash JDocs / KDocs of all classes / methods that share same name. For convenient results pass single flavor that you want to generate documentation of.

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/ConfigurationExtractor.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/ConfigurationExtractor.kt
@@ -24,13 +24,13 @@ import java.io.Serializable
 
 class ConfigurationExtractor(private val project: Project) {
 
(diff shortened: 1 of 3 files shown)


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/ConfigurationExtractor.kt
```diff
@@ -24,13 +24,13 @@ import java.io.Serializable
 
 class ConfigurationExtractor(private val project: Project) {
 
-    fun extractConfiguration(targetName: String, variantName: String?) = if (project.isMultiplatformProject()) {
-        extractFromMultiPlatform(targetName, variantName)
+    fun extractConfiguration(targetName: String, variantNames: List<String?>) = if (project.isMultiplatformProject()) {
```
kamilok1965: I think we don't need `List<String?>` here, `List<String>` would be sufficient, because the nullability here doesn't have any meaning

PR #530, opened 2019-11-27
Ignore samples.Sample import when generating runnable samples

TillmannBerg:
Some of the runnable samples for the Kotlin standard library fail to run on the website, because the `import samples.Sample` leaks through from the Kotlin project onto the website and cannot be found there.

Affected are for example https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/with-index.html and https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/as-reversed.html .

This pull request fixes that by not only ignoring imports of `samples.*` but also `samples.Sample`.

### DIFF
--- a/core/src/main/kotlin/Samples/KotlinWebsiteSampleProcessingService.kt
+++ b/core/src/main/kotlin/Samples/KotlinWebsiteSampleProcessingService.kt
@@ -165,7 +165,7 @@ open class KotlinWebsiteSampleProcessingService
         return sampleBuilder.text
     }
 
-    val importsToIgnore = arrayOf("samples.*").map { ImportPath.fromString(it) }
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/Samples/KotlinWebsiteSampleProcessingService.kt
```diff
@@ -165,7 +165,7 @@ open class KotlinWebsiteSampleProcessingService
         return sampleBuilder.text
     }
 
-    val importsToIgnore = arrayOf("samples.*").map { ImportPath.fromString(it) }
```
ilya-g: LGTM. Note however, that there are some other declarations in the `samples` package, that could leak into the generated docs. Thus I'd prefer if the samples kept using a star-import of the `samples` package and dokka reported a warning otherwise (in case of non-star import).

PR #536, opened 2019-12-06
Fix task method and plugin class signature for extension

Den-Rimus:


Den-Rimus:
Sorry - mixed up the base project

PR #538, opened 2019-12-11
AnalysisExtensions

i-walker:
AnalysisExtensions allow plugin Devs to manipulate ResolutionFacade, which currently 
`DokkaResolutionFacade`.
We may want to consider refactoring `org.jetbrains.dokka.EnvironmentAndFacade` so it allows
Any given ResolutionFacade. I am open to feedback on any matter. If you have questions about the code feel free to ask.

### DIFF
--- a/core/src/main/kotlin/analysis/AnalysisTransformer.kt
+++ b/core/src/main/kotlin/analysis/AnalysisTransformer.kt
@@ -1,0 +1,11 @@
+package org.jetbrains.dokka.analysis
+
+import org.jetbrains.dokka.EnvironmentAndFacade
+import org.jetbrains.dokka.pages.PlatformData
+
+/**
+ * [AnalysisTransformer] allows to map over
+ */
+interface AnalysisTransformer {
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/analysis/AnalysisTransformer.kt
```diff
@@ -0,0 +1,11 @@
+package org.jetbrains.dokka.analysis
+
+import org.jetbrains.dokka.EnvironmentAndFacade
+import org.jetbrains.dokka.pages.PlatformData
+
+/**
+ * [AnalysisTransformer] allows to map over
```
raulraja: This description seems incomplete. `to map over...` what is it mapping over?

i-walker: Yes, sure I forgot to commit that change

sellmair:
@i-walker Hey! Could please check whether #1008 will also give you all the power you need?

Please note that #1008 was not built for that use case in mind, but happens to solve your problem :)

sellmair:
I am closing this PR for now. Feel free to re-open this if you feel like #1008 does not fit your needs (and cannot be adapted towards it)

Issue #160, opened 2017-05-18
Javadoc is missing a lot of information


joserobjr:
With this simple test class:
```java
package test;

import org.jetbrains.annotations.NotNull;
import org.jetbrains.annotations.Nullable;

public class Test
{
    @NotNull
    public static String test(@Nullable String test) {
        if(test == null)
            return "null";
        return test.toLowerCase();
    }
}
```
Dokka is generating an incomplete documentation page for it in javadoc mode.

It is missing the "extends Object" information, the "Methods inherited from class java.lang.Object", the default constructor "Test()", the method annotations and the method parameter.

---

## Page generated by dokka
![image](https://cloud.githubusercontent.com/assets/3679022/26227309/fe5830e2-3c07-11e7-85c2-dbb484ef2234.png)

---

## Page generated by javadoc
![image](https://cloud.githubusercontent.com/assets/3679022/26227296/e4c26d82-3c07-11e7-9857-6fd8dd2186af.png)

---

The same is applied to Kotkin classes but it is missing less information then Java classes.

Example Kotlin class:
```kotlin
class TestKotlin {
    fun test(test: String?): String {
        if (test == null)
            return "null"
        return test.toLowerCase()
    }
```
---

## Page generated by dokka
![image](https://cloud.githubusercontent.com/assets/3679022/26227494/495d1e6c-3c09-11e7-89a9-607e612c4113.png)

---

The most critical issues are the arguments missing in method information, the default constructor being omitted and the annotations which contains the `@Documented` decoration not being documented as they should.

JJK3:
It looks like javadocs are still missing the method summary, has there been any progress on this? I'm willing to look into this and try to fix it, I just don't want to put needless effort into it if someone else is working on it.

kamilok1965:
Yes and no. We'd really appreciate your help with this, @JJK3, just not now, because we're refactoring dokka quite extensively. We're changing the whole internal model, so fixing this javadoc format would be a waste of time, because it will have to be rewritten from scratch before the next big release (0.11.0).

AfzalivE:
@kamildoleglo Do you mean that you need someone to implement `CommonLanguageService` in `JavaLanguageService`?

Kordyjan:
As far as I know, this is fixed in new javadoc that was added in 1.4.0

PR #542, opened 2019-12-13
Implement display of Java default constructors

asfalcone:
PsiClass does not include methods which are not present in the text
of the code. So when there are no constructors in a java class this
represents a class that only has a default constructor. The fix is
to create one for documentation.

### DIFF
--- a/core/src/main/kotlin/Java/JavaPsiDocumentationBuilder.kt
+++ b/core/src/main/kotlin/Java/JavaPsiDocumentationBuilder.kt
@@ -29,7 +29,7 @@ fun getSignature(element: PsiElement?) = when(element) {
 }
 
 private fun methodSignature(method: PsiMethod): String {
-    return method.containingClass!!.qualifiedName + "$" + method.name + "(" +
+    return method.containingClass?.qualifiedName + "$" + method.name + "(" +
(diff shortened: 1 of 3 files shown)


asfalcone:
Partially solves #160

core/src/main/kotlin/Java/JavaPsiDocumentationBuilder.kt
```diff
@@ -167,7 +167,9 @@ class JavaPsiDocumentationBuilder : JavaDocumentationBuilder {
                         element.isInternal())
 
     private fun skipElementBySuppressedFiles(element: Any): Boolean =
-            element is PsiElement && element.containingFile.virtualFile.path in passConfiguration.suppressedFiles
+        element is PsiElement &&
+            element.containingFile.virtualFile != null &&
+            File(element.containingFile.virtualFile.path).absoluteFile in options.suppressedFiles
```
kamilok1965: `File` lacks an import and `options` have been replaced with `passConfiguration`. Also the type of `suppressedFiles` changed from `List<File>` to `List<String>`

asfalcone: Sorry about that, merging from our fork did not work very well!

PR #544, opened 2019-12-16
Dev 0.11.0

sswistun-vl:
Few fixes for Kotlin stdlib Dokka generation.

### DIFF
--- a/core/src/main/kotlin/DokkaGenerator.kt
+++ b/core/src/main/kotlin/DokkaGenerator.kt
@@ -24,14 +24,14 @@ class DokkaGenerator(
     fun generate() {
         logger.debug("Setting up analysis environments")
         val platforms: Map<PlatformData, EnvironmentAndFacade> = configuration.passesConfigurations.map {
-            PlatformData(it.analysisPlatform, it.targets) to createEnvironmentAndFacade(it)
+            PlatformData(it.moduleName, it.analysisPlatform, it.targets) to createEnvironmentAndFacade(it)
         }.toMap()
 
         logger.debug("Initializing plugins")
         val context = DokkaContext.create(configuration.pluginsClasspath, logger, platforms)
 
         logger.debug("Creating documentation models")
-        val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(configuration.passesConfigurations.first().moduleName, pdata, context) }
+        val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(pdata, context) }
 
         logger.debug("Merging documentation models")
         val documentationModel = context.single(CoreExtensions.documentationMerger)
@@ -75,7 +75,7 @@ class DokkaGenerator(
             EnvironmentAndFacade(environment, facade)
         }
 
-    private fun translateDescriptors(moduleName: String, platformData: PlatformData, context: DokkaContext): Module {
+    private fun translateDescriptors(platformData: PlatformData, context: DokkaContext): Module {
         val (environment, facade) = context.platforms.getValue(platformData)
 
         val packageFragments = environment.getSourceFiles().asSequence()
@@ -85,7 +85,7 @@ class DokkaGenerator(
             .toList()
 
         return context.single(CoreExtensions.descriptorToDocumentationTranslator)
-            .invoke(moduleName, packageFragments, platformData, context)
+            .invoke(platformData.name, packageFragments, platformData, context)
     }
 
     private class DokkaMessageCollector(private val logger: DokkaLogger) : MessageCollector {
--- a/core/src/main/kotlin/DokkaGenerator.kt
+++ b/core/src/main/kotlin/DokkaGenerator.kt
@@ -31,7 +31,7 @@ class DokkaGenerator(
         val context = DokkaContext.create(configuration.pluginsClasspath, logger, platforms)
 
         logger.debug("Creating documentation models")
-        val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(pdata, context) }
+        val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(configuration.passesConfigurations.first().moduleName, pdata, context) }
 
         logger.debug("Merging documentation models")
         val documentationModel = context.single(CoreExtensions.documentationMerger)
@@ -75,7 +75,7 @@ class DokkaGenerator(
             EnvironmentAndFacade(environment, facade)
         }
 
-    private fun translateDescriptors(platformData: PlatformData, context: DokkaContext): Module {
+    private fun translateDescriptors(moduleName: String, platformData: PlatformData, context: DokkaContext): Module {
         val (environment, facade) = context.platforms.getValue(platformData)
 
         val packageFragments = environment.getSourceFiles().asSequence()
@@ -85,7 +85,7 @@ class DokkaGenerator(
             .toList()
 
         return context.single(CoreExtensions.descriptorToDocumentationTranslator)
-            .invoke(packageFragments, platformData, context)
+            .invoke(moduleName, packageFragments, platformData, context)
     }
 
     private class DokkaMessageCollector(private val logger: DokkaLogger) : MessageCollector {
--- a/core/src/main/kotlin/links/DRI.kt
+++ b/core/src/main/kotlin/links/DRI.kt
@@ -94,10 +94,11 @@ data class TypeReference(val classNames: String, val typeBounds: List<TypeRefere
             TypeReference("", tp.upperBounds.map { from(it) })
 
         private fun from(t: KotlinType): TypeReference =
-            when (val d = t.constructor.declarationDescriptor) {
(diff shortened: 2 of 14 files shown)


core/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -74,20 +74,20 @@ class MarkdownParser (
...
-                        resolveKDocLink(
+                    resolveKDocLink(
                         resolutionFacade.resolveSession.bindingContext,
                         resolutionFacade,
                         declarationDescriptor,
                         null,
                         link.split('.')
-                    ).single()
-                )
+                    ).also { if (it.size > 1) println("Markdown link resolved more than one element: $it") }.firstOrNull()//.single()
```
sellophane: Please use logger instead

core/src/main/kotlin/DokkaGenerator.kt
```diff
@@ -31,7 +31,7 @@ class DokkaGenerator(
         val context = DokkaContext.create(configuration.pluginsClasspath, logger, platforms)
 
         logger.debug("Creating documentation models")
-        val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(pdata, context) }
+        val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(configuration.passesConfigurations.first().moduleName, pdata, context) }
```
sellophane: That is wrong. 

A bit earlier we are doing 
`val platforms: Map<PlatformData, EnvironmentAndFacade> = configuration.passesConfigurations.map {
            PlatformData(it.analysisPlatform, it.targets) to createEnvironmentAndFacade(it)
        }.toMap()`

so now we are actually iterating over a list of passesConfigurations after mapping. I would rather suggest adding `moduleName` to `PlatformData` class, so this lines would look as follows:

`val platforms: Map<PlatformData, EnvironmentAndFacade> = configuration.passesConfigurations.map {
            PlatformData(it.analysisPlatform, it.targets, it.moduleName) to createEnvironmentAndFacade(it)
        }.toMap()`

and then

`val modulesFromPlatforms = platforms.map { (pdata, _) -> translateDescriptors(pdata.moduleName, pdata, context) }`
You can even skip passing `moduleName` here, you can simply retrieve it from pdata later, inside `translateDescriptors` function.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -95,7 +95,7 @@ data class TypeReference(val classNames: String, val typeBounds: List<TypeRefere
 
         private fun from(t: KotlinType): TypeReference =
             when (val d = t.constructor.declarationDescriptor) {
-                is TypeParameterDescriptor -> from(d)
+                is TypeParameterDescriptor -> TypeReference("todo:${d.name}")
```
sellophane: What is the purpose of this?

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -5,7 +5,7 @@ import org.jetbrains.dokka.transformers.descriptors.KotlinTypeWrapper
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.pages.PlatformData
 
-class Module(val packages: List<Package>) : Documentable() {
+class Module(override val name: String?, val packages: List<Package>) : Documentable() {
```
sellophane: Correct me if I'm wrong, but do we ever create `Module` with nulled name? Because I failed to find this situation, and if so, we do not need a nullability here.

core/src/main/kotlin/links/DRI.kt
sswistun-vl: Right now there's infinite recursion when used on bigger project like Kotlin stdlib. It's temporary "solution" to get it to finish.

core/src/main/kotlin/DokkaGenerator.kt
sswistun-vl: Changed.

core/src/main/kotlin/parsers/MarkdownParser.kt
sswistun-vl: Changed.

core/src/main/kotlin/model/Documentable.kt
sswistun-vl: Changed.

core/src/main/kotlin/pages/PageBuilder.kt
```diff
@@ -9,13 +9,12 @@ class DefaultPageBuilder(
 ) : PageBuilder {
 
     override fun pageForModule(m: Module): ModulePageNode =
-        ModulePageNode("root", contentForModule(m), m, m.packages.map { pageForPackage(it) })
+        ModulePageNode(m.name ?: "root", contentForModule(m), m, m.packages.map { pageForPackage(it) })
```
sellophane: It won't be null anymore

sellophane: But you might actually want to check for emptiness, and then put "root" as a name

core/src/main/kotlin/transformers/documentation/DefaultDocumentationNodeMerger.kt
```diff
@@ -7,6 +7,7 @@ import org.jetbrains.dokka.plugability.DokkaContext
 internal object DefaultDocumentationNodeMerger : DocumentationNodeMerger {
     override fun invoke(modules: Collection<Module>, context: DokkaContext): Module =
         Module(
+            modules.first().name,
```
kamilok1965: This should check whether all the modules have the same name and log an error if not

core/src/main/kotlin/pages/PageBuilder.kt
sswistun-vl: Done.

core/src/main/kotlin/transformers/documentation/DefaultDocumentationNodeMerger.kt
sswistun-vl: Done.

kamilok1965: I think that the behavior of `all{}` on an empty list is unintuitive and we should remove `drop(1)` even though it  works correctly. WDYT @Kordyjan?

sswistun-vl: Removed.

Kordyjan:
@kamilok1965 @sellophane Any updates? Can we merge it?

core/src/main/kotlin/renderers/FileWriter.kt
```diff
@@ -1,5 +1,6 @@
 package org.jetbrains.dokka.renderers
 
+import org.jetbrains.dokka.utilities.DokkaConsoleLogger
```
kamilok1965: Why do we need this import?

core/src/main/kotlin/pages/PageNodes.kt
```diff
@@ -119,7 +119,7 @@ class MemberPageNode(
         else MemberPageNode(name, content, dri, documentable, children, embeddedResources)
 }
 
-data class PlatformData(val platformType: Platform, val targets: List<String>) {
+data class PlatformData(val name: String, val platformType: Platform, val targets: List<String>) {
```
kamilok1965: I wonder whether we should pass the `moduleName` using `PlatformData` or should we have a "global" one in the `context`? @sellophane ?

core/src/main/kotlin/links/DRI.kt
```diff
@@ -94,10 +94,11 @@ data class TypeReference(val classNames: String, val typeBounds: List<TypeRefere
...
         private fun from(t: KotlinType): TypeReference =
-            when (val d = t.constructor.declarationDescriptor) {
-                is TypeParameterDescriptor -> from(d)
-                else -> TypeReference(t.constructorName.orEmpty(), t.arguments.map { from(it) })
-            }
+            TypeReference(t.constructorName.orEmpty(), t.arguments.map { from(it) })
+//            when (val d = t.constructor.declarationDescriptor) {
+//                is TypeParameterDescriptor -> TypeReference("todo:${d.name}")
+//                else -> TypeReference(t.constructorName.orEmpty(), t.arguments.map { from(it) })
+//            }
```
sellophane: Please try not to leave commented code if it's not necessary.

core/src/main/kotlin/pages/PageNodes.kt
sellophane: Well, it might be a good idea actually

core/src/main/kotlin/links/DRI.kt
sswistun-vl: Removed. I've left it as a reminder that it's not permanent solution.

core/src/main/kotlin/renderers/FileWriter.kt
sswistun-vl: Removed.

PR #545, opened 2019-12-18
[WIP] Type Checking Plugin

i-walker:
It follows the same style as the DefaultDescriptionToDocumentationTranslator.
With the difference, that it intercepts Code Fences, Code Blocks, Code Span and compiles it. 

I am still investigating, which way makes sense, because `DokkaResolutionFacade`, does not implement `analyzeWithAllCompilerChecks` and using `KotlinCacheServiceImpl`, bails bc SdkTypes are not initialized at RunTime.


i-walker:
Closing this in favor of PostProcess Extensions: #571

PR #547, opened 2019-12-18
Gh actions

sswistun-vl:
CI based on Github Actions.

### DIFF
--- a/bin/example-dokka.sh
+++ b/bin/example-dokka.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+patch='Index: build.gradle.kts
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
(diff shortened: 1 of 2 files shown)


bin/example-dokka.sh
```diff
@@ -0,0 +1,25 @@
...
+<+>UTF-8
+===================================================================
+--- build.gradle.kts	(revision 58f85a4dce464cb5eea3d9dba4d8337c90ba3d9d)
++++ build.gradle.kts	(date 1576160110756)
+@@ -3,7 +3,7 @@
+ plugins {
+     id("java")
+     id("org.jetbrains.kotlin.multiplatform") version "1.3.50"
+-    id("org.jetbrains.dokka") version "0.10.1-SNAPSHOT"
++    id("org.jetbrains.dokka") version "0.11.0-SNAPSHOT"
```
kamilok1965: I think we can just update dokka version in the example repo

kamilok1965: Did that

PR #548, opened 2019-12-19
Resolved problems with plugins loading ordering and added type safe ordering function

BarkingBad:


core/src/main/kotlin/plugability/DokkaContext.kt
```diff
@@ -68,6 +69,49 @@ private class DokkaContextConfigurationImpl(
...
+        VISITING,
+        VISITED;
+    }
+
+    internal val verticesWithState = mutableMapOf<Extension<*>, State>()
+
+    internal val adjacencyList: MutableMap<Extension<*>, MutableList<Extension<*>>> = mutableMapOf()
+
+    private fun topologicalSort() {
+
```
sellophane: Why so many whitespaces?

core/src/main/kotlin/plugability/DokkaContext.kt
```diff
@@ -68,6 +69,49 @@ private class DokkaContextConfigurationImpl(
...
+            val state = verticesWithState[n]
+
+            if(state == State.VISITED)
+                return
+            if(state == State.VISITING)
+                throw Error("Detected cycle in plugins graph")
+
+            verticesWithState[n] = State.VISITING
+
+            for(m in adjacencyList[n]!!)
```
sellophane: adjacencyList[n]?.forEach { visit(it) }

core/src/main/kotlin/plugability/DokkaContext.kt
BarkingBad: Hard to tell, was more readable while writing

core/src/main/kotlin/plugability/DokkaContext.kt
BarkingBad: Will change

core/src/main/kotlin/plugability/DokkaContext.kt
BarkingBad: Will change

PR #557, opened 2020-01-08
testing utils and examples

sswistun-vl:
Tools for Dokka testing.


core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    fun generateTestDokka(
+        name: String,
+        query: String,
+        checks: Map<DokkaTestGenerator.Companion.Stage, (DokkaTestGenerator) -> Boolean>,
+        targets: List<String> = listOf("jvm"),
+        logger: DokkaLogger = DokkaConsoleLogger
+    ) {
+        val root = Paths.get(".").resolve("build").resolve("tests").resolve(name)
+        root.deleteRec()
```
kamilok1965: Do not hardcode paths, we should probably use temporary directories for this

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,93 @@
...
+        val config = DokkaConfigurationImpl(
+            outputDir = output.toString(), format = "html", generateIndexPages = true, passesConfigurations = listOf(
+                PassConfigurationImpl(
+                    moduleName = name,
+                    sourceRoots = listOf(root).map { SourceRootImpl(it.toString()) },
+                    analysisPlatform = Platform.jvm,
+                    targets = targets
+                )
+            )
+        )
```
kamilok1965: This prevents us from testing with different configurations as the only way to change the config is to create another method

core/src/test/kotlin/generator/DokkaTestGenerator.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.jetbrains.dokka.renderers.FileWriter
+import org.jetbrains.dokka.utilities.DokkaLogger
+
+class DokkaTestGenerator(
+    private val configuration: DokkaConfiguration,
+    private val logger: DokkaLogger,
+    private val checks: Map<Stage, (DokkaTestGenerator) -> Boolean> = mapOf()
+) : DokkaGenerator(configuration, logger) {
+    companion object {
+        enum class Stage {
```
kamilok1965: We can just do it using types

core/src/main/kotlin/DokkaBootstrapImpl.kt
```diff
@@ -19,10 +19,30 @@ fun parsePerPackageOptions(arg: String): List<PackageOptions> {
...
+    if (!config.noStdlibLink)
+        links += DokkaConfiguration.ExternalDocumentationLink
+            .Builder("https://kotlinlang.org/api/latest/jvm/stdlib/")
+            .build() as ExternalDocumentationLinkImpl
+    return links
+}
+
 class DokkaBootstrapImpl : DokkaBootstrap {
 
     private class DokkaProxyLogger(val consumer: BiConsumer<String, String>) : DokkaLogger {
```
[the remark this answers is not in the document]
sswistun-vl: Done.

integration/src/main/kotlin/org/jetbrains/dokka/defaultConfiguration.kt
```diff
@@ -5,38 +5,38 @@ import java.net.URL
...
-    override val cacheRoot: String?,
-    override val impliedPlatforms: List<String>,
-    override val passesConfigurations: List<PassConfigurationImpl>,
-    override var pluginsClasspath: List<File>
+    override val format: String = "html",
+    override val generateIndexPages: Boolean = true,
+    override val cacheRoot: String? = null,
+    override val impliedPlatforms: List<String> = listOf(),
+    override val passesConfigurations: List<PassConfigurationImpl> = listOf(),
+    override var pluginsClasspath: List<File> = listOf()
```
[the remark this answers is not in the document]
sswistun-vl: Done.

core/src/test/kotlin/generator/DokkaTestGenerator.kt
```diff
@@ -0,0 +1,88 @@
...
+            SETUP, PLUGINS, MODELS_CREATE, MODELS_MERGE, MODELS_TRANSFORM, PAGES_CREATE, PAGES_TRANSFORM, RENDER
```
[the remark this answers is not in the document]
sswistun-vl: Done.

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,93 @@
...
+
+                    @Throws(IOException::class)
+                    override fun visitFile(
+                        file: Path, attrs: BasicFileAttributes
+                    ): FileVisitResult {
+                        Files.delete(file)
+                        return FileVisitResult.CONTINUE
+                    }
+                })
+    }
```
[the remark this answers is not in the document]
sswistun-vl: Done.

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,93 @@
...
+                    if (path.isEmpty() || content.isEmpty()) throw IllegalArgumentException(
+                        "Unable to split argument info path/contents! \n"
+                    )
+                    else path.removePrefix("/") to content
+                }
+                .toMap()
+        } else {
+            mapOf()
+        }
+    }
```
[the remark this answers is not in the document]
sswistun-vl: Done.

core/src/test/kotlin/generator/TestRunner.kt
sswistun-vl: Done.

core/src/test/kotlin/generator/TestRunner.kt
sswistun-vl: Done.

core/src/test/kotlin/generator/DokkaTestGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    data class State(var platforms: Map<PlatformData, EnvironmentAndFacade> = mapOf(),
+                     var context: DokkaContext? = null,
+                     var modulesFromPlatforms: List<Module> = listOf(),
+                     var documentationModel: Module? = null,
+                     var transformedDocumentation: Module? = null,
+                     var pages: ModulePageNode? = null,
+                     var transformedPages: ModulePageNode? = null)
```
kamilok1965: Still nullable vars that can be avoided easily by changing `enum class Stage` to specific classes and passing the result explicitly to each stage

core/src/test/kotlin/generator/DokkaTestGenerator.kt
```diff
@@ -0,0 +1,101 @@
...
+            .invoke(fileWriter, locationProvider, context)
+
+        renderer.render(transformedPages)
+
+        checkStage(Companion.Stage.RENDERING, state)
+    }
+
+    private fun checkStage(stage: Stage, state: State) {
+        checks[stage]?.invoke(state)
+    }
```
kamilok1965: Same here, there's no need to use `invoke` this way

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,101 @@
...
+                    generateIndexPages = true,
+                    passesConfigurations = listOf(
+                        PassConfigurationImpl(
+                            moduleName = name,
+                            sourceRoots = listOf(root).map { SourceRootImpl(it.toString()) },
+                            analysisPlatform = it,
+                            targets = targets
+                        )
+                    )
+                )
```
kamilok1965: Still no way to change config other than writing a new `generateTestDokka` method

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.jetbrains.dokka.utilities.DokkaLogger
+import java.io.File
+import java.nio.charset.Charset
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+
+
+object TestRunner {
+    private var testDir: Path? = null
```
kamilok1965: Looks like we don't need this at all, since it's only used in one method locally

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,101 @@
...
+                    it.split("\n".toRegex(), 2)
+                        .let{
+                            it.first().trim().removePrefix("/") to it.last().trim()
+                        }
+                }
+                .toMap()
+        } else {
+            mapOf()
+        }
+    }
```
kamilok1965: I'm almost sure you can still minify that by using `split("\n").chunked(2) { (k, v) -> k.trim() to v.trim() }.toMap()`

kamilok1965: Futhermore, you can always replace 
```
fun x() {
    return y
} 
```
with 
``` 
fun x() = y 
```
here and in any other place

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    fun String.toPlatform(): Platform? {
+        return try {
+            Platform.fromString(this)
+        }
+        catch (e: IllegalArgumentException) {
+            null
+        }
+    }
```
kamilok1965: IMHO we don't need this at all. If a platform cannot be parsed we want the test to fail because if we silently ignore it, we'll have a hard time debugging

core/src/test/kotlin/generator/test/GeneratorTest.kt
```diff
@@ -0,0 +1,38 @@
...
+            DokkaTestGenerator.Companion.Stage.DOCUMENTABLES_CREATION to { state: DokkaTestGenerator.State ->
+                state.context?.platforms?.forEach { (pd, _) ->
+                    logger.info(pd.name)
+                }
+                true
+            },
+            DokkaTestGenerator.Companion.Stage.RENDERING to { state: DokkaTestGenerator.State ->
+                val test = state.pages?.parentMap?.size == 7
+                assert(test)
+                test
```
kamilok1965: I think the assert should be lifted from here as we're returning the test result up

core/src/test/kotlin/generator/TestRunner.kt
kamilok1965: Basically you're just grouping the paths here in sets of two

sswistun-vl: String.split(r: Regex, n: Int) splits it to n parts, last one may contain r occurrences. In this case it splits it to first line -> rest.

core/src/test/kotlin/generator/test/GeneratorTest.kt
sswistun-vl: I've changed return type to Unit? because in my opinion it's better to keep asserts inside tests.

core/src/test/kotlin/generator/TestRunner.kt
sswistun-vl: I've removed try catch.

core/src/test/kotlin/generator/TestRunner.kt
sswistun-vl: I've changed it to lazy init - main tests directory is common for all tests to make it easier to clean up after them.

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,89 @@
...
+                    .let {
+                        it.first().trim().removePrefix("/") to it.last().trim()
+                    }
+            }
+            .toMap()
+    } else {
+        mapOf()
+    }
+
+    fun String.toPlatform(): Platform = Platform.fromString(this)
```
kamilok1965: Now it's pointless, as you can just use `Platform.fromString()` instead

core/src/test/kotlin/generator/TestRunner.kt
kamilok1965: `split("\n", limit = 2).let { mapOf(it.first().trim to it.last().trim()) }`

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,89 @@
...
+            .map {
+                it.split("\n".toRegex(), 2)
+                    .let {
+                        it.first().trim().removePrefix("/") to it.last().trim()
+                    }
+            }
+            .toMap()
+    } else {
+        mapOf()
+    }
```
kamilok1965: `split("\n", limit = 2).let { mapOf(it.first() to it.last().trim()) }`

core/src/test/kotlin/generator/test/GeneratorTest.kt
```diff
@@ -0,0 +1,52 @@
...
+            passesConfigurations = listOf(
+                PassConfigurationImpl(
+                    moduleName = name,
+                    sourceRoots = listOf(root).map { SourceRootImpl(it.toString()) },
+                    analysisPlatform = Platform.jvm,
+                    targets = listOf("jvm")
+                )
+            )
+        )
+    }
```
kamilok1965: I don't like passing lambdas but I don't have a better way to do it right now, let's discuss it tomorrow

kamilok1965: Let's leave it like this but please extract those lambda params to some data class and pass that instead

sswistun-vl: Done.

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.nio.charset.Charset
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+
+
+object TestRunner {
+    private val testDir: Path by lazy {
+        Files.createTempDirectory("tests").also { it.toFile().deleteOnExit() }
+    }
```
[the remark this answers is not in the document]
sswistun-vl: It isn't anymore.

core/src/test/kotlin/generator/DokkaTestGenerator.kt
kamilok1965: Bump

core/src/test/kotlin/generator/TestRunner.kt
```diff
@@ -0,0 +1,67 @@
...
+            .let { it to it.resolve("out").also { it.createDirectories() } }
+
+    fun generateTestDokka(
+        name: String,
+        query: String,
+        checks: Map<DokkaTestGenerator.Companion.Stage, (DokkaTestGenerator.State) -> Unit?>,
+        config: (String, Path, Path) -> DokkaConfiguration,
+        logger: DokkaLogger = DokkaConsoleLogger
+    ) {
+        val (root, out) = getTestDirs(name)
```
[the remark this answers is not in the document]
kamilok1965: Or we can extract the test name automatically

core/src/test/kotlin/generator/test/GeneratorTest.kt
```diff
@@ -0,0 +1,53 @@
...
+                    }
+                },
+                DokkaTestGenerator.Companion.Stage.RENDERING to { state: DokkaTestGenerator.State ->
+                    val test = state.pages?.parentMap?.size == 7
+                    if (!test) println(state.pages?.parentMap?.size)
+                    assert(test)
+                }
+            )
+
+            TestRunner.generateTestDokka(name = "test1", query = """
```
kamilok1965: `Name` should be extracted automatically

core/src/test/kotlin/generator/TestRunner.kt
sswistun-vl: Removed.

core/src/test/kotlin/generator/TestRunner.kt
sswistun-vl: Removed.

core/src/test/kotlin/generator/DokkaTestGenerator.kt
sswistun-vl: Current version uses lambdas passed as parameters to testing function.

core/src/test/kotlin/generator/test/GeneratorTest.kt
sswistun-vl: Done.

PR #564, opened 2020-01-13
Resolved problems with DRI generics

BarkingBad:
Known front-end bugs
- If generic type is nullable it should be noted on front-end (for now it holds unused property isNullable)
- If name of function/property/class etc. is in backticks with some prohibited characters it causes front-end links to work improperly, e. g. `any? inheritance` (actually it's not related to DRI changes)


core/src/main/kotlin/links/DRI.kt
```diff
@@ -68,44 +68,102 @@ data class Callable(
...
+//        private fun from(t: TypeProjection): TypeReference =
+//            if (t.isStarProjection) {
+//                starProjection
+//            } else {
+//                from(t.type)
+//            }
+//
+//        val starProjection = TypeReference("*")
+//    }
+//}
```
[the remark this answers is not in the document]
BarkingBad: Will do

Kordyjan:
`fun <T: List<Int>> T.foo() {}` is generating following callable part of DRI (assuming package is called `example`): `Callable(name=foo, receiver=example.foo.T, params=[])`

It has two problems:
- `example.foo.T` is not correct fully qualifiad name and it results in situation where different DRIs can be assigned to two methods that are not distinctible from kotlin's perspective.
- there is no information about receiver being bound to `List<Int>`.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -68,44 +68,102 @@ data class Callable(
...
-                from(t.type)
+                from(t.type, r)
             }
 
-        val starProjection = TypeReference("*")
+        private val starProjection = Param(listOf(Bound("*", emptyList(), true)))
+    }
+}
+data class Param(val bounds: List<TypeReference>) : TypeReference()
+data class Bound(val fullyQualifiedName: String, val params: List<TypeReference>, val isNullable: Boolean) : TypeReference() {
```
kamilok1965: `Bound` -> `ConcreteType` ? or `InstantiatedType`

core/src/main/kotlin/links/DRI.kt
```diff
@@ -68,44 +68,102 @@ data class Callable(
...
-    }
+sealed class TypeReference {
 
     companion object {
         fun from(d: ReceiverParameterDescriptor): TypeReference? =
             when (val value = d.value) {
-                is ExtensionReceiver -> TypeReference(
-                    classNames = value.type.constructorName.orEmpty(),
-                    typeBounds = value.type.arguments.map { from(it) }
+                is ExtensionReceiver -> Bound(
```
kamilok1965: A `ReceiverParameter` is not always a `Bound` I think

PR #565, opened 2020-01-13
Rewriting HtmlRenderer to use kotlinx.html

sellophane:


core/src/main/kotlin/renderers/DefaultRenderer.kt
```diff
@@ -4,62 +4,67 @@ import org.jetbrains.dokka.pages.*
 import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.resolvers.LocationProvider
 
-abstract class DefaultRenderer(
+abstract class DefaultRenderer<T>(
```
kamilok1965: Can you explain why do you make it generic? It looks like we only have `FlowContent` right now and if we wanted to create another type of content we could probably just subtype it

Issue #580, opened 2020-01-15
Fix repeating comments for multiplatform Documentables


Kordyjan:
For some reason comments for functions and classes defined on multiple platforms (for example in common source set) are repeated and rendered without any spaces between them.

Example:
```kt
/**
 * foo comment
 *
 * @receiver receiver comment
 * @param bar bar comment
 */
fun <T> T.foo(bar:T) {}
```
creates something that looks like: 
![this](https://user-images.githubusercontent.com/11855167/72447700-1559be80-37b6-11ea-87cc-46c2afd5eb4e.png)
but should be:
![obraz](https://user-images.githubusercontent.com/11855167/72448024-a3ce4000-37b6-11ea-800f-0b2b37927ae9.png)

PR #585, opened 2020-01-16
Resolved problems with doubling comments

BarkingBad:


core/src/main/kotlin/model/Documentable.kt
```diff
@@ -113,11 +113,6 @@ abstract class Documentable {
 
     override fun hashCode() = dri.hashCode()
 
-
-
-    val commentsData: List<DocumentationNode>
-        get() = platformInfo.map { it.documentationNode }
```
BarkingBad: We don't need this anymore as by referencing documentation nodes with this shortcut we lose information about specific platform data

kamilok1965:
Fixes #580

core/src/main/kotlin/transformers/descriptors/DefaultDescriptorToDocumentationTranslator.kt
```diff
@@ -27,119 +30,170 @@ object DefaultDescriptorToDocumentationTranslator: DescriptorToDocumentationTran
...
 }
 
+
+data class DRIWithPlatformInfo(
+    val dri: DRI,
+    val expected: PlatformInfo?,
+    val actual: List<PlatformInfo>
+)
+
 class DokkaDescriptorVisitor(
```
Kordyjan: If this two classes can be made private or internal - do so. `DRIWithPlatformInfo` may be confusing and i don't think we want to make them part of the api.

PR #596, opened 2020-01-27
Update kotlin compiler version to 1.3.61

### LINKED ISSUE #531: Update bundled compiler version to 1.3.61
To support klib it's necessary to update dokka bundled compiler version to 1.3.60

kamilok1965:
Fixes #531


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
```
martinbonnin: Is that intentional ?

.idea/compiler.xml
```diff
@@ -11,14 +11,6 @@
...
       <entry name="!?*.clj" />
     </wildcardResourcePatterns>
-    <annotationProcessing>
-      <profile name="Maven default annotation processors profile" enabled="true">
-        <sourceOutputDir name="target/generated-sources/annotations" />
-        <sourceTestOutputDir name="target/generated-test-sources/test-annotations" />
-        <outputRelativeToContentRoot value="true" />
-        <module name="dokka-maven-plugin" />
-      </profile>
-    </annotationProcessing>
```
martinbonnin: Should this file be removed from source control ?

kamilok1965: Probably, but the decision was to keep the entire `.idea` directory to help setup this project, so we sometimes just update it

.idea/codeStyles/codeStyleConfig.xml
kamilok1965: Not sure TBH. Intellij changed it, doesn't look harmful

.idea/compiler.xml
martinbonnin: I usually only commit the codestyle settings (`.idea/codeStyles/`). The problem of adding other files is that it creates noise in the diffs.

PR #597, opened 2020-01-27
Add a CONTRIBUTING.md

martinbonnin:
Move the **Building** section out of the README.md to CONTRIBUTING.md.

Also add a section about the different branches. I'm pretty new to the project so please correct if I'm wrong.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,40 @@
+## Branches
+
+Dokka follows [semantic versioning](https://semver.org/).
+
+Development happens in several branches:
+
+* master is the latest released version (0.x.y) 
+* dev-0.x.{y + 1} is the development branch for the next patch version
+* dev-0.{x + 1}.0 is the development branch for the next minor version
+
+If you're doing a pull request for a bug fix, use the patch branch. For new features, 
+use the minor branch.  
+
+## Building dokka
+
+Dokka is built with Gradle. To build it, use `./gradlew build`.
+Alternatively, open the project directory in IntelliJ IDEA and use the IDE to build and run dokka.
+
+Here's how to import and configure Dokka in IntelliJ IDEA:
+ * Select "Open" from the IDEA welcome screen, or File > Open if a project is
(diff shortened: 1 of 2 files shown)


Kordyjan:
Thanks for this contribution. It is fact that we need some basic guides for contributors.
Nevertheless I think it is worth to mention that Semantic Versioning does not specify meaning of second and third component of version string if the first component is "0". We will definitely use second component as minor version number after dokka 1.0.0. 
Right now  it is not the case because 0.11.0 is complete reimplementation of dokka and shares very actual usefull code with earlier versions. It even possible that in the future we decide to release what is now developed as 0.11.0 as dokka 1.0.0.

martinbonnin:
Got it, thanks for the explanation ! I'll try to rephrase a bit. As an external contributor, what I really want to know is what branch to base my pull requests on. 
Is it a good idea to advise external developers to work in `dev-0.11.0` or is it moving too fast/breaking too often to onboard new devs ?

Kordyjan:
So the situation basically looks like this: 
After release of 0.10.1 (which we plan to do soon) we will focus on 0.11.0 branch. 0.10.X will be updated only if tbreaking bugs appear. We definitely not plan to add any new features to this branch.
This means that all new development should be based on 0.11.0. As it is nearly complete reimplementation, it is still far from being feature complete and breaking changes still happens very often. So indeed first few contributions may be really challenging for new devs.
The upside is that we plan to stabilise core model in week or two so contributions may soon become easier.

martinbonnin:
Nice !

I updated the CONTRIBUTING.md a bit. Again, feel free to close and/or edit this PR, you know best. 

For my other PR (https://github.com/Kotlin/dokka/pull/598) which adds regex support to perPackageOptions, I guess this should go in 0.11.0 so I'll rebase it in a few weeks once things have settled a bit if it's ok for you ?

kamilok1965:
Sure, that's a good idea, however it'll probably require rewriting it completely as package filtering should probably be a plugin

Issue #593, opened 2020-01-27
Allow to apply perPackageOption based on a regex, not only prefix


martinbonnin:
Our internal packages are scattered in various places in the package hierarchy so I'd like to be able to do something like this:

```kotlin
    perPackageOption {
        matchingRegex = ".*\\.internal"
        suppress = true
    }
```

PR #598, opened 2020-01-27
Add matchingRegex

martinbonnin:
Closes https://github.com/Kotlin/dokka/issues/593

This allows to match a package based on a regex and not only a prefix.

```
            perPackageOption {
                matchingRegex = ".*\\.internal"
                suppress = true
            }
```

### DIFF
--- a/core/src/main/kotlin/DokkaBootstrapImpl.kt
+++ b/core/src/main/kotlin/DokkaBootstrapImpl.kt
@@ -8,10 +8,6 @@ import java.util.function.BiConsumer
 
 fun parsePerPackageOptions(args: List<String>): List<PackageOptions> = args.map { it.split(",") }.map {
     val matchingRegex = it.first()
-    if (matchingRegex == ".*")
-        throw IllegalArgumentException(
-            "Please do not register packageOptions with all match pattern, use global settings instead"
-        )
 
     val args = it.subList(1, it.size)
 
--- a/core/src/main/kotlin/DokkaBootstrapImpl.kt
+++ b/core/src/main/kotlin/DokkaBootstrapImpl.kt
@@ -7,8 +7,8 @@ import java.util.function.BiConsumer
 
 
 fun parsePerPackageOptions(args: List<String>): List<PackageOptions> = args.map { it.split(",") }.map {
-    val prefix = it.first()
-    if (prefix == "")
+    val matchingRegex = it.first()
+    if (matchingRegex == ".*")
         throw IllegalArgumentException(
             "Please do not register packageOptions with all match pattern, use global settings instead"
         )
@@ -28,7 +28,7 @@ fun parsePerPackageOptions(args: List<String>): List<PackageOptions> = args.map 
         ?: DokkaDefaults.suppress
 
     PackageOptionsImpl(
-        prefix,
+        matchingRegex,
         includeNonPublic = privateApi,
         reportUndocumented = reportUndocumented,
         skipDeprecated = !deprecated,
--- a/core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
+++ b/core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
@@ -28,8 +28,8 @@ fun PreMergeDocumentableTransformer.sourceSet(documentable: Documentable): Dokka
 fun PreMergeDocumentableTransformer.perPackageOptions(documentable: Documentable): PackageOptions? {
     val packageName = documentable.dri.packageName ?: return null
     return sourceSet(documentable).perPackageOptions
-        .sortedByDescending { packageOptions -> packageOptions.prefix.length }
-        .firstOrNull { packageOptions -> packageName.startsWith(packageOptions.prefix) }
+        .sortedByDescending { packageOptions -> packageOptions.matchingRegex.length }
+        .firstOrNull { packageOptions -> Regex(packageOptions.matchingRegex).matches(packageName) }
 }
 
 fun <T> PreMergeDocumentableTransformer.source(documentable: T) where T : Documentable, T : WithSources =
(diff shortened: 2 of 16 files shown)


core/src/main/kotlin/DokkaBootstrapImpl.kt
```diff
@@ -18,7 +18,13 @@ fun parsePerPackageOptions(arg: String): List<PackageOptions> {
         val reportUndocumented = args.find { it.endsWith("warnUndocumented") }?.startsWith("+") ?: true
         val privateApi = args.find { it.endsWith("privateApi") }?.startsWith("+") ?: false
         val suppress = args.find { it.endsWith("suppress") }?.startsWith("+") ?: false
-        PackageOptionsImpl(prefix, includeNonPublic = privateApi, reportUndocumented = reportUndocumented, skipDeprecated = !deprecated, suppress = suppress)
+        PackageOptionsImpl(
+            prefix = prefix,
+            matchingRegex = null,
```
kamilok1965: IMHO regex could be a non-nullable String as an empty regex matches only empty Strings

martinbonnin: I was thinking about making prefix nullable actually. That would allow to validate the user parameters. Something like:

```kotlin
    require(prefix != null || matchingRegex != null) {
         "perPackageOption does not specify prefix neither matchingRegex, it will never be used"
    }
```

martinbonnin: Damn, I just missed 0.10.1 :-/. Should I rebase this PR somewhere else and make `regex` non-nullable ?

MarcinAman:
Hi, since this PR was created a lot of dokka has changed. Would you mind rebasing it onto master?

martinbonnin:
> Hi, since this PR was created a lot of dokka has changed. Would you mind rebasing it onto master?

Yup, give me a bit of time to switch back to this and I'll rebase.

MarcinAman:
Right now dokka has a prefix option in `PackageOptions`. I think it would be worth to change that to your regex-based approach to make it more powerfull, i dont feel like we need 2 options to configure more or less same thing

martinbonnin:
@MarcinAman PR is rebased. Removing `prefix` is technically a breaking change, let me know if that can be shipped at this stage or if you want me to add it again with a deprecation warning (which will complexify both the documentation and the handling of default values but hey, it happens!)

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
```diff
@@ -28,8 +28,8 @@ fun PreMergeDocumentableTransformer.sourceSet(documentable: Documentable): Dokka
 fun PreMergeDocumentableTransformer.perPackageOptions(documentable: Documentable): PackageOptions? {
     val packageName = documentable.dri.packageName ?: return null
     return sourceSet(documentable).perPackageOptions
-        .sortedByDescending { packageOptions -> packageOptions.prefix.length }
-        .firstOrNull { packageOptions -> packageName.startsWith(packageOptions.prefix) }
+        .sortedByDescending { packageOptions -> packageOptions.matchingRegex.length }
```
martinbonnin, line 31: This feels a bit awkward with regexes. I think I would prefer the order of the rules in the Gradle file to determine priority but I'll let people more familiar with the subject decide.

MarcinAman:
> @MarcinAman PR is rebased. Removing `prefix` is technically a breaking change, let me know if that can be shipped at this stage or if you want me to add it again with a deprecation warning (which will complexify both the documentation and the handling of default values but hey, it happens!)

It can be shipped, as for now Dokka doesn't have any guarantees about stability.

Thanks, i'll check out the changes tomorrow 👍

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
MarcinAman: @kamildoleglo ?

core/src/main/kotlin/DokkaBootstrapImpl.kt
```diff
@@ -7,8 +7,8 @@ import java.util.function.BiConsumer
 
 
 fun parsePerPackageOptions(args: List<String>): List<PackageOptions> = args.map { it.split(",") }.map {
```
kamildoleglo, line 9: I think this check is now unnecessary as we dropped global settings and I see no reason why the regex could not be an `all match`

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
kamildoleglo: I think ordering in Gradle might be more problematic and potentially could break easier. Also it will cause some confusion especially when having split logic for dokka configuration (eg. some packageOptions in buildSrc and some in Gradle build files), so I'd personally stick with the regex for now

kamildoleglo (changes requested):
Just one request to remove the unnecessary (I believe) check and it should be good to go

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
martinbonnin: Fair point about split logic, especially since it's hard to know in advance the order of execution of Gradle scripts. I added a [small note](https://github.com/Kotlin/dokka/pull/598/commits/f97135b49ae5008c05b33f690d81e2d565f43c42) to the Gradle documentation to tell users that the longuest matchingRegexp will be used. 

For the `cli` usage, maybe the order of arguments would still make sense? Or ultimately, maybe an explicit `priority` option will be needed. But this can be all be added later depending the needs and use cases

kamildoleglo (approved):

MarcinAman:
Thanks ☺️

MarcinAman:
You can check out the build with your changes by adding a new repository: `maven("https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev")` and updating version to: `1.4.20-dev-23`

PR #604, opened 2020-01-31
Unifying structure for content pages and technical rednerer specific pages

Kordyjan:
This PR introduces concept of renderer specific page, that can be used and transformed as any other page but does not carry any content inside.
I also included here parts of @sellophane work on new html format, applying new page model to them.

It is worth mentioning that html format is still work in progress and it will be updated in later prs.


core/src/main/kotlin/renderers/FileWriter.kt
```diff
@@ -15,19 +16,46 @@ class FileWriter(val root: String, override val extension: String): OutputWriter
...
+
     private fun File.mkdirsOrFail() {
         if (!mkdirs() && !exists()) {
             throw IOException("Failed to create directory $this")
         }
     }
+
+    private fun getFileSystemForURI(uri: URI): FileSystem =
+        try {
+            FileSystems.newFileSystem(uri, emptyMap<String, Any>())
```
kamilok1965: This will fail for regular directories when running tests, see 280859d8c9bdd6d5b5417a97edf1b1378bd9a9c7

core/src/main/kotlin/renderers/HtmlRenderer.kt
```diff
@@ -1,26 +1,52 @@
...
 
-    override fun FlowContent.buildList(node: ContentList, pageContext: PageNode) =
+    private val pageList = mutableListOf<String>()
+
+    private var idCounter = 0
+        get() = ++field
+
+    private fun FlowContent.buildSideMenu(context: PageNode, node: PageNode) {
+        val children = node.children.filter { it !is MemberPageNode }
+        val className = children.ifNotEmpty { "nav$idCounter" }
```
kamilok1965: Not sure about `ifNotEmpty` as it is in `addToStdlib` package which proved to be unstable in the past. I don't have a strong opinion though

core/src/main/kotlin/renderers/FileWriter.kt
Kordyjan: I've forgotten about this problem. Now the fix is also on this branch.

core/src/main/kotlin/renderers/HtmlRenderer.kt
Kordyjan: I've removed this function call. Also I've refactored this piece of code slightly.

Issue #583, opened 2020-01-16
Clean up the way file extension is handled


Kordyjan:
Right now file extensions lives both as extension point and as a property of `OutputWriter`. It is also passed as a parameter to `write` function of `OutputWriter`. To complete this task make sure that default file extension fro given format is handled by extension point or propose alternative way to get rid of this redundancy.

PR #615, opened 2020-02-05
Cleans up file extension redundancy

BarkingBad:
#583

### DIFF
--- a/core/src/main/kotlin/renderers/OutputWriter.kt
+++ b/core/src/main/kotlin/renderers/OutputWriter.kt
@@ -1,7 +1,15 @@
 package org.jetbrains.dokka.renderers
 
-interface OutputWriter{
+import org.jetbrains.dokka.CoreExtensions
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.plugability.single
+
+interface OutputWriter {
+
+    val context: DokkaContext
     val extension: String
+        get() = context.single(CoreExtensions.fileExtension)
+    
     fun write(path: String, text: String, ext: String = extension)
     fun writeResources(pathFrom: String, pathTo: String)
 }
--- a/core/src/main/kotlin/renderers/OutputWriter.kt
+++ b/core/src/main/kotlin/renderers/OutputWriter.kt
@@ -6,10 +6,6 @@ import org.jetbrains.dokka.plugability.single
 
 interface OutputWriter {
 
-    val context: DokkaContext
-    val extension: String
-        get() = context.single(CoreExtensions.fileExtension)
-    
-    fun write(path: String, text: String, ext: String = extension)
+    fun write(path: String, text: String, ext: String)
     fun writeResources(pathFrom: String, pathTo: String)
 }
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/renderers/OutputWriter.kt
```diff
@@ -1,7 +1,15 @@
 package org.jetbrains.dokka.renderers
 
-interface OutputWriter{
+import org.jetbrains.dokka.CoreExtensions
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.plugability.single
+
+interface OutputWriter {
+
+    val context: DokkaContext
```
BarkingBad: I think I could move this property to super interface called something like "Contextual" and make some interfaces which requires DokkaContext inherit from it. In my opinion it would be logical to enforce inclusion DokkaContext as a property of class. The question is whether `HtmlRenderer`, `DocTagToContentConverter` and `DefaultLocationProviderFactory` always need context as they need it in default implementations.

core/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -43,7 +43,7 @@ object NavigationPageInstaller : PageNodeTransformer {
 
     private fun visit(page: ContentPage): NavigationNode = NavigationNode(
         page.name,
-        page.dri,
+        page.dri.first(),
```
kamilok1965: Not sure about this (you don't change pages in this PR)

core/src/main/kotlin/renderers/OutputWriter.kt
kamilok1965: I think this is reasonable. Thoughts @Kordyjan ?

core/src/main/kotlin/renderers/html/htmlPreprocessors.kt
BarkingBad: Some leftover after fixing the build proccess

BarkingBad: Will remove

Kordyjan:
The more I think about this problem the more I'm convinced that making separate extension point for file extension was a mistake. Here we are assuming that there are many correct combinations of file extensions and renderers while in fact usually for each renderer there is exactly one correct default file extension. Even if some exotic renderer can operate for multiple default file extensions it's its author's responsibility to provide extension point or config option for that.
Right now I see two reasonable solutions for this problem:
* to remove `fileExtenison` extension point and let renderer instantiate its `OutputWriter` with arbitrary default extension, similarly how `locationProvider` is instantiated with arbitrary page tree
* to remove entire concept of deafult extension as it brings little to no benefits and just specify extension for every write instruction.

I'm slightly leaning toward second solution.
@kamilok1965 @BarkingBad am I missing something? any thoughts?

kamilok1965:
I'm also for the second solution. Each renderer can just wrap the calls to `FileWriter` in private methods with the correct extension. This should be more maintainable and cleaner

BarkingBad:
I think we can opt for the second solution, however, I think we can just leave default extension as an OutputWriter property, and make it as a default parameter if none is specified in write function. If somebody doesn't want to use the default file extension, he has to be explicit every time calling write function.

PR #616, opened 2020-02-05
Page merger

### LINKED ISSUE #577: Create PageMerger
To complete this task, register extension to pageTransformer extension point that will traverse tree of pages and for each parent it will group its children by name and produce one output page from each group.
For now it should take the first page from group and log warning if there are other nodes.

sswistun-vl:
Added merger for pages, fixes #577


Kordyjan:
Already merged.

PR #619, opened 2020-02-06
Page merger strategy

sswistun-vl:
Implemented strategies for page merger.


core/src/main/kotlin/transformers/psi/DefaultPsiToDocumentationTranslator.kt
```diff
@@ -100,14 +111,15 @@ object DefaultPsiToDocumentationTranslator : PsiToDocumentationTranslator {
...
+                getComment(psi),
+                visibility = mapOf(platformData to psi.getVisibility())
             )
         }
 
         private fun parseField(psi: PsiField, parent: DRI): Property {
             val dri = parent.copy(
                 callable = Callable(
-                    psi.name,
+                    psi.name ?: "",
```
kamilok1965: I think `psi.name` returns `String`, so the elvis is redundant

gradle.properties
```diff
@@ -6,6 +6,7 @@ kotlin_version=1.3.61
 kotlin_plugin_version=1.3.61-release-180
 idea_version=192.5728.98
 language_version=1.3
+org.gradle.jvmargs=-Xmx2g
```
kamilok1965: Do we need this? If so, it should be at the bottom with a new comment

plugins/kotlin-as-java/build.gradle.txt
```diff
@@ -0,0 +1,43 @@
+import javax.tools.ToolProvider
```
kamilok1965: What's the purpose of this file?

plugins/kotlin-as-java/src/main/kotlin/KotlinAsJavaPageContentBuilder.kt
```diff
@@ -0,0 +1,185 @@
...
+
+class KotlinAsJavaPageContentBuilder(
+    private val dri: Set<DRI>,
+    private val platformData: Set<PlatformData>,
+    private val kind: Kind,
+    private val commentsConverter: CommentsToContentConverter,
+    val logger: DokkaLogger,
+    private val styles: Set<Style> = emptySet(),
+    private val extras: Set<Extra> = emptySet()
+) : PageContentBuilder {
```
kamilok1965: Subclassing `DefaultPageContentBuilder` is a good idea there IMHO, as some methods are just duplicates

settings.gradle.kts
```diff
@@ -8,8 +8,8 @@ include("runners:cli")
 include("runners:maven-plugin")
 include("plugins:xml")
 include("plugins:mathjax")
-include("integration-tests:gradle-integration-tests")
-
+include("plugins:kotlin-as-java")
+//include("integration-tests:gradle-integration-tests")
```
kamilok1965: Don't comment it, if you want to skip tests better to uncomment a line in integration-tests' build.gradle

core/src/main/kotlin/transformers/pages/PageNodeTransformer.kt
```diff
@@ -4,6 +4,6 @@ import org.jetbrains.dokka.pages.ModulePageNode
 import org.jetbrains.dokka.pages.RootPageNode
 import org.jetbrains.dokka.plugability.DokkaContext
 
-interface PageNodeTransformer {
-    operator fun invoke(input: RootPageNode): RootPageNode
+abstract class PageNodeTransformer(context: DokkaContext? = null) {
+    abstract operator fun invoke(input: RootPageNode): RootPageNode
 }
```
kamilok1965: This change is probably unnecessary

core/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -43,20 +43,20 @@ object NavigationPageInstaller : PageNodeTransformer {
 
     private fun visit(page: ContentPage): NavigationNode = NavigationNode(
         page.name,
-        page.dri,
+        page.dri.first(),
```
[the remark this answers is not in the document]
Kordyjan: It works as long as there is one-to-many relation between pages and dris. If we have many-to-one or many-to-many, our whole location resolution breaks down, so it right here is not the biggest problem.

core/src/main/kotlin/transformers/pages/PageMergerStrategy.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.dokka.transformers.pages
+
+import org.jetbrains.dokka.pages.PageNode
+
+abstract class PageMergerStrategy {
```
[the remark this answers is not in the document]
Kordyjan: Moreover interfaces don't restrict adding any further supertypes while classes do. It is strictly better to use interface when possible.

core/src/main/kotlin/transformers/pages/PageNodeTransformer.kt
Kordyjan: Yup. Extension points' interfaces(!) shouldn't be aware of context. Only their implementation should. It is done this way in every other EP.

core/src/main/kotlin/transformers/pages/DefaultPageMergerStrategy.kt
```diff
@@ -0,0 +1,30 @@
+package org.jetbrains.dokka.transformers.pages
+
+import org.jetbrains.dokka.pages.*
+
+object DefaultPageMergerStrategy : PageMergerStrategy() {
```
Kordyjan: The idea of this class was that it "merges" pages by just returning the first one and reporting warning if group contains more than one page.

sswistun-vl: Now there are Default and SamePage strategies.

core/src/main/kotlin/pages/PageNodes.kt
```diff
@@ -135,11 +138,12 @@ class MemberPageNode(
     override val embeddedResources: List<String> = listOf()
 ) : ContentPage {
     override fun modified(name: String, children: List<PageNode>): MemberPageNode =
-        modified(name = name, content = this.content, children = children)
+        modified(name = name, content = this.content, children = children) as MemberPageNode
```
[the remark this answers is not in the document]
kamilok1965: Bump

PR #608, opened 2020-02-03
Full signature

sswistun-vl:
Additional informations in function signatures. Fixes #567

PR #559, opened 2020-01-09
Enums

sswistun-vl:
Support for enum classes

PR #588, opened 2020-01-20
kotlin-as-java plugin

sswistun-vl:
Plugin for generating documentation that reflects Kotlin code from Java viewpoint. Fix #563, fix #574

PR #622, opened 2020-02-10
Mass merge (prs 559, 588, 608, 616, 619)

### LINKED ISSUE #563: Kotlin-as-java output format in 0.11.0

### LINKED ISSUE #567: Render full signatures (eg. with generic type parameters)

### LINKED ISSUE #574: Information about visibility modifiers and accessors needs to be stored in Documentable Model.

### LINKED ISSUE #575: Page nodes needs to hold set of DRIs, not a single DRI.
Right now PageNodes are holding single DRI. This is incorrect as single page can hold informations from multiple different Documentables.
To complete this task, change single DRI stored in PageNode to set of DRIs and fix all places in code that cease to compile.

### LINKED ISSUE #577: Create PageMerger

### LINKED ISSUE #578: Add pageMergerStrategy extension point
Requires #577

To complete this task:
- add extension point named `pageMergerStrategy` to core extensions. Implementations of this extension should have one method `tryMerge` that takes list of pages and returns list of pages.
- change behavior of PageMerger: it should now have list of merger strategies from context and for each group it merges it should apply them one by one, feeding output of one strategy to the input of next, stopping when some strategy returns list containing only one element
- move current behavior of PageMerger (choosing first page from group and reporting warnings if there are more) to newly created default strategy.

### LINKED ISSUE #579: Implement PageMergerStrategy for merging methods with the same name but different signature
Requires #578

To complete this task:
- create extension to pageMergerStrategy that takes all pages for functions with the same name from the group and returns them as single page
- assure that newly created extension is necessarily executed BEFORE default extension for this ep.

As a reference https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/all.html can be used. Note that all functions are on the same page, but they are grouped in such a way that no documentation is lost nor repeated.

Kordyjan:
Contains https://github.com/Kotlin/dokka/pull/559, https://github.com/Kotlin/dokka/pull/588, https://github.com/Kotlin/dokka/pull/608, https://github.com/Kotlin/dokka/pull/616, https://github.com/Kotlin/dokka/pull/619

fix #574, fix #563, fix #608, fix #567, fix #559, fix #575, fix #577, fix #578, fix #579


Kordyjan:
Turned out it is not very smart idea.

PR #623, opened 2020-02-10
Extended tests

### LINKED ISSUE #601: Extend `testApi` to handle end-to-end tests with comparing output result
`testApi` allows us to generate docs from both files and inline sources and write unit tests for each stage. To complete this task add a possibility to compare the output with supplied result ("golden tests")

sswistun-vl:
Added test point after rendering. Fixes #601


testApi/src/main/kotlin/testApi/testRunner/TestRunner.kt
```diff
@@ -95,6 +95,7 @@ abstract class AbstractCoreTest {
         var documentablesTransformationStage: (Module) -> Unit = {}
         var pagesGenerationStage: (ModulePageNode) -> Unit = {}
         var pagesTransformationStage: (RootPageNode) -> Unit = {}
+        var renderingStage: (RootPageNode, DokkaContext) -> Unit = { a, b -> }
```
kamilok1965: Lambda params are unused here, they can be renamed to: `{ _, _ -> }`

PR #625, opened 2020-02-10
Implements tests for MarkdownParser and fixes some bugs related to its

BarkingBad:


core/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -77,34 +79,63 @@ class MarkdownParser (
...
-            } else {
-                    resolveKDocLink(
-                        resolutionFacade.resolveSession.bindingContext,
-                        resolutionFacade,
-                        declarationDescriptor,
-                        null,
-                        link.split('.')
-                    ).also { if (it.size > 1) DokkaConsoleLogger.warn("Markdown link resolved more than one element: $it") }.firstOrNull()//.single()
-                        ?.let { DRI.from(it) }
+        private fun resolveDRI(link: String): DRI? = if (link.startsWith("http") || link.startsWith("www")) {
```
kamilok1965: This should be done in another method I think, as the list of URI prefixes can be longer (`file`, `ftp`, `mailto` and so on). Moreover I think we could recognise it somehow (maybe by a colon after the first word or by creating a `java.net.URL` and catching `MalformedURLException`)

core/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -171,20 +207,84 @@ class MarkdownParser (
                 MarkdownElementTypes.IMAGE                  -> imagesHandler(node)
                 MarkdownTokenTypes.CODE_FENCE_CONTENT,
                 MarkdownTokenTypes.CODE_LINE,
-                MarkdownTokenTypes.TEXT                     -> DocNodesFromIElementFactory.getInstance(MarkdownTokenTypes.TEXT, body = text.substring(node.startOffset, node.endOffset))
+                MarkdownTokenTypes.TEXT                     -> DocNodesFromIElementFactory.getInstance(
+                    MarkdownTokenTypes.TEXT,
+                    body = text
+                        .substring(node.startOffset, node.endOffset).transform()
```
kamilok1965: Could you please add a comment in code explaining why this is doing what this is doing?

core/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -171,20 +207,84 @@ class MarkdownParser (
...
+                    }
+                }
+                index++
+            }
+            return children
+        }
+
+        private fun String.transform() = this
+            .replace(Regex("\n\n+"), "")
+            .replace(Regex("\n>+ "), "\n")
```
kamilok1965: May Windows line endings be present there? If so they should also be replaced if I understand this correctly

testApi/src/main/kotlin/testApi/testRunner/TestRunner.kt
```diff
@@ -197,6 +200,48 @@ abstract class AbstractCoreTest {
     }
 }
 
+abstract class AbstractKDocTest : AbstractCoreTest() {
```
kamilok1965: IMHO this should not be in the testAPI as those methods are meant for a specific type of KDoc tests. That's why this should be moved to a separate file in core/test/markdown

core/src/main/kotlin/parsers/MarkdownParser.kt
BarkingBad: Second option seems nice

BarkingBad: Done

core/src/main/kotlin/parsers/MarkdownParser.kt
BarkingBad: Tested on CRLF, did fine. We can add in the future source file test with CRLF encoded file to automate and ensure this case

testApi/src/main/kotlin/testApi/testRunner/TestRunner.kt
BarkingBad: Moved

kamilok1965:
One more thing, parser should probably throw something more specific than just `Error`s

BarkingBad:
> One more thing, parser should probably throw something more specific than just `Error`s

Perfectly, it should be handled by error logger, but I wasn't sure can I make parsers dependent from logger/context

Issue #620, opened 2020-02-07
Renderer for Markdown for 0.11


Kordyjan:

kamilok1965:
Please note that tables in markdown are only an optional extension

PR #647, opened 2020-02-14
Markdown renderer plugins to GFM and Jekyll

BarkingBad:
#620


BarkingBad:
There is a major problem with common markdown renderer parts in both Jekyll and Gfm. There are two aproaches available with current API.
- Both plugins are in one module. Then we create one open class, called GfmRenderer which renders Markdown. The JekyllRenderer extends GfmRenderer just to overwrite renderPage (need of custom headers). Then they would be compiled as a one module, but actually would produce two different plugins.
- There are three plugins. One, base plugin is to generate Markdown files with html Location Provider. Then GFM plugin depends on base plugin but extends it with custom Location Provider. The Jekyll plugin extends base plguin by overriding renderPage (need of custom header).

For now we can't make two plugins, one dependent from the other, because neither one strongly includes semantic of the other one

core/src/main/kotlin/pages/ContentNodes.kt
```diff
@@ -101,7 +101,8 @@ data class ContentList(
     override val dci: DCI,
     override val platforms: Set<PlatformData>,
     override val style: Set<Style>,
-    override val extras: Set<Extra>
+    override val extras: Set<Extra>,
+    val start: Int = 1
```
kamilok1965: This only makes sense when the list is ordered. It should either be an `extra` or we should create separate classes for ordered and unordered lists. I'm slightly leaning towards the first solution, @Kordyjan ?

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+import org.jetbrains.dokka.pages.*
+import org.jetbrains.dokka.plugability.DokkaContext
+import java.lang.StringBuilder
+
+
+open class GfmRenderer(context: DokkaContext) : DefaultRenderer<StringBuilder>(context) {
+
+    override fun StringBuilder.buildHeader(level: Int, content: StringBuilder.() -> Unit) {
+        buildParagraph()
+        append("${"#".repeat(level)} ")
```
kamilok1965:
```suggestion
        append("#".repeat(level) + " ")
```
better expresses the intention IMHO

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    private val indent = " ".repeat(4)
+
+    private fun StringBuilder.buildListItem(items: List<ContentNode>, pageContext: ContentPage, bullet: String = "*") {
+        items.forEach {
+            if(it is ContentList) {
+                val builder = StringBuilder()
+                builder.append(indent)
+                builder.buildListLevel(it, pageContext)
+                append(builder.toString().replace(Regex("  \n(?!$)"), "  \n$indent"))
```
kamilok1965:
```suggestion
                append(builder.toString().replace(Regex("\\s{2}\n(?!$)"), "  \n$indent"))
```
looks better IMHO, also I'd add a comment explaining why it is there

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+    private fun StringBuilder.buildListLevel(node: ContentList, pageContext: ContentPage) {
+        if(node.ordered) {
+            buildListItem(node.children, pageContext, "${node.start}.")
+        } else {
+            buildListItem(node.children, pageContext, "*")
+        }
+    }
+
+    override fun StringBuilder.buildNewLine() {
+        this.append("  \n")
```
kamilok1965:
```suggestion
       append("  \n")
```
no need for an explicit `this` I think

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    fun StringBuilder.buildParagraph() {
+        this.append("\n\n")
+    }
```
kamilok1965:
```suggestion
   private fun StringBuilder.buildParagraph() {
       append("\n\n")
    }
```

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    override fun StringBuilder.buildResource(node: ContentEmbeddedResource, pageContext: ContentPage) {
+        this.append("Resource")
```
kamilok1965:
```suggestion
       append("Resource")
```

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    private fun String.escapeIllegalCharacters() = this
+        .replace("<", "\\<")
+        .replace(">", "\\>")
+
+    override fun StringBuilder.buildText(textNode: ContentText) {
+        val decorators = decorators(textNode.style)
+        this.append(decorators)
+        this.append(textNode.text.escapeIllegalCharacters())
+        this.append(decorators.reversed())
```
kamilok1965:
```suggestion
        append(decorators)
        append(textNode.text.escapeIllegalCharacters())
        append(decorators.reversed())
```

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+            else append(node.name)
+        }
+        buildParagraph()
+    }
+
+    override fun buildPage(page: ContentPage, content: (StringBuilder, ContentPage) -> Unit): String {
+        val builder = StringBuilder()
+        content(builder, page)
+        return builder.toString()
+    }
```
kamilok1965:
```suggestion
    override fun buildPage(page: ContentPage, content: (StringBuilder, ContentPage) -> Unit) =
        content(StringBuilder(), page).toString()
```

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    override fun buildError(node: ContentNode) {
+        println("Error")
+    }
```
kamilok1965: This should the `DokkaLogger` instead of `println`

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
```diff
@@ -0,0 +1,185 @@
...
+            when(it) {
+                TextStyle.Bold          -> decorators.append("**")
+                TextStyle.Italic        -> decorators.append("*")
+                TextStyle.Strong        -> decorators.append("**")
+                TextStyle.Strikethrough -> decorators.append("~~")
+                else                    -> Unit
+            }
+        }
+        return decorators.toString()
+    }
```
kamilok1965:
```suggestion
    private fun decorators(styles: Set<Style>) = StringBuilder().let { decorators ->
        styles.forEach {
            when (it) {
                TextStyle.Bold -> decorators.append("**")
                TextStyle.Italic -> decorators.append("*")
                TextStyle.Strong -> decorators.append("**")
                TextStyle.Strikethrough -> decorators.append("~~")
                else -> Unit
            }
        }
    }.toString()
```
Looks better IMHO, also please use the formatter

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: Ok

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: Ok

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: Ok

core/src/main/kotlin/pages/ContentNodes.kt
BarkingBad: I haven't thought about reusing extras. I guess I have mistakenly taken `Extra` to be enum class so it couldn't store any value.

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: It's not a content() we won't call toString() but StringBuilder object.

BarkingBad: In my opinion there is barely anything we can do to zip it. Either way we need to keep reference to `StringBuilder` to have possibility to call later `toString()` on it.

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: It may look better but unfortunately it is not equal. However it may work in our scope. I will check that.

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: It may look better but once again it's not equal, cause `let` returns kotlin.Unit ('`forEach {}`), and whole result is `"kotlin.Unit"`

BarkingBad: It may look better but once again it's not equal, cause `let` returns kotlin.Unit (`forEach {}`), and whole result is `"kotlin.Unit"`

BarkingBad: Of course it can be mended with returning at the end of lambda decorators

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: I just come up with 
```    
override fun buildPage(page: ContentPage, content: (StringBuilder, ContentPage) -> Unit): String =
        StringBuilder().let {
            content(it, page)
            it
        }.toString()
```
Both version suits me so decide which one you prefer

BarkingBad: I just come up with 
```    
override fun buildPage(page: ContentPage, content: (StringBuilder, ContentPage) -> Unit): String =
        StringBuilder().apply {
            content(this, page)
        }.toString()
```
Of course we can use also instead
Both version suits me so decide which one you prefer

plugins/base/src/main/kotlin/renderers/gfm/GfmRenderer.kt
BarkingBad: It may look better but once again it's not equal, cause `let` returns lambda result, which is `kotlin.Unit` (`forEach {}`). It can of course be easily mended by changing scope function to apply/also.

BarkingBad:
@kamilok1965 We had been discussing about changes and agreed everything is OK. Can you look into it to once more and give approve? @sellophane can you also look into it, you are experienced with renderers

BarkingBad:
Reopen due to my fail while rebasing

PR #649, opened 2020-02-14
Page transformer builders

### LINKED ISSUE #633: Prepare builders for page transformers
The goal of this task is to provide convenient dsl for creating page transformers. There should exist at least three separate builders:
* for transformers that only scans the page tree not modifying it
* for transformers  that maps nodes withut changing structure of entire tree
* for transformers that changes structure of the the tree

Example of builder dsl:
```kt
val transformer: PageTransformer = mappingTransformer {
    content += header(name)
}
```

sswistun-vl:
Added builders for page transformers. Fixes #633


kamilok1965:
I think you misinterpreted the issue as these are only some helper methods to alter pages, not builders

PR #652, opened 2020-02-14
Declaring some build dependencies

mathjeff:
To make it more convenient to build Dokka from source and use the result in another build


sellmair:
I think this is not necessary anymore with dokka 0.11.0. Thanks filing this PR. 
Please let us know, if you encounter any further issues in that direction!

PR #662, opened 2020-02-18
Rename extensions to make them more uniform

kamilok1965:


plugins/base/src/main/kotlin/DokkaBase.kt
```diff
@@ -25,20 +25,20 @@ class DokkaBase : DokkaPlugin() {
...
     val psiToDocumentationTranslator by extending(isFallback = true) {
-        CoreExtensions.psiToDocumentationTranslator with DefaultPsiToDocumentationTranslator
+        CoreExtensions.psiToDocumentableTranslator with DefaultPsiToDocumentableTranslator
     }
 
     val documentableMerger by extending(isFallback = true) {
         CoreExtensions.documentableMerger with DefaultDocumentableMerger
     }
 
     val documentablesToPageTranslator by extending(isFallback = true) {
```
Kordyjan: I think you should also change extension names.

PR #664, opened 2020-02-19
F/missing generic types for functions

### LINKED ISSUE #659: Missing information about generic functions in documentables
Informations about generic types are missing from documentables.

sswistun-vl:
Added information about generic types for functions and classes. Fixes #659


core/src/main/kotlin/model/Generics.kt
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    abstract fun print(t: Type? = null): String
+}
+
+open class TypeConstructor(override val name: String, override val bounds: List<Type>) : Type(name, bounds) {
+    override fun genericCount(root: String): Int = 0
+    override fun toString(): String = "$name<${bounds.joinToString(separator = "") { "$it" }}>"
+    override fun print(t: Type?): String =
+        if (this == t) "" else "$name<${bounds.joinToString(separator = "") { it.print(t ?: this) }}>"
```
Kordyjan: It is never possible for `this` to be equal `t`.

core/src/main/kotlin/model/Generics.kt
```diff
@@ -0,0 +1,135 @@
...
+}
+
+data class TypeConstructorMutable(override val name: String, override var bounds: List<Type>) :
```
Kordyjan: Data classes with vars are confusing. Does this even need to be declared as data class?

plugins/base/src/test/kotlin/model/ModelUtils.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    fun List<DocumentationNode>.commentsToString(): String =
+        this.flatMap { it.children }.joinToString(separator = "\n") { it.root.docTagSummary() }
+
+    fun <T> Collection<T>.assertCount(n: Int, prefix: String = "") =
+        assert(count() == n) { "${prefix}Expected $n, got ${count()}" }
+
+    fun <T> T?.assertEqual(expected: T, prefix: String = "") = assert(this == expected) {
+        "${prefix}Expected $expected, got $this"
+    }
```
Kordyjan: This shouldn't be inside of this class.

plugins/base/src/test/kotlin/model/ModelUtils.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    operator fun List<Documentable?>?.div(name: String): Documentable? = this?.let { d -> d.find { it?.name == name } }
+
+    operator fun Documentable?.div(name: String): Documentable? = this?.children / name
+    fun <T> T?.assertNotNull(name: String): T = this.also { assert(this != null) { "$name should not be null" } }
+        .let { it!! }
+
+    fun <T> Any?.assertIsInstance(name: String): T =
+        this.let { it as? T }.also { assert(this != null) { "$name should not be null" } }.let { it!! }
```
Kordyjan: This is not working as you think it is. Parameter `T` needs to be reified for cast `it as? T` to work and IDE is telling you that!

Kordyjan:
```kt
inline fun <T> Any?.assertIsInstance(name: String): T = 
    this as T ?: throw AssertionError("$name is not ${T::class.name}")
```
is way simpler and thas what you think it does.

plugins/base/src/test/kotlin/model/FunctionsTest.kt
```diff
@@ -0,0 +1,180 @@
...
+                val T = typeParameters.find { it.name == "T" }.assertIsInstance<GenericType>("T")
+                val R = typeParameters.find { it.name == "R" }.assertIsInstance<GenericType>("R")
+
+                with(T) {
+                    name.assertEqual("T", "First type parameter name: ")
+                    bounds.assertCount(1, "First type parameter bounds: ")
+                    with(bounds.single().assertIsInstance<TypeConstructor>("Comparable")) {
+                        name.assertEqual("kotlin.Comparable", "Comparable TypeConstructor: ")
+                        bounds.assertCount(1, "Comparable bounds: ")
+                        bounds.single().name.assertEqual("R", "Comparable bound: ")
```
kamilok1965: Comparable has no type bounds, it only has a type parameter (which is `R` in this case)

core/src/main/kotlin/model/Generics.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    open fun genericCount(root: String): Int = 0
+    protected open fun setRef(cache: Map<String, GenericType>) {
+        bounds.filter { !(it is GenericType) || it is GenericType.GenericReference }.forEach { it.setRef(cache) }
+    }
+
+    abstract fun print(t: Type? = null): String
+}
+
+open class TypeConstructor(override val name: String, override val bounds: List<Type>) : Type(name, bounds) {
```
kamilok1965: TypeConstructor doesn't have bounds, only parameters

PR #674, opened 2020-02-23
Refactor PageBuilder and PageContentBuilder

kamilok1965:
PageBuilder has been renamed to DefaultPageCreator. While creating this commit it turned out that `Classlike` should be an `abstract class` that extends `Documentable`. This will remove all of the casts that this commit introduces


kamilok1965:
For some reason this PR didn't update properly with new commits, closing and reopening

PR #675, opened 2020-02-24
Descriptor to Documentable rewrite

sswistun-vl:
Rewrite of DefaultDescriptorToDocumentableTranslator

### DIFF
--- a/core/src/main/kotlin/model/Documentable.kt
+++ b/core/src/main/kotlin/model/Documentable.kt
@@ -47,7 +47,8 @@ data class PlatformDependent<out T>(
         get() = map.values.distinct().singleOrNull()
 
     companion object {
-        fun <T> empty() = PlatformDependent(mapOf<PlatformData, T>())
+        fun <T> empty(): PlatformDependent<T> = PlatformDependent(emptyMap())
+        fun <T> from(platformData: PlatformData, element: T) = PlatformDependent(mapOf(platformData to element))
     }
 }
 
@@ -77,7 +78,7 @@ interface WithAbstraction {
     val modifier: Modifier?
 
     enum class Modifier {
-        Abstract, Open, Final, Static
+        Abstract, Open, Final, Sealed, Empty
     }
 }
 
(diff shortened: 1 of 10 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -298,38 +349,74 @@ open class DokkaDescriptorVisitor( // TODO: close this class and make it private
...
-    private fun List<PlatformInfo>.filterTagWrappers(
-        types: List<KClass<out TagWrapper>>,
-        name: String? = null
-    ): List<PlatformInfo> =
-        this.map { it.filterTagWrappers(types, name)!! }
-}
+    private fun TypeProjection.toProjection(): Projection =
+        if (isStarProjection) Projection.Star else fromPossiblyNullable(type)
 
+    private fun fromPossiblyNullable(t: KotlinType): Projection =
```
sellophane: You are using this function once, while you use `let { if (t.isMarkedNullable) Projection.Nullable(it) else it }` twice.  Maybe it's better to functionize this instead?

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -75,7 +76,7 @@ interface WithAbstraction {
     val modifier: Modifier
 
     enum class Modifier {
-        Abstract, Open, Final
+        Abstract, Open, Final, Sealed, Empty
```
sellophane: I went with nullabling Modifiler and it was merged, so you can remove empty and go the same way

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -68,155 +69,198 @@ open class DokkaDescriptorVisitor( // TODO: close this class and make it private
...
+        val info = descriptor.resolveClassDescriptionData(platformData)
+
+        return Interface(
+            dri = driWithPlatform.dri,
+            name = descriptor.name.asString(),
+            functions = scope.functions(driWithPlatform),
+            properties = scope.properties(driWithPlatform),
+            classlikes = scope.classlikes(driWithPlatform),
+            actual = descriptor.actual(platformData, descriptor),
+            visibility = PlatformDependent.from(platformData, descriptor.visibility),
```
sellophane: For psi parsing I created an extension function for simplifying `PlatformDependent.from(platformData, <something>)`, you can reuse it here

PR #693, opened 2020-02-28
Stdlib npe fix

sswistun-vl:
JavaPropertyDescriptor has no accessors, so I've altered descriptorToDocumentableTranslator to work with them.


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -17,6 +17,7 @@ import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.impl.DeclarationDescriptorVisitorEmptyBodies
 import org.jetbrains.kotlin.idea.kdoc.findKDoc
 import org.jetbrains.kotlin.idea.refactoring.fqName.fqName
+import org.jetbrains.kotlin.library.metadata.KlibMetadataProtoBuf.fqName
```
kamilok1965: This import looks wrong

sswistun-vl: Cleaned up.

PR #694, opened 2020-02-28
Enum fix

sswistun-vl:
Fixed small problem with enums and enum entries - enums' companions weren't separated from enum entries during its descriptor processing.


sswistun-vl:
I've changed commit message.

PR #701, opened 2020-03-03
Rewrite of signature providing & other minor changes to content model

Kordyjan:
Creates mechanism for signature providers. Also adds wrapping for group contents and hints about platform dependent content.

### DIFF
--- a/plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
+++ b/plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
@@ -48,7 +48,7 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
 
     private fun signature(f: Function) = contentBuilder.contentFor(f, ContentKind.Symbol) {
         platformText(f.visibility) { it.externalDisplayName + " " }
-        text(f.modifier.toString() + " ")
+        text(f.modifier.toString().toLowerCase() + " ")
         text("fun ")
         f.receiver?.also {
             type(it.type)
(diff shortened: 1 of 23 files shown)


plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private fun signature(t: TypeParameter) = contentBuilder.contentFor(t, ContentKind.Symbol) {
+        link(t.name, t.dri)
+        list(t.bounds, prefix = " : ") {
+            signatureForProjection(it)
+        }
+    }
+
+    private fun PageContentBuilder.DocumentableContentBuilder.signatureForProjection(p: Projection): Unit = when (p) {
```
kamilok1965: I think you can just wrap this in `contentBuilder.contentFor(){}` and then use it with plus: `+signatureForProjection(it)`, like `buildSignature`

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private fun <T : Documentable> Collection<T>.filterOnPlatform(platformData: PlatformData) =
+        this.filter { it.platformData.contains(platformData) }
+}
+
+private fun <T> PageContentBuilder.DocumentableContentBuilder.platformText(
+    value: PlatformDependent<T>,
+    transform: (T) -> String
+) = value.entries.forEach { (p, v) -> text(transform(v), platformData = setOf(p)) }
```
kamilok1965: Maybe we can move this to the `DocumentableContentBuilder` altogether?

Issue #639, opened 2020-02-13
`applyIf` block in Extension DSL overwrites each other


Kordyjan:
Right now user can declare multiple `applyIf` blocks inside of Extension DSL. As this blocks overrwrites each other, only the last one will be checked.

There are three possible solutions:
* Leverage type system to make it compile time error to declare multiple `applyIf` blocks in single `extending` block.
* Evaluate multiple conditions and apply plugin only if all of them are satisfied
* Make it runtime error if there are multiple `applyIf`s

PR #705, opened 2020-03-04
Multiple apply if

sswistun-vl:
Conditions from applyIf for plugins are combined instead of getting overwritten. Closing #639

### DIFF
--- a/core/src/main/kotlin/model/properties/properties.kt
+++ b/core/src/main/kotlin/model/properties/properties.kt
@@ -1,5 +1,8 @@
 package org.jetbrains.dokka.model.properties
 
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.ExtraModifiers
+
 interface ExtraProperty<in C : Any> {
     interface Key<in C : Any, T : Any> {
         fun mergeStrategyFor(left: T, right: T): MergeStrategy<C> = MergeStrategy.Fail {
--- a/core/src/main/kotlin/model/properties/properties.kt
+++ b/core/src/main/kotlin/model/properties/properties.kt
@@ -1,8 +1,5 @@
 package org.jetbrains.dokka.model.properties
 
-import org.jetbrains.dokka.model.Documentable
-import org.jetbrains.dokka.model.ExtraModifiers
-
 interface ExtraProperty<in C : Any> {
     interface Key<in C : Any, T : Any> {
         fun mergeStrategyFor(left: T, right: T): MergeStrategy<C> = MergeStrategy.Fail {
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/plugability/extensions.kt
```diff
@@ -96,7 +96,9 @@ class ExtendingDSL(private val pluginClass: String, private val extensionName: S
         ExtensionOrdered(extensionPoint, pluginClass, extensionName, action, block)
 
     infix fun <T : Any> Extension<T>.applyIf(condition: DokkaConfiguration.() -> Boolean): Extension<T> =
-        this.setCondition(condition)
+        this.setCondition(this.condition + condition)
+
+    infix operator fun <T: Any> ((T) -> Boolean).plus(other: (T) -> Boolean): (T) -> Boolean = {this(it) && other(it)}
```
Kordyjan: If your goal was to make multiple `applyIf`s combine with each other to form the new predicate, this plus operator shouldn't be part of the dsl. As it is right now it will be only confusing for end users. Inline it inside of `applyIf`.

sswistun-vl: Changed.

core/src/main/kotlin/model/properties/properties.kt
```diff
@@ -1,5 +1,8 @@
 package org.jetbrains.dokka.model.properties
 
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.ExtraModifiers
+
```
kamilok1965: Looks like unused imports

core/src/main/kotlin/plugability/extensions.kt
```diff
@@ -96,7 +96,8 @@ class ExtendingDSL(private val pluginClass: String, private val extensionName: S
         ExtensionOrdered(extensionPoint, pluginClass, extensionName, action, block)
 
     infix fun <T : Any> Extension<T>.applyIf(condition: DokkaConfiguration.() -> Boolean): Extension<T> =
-        this.setCondition(condition)
+        this.setCondition { this@applyIf.condition(this) && condition(this) }
```
kamilok1965: IMHO it would be better to just have a `addCondition` or something similar to have a list of conditions. This way you'll just call `conditions.all { it == somepredicate }`  when you need this

kamilok1965:
Also commit names do not conform to commit naming guidelines

core/src/main/kotlin/plugability/extensions.kt
sswistun-vl: Changed.

core/src/main/kotlin/model/properties/properties.kt
sswistun-vl: Removed.

PR #712, opened 2020-03-09
Adds Dokka Base plugin to runners and add dokkaPlugins rule to CLI and Maven runners

BarkingBad:


runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -190,16 +226,76 @@ abstract class AbstractDokkaMojo : AbstractMojo() {
...
+        locator.addService(RepositoryConnectorFactory::class.java, BasicRepositoryConnectorFactory::class.java)
+        locator.addService(TransporterFactory::class.java, FileTransporterFactory::class.java)
+        locator.addService(TransporterFactory::class.java, HttpTransporterFactory::class.java)
+        return locator.getService(RepositorySystem::class.java)
+    }
+
+    private fun newSession(system: RepositorySystem): RepositorySystemSession {
+        val session: DefaultRepositorySystemSession =
+            MavenRepositorySystemUtils.newSession()
+        val localRepo = LocalRepository(System.getProperty("user.home") + "/.m2/repository")
```
BarkingBad: This path should cover Windows + MacOS/Unix systems.

PR #718, opened 2020-03-10
Collapsing navigation menu properly to opened page

pikinier20:
Added IDs to navigation elements, tagged each contentpage with ID. Added js function, which collapses navigation based on opened content

### DIFF
--- a/core/src/main/resources/dokka/scripts/navigationLoader.js
+++ b/core/src/main/resources/dokka/scripts/navigationLoader.js
@@ -1,12 +1,41 @@
 onload = () => {
     fetch(pathToRoot + "navigation.html")
-    .then(response => response.text())
-    .then(data => {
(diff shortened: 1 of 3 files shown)


core/src/main/resources/dokka/scripts/navigationLoader.js
```diff
@@ -8,5 +8,34 @@ onload = () => {
...
+        })
+    }).then(() => {
+        collapseNavigation()
+ })
+}
+
+collapseNavigation = () => {
+        var pageId = document.getElementById("content").attributes["pageIds"].value.toString()
+        var parts = document.querySelectorAll(".sideMenuPart")
+        var found = 0
```
kamilok1965: I think we should replace `var`s with `let`s

PR #723, opened 2020-03-11
Fixing signatures font, incorrect links and line breaking

sellophane:


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -27,11 +27,15 @@ open class HtmlRenderer(
         node: ContentGroup,
         pageContext: ContentPage,
         childrenCallback: FlowContent.() -> Unit
-    ) = when {
-        node.dci.kind == ContentKind.BriefComment -> div("brief") { childrenCallback() }
-        node.style.contains(TextStyle.Paragraph) -> p { childrenCallback() }
-        node.style.contains(TextStyle.Block) -> div { childrenCallback() }
-        else -> childrenCallback()
+    ) {
+        val additionalClasses = if (node.style.contains(TextStyle.Monospace)) "monospace" else ""
```
kamilok1965: Instead of this, I'd suggest doing this:
```suggestion
        val additionalClasses =  node.style.joinToString(" ") { it.toString().toLowerCase() }
```
This way we'll have all the classes and we can always not use them in the css

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -61,6 +61,7 @@ open class DefaultPageCreator(
             link(it.name.orEmpty(), it.dri)
             group {
                 +buildSignature(it)
+                +ContentBreakLine(platformData.toSet())
```
kamilok1965: I'd make this a helper function in the content builder

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
sellophane: ok

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
sellophane: ok

plugins/base/src/main/kotlin/translators/documentables/PageContentBuilder.kt
```diff
@@ -229,6 +229,10 @@ open class PageContentBuilder(
             block: DocumentableContentBuilder.() -> Unit
         ): ContentGroup = contentFor(dri, platformData, kind, styles, extra, block)
 
+        fun breakLine(platformData: Set<PlatformData>) {
```
kamilok1965:
```suggestion
        fun breakLine(platformData: Set<PlatformData> = mainPlatformData) {
```
and the call looks like this:
`breakLine()`

PR #735, opened 2020-03-12
Adding documentable filtering based on visibility

sellophane:
Please do merge now

### DIFF
--- a/plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilter.kt
+++ b/plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilter.kt
@@ -1,0 +1,316 @@
+package org.jetbrains.dokka.base.transformers.documentables
+
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.model.DAnnotation
+import org.jetbrains.dokka.model.DEnum
+import org.jetbrains.dokka.model.DFunction
+import org.jetbrains.dokka.pages.PlatformData
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.transformers.documentation.PreMergeDocumentableTransformer
+
+internal object DocumentableVisibilityFilter : PreMergeDocumentableTransformer {
+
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/model/Documentable.kt
```diff
@@ -322,6 +327,22 @@ data class TypeParameter(
...
+            if (intersection.isEmpty()) null
+            else TypeParameter(
+                dri,
+                name,
+                documentation.filtered(intersection),
+                bounds,
+                intersection,
+                extra
+            )
+        }
```
kamilok1965: IMHO it would be better to move this to some utils class also

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -37,6 +37,11 @@ data class PlatformDependent<out T>(
         yieldAll(map.values)
     }
 
+    fun filtered(platformDataList: List<PlatformData>) = PlatformDependent(
+        map.filter { it.key in platformDataList },
+        expect
+    )
```
kamilok1965: IMHO it would be better to move this to some utils class

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilter.kt
```diff
@@ -0,0 +1,299 @@
...
+            }
+        }
+        return Pair(classlikesListChanged, filteredClasslikes)
+    }
+
+    private val allowedVisibilities = arrayOf(
+        JavaVisibility.Public,
+        JavaVisibility.Default,
+        JavaVisibility.Protected,
+        KotlinVisibility.Protected,
```
kamilok1965: I'd personally filter out `Protected` members by default as they are not a part of the API most of the time. 
This list should be configurable by user though or at least we should have some options to choose from

core/src/main/kotlin/pages/PageNodes.kt
```diff
@@ -1,8 +1,8 @@
 package org.jetbrains.dokka.pages
 
-import org.jetbrains.dokka.model.Documentable
 import org.jetbrains.dokka.Platform
 import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.Documentable
```
kamilok1965: Looks like an unneeded change

PR #739, opened 2020-03-13
Fixed data class rendering. Added information about primary construct…

pikinier20:
…or and changed its rendering.

### DIFF
--- a/plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
+++ b/plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
@@ -33,7 +33,11 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
     private fun signature(c: DClasslike) = contentBuilder.contentFor(c, ContentKind.Symbol, setOf(TextStyle.Monospace)) {
         platformText(c.visibility) { (it.takeIf { it !in ignoredVisibilities }?.name ?: "") + " " }
         if (c is DClass) {
-            text(c.modifier.name + " ")
+            if (c.extra[AdditionalModifiers.AdditionalKey]?.content?.contains(ExtraModifiers.DATA) == true) {
+                text("data ")
+            } else {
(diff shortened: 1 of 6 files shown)


plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -43,6 +47,17 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
             is Annotation -> text("annotation class ")
         }
         link(c.name!!, c.dri)
+        if(c is Class){
+            val pConstructor = c.constructors.singleOrNull() { it.extra[PrimaryConstructorExtra.PrimaryConstructorExtraKey] != null }
+            if(pConstructor != null) {
+                list(pConstructor.parameters, "(", ")", ",", pConstructor.platformData.toSet()){
+                    breakLine()
+                    text(it.name ?: "", styles = mainStyles.plus(TextStyle.Bold))
```
sellophane: There is Kotlin function String?.orEmpty() you can use here

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -458,6 +458,13 @@ private class DokkaDescriptorVisitor( // TODO: close this class and make it priv
         ExtraModifiers.OVERRIDE.takeIf { DescriptorUtils.isOverride(this) }
     ).toContainer()
 
+    inline fun <reified D : Documentable> ConstructorDescriptor.additionalExtras(): PropertyContainer<D> =
+        if(this.isPrimary)
+            (this as FunctionDescriptor).additionalExtras<D>() + PrimaryConstructorExtra
```
sellophane: If you are casting ConstructorDescriptor to FunctionDescriptor either way, you can refactor the whole extension into `FunctionDescriptor.additionalExtras()`

PR #742, opened 2020-03-16
Add default values for parameters

### LINKED ISSUE #741: Add extras for parameter values

sswistun-vl:
Added documenting default values of parameters in extras. Closes #741


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -349,7 +349,12 @@ private class DokkaDescriptorVisitor( // TODO: close this class and make it priv
             type = descriptor.type.toBound(),
             documentation = descriptor.resolveDescriptorData(platformData),
             platformData = listOf(platformData),
-            extra = descriptor.additionalExtras()
+            extra = descriptor.getDefaultValue().let {
+                if (it == null)
+                    descriptor.additionalExtras()
+                else
+                    descriptor.additionalExtras() + DefaultValue(it)
+            }
```
kamilok1965:
```suggestion
            extra = descriptor.getDefaultValue()?.let {
                    descriptor.additionalExtras() + DefaultValue(it)
            } ?: descriptor.additionalExtras()
```
Checking for nullability in `.let` cannot succeed as `.let` will just fail when `getDefaultValue()` returns null

sswistun-vl: Changed.

PR #744, opened 2020-03-17
Gradle Task supporting multimodular projects

BarkingBad:


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaCollectorTask.kt
```diff
@@ -0,0 +1,55 @@
...
+    @Input
+    var outputDirectory: String = ""
+
+    private lateinit var configuration: GradleDokkaConfigurationImpl
+
+    @TaskAction
+    fun collect() {
+        val passesConfigurations = getProjects(project).filter { it.name in modules }.map {
+            val task = try {
+                it.tasks.getByName(DOKKA_TASK_NAME) as DokkaTask
```
Kordyjan: if you add `compileOnly(gradleKotlinDsl())` dependency you can use typesafe variants here.

PR #750, opened 2020-03-18
Update README.md to specify import statement needed in build.gradle.kts

vsajip:
Without this, you get errors when Gradle tries to compile `build.gradle.kts`. The fully-qualified name isn't obvious, so adding this information will save at least an Internet search for `DokkaTask` to find where it's defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ tasks {
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


kamilok1965:
I agree. Thanks for this contribution!

PR #758, opened 2020-03-20
Fix default values

sswistun-vl:
Default values are now platform dependent and merged correctly.


Kordyjan:
Did you test if actual functions parameters are able to have different default values than parameters from expected function?

PR #781, opened 2020-03-31
Parameters rendering

Kordyjan:


plugins/base/build.gradle.kts
```diff
@@ -6,6 +6,7 @@ plugins {
 
 dependencies {
     implementation("org.jsoup:jsoup:1.12.1")
+    testImplementation(project(":test-tools"))
 }
```
kamilok1965: While we're here can we unify those two `dependencies{}` blocks that are in this file?

PR #783, opened 2020-03-31
Includes, SourceLinks, Samples

BarkingBad:
Handles configuration inputs: Includes, SourceLinks, Samples.

Please look into the EP for Samples. I don't know if I got it right.

I also think the API for Samples (actually annotations type aliases) should be included in dokka. Right now I have seen them losely attached to stdlib.

#747

### DIFF
--- a/plugins/base/src/main/kotlin/DokkaBase.kt
+++ b/plugins/base/src/main/kotlin/DokkaBase.kt
@@ -22,6 +22,7 @@ import org.jetbrains.dokka.base.transformers.pages.merger.FallbackPageMergerStra
 import org.jetbrains.dokka.base.transformers.pages.merger.PageMerger
 import org.jetbrains.dokka.base.transformers.pages.merger.PageMergerStrategy
 import org.jetbrains.dokka.base.transformers.pages.merger.SameMethodNamePageMergerStrategy
+import org.jetbrains.dokka.base.transformers.pages.sourcelinks.SourceLinksTransformer
 import org.jetbrains.dokka.base.translators.descriptors.DefaultDescriptorToDocumentableTranslator
 import org.jetbrains.dokka.base.translators.documentables.DefaultDocumentableToPageTranslator
 import org.jetbrains.dokka.base.translators.psi.DefaultPsiToDocumentableTranslator
@@ -125,6 +126,10 @@ class DokkaBase : DokkaPlugin() {
 
     val rootCreator by extending {
         htmlPreprocessors with RootCreator
+    }
+
+    val sourceLinksTransformer by extending {
+        htmlPreprocessors providing ::SourceLinksTransformer order { after(rootCreator) }
     }
 
     val navigationPageInstaller by extending {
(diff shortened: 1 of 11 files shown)


plugins/base/src/main/kotlin/transformers/pages/samples/SamplesTransformer.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.jetbrains.kotlin.resolve.DescriptorToSourceUtils
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+import java.io.File
+
+abstract class SamplesTransformer(val context: DokkaContext) : PageTransformer {
+
+    abstract fun processBody(psiElement: PsiElement): String
+    abstract fun processImports(psiElement: PsiElement): String
+
+    override fun invoke(input: RootPageNode): RootPageNode {
```
Kordyjan, line 28: Maybe make this final override.

plugins/base/src/main/kotlin/transformers/documentables/ModuleAndPackageDocumentationTransformer.kt
```diff
@@ -0,0 +1,81 @@
...
+internal object ModuleAndPackageDocumentationTransformer : PreMergeDocumentableTransformer {
+
+    override fun invoke(original: List<DModule>, context: DokkaContext): List<DModule> {
+
+        val modulesAndPackagesDocumentation =
+            context.configuration.passesConfigurations
+                .map {
+                    Pair(it.moduleName, it.platformData) to
+                        it.includes.map { Paths.get(it) }
+                            .filter { Files.exists(it) }
```
kamilok1965, line 22: I'd log warning about nonexistent file here. Otherwise it'll be hard to debug

plugins/base/src/main/kotlin/transformers/pages/samples/SamplesTransformer.kt
BarkingBad: Good idea

plugins/base/src/main/kotlin/DokkaBase.kt
```diff
@@ -51,6 +56,10 @@ class DokkaBase : DokkaPlugin() {
         CoreExtensions.preMergeDocumentableTransformer with DocumentableVisibilityFilter
     }
 
+    val modulesAndPackagesDocumentation by extending(isFallback = true) {
```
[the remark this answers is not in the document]
BarkingBad: Ok

BarkingBad:
> I was imagining this extension point more as a separate interface that is used by some

I made it compatible with other PageTransformers, but can chagne it. Just let me know

plugins/base/src/main/kotlin/transformers/documentables/ModuleAndPackageDocumentationTransformer.kt
BarkingBad: Ok

PR #788, opened 2020-04-02
Fix to Authors comment section - added handling for multiple author tags

pikinier20:


pikinier20:
> Please rebase and check if tests are passing.

Rebased. Few tests failed because authors are now named tag and produce table instead of group, so I modified them.

core/src/main/kotlin/model/doc/TagWrapper.kt
```diff
@@ -19,7 +19,7 @@ sealed class NamedTagWrapper(root: DocTag, val name: String) : TagWrapper(root)
 }
 
 class Description(root: DocTag) : TagWrapper(root)
-class Author(root: DocTag) : TagWrapper(root)
+class Author(root: DocTag, name: String) : NamedTagWrapper(root, name)
```
BarkingBad, line 22: Do we really need this as a named Tag? In parser you just duplicate this property.

BarkingBad: There is possibility to process this tag later via `GroupedTags.withTypeUnnamed()`

pikinier20: Not exactly, because `GroupedTags.withTypeUnnamed()`, as the name states, does grouping which in result gives you only one tag of specific type. This solution would fail if you had more than one author in KDoc (which is possible and justified). Named tag is grouped additionally by name.

BarkingBad: Of course @pikinier20, but this will lead to unwanted behaviour with expect/actuals, because name in expect will be treated as fallback. So in expect class you can have "Foo" and "Bar" authors, in actual have only "Bar", but both will be printed in generated documentation

pikinier20: But how unnamed tag would avoid that while preserving possibility of adding multiple authors?

PR #795, opened 2020-04-03
DModuleView

sswistun-vl:


core/src/main/kotlin/model/Documentable.kt
```diff
@@ -121,18 +121,38 @@ interface Callable : WithVisibility, WithType, WithAbstraction, WithExpectActual
...
+}
+
+data class DPass(
+    override val name: String,
+    override val packages: List<DPackage>,
+    override val documentation: PlatformDependent<DocumentationNode>,
+    override val platformData: List<PlatformData>,
+    override val extra: PropertyContainer<DModuleView> = PropertyContainer.empty()
+) : DModuleView(name, packages, documentation, platformData, extra) {
+    override fun withNewExtras(newExtras: PropertyContainer<DModuleView>): DModuleView  = copy(extra = newExtras)
```
Kordyjan, line 153: We do not need to override each property. We can just use them as constructor parameters, passed to superclass constructor. The other way is to make every property in base class abstract. The way that it is now we are creating an additional property for each of the properties from the base class.

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -121,18 +121,38 @@ interface Callable : WithVisibility, WithType, WithAbstraction, WithExpectActual
 
 abstract class DClasslike : Documentable(), WithScope, WithVisibility, WithExpectActual
 
-data class DModule(
+sealed class DModuleView(
     override val name: String,
-    val packages: List<DPackage>,
+    open val packages: List<DPackage>,
```
Kordyjan, line 126: We do not need this to be `open`.

core/src/main/kotlin/model/properties/PropertyContainer.kt
```diff
@@ -56,4 +56,6 @@ fun <C> C.mergeExtras(left: C, right: C): C where C : Any, C : WithExtraProperti
     val newExtras = PropertyContainer((unambiguous.flatten() + replaces).associateBy { it.key })
 
     return needingFullMerge.fold(withNewExtras(newExtras)) { acc, merger -> merger(acc, left, right) }
-}
\ No newline at end of file
+}
+
+fun <C> C.mergeExtras(elems: Collection<C>): C where C : Any, C : WithExtraProperties<C> = elems.fold(this, ::mergeExtras)
```
Kordyjan, line 61: Sadly it is not that simple. This will lead to the case where we merge three objects - let's call them A, B, C. We merge ordinary properties of A and B, then merge this new (A+B) object's ordinary properties with ordinary properties of C, but this may lead to unexpected behavior since we extras for (A+B) are empty. I know it is quite subtle, but it may escalate, resulting in really hard to find bugs.

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
```diff
@@ -9,25 +9,22 @@ import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
...
-        }
+        return DModule(
+            name = projectName,
+            packages = merge(
+                passes.flatMap { it.packages },
+                DPackage::mergeWith
+            ),
+            documentation = passes.platformDependentFor { documentation },
+            platformData = passes.flatMap { it.platformData }.distinct()
+        ).mergeExtras(passes) as DModule
```
Kordyjan, line 27: As I mentioned in one of the previous comments we cannot merge all objects at once.

PR #798, opened 2020-04-06
Changed PlatformHintedContent rendering, js and css to match figma st…

pikinier20:
…yle.

### DIFF
--- a/core/src/main/resources/dokka/styles/style.css
+++ b/core/src/main/resources/dokka/styles/style.css
@@ -487,6 +487,56 @@ td.content {
     color: #5B5DEF;
 }
 
+.platform-hinted {
+    display: block;
+}
+
+.platform-hinted > .platform-bookmarks-row {
+    display: flex;
+    flex-direction: row;
+    padding: 0px;
+}
+
+.platform-hinted > .platform-bookmarks-row > .platform-bookmark {
+    min-width: 64px;
+    height: 36px;
+    border: 2px solid white;
+    background: white;
+    outline: none;
+    flex: none;
+    order: 5;
+    align-self: flex-start;
+    margin: 0;
+}
+
+.platform-hinted > .platform-bookmarks-row > .platform-bookmark:nth-child(1):hover {
+    border-top: 2px solid gray;
+}
+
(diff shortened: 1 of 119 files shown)


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -44,7 +44,7 @@ open class HtmlRenderer(
         div("platform-tagged"){
             node.platforms.forEach {
                 div("platform-tag ${it.platformType.name}"){
-                    text(it.platformType.key.toUpperCase())
+                    text(it.platformType.toPrettyString())
```
kamilok1965, line 47: We should use `targets` for that, not `analysisPlatform`

core/src/main/kotlin/configuration.kt
```diff
@@ -10,6 +10,13 @@ enum class Platform(val key: String) {
     native("native"),
     common("common");
 
+    fun toPrettyString() = when(this){
+            jvm -> "JVM"
+            js -> "JS"
+            common -> "Common"
+            native -> "Native"
+    }
```
kamilok1965, line 18: Unneeded, `Platform` should only be used to identify passes and to configure analysis properly

core/src/main/resources/dokka/styles/style.css
```diff
@@ -427,6 +427,59 @@ tr.platform-tagged {
...
+    border-top: 2px solid crimson;
+}
+
+.platform-hinted > .platform-bookmarks-row > .platform-bookmark[index='common']:hover{
+    border-top: 2px solid gray;
+}
+
+.platform-hinted > .platform-bookmarks-row > .platform-bookmark[index='native']:hover{
+    border-top: 2px solid blue;
+}
```
kamilok1965, line 469: Hardcoding colors for specific platforms in CSS is not a good idea IMHO.  We should have a class with a color for common elements and then some classes that we choose from depending on the order I think? Or even randomly

kamilok1965:
Also it's called Figma, but I appreciate the wordplay

PR #803, opened 2020-04-08
Frontend fixes (mostly)

kamilok1965:


plugins/base/src/main/kotlin/translators/documentables/PageContentBuilder.kt
```diff
@@ -148,7 +148,6 @@ open class PageContentBuilder(
                     emptyList(),
                     elements.map {
                         buildGroup(it.dri, it.platformData.toSet(), kind, styles, extra) {
-                            // TODO this will fail
```
Kordyjan, line 151: So this won't fail?

kamilok1965: I'm baffled too

PR #807, opened 2020-04-14
Fix stdlib generation

kamilok1965:


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -72,7 +72,14 @@ private class DokkaDescriptorVisitor(
         descriptor: PackageFragmentDescriptor,
         parent: DRIWithPlatformInfo
     ): DPackage {
-        val driWithPlatform = DRI(packageName = descriptor.fqName.asString()).withEmptyInfo()
+        val name = descriptor.fqName.asString().let {
```
sellophane, line 75: .takeUnless { it.isBlank() } ?:

kamilok1965: Yesss, that's what I needed

PR #808, opened 2020-04-15
Adjustments to css: font-families, font size, active navigation part style

pikinier20:


core/src/main/resources/dokka/styles/style.css
```diff
@@ -193,7 +197,7 @@ td:first-child {
 }
 
 .symbol {
-    font-family: Monaco, Bitstream Vera Sans Mono, Lucida Console, Terminal;
+    font-family: JetBrains Mono;
```
kamilok1965, line 200: We should have a fallback font if this one can't be found

pikinier20: Added.

PR #810, opened 2020-04-15
Fix missing expect KDoc when all actual implementations don't override KDoc

BarkingBad:
Tested on multiplatfrom with jvm, js and linuxX64 actuals. Works fine now


BarkingBad:
After further investigation there is more problems to be fixed

PR #814, opened 2020-04-17
All modules page

sswistun-vl:
Task for generating top level page with links to submodules.


plugins/base/src/main/kotlin/allModulePage/MultimodulePageCreator.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.jetbrains.dokka.transformers.pages.PageCreator
+import org.jetbrains.dokka.utilities.DokkaLogger
+import java.io.File
+
+class MultimodulePageCreator(
+    context: DokkaContext
+) : PageCreator {
+    private val logger: DokkaLogger = context.logger
+
+    override fun invoke(context: DokkaContext): RootPageNode {
```
Kordyjan, line 24: I don't think we should pass context here. It is already available in this class.

PR #821, opened 2020-04-22
Platform tags adjustments

pikinier20:
Fixed alignment and colors of platform tags. Common tag is always #F4F4F4. Platform tags are now based on target instead of analysis platform

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -49,8 +49,10 @@ open class HtmlRenderer(
     ) {
         div("platform-tagged") {
             node.platforms.forEach {
-                div("platform-tag ${it.platformType.name}") {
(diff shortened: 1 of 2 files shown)


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -49,8 +49,10 @@ open class HtmlRenderer(
     ) {
         div("platform-tagged") {
             node.platforms.forEach {
-                div("platform-tag ${it.platformType.name}") {
-                    text(it.platformType.key.toUpperCase())
+                val targets = it.targets.joinToString(", ")
+                div("platform-tag") {
+                    if( targets == "common" ) classes = classes + "common"
```
kamilok1965, line 54: "common" can be upper-case and this won't work

pikinier20: Changed, is it ok?

PR #843, opened 2020-05-11
Developer's guide to dokka

sellophane:


plugins/developer_guide.md
```diff
@@ -0,0 +1,305 @@
...
+}
+```
+
+## Dokka Data Model
+
+There a four data models that dokka uses: Documentable Model, Documentation Model, Page Model and Content Model.
+
+### Documentable Model
+
+Documentable model represents parsed data, returned by compilator analysis. It retains basic order structure of parsed `Psi` or `Descriptor` models.
```
MarcinAman, line 63: compiler?

plugins/developer_guide.md
```diff
@@ -0,0 +1,305 @@
...
+Merge strategy for extras is invoked only if merged objects have different values for same Extra. If you don't expect it to happen, you can ommit implementing `mergeStrategyFor` function.
+All extras are stored in `PropertyContainer`.
+
+### Documentation Model
+
+Documentation model is used along Documentable Model to store data obtained by parsing code commentaries.
+
+There are three important classes here:
+
+* `DocTag` describes a specific documentation syntax element, for example header, footer, list, link, raw text, paragraph, etc.
```
MarcinAman, line 102: i think there should be `:` after `for example`

plugins/developer_guide.md
```diff
@@ -0,0 +1,305 @@
...
+* `TagWrapper` described a whole comment description or a specific comment tag, for example @See, @Returns, @Author, and holds consisting `DocTag` elements 
+* `DocumentationNode` acts as a container for `TagWrappers` for a specific `Documentable`
+
+DocumentationNodes are references by a specific `Documentable`
+
+### Page Model
+
+Page Model represents the structure of future generated documentation pages and is independent of the final output format, which each node corresponding to exactly one output file. `Rendered` is processing each page separately.
```
MarcinAman, line 110: What is `Rendered`? 

Did you mean `Renderer`?

plugins/developer_guide.md
```diff
@@ -0,0 +1,305 @@
...
+    fun render(root: RootPageNode)
+}
+```
+
+By default, only `HTMLRenderer`, that extends basic `DefaultRenderer`, is created, but it will be registered only if configuration parameter `format` is set to `html`.
+
+### Default extensions' extension points
+
+Default core extension points already have an implementation for providing basic dokka functionality. All of them are declared in `DokkaBase` plugin.
+If you don't want this default extensions to load, all you need to do is write a custom configuration to disable loading said plugin. You will then need to implement extensions for all core extension points.
```
MarcinAman, line 262: What if I want only some to load? Or add a new extension to a current list of extensions for a extension point?

Also: how to disable them?

plugins/developer_guide.md
```diff
@@ -0,0 +1,305 @@
...
+
+tasks.withType<KotlinCompile> {
+    kotlinOptions.jvmTarget = "1.8"
+}
+```
+
+## Building sample plugin
+
+In order to load a plugin into dokka, your class must extend `DokkaPlugin` class. All instances are automatically 
+loaded during dokka setup using `ServiceLoader`.
```
MarcinAman, line 24: it might be worth to mention that `ServiceLoader` is from `java.util` and it not a custom implementation. I'd change `ServiceLoader` to `java.util.ServiceLoader`

plugins/developer_guide.md
sellophane: I added a paragraph about adding extension points and obtaining extension instances.

You can not load only some of them.

kamilok1965:
We should also include information about `META-INF.services`  that is needed when creating a plugin

plugins/developer_guide.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Documentable Model
+
+Documentable model represents parsed data, returned by compiler analysis. It retains basic order structure of parsed `Psi` or `Descriptor` models.
+
+After creation, it is a collection of trees, each with `DModel` as a root. After the Merge step, all trees are folded into one. 
+
+The main building block of this model is `Documentable` class, that is a base class for all more specific types that represents elements of parsed Kotlin and Java classes with pretty self-explanatory names: `DPackage`, `DFunction` and so on. `DClasslike` is a base for class-like elements, such as Classes, Enums, Interfaces and so on.
+
+There are three non-documentable classes important for the model: `DRI`, `SourceSetDependent` and `Extra`.
```
Kordyjan, line 112:
```suggestion
There are three non-documentable classes important for the model: `DRI`, `SourceSetDependent` and `ExtraProperty`.
```

plugins/developer_guide.md
```diff
@@ -0,0 +1,348 @@
...
+    interface Key<in C : Any, T : Any> {
+        fun mergeStrategyFor(left: T, right: T): MergeStrategy<C> = MergeStrategy.Fail {
+            throw NotImplementedError("Property merging for $this is not implemented")
+        }
+    }
+
+    val key: Key<C, *>
+}
+```
+It is advised to use following pattern when declaring new extras:
```
Kordyjan, line 128: I think we need here three things:
1. Example of accessing an extra using its class name
2. Explanations of what generic parameter of extra means and how it can be used to narrow down possible targets.
3. Some mention of marker extras (so extras that are objects that are keys to themselves).

PR #844, opened 2020-05-11
Missing generics on class #834

MarcinAman:
- Added displaying generics on classes
- Added generics to Annotations
- Removed inherited types from inner classes (now there are only types that are declared by the inner class and not inherited from parent), just like it is in code
- Refactored DRI Target to use a sealed class instead of `int?`
- Added a reference to a point of declaration for types. So in case of:
```
class Outer<T> {
    inner class Inner<G> {
        fun foo(): T = { TODO() }
    }
}
```
While user is on Inner class, link on the T return type forwards to Outer class.

### DIFF
--- a/core/src/main/kotlin/links/DRI.kt
+++ b/core/src/main/kotlin/links/DRI.kt
@@ -1,9 +1,6 @@
 package org.jetbrains.dokka.links
 
-import com.intellij.psi.PsiClass
-import com.intellij.psi.PsiElement
-import com.intellij.psi.PsiMethod
-import com.intellij.psi.PsiParameter
+import com.intellij.psi.*
 import org.jetbrains.kotlin.descriptors.*
 import org.jetbrains.kotlin.psi.psiUtil.parentsWithSelf
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
@@ -20,17 +17,15 @@ data class DRI(
     val packageName: String? = null,
     val classNames: String? = null,
     val callable: Callable? = null,
-    val target: Int? = null,
-    val genericTarget: Int? = null,
+    val target: DriTarget = PointingToDeclaration,
     val extra: String? = null
 ) {
     override fun toString(): String =
-        "${packageName.orEmpty()}/${classNames.orEmpty()}/${callable?.name.orEmpty()}/${callable?.signature().orEmpty()}/${target?.toString().orEmpty()}/${extra.orEmpty()}"
+        "${packageName.orEmpty()}/${classNames.orEmpty()}/${callable?.name.orEmpty()}/${callable?.signature().orEmpty()}/$target/${extra.orEmpty()}"
 
     companion object {
         fun from(descriptor: DeclarationDescriptor) = descriptor.parentsWithSelf.run {
             val callable = firstIsInstanceOrNull<CallableDescriptor>()
-            val params = callable?.let { listOfNotNull(it.extensionReceiverParameter) + it.valueParameters }.orEmpty()
             DRI(
                 firstIsInstanceOrNull<PackageFragmentDescriptor>()?.fqName?.asString(),
                 (filterIsInstance<ClassDescriptor>() + filterIsInstance<TypeAliasDescriptor>()).toList()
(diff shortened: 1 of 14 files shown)


Kordyjan:
FYI: If you add `fixes #834` in your initial PR message, that PR will be linked and then automatically marked as solved when this PR is merged.

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -364,7 +365,7 @@ data class DTypeAlias(
 
 sealed class Projection
 sealed class Bound : Projection()
-data class OtherParameter(val name: String) : Bound()
+data class OtherParameter(val dri: DRI, val declarationDRI: DRI, val name: String) : Bound()
```
Kordyjan, line 368: I'm not sure why we need both DRIs here.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -167,8 +161,66 @@ object StarProjection : TypeReference() {
...
+                is PropertyDescriptor -> PointingToProperty
+                else -> parameterDescriptor?.let { PointingToAdditionalParameters(params.indexOf(it)) }
+                    ?: PointingToOther(descriptor.name.toString())
+            }
+        }
+
+        fun from(psi: PsiElement): DriTarget = psi.parentsWithSelf.run {
+            val callable = firstIsInstanceOrNull<PsiMethod>()
+            val params = (callable?.parameterList?.parameters).orEmpty()
+            firstIsInstanceOrNull<PsiParameter>()?.let { params.indexOf(it) }
```
Kordyjan, line 189: This line isn't doing anything.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -167,8 +161,66 @@ object StarProjection : TypeReference() {
...
+                is ClassDescriptor, is FunctionDescriptor -> PointingToFunctionOrClasslike
+                is PropertyDescriptor -> PointingToProperty
+                else -> parameterDescriptor?.let { PointingToAdditionalParameters(params.indexOf(it)) }
+                    ?: PointingToOther(descriptor.name.toString())
+            }
+        }
+
+        fun from(psi: PsiElement): DriTarget = psi.parentsWithSelf.run {
+            val callable = firstIsInstanceOrNull<PsiMethod>()
+            val params = (callable?.parameterList?.parameters).orEmpty()
```
Kordyjan, line 188: same as above.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -167,8 +161,66 @@ object StarProjection : TypeReference() {
...
+    override fun toString(): String = "PointingToGenericParameters($parameterIndex)"
+}
+
+object PointingToFunctionOrClasslike: DriTarget()
+
+object PointingToReceiver: DriTarget()
+
+object PointingToProperty: DriTarget()
+
+data class PointingToAdditionalParameters(val parameterIndex: Int): DriTarget(){
```
Kordyjan, line 213: I'm not sure about the word `Additional`. It may turn out as misleading.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -167,8 +161,66 @@ object StarProjection : TypeReference() {
...
+            }
+        }
+    }
+}
+
+data class PointingToGenericParameters(val parameterIndex: Int) : DriTarget() {
+    override fun toString(): String = "PointingToGenericParameters($parameterIndex)"
+}
+
+object PointingToFunctionOrClasslike: DriTarget()
```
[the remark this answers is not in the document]
MarcinAman: Pointing to a declaration is not the best solution, since everything that is passed to a DRI constructing function is a declaration. 

I've split this to make it more readable from the perspective of somebody reading those DRIs.

core/src/main/kotlin/model/Documentable.kt
MarcinAman: Changed to having only the declaration

core/src/main/kotlin/links/DRI.kt
MarcinAman: `PointingToCallableParameters` looks a little bit better

PR #861, opened 2020-05-14
Add github actions CI

MarcinAman:


plugins/base/src/test/kotlin/basic/DRITest.kt
```diff
@@ -10,7 +10,7 @@ import org.junit.jupiter.api.Test
 
 class DRITest : AbstractCoreTest() {
     @Test
-    fun `#634`() {
+    fun issue634() {
```
sellophane, line 13: what was wrong with previous naming convention?

MarcinAman: it was failing to build on windows

PR #868, opened 2020-05-19
Tabs for sections v1

MarcinAman:


core/src/main/kotlin/pages/contentNodeProperties.kt
```diff
@@ -5,4 +5,8 @@ import org.jetbrains.dokka.model.properties.ExtraProperty
 class SimpleAttr(val extraKey: String, val extraValue: String) : ExtraProperty<ContentNode> {
     data class SimpleAttrKey(val key: String) : ExtraProperty.Key<ContentNode, SimpleAttr>
     override val key: ExtraProperty.Key<ContentNode, SimpleAttr> = SimpleAttrKey(extraKey)
+
+    companion object {
+        fun header(value: String) = SimpleAttr("data-togglable", value)
```
kamilok1965, line 10: This can just be an extension function, it'll be cleaner this way IMHO

MarcinAman: ok

PR #871, opened 2020-05-20
Refactor type aliases

BarkingBad:


BarkingBad:
Do not merge, as the DocumentableMerger current behaviour causes dokka-stdlib-build fail

PR #872, opened 2020-05-20
Tabs for sections - update tests

MarcinAman:


plugins/base/src/test/kotlin/renderers/html/DivergentTest.kt
```diff
@@ -320,11 +324,8 @@ class DivergentTest : RenderingOnlyTestBase() {
 
         HtmlRenderer(context).render(page)
         renderedContent.match(
-            "a-",
-            Div(Div(Div("a")), Div(Div("NATIVE"))),
-            "ab+b-",
-            Div(Div(Div("b")), Div(Div("NATIVE"))),
-            "ab+"
+            Div("a-", Div(Div(Div("a")), Div(Div("NATIVE"))), "ab+"),
```
MarcinAman, line 327: @kamilok1965 could you look at those lines? I think they are correct and previous tests were not wrapped in additional `div` so they merged into one like strings

kamilok1965: I agree, it looks better now

plugins/kotlin-as-java/src/test/kotlin/KotlinAsJavaPluginTest.kt
```diff
@@ -2,12 +2,12 @@ package kotlinAsJavaPlugin
 
 import org.jetbrains.dokka.pages.*
 import org.jetbrains.dokka.testApi.testRunner.AbstractCoreTest
+import org.jetbrains.kotlin.utils.addToStdlib.cast
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
```
kamilok1965, line 6: Unnecessary import

PR #877, opened 2020-05-21
Return types links in inner classes with generic parents

MarcinAman:
The issue was that given a class without a package:
```
class Sample<S>{
    inner class SampleInner {
        fun foo(): S = TODO()
    }
}
```

We would generate package name from a sourceRoot, but while creating a DRI for `foo`'s return type, the package would be empty. Then the location provider wouldn't be able to give a correct value so we end up with an empty `href` on frontend.

### DIFF
--- a/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
@@ -101,8 +101,7 @@ private class DokkaDescriptorVisitor(
         descriptor: PackageFragmentDescriptor,
         parent: DRIWithPlatformInfo
     ): DPackage {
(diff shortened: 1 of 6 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -640,4 +641,14 @@ private class DokkaDescriptorVisitor(
         Visibilities.PRIVATE -> KotlinVisibility.Private
         else -> KotlinVisibility.Public
     }
+
+    private fun fallbackPackageName(): String = "[${sourceSet.sourceSetName} root]"// TODO: error-prone, find a better way to do it
+}
+
+fun DRI.withPackageFallbackTo(fallbackPackage: String): DRI {
```
Kordyjan, line 648: Move it to DRI.kt or make it private if you want to treat it as a local utility, to avoid scope pollution.

PR #880, opened 2020-05-22
Trim windows carrage return sign in importing includes

MarcinAman:


plugins/base/src/test/kotlin/linkableContent/LinkableContentTest.kt
```diff
@@ -29,20 +30,20 @@ class LinkableContentTest : AbstractCoreTest() {
                     analysisPlatform = "js"
                     targets = listOf("js")
                     sourceRoots = listOf("jsMain", "commonMain", "jvmAndJsSecondCommonMain").map {
-                        "$testDataDir/$it/kotlin"
+                        Paths.get("$testDataDir/$it/kotlin").toString()
```
MarcinAman, line 33: Wrapping everything in `Paths.get` is not required but nice to have singe you are not debugging a path like `D:\somefolder/test/it/kotlin`

PR #890, opened 2020-05-27
Rework dokka configuration and Gradle plugin

kamildoleglo:
Probably needs more testing before merging


build.gradle.kts
```diff
@@ -39,7 +40,11 @@ allprojects {
 subprojects {
     apply {
         plugin("org.jetbrains.kotlin.jvm")
+        plugin("java")
     }
+
+    // Gradle metadata
+    extensions.findByType(JavaPluginExtension::class)?.targetCompatibility = JavaVersion.VERSION_1_8
```
sellmair, line 47: Maybe something like 
```kotlin
    java {
        targetCompatibility = JavaVersion.VERSION_1_8
    }
```

would be more idiomatic?

sellmair:
The current state of this PR broke some links in my sample project. @kamildoleglo , maybe lets have a look at this together?

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
```diff
@@ -57,63 +58,54 @@ open class DokkaTask : DefaultTask(), Configurable {
...
-            ClassloaderContainer.fatJarClassLoader = URLClassLoader(
+            ClassloaderContainer.coreClassLoader = URLClassLoader(
                 jars.map { it.toURI().toURL() }.toTypedArray(),
                 ClassLoader.getSystemClassLoader().parent
             )
         }
     }
 
-    private fun extractKotlinCompileTasks(collectTasks: () -> List<Any?>?): List<Task> {
+    protected fun extractKotlinCompileTasks(collectTasks: () -> List<Any?>?): List<Task> {
```
sellmair, line 103: Is this lambda still necessary, if the first thing this function does is evaluating it? 🙄

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
```diff
@@ -57,63 +58,54 @@ open class DokkaTask : DefaultTask(), Configurable {
...
-
-    var config: GradleDokkaConfigurationImpl? = null
-
-    // Configure Dokka with closure in Gradle Kotlin DSL
-    fun configuration(action: Action<in GradlePassConfigurationImpl>) = action.execute(configuration)
-
     private val kotlinTasks: List<Task> by lazy {
-        extractKotlinCompileTasks(
-            configuration.collectKotlinTasks ?: { defaultKotlinTasks() })
+        extractKotlinCompileTasks({
```
sellmair, line 76: For me: This is not suuuper easy to understand. Can we make this part a little bit easier?

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/utils.kt
```diff
@@ -16,4 +16,17 @@ fun Project.isAndroidProject() = try {
     false
 }
 
+fun Project.isNotMultiplatformProject() = !isMultiplatformProject()
```
sellmair, line 19: Functions like this should be internal IMO!

build.gradle.kts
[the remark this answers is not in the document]
kamildoleglo: Oh, yes. I don't know why I haven't thought of that, changed

core/src/main/kotlin/model/SourceSetData.kt
```diff
@@ -6,22 +6,33 @@ import org.jetbrains.dokka.plugability.DokkaContext
...
     private val sourceSets = HashMap<String, SourceSetData>()
 
     val allSourceSets: List<SourceSetData>
-    get() = sourceSets.values.toList()
+        get() = sourceSets.values.toList()
 
     fun getSourceSet(pass: DokkaConfiguration.PassConfiguration) =
-        sourceSets.getOrPut("${pass.moduleName}/${pass.sourceSetName}",
-            { SourceSetData(pass.moduleName, pass.sourceSetName, pass.analysisPlatform, pass.sourceRoots, pass.dependentSourceSets) }
+        sourceSets.getOrPut("${pass.moduleName}/${pass.sourceSetID}",
```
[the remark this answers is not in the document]
kamildoleglo: > Maybe we should give the string "moduleName/sourceSetID" a formal name, since it is used multiple times? 
Sure, what do you suggest? 

> Also, can't we enforce a true unique sourceSetID? In that case, this would be simpler, because moduleName would not be required.
I'm not quite sure how to do it right. We could theoretically generate a unique string, but then we would have to do some translation from user input for dependent source sets to those internal ids, which basically boils down to the same thing (name used as an identifier), unless we force users to use other sourceSets reference and I'm not sure if that's a good idea

plugins/base/src/main/kotlin/resolvers/local/DefaultLocationProvider.kt
```diff
@@ -98,26 +98,19 @@ open class DefaultLocationProvider(
...
-                        info
-                    )
+                    return link.url.toExternalForm() + getLink(dri, info)
                 }
             }
         }
         // Not in cache, resolve packageLists
         for ((jdk, links) in toResolve) {
             for (link in links) {
+                if(dokkaContext.configuration.offlineMode && link.packageListUrl.protocol.toLowerCase() != "file")
```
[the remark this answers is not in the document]
kamildoleglo: I doubt that unless you used the `offlineMode = true`. We shall check it though

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
kamildoleglo: Yes, because you can't pass it from Gradle otherwise

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
kamildoleglo: Gladly, I'm just not sure how to do it better, considering that the `collectKotlinTasks` have to be a nullable lambda to nullable list of nullable Any, because of Gradle's lifecycle

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/utils.kt
kamildoleglo: Agreed, fixed

build.gradle.kts
```diff
@@ -5,6 +5,7 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 plugins {
     kotlin("jvm") apply false
     id("com.jfrog.bintray") apply false
+    id("java")
```
Kordyjan, line 8: I don't think wee need to apply it to top level project as it is only an aggregate.

plugins/base/src/main/kotlin/resolvers/local/DefaultLocationProvider.kt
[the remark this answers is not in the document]
Kordyjan: Was that resolved?

build.gradle.kts
kamildoleglo: I think we do, otherwise we can't configure it in the `subprojects` block, but maybe I'm missing something

PR #921, opened 2020-06-03
Added persistence in window scope for filtering and tabs state.

pikinier20:
#917

### DIFF
--- a/plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
+++ b/plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
@@ -4,12 +4,12 @@ filteringContext = {
     activeFilters: []
 }
 window.addEventListener('load', () => {
-    initializeFiltering()
     document.querySelectorAll("div[data-platform-hinted]")
         .forEach(elem => elem.addEventListener('click', (event) => togglePlatformDependent(event,elem)))
     document.querySelectorAll("div[tabs-section]")
         .forEach(elem => elem.addEventListener('click', (event) => toggleSections(event)))
     document.getElementById('filter-section').addEventListener('click', (event) => filterButtonHandler(event))
+    initializeFiltering()
     initTabs()
 })
 
@@ -19,6 +19,14 @@ function initTabs(){
             showCorrespondingTabBody(element)
             element.addEventListener('click', (event) => toggleSections(event))
         })
+    let cached = localStorage.getItem("active-tab")
+    if (cached) {
+        let parsed = JSON.parse(cached)
+        let tab = document.querySelector('div[tabs-section] > button[data-togglable="' + parsed + '"]')
+        if(tab) {
+            tab.click()
+        }
+    }
 }
 
 function showCorrespondingTabBody(element){
@@ -37,8 +45,6 @@ function filterButtonHandler(event) {
                 unfilterSourceset(sourceset)
             }
         }
-        refreshFilterButtons();
-        refreshPlatformTabs()
 }
 
 function initializeFiltering() {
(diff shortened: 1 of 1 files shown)


plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
```diff
@@ -19,6 +19,11 @@ function initTabs(){
             showCorrespondingTabBody(element)
             element.addEventListener('click', (event) => toggleSections(event))
         })
+    let cached = localStorage.getItem("active-tab")
```
MarcinAman, line 22: Hmm that means that there is only one entry for all pages, so when i change tab on 1 page it will be the same on the others.

I don't think it should work like that. You can add a location to the end of the key to make it page-specific.

Same thing goes for the sourceset filters, but imho persisting them between pages i nice.

MarcinAman: Also you can just do `if(cached)` and why do you need to `JSON.parse` it when it is a string?

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
```diff
@@ -49,38 +52,58 @@ function initializeFiltering() {
...
-                f.setAttribute("data-active","")
-            } else {
-                f.removeAttribute("data-active")
-            }
-        })
+    let cached = window.localStorage.getItem('inactive-filters')
+    if (cached !== null) {
+        let parsed = JSON.parse(cached)
+        filteringContext.activeFilters = filteringContext.restrictedDependencies
+            .filter(q => -1 == parsed.indexOf(q))
```
MarcinAman, line 59: i find it weirdly disturbing to have a reversed equals XD

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
```diff
@@ -19,6 +19,11 @@ function initTabs(){
             showCorrespondingTabBody(element)
             element.addEventListener('click', (event) => toggleSections(event))
         })
+    let cached = localStorage.getItem("active-tab")
+    if (cached !== null) {
+        let parsed = JSON.parse(cached)
+        document.querySelector("div[tabs-section] > button[data-togglable=" + parsed + "]").click()
```
MarcinAman, line 25: Small thing but this will cause an error when on the first page we click on a tab and then move to another page that doesn't have this tab.

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
pikinier20: I have spoken with @Kordyjan about scope of cache and if I understood correctly, we decided to make filters cache to work between pages.

About comparison - I never know how to check nulls and undefined's correctly.

About JSON.parse - It's true, I didn't realized that it is just string and my default behavior is to call JSON.parse when retrieving data from localStorage.

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
pikinier20: Fixed

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
pikinier20: Fixed

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
pikinier20: Fixed null-checks

PR #936, opened 2020-06-04
Changed anchors handling to work properly with tabs

pikinier20:


plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
```diff
@@ -7,24 +7,47 @@ window.addEventListener('load', () => {
...
+        .forEach(elem => elem.addEventListener('click', (event) => toggleSectionsEventHandler(event)))
     document.getElementById('filter-section').addEventListener('click', (event) => filterButtonHandler(event))
     initializeFiltering()
     initTabs()
+    handleAnchor()
 })
 
+function handleAnchor() {
+    let searchForTab = function(element) {
+        if(element && element.hasAttribute) {
```
MarcinAman, line 19: what `element.hasAttribute` does in this case?
imho as it is an existing property it evaluates to true

plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
```diff
@@ -7,24 +7,47 @@ window.addEventListener('load', () => {
...
+            if(element.hasAttribute("data-togglable")) return element;
+            else return searchForTab(element.parentNode)
+        } else return null
+    }
+    let anchor = window.location.hash
+    if (anchor != "") {
```
MarcinAman, line 25: you can just do `if(anchor)` or i'd recommend to go for triple equals so: `!==` since it checks the type

PR #938, opened 2020-06-04
Make enum entry constructor sourceset dependent

MarcinAman:
Jun-04-2020 16-54-13

### DIFF
--- a/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
@@ -219,7 +219,7 @@ private class DokkaDescriptorVisitor(
             extra = PropertyContainer.withAll(
                 descriptor.additionalExtras(),
                 descriptor.getAnnotations(),
-                ConstructorValues(descriptor.getAppliedConstructorParameters())
+                ConstructorValues(descriptor.getAppliedConstructorParameters().toSourceSetDependent())
             )
         )
     }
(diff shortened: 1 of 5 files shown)


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -141,8 +141,10 @@ open class DefaultPageCreator(
     protected open fun contentForEnumEntry(e: DEnumEntry) = contentBuilder.contentFor(e) {
         group(kind = ContentKind.Cover) {
             header(1, e.name)
-            +buildSignature(e)
-            +contentForDescription(e)
+            sourceSetDependentHint(e.dri, e.sourceSets.toSet()){
```
kamildoleglo, line 144: Please run the formatter here

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -219,7 +219,7 @@ private class DokkaDescriptorVisitor(
             extra = PropertyContainer.withAll(
                 descriptor.additionalExtras(),
                 descriptor.getAnnotations(),
-                ConstructorValues(descriptor.getAppliedConstructorParameters())
+                ConstructorValues(mapOf(sourceSet to descriptor.getAppliedConstructorParameters()))
```
BarkingBad, line 222: Let's use the predefined util function to keep consistency `private fun <T> T.toSourceSetDependent()` defined above in the class body.

Source:
https://github.com/Kotlin/dokka/blob/dev-0.11.0/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt#L102

MarcinAman: Thanks, fixed

PR #939, opened 2020-06-05
Report undocumented code

### LINKED ISSUE #853: Support for `reportUndocumented` in `perPackageOptions` and toplevel
What do we want to accomplish?
 We want to have every declaration in code that has empty documentation logged in the console with the fully qualified name of the declaration. 

* implement a DocumentableTransformer for reporting undocumented declarations 
* undocumented packages and modules should not be reported (only classes, functions and so on) 
* parameters should not be reported 

optional: 
* have different levels of reporting undocumented declarations (like full for all parameters and packages, regular only for classes and functions, ...)

sellmair:
Closes #853


plugins/base/src/main/kotlin/transformers/documentables/ReportUndocumentedTransformer.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+internal class ReportUndocumentedTransformer : DocumentableTransformer {
+
+    override fun invoke(original: DModule, context: DokkaContext): DModule = original.apply {
+        flatTree().forEach { documentable -> invoke(documentable, context) }
+    }
+
+    private fun invoke(documentable: Documentable, context: DokkaContext) {
+        documentable.sourceSets.forEach { sourceSet ->
```
sellmair, line 19: @Kordyjan I think this was trick: Inspect the documentable in the perspective of the different source sets. Behaves exactly the way I would expect now. Thanks for your help 🤜 🤛

core/src/main/kotlin/configuration.kt
```diff
@@ -83,7 +83,7 @@ interface DokkaConfiguration {
     interface PackageOptions {
         val prefix: String
         val includeNonPublic: Boolean
-        val reportUndocumented: Boolean
+        val reportUndocumented: Boolean?
```
BarkingBad, line 86: What's the point of having three-states flag instead of binary option?

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -46,12 +41,13 @@ import org.jetbrains.kotlin.resolve.source.PsiSourceElement
 import org.jetbrains.kotlin.types.DynamicType
 import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.TypeProjection
-import org.jetbrains.kotlin.utils.addToStdlib.firstIsInstance
 import org.jetbrains.kotlin.utils.addToStdlib.firstIsInstanceOrNull
 import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 import java.nio.file.Paths
-import kotlin.IllegalArgumentException
-import kotlin.reflect.jvm.internal.impl.resolve.constants.KClassValue
+import org.jetbrains.kotlin.resolve.constants.AnnotationValue as ConstantsAnnotationValue
```
BarkingBad, line 47: What's the point of shuffling imports?

core/src/main/kotlin/configuration.kt
sellmair: Package options have to override given "global" option from the PassConfiguration. So if it is "not set" (third option), then we have respect only what the PassConfiguration says.

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
sellmair: No intention. Probably just hit auto-format inside the IDE. Will cleanup all unnecessary changes like this before removing the "Draft" status!

core/src/main/kotlin/configuration.kt
BarkingBad: Make sense

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
BarkingBad: No problem, I was just curious

README.md
```diff
@@ -604,7 +604,7 @@ Dokka supports the following command line arguments:
     * `-skipDeprecated` - if set, deprecated elements are not included in the generated documentation
     * `-reportUndocumented` - warn about undocumented members
     * `-skipEmptyPackages` - do not create index pages for empty packages
-    * `-packageOptions` - list of package options in format `prefix,-deprecated,-privateApi,+warnUndocumented;prefix, ...`
```
sellmair, line 607: Personally, I think "warnUndocumented" was also pretty fine. Just wanted to align this with the actual option.

core/src/main/kotlin/links/DRI.kt
```diff
@@ -169,11 +179,12 @@ sealed class DriTarget {
 
     companion object {
         fun from(descriptor: DeclarationDescriptor): DriTarget = descriptor.parentsWithSelf.run {
-            return when(descriptor){
+            return when (descriptor) {
```
sellmair, line 182: Probably hit the auto-format inside the IDE. Should I revert?

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -378,6 +378,21 @@ fun Documentable.dfs(predicate: (Documentable) -> Boolean): Documentable? =
         this.children.asSequence().mapNotNull { it.dfs(predicate) }.firstOrNull()
     }
 
+operator fun Documentable.iterator(): Iterator<Documentable> {
+    return flatTree().iterator()
+}
+
+fun Documentable.flatTree(): List<Documentable> {
```
sellmair, line 385: @Kordyjan .
I am not sure if the team likes something like this as public API?

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
```diff
@@ -4,5 +4,5 @@ import org.jetbrains.dokka.model.DModule
 import org.jetbrains.dokka.plugability.DokkaContext
 
 interface PreMergeDocumentableTransformer {
-    operator fun invoke(modules: List<DModule>): List<DModule>
+    operator fun invoke(modules: List<DModule>, context: DokkaContext): List<DModule>
```
sellmair, line 7: This change was necessary while I tried implementing the reporting mechanism using a `PreMergeDocumentableTransformer`. I think this might make sense for the public API. Should I revert?

plugins/base/src/main/kotlin/translators/psi/JavadocParser.kt
```diff
@@ -43,6 +44,40 @@ class JavadocParser(
         return DocumentationNode(nodes)
     }
 
+    private fun findClosestDocComment(element: PsiNamedElement): PsiDocComment? {
```
sellmair, line 47: @sellophane 
This turned out to be not as simple as we might have hoped. I am here reporting a potential conflict in documentation when there is a Interface and a base class that would provide proper documentation for the same method.

plugins/base/src/test/kotlin/transformers/ReportUndocumentedTransformerTest.kt
```diff
@@ -0,0 +1,832 @@
...
+            """.trimIndent(),
+            configuration
+        ) {
+            pagesTransformationStage = {
+                assertNoUndocumentedReport()
+            }
+        }
+    }
+
+    @Disabled
```
sellmair, line 153: I am not sure if a secondary constructor should *always* be documented by default 🤔

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
BarkingBad: Actually you can pass context/logger to any extension point, including `PreMergeDocumentableTransformer` while registering. Check out how it's done with `ModuleAndPackageDocumentationTransformer` which previously had been provided with context via constructor.

plugins/base/src/main/kotlin/transformers/documentables/ModuleAndPackageDocumentationTransformer.kt
```diff
@@ -14,7 +14,7 @@ import java.nio.file.Paths
 
 internal class ModuleAndPackageDocumentationTransformer(val context: DokkaContext) : PreMergeDocumentableTransformer {
 
-    override fun invoke(original: List<DModule>): List<DModule> {
+    override fun invoke(modules: List<DModule>, context: DokkaContext): List<DModule> {
```
BarkingBad, line 17: Another context is redundant here, as it is passed through constructor

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilter.kt
```diff
@@ -10,7 +10,7 @@ import org.jetbrains.dokka.transformers.documentation.PreMergeDocumentableTransf
 
 internal class DocumentableVisibilityFilter(val context: DokkaContext) : PreMergeDocumentableTransformer {
 
-    override fun invoke(modules: List<DModule>): List<DModule> = modules.map { original ->
+    override fun invoke(modules: List<DModule>, context: DokkaContext): List<DModule> = modules.map { original ->
```
BarkingBad, line 13: Again redundant context

plugins/base/src/main/kotlin/translators/psi/JavadocParser.kt
sellophane: I did some testing and it seems it returns only the most immediate ancestor - so if there is a structure like:
classA <- classB <- classC <- InterfaceC1, only classB method will be returned. For this, much simpler recursive function should be sufficient. 
In what scenario you encountered yours behaviour?

plugins/base/src/main/kotlin/transformers/documentables/ReportUndocumentedTransformer.kt
sellmair: Reverted this change. I changed this together with @kamildoleglo, because we thought this is consistent. See slack for further details ☺️

core/src/main/kotlin/transformers/documentation/PreMergeDocumentableTransformer.kt
sellmair: See comment about: changes reverted ☺️

plugins/base/src/main/kotlin/translators/psi/JavadocParser.kt
sellmair: I also cover the following hierarchies:

1.

```kotlin
BaseClass {
    /** Documentation */
    fun x() 
}

InterfaceX {
    /** Other Documentation */
    fun x()
}

class X: BaseClass(), InterfaceX {
    fun x() // <-- which documentation to pick now? Interface or class?
}
```

2.

```kotlin

class BaseClass1 {
    /** Documented */
    fun x() 
}

class BaseClass2: BaseClass1() {
    fun x() // <- Not explicitly documented
}

class X : BaseClass2() {
    fun x() // <- Not explicitly documented. Implicitly by BaseClass1
}

```

sellmair: The warning, as well as the convention class > interface was discussed with @Kordyjan

plugins/base/src/main/kotlin/transformers/documentables/ModuleAndPackageDocumentationTransformer.kt
sellmair: See discussion above ☺️

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilter.kt
sellmair: See discussion above ☺️

core/src/main/kotlin/model/Documentable.kt
Kordyjan: I'm totally ok with `flatTree` extension. I would only prefer a slightly different name for it, as it is not returning tree. `.withDescendats()` is what I would choose instead.
On the other hand, I'm not a fan of implementing the `iterator` operator on documentables. This allows for code like `for (d in someDocumentable) {...}` which may be confusing and is not much shorter than `for (d in someDocumentable.withDescendants()) {...}`.
Also, returning this as a lazily evaluated sequence instead of a list may be beneficial in some cases. Something like that is done for [pages](https://github.com/Kotlin/dokka/blob/77c8777b7f66bddd374d68decd507547d356d602/core/src/main/kotlin/pages/PageNodes.kt#L160), but I think it will be slightly more performant if we implement that as a recursive method having `SequenceScope` as a receiver, instead of sequence nesting.

plugins/base/src/test/kotlin/transformers/ReportUndocumentedTransformerTest.kt
```diff
@@ -0,0 +1,832 @@
+package transformers
```
Kordyjan, line 1: Those are some really well-done tests. 👍

core/src/main/kotlin/model/Documentable.kt
sellmair: "withDescendats" 👌 Very nice! Thank you ☺️ 

Regarding the "iterator" API: Very good argument. I removed it ✅ 

Now to the "sequence" API: I do not think this will be always faster. For smaller trees, I would bet that just creating that list is much faster. However I am uncertain which dimensions this will have. So I switched to use the Sequence API. ✅

plugins/base/src/test/kotlin/transformers/ReportUndocumentedTransformerTest.kt
sellmair: Thanks a lot 😌 ☺️

PR #955, opened 2020-06-09
Option to fail on warning

### LINKED ISSUE #459: Option to fail on warning
It would be useful to have an option to fail the build when warnings are found so this can be enforced by CI builds.

sellmair:
Closes #459


runners/cli/src/main/kotlin/cli/main.kt
```diff
@@ -48,6 +48,10 @@ open class GlobalArguments(parser: DokkaArgumentsParser) : DokkaConfiguration {
 
     override val impliedPlatforms: List<String> = emptyList()
 
+    override val failOnWarning: Boolean by parser.singleFlag(
+        listOf("-failOnWarning"), "Fail dokka task if at least one warning was reported"
```
sellmair, line 52: Alternative: "-warningPolicy=strict"

PR #959, opened 2020-06-10
Update Gradle version to 6.5

sellmair:


kamildoleglo:
Yeah, certainly. @sellmair if you could just remove all unneccessary files for now and we'll merge this

sellmair:
@kamildoleglo Done! We should really start ignoring this files 🙈

PR #965, opened 2020-06-11
First batch of fixes for javadoc

sellophane:


plugins/base/.gitignore
```diff
@@ -0,0 +1,2 @@
+src/main/resources/dokka/scripts/main.js
```
sellmair, line 1: It is still getting pushed. You might also want to remove the file from git 🙄

plugins/base/src/main/kotlin/transformers/documentables/ModuleAndPackageDocumentationTransformer.kt
```diff
@@ -73,8 +73,7 @@ internal class ModuleAndPackageDocumentationTransformer(val context: DokkaContex
                     val doc = modulesAndPackagesDocumentation[Pair(module.name, pd)]
                     val facade = context.platforms[pd]?.facade
                         ?: return@mapNotNull null.also { context.logger.warn("Could not find platform data for ${pd.moduleName}/${pd.sourceSetName}") }
-                    val descriptor = facade.resolveSession.getPackageFragment(FqName(it.name))
-                        ?: return@mapNotNull null.also { _ -> context.logger.warn("Could not find descriptor for ${it.name}") }
+                    val descriptor = facade.moduleDescriptor.getPackage(FqName(it.name.let { if(it == "[JS root]") "" else it }))
```
sellmair, line 76: This class is already very, very hard to understand intuitively. This change just adds on top. 
Could you extract the idea of this `it == "[JS root]` logic into some function with a name, or add some kind of comment? 🤔

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
```diff
@@ -46,12 +51,18 @@ open class JavadocPageCreator(
         ) {
             title(m.name, "0.0.1", dri = setOf(m.dri), kind = ContentKind.Main)
             list("Packages", "Package", setOf(m.dri), ContentKind.Packages, m.packages.map { p ->
-                val doc = p.documentation.entries.find { (k, _) -> k.platform == Platform.jvm }?.value?.let {
-                    it.children.joinToString("\n") { it.root.toString() }
+                val description = p.documentation.entries.find { (k, _) -> k.platform == Platform.jvm }?.value?.let {
```
sellmair, line 54: This is not intuitively easy for me to understand (because of me lacking some contextual knowledge here). Would you mind explaining this part to me?

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
```diff
@@ -64,12 +75,9 @@ open class JavadocPageCreator(
         ) {
             title(p.name, "0.0.1", dri = setOf(p.dri), kind = ContentKind.Packages)
             list("Packages", "Package", setOf(p.dri), ContentKind.Packages, p.classlikes.map { c ->
-                val doc = c.documentation.entries.find { (k, _) -> k.platform == Platform.jvm }?.value?.let {
-                    it.children.joinToString("\n") { it.root.toString() }
-                }.orEmpty()
                 RowJavadocListEntry(
                     LinkJavadocListEntry(c.name.orEmpty(), setOf(c.dri), JavadocContentKind.Class, sourceSets),
-                    doc
+                    listOf(signatureProvider.signature(c))
```
sellmair, line 80: Same as above: Could you explain this to me?

PR #969, opened 2020-06-12
Disable "reportUndocumented" by default

sellmair:


sellmair:
@sellophane I agree, but this is not the only flag that works like this. Should we change the default for this flag with this PR and change the mechanism for all settings in another PR?

sellmair:
Closed for now because of #975

PR #972, opened 2020-06-15
Rebundle the compiler to make possible to add kotlinx.coroutines to dependencies

Kordyjan:
Making one completely independent change in the codebase, I wanted to use flow api from `kotlinx.coroutines`. I was sure that we have a dependency on it defined somewhere, as we are using coroutines extensively in renderers. Then I realized that some newer functions were missing and some others which are definitely stable were marked as experimental. This suggested that we were using some older coroutines version and it needs updating. I went through all of the modules and was not able to find any place where we specify the coroutines version. After further investigation, it turned out that the old version of coroutines is bundled together with a kotlin compiler artifact, which is a transitive dependency of every Dokka plugin through the core module.

This is serious, as it means that no matter what version of `kotlinx.coroutines` is specified by plugin creator, due to how gradle sorts classpath, the old version bundled with compiler will be always preferred. As it affects all plugins, I tried to fix it asap.

After a few hours of different approaches and reading sources of gradle, I realized that it is not possible to provide any solution that would be transparent for plugin creators and would assure that the correct version is always preferred both during compilation and runtime.

I really don't like the idea of bundling such big dependency as kotlin compiler, but here it was necessary.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -8,6 +8,9 @@ plugins {
 dependencies {
     api(project(":coreDependencies", configuration = "shadow"))
 
+    val coroutines_version: String by project
+    api("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version")
+
     val kotlin_version: String by project
     implementation("org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version")
     implementation("org.jetbrains.kotlin:kotlin-reflect:$kotlin_version")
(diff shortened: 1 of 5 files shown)


sellmair:
I think this could be closed until we can verify whether or not Kotlin 1.4 M3 resolves our issue, right?

Kordyjan:
It will be no longer needed after update to 1.4-M3 🎉

PR #973, opened 2020-06-15
Update to Kotlin 1.4 M3

sellmair:


Kordyjan:
We need to remember to also update `kotlin-plugin-version` as soon as M3 for it is published on bintray.

PR #975, opened 2020-06-16
Implement default configurations

sellmair:


core/src/main/kotlin/configuration.kt
```diff
@@ -3,6 +3,25 @@ package org.jetbrains.dokka
 import java.io.File
 import java.net.URL
 
+object DokkaDefaults {
```
sellmair, line 6: Probable questions:

____ 
_Why not splitting defaults for `DokkaConfiguration`, `PassConfiguration` and `packageConfiguration` into its own defaults?_

That was what I actually did first, because it seems intuitive. However, this leads to potential inconsistencies, minor maintenance overhead and overall confusion inside the `DokkaMojo` class. 

____

_Why not calling it Dokka**Configuration**Defaults or putting DokkaConfiguration.Defaults_

This is to avoid the potential misconception that this Defaults are just used for types implementing `DokkaConfiguration`

core/src/main/kotlin/DokkaBootstrapImpl.kt
```diff
@@ -11,12 +11,26 @@ import java.util.function.BiConsumer
...
+        )
+
     val args = it.subList(1, it.size)
-    val deprecated = args.find { it.endsWith("deprecated") }?.startsWith("+") ?: true
-    val reportUndocumented = args.find { it.endsWith("reportUndocumented") }?.startsWith("+") ?: true
-    val privateApi = args.find { it.endsWith("privateApi") }?.startsWith("+") ?: false
-    val suppress = args.find { it.endsWith("suppress") }?.startsWith("+") ?: false
+
+    val deprecated = args.find { it.endsWith("deprecated") }?.startsWith("+")
+        ?: args.find { it.endsWith("skipDeprecated") }?.startsWith("+")
```
kamildoleglo, line 21: Since we're changing the configuration quite extensively I think we can just drop the old argument names like "deprecated" and "privateApi"

sellmair: ✅ Removed support for those "old" argument names with additional commit

Issue #971, opened 2020-06-13
Functional interfaces introduced in Kotlin 1.4 are reported in a wrong way


alexey-anufriev:
**Describe the bug**
Dokka generates documentation for functional interfaces with the wrong format.

For functional interface like this:
```
fun interface Test {
    fun test()
}
```

Dokka will generate the following line:
```
fun <no name provided>(): Unit
```

**Expected behaviour**
Ideally, Dokka should know that this is not a function but rather interface and generate documentation properly.

**To Reproduce**
1. Define functional interface
2. Generate documentation

**Dokka configuration**
Default

**Installation**
- Operating system: Ubuntu 20.04
- Build tool: Gradle v5.5.1
- Dokka version: 0.10.1

Kordyjan:
Released versions of Dokka officially do not support kotlin 1.4. We added 1.4 support in 0.11. I'll leave this issue for checking if there are no problems caused by `fun interface`s in 0.11.

PR #981, opened 2020-06-17
KotlinSignatureProvider: Render 'fun interface' into signature

sellmair:
#971


plugins/base/src/main/kotlin/translators/documentables/PageContentBuilder.kt
```diff
@@ -353,11 +353,11 @@ open class PageContentBuilder(
         ) =
             ContentText(text, DCI(mainDRI, kind), sourceSets, styles, extra)
 
-        fun <T> platformText(
```
sellmair, line 356: Please tell me, if you want me to change this in another commit/PR ☺️

PR #993, opened 2020-06-22
Separate "Content" from "TemplateMap" and move creation of "TemplateM…

sellmair:


plugins/javadoc/src/main/kotlin/javadoc/KorteJavadocRenderer.kt
```diff
@@ -206,93 +204,167 @@ class KorteJavadocRenderer(val outputWriter: OutputWriter, val context: DokkaCon
...
+
+    private fun renderTitleNode(node: TitleNode): TemplateMap {
+        return mapOf(
+            "title" to node.title,
+            "version" to node.version,
+            "packageName" to node.parent
+        )
+    }
+
+    private fun renderJavadocContentGroup(note: JavadocContentGroup): TemplateMap {
```
MarcinAman, line 351: node

plugins/javadoc/src/main/kotlin/javadoc/KorteJavadocRenderer.kt
```diff
@@ -206,93 +204,167 @@ class KorteJavadocRenderer(val outputWriter: OutputWriter, val context: DokkaCon
...
+        return mapOf("text" to node.text)
+    }
+
+    private fun renderListNode(node: ListNode): TemplateMap {
+        return mapOf(
+            "tabTitle" to node.tabTitle,
+            "colTitle" to node.colTitle,
+            "list" to node.children
+        )
+    }
```
kamildoleglo, line 367: I guess those can be written with `=` instead of `: Type { return ... }` ( `=` looks better IMHO)

sellophane:
I merged this commit into f/new_javadoc with dev-0.11.0 rebase

PR #997, opened 2020-06-22
DokkaTask: Replace exitProcess with exception and helpful message if …

sellmair:
…no source sets are configured


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
```diff
@@ -143,7 +143,7 @@ open class DokkaTask : DefaultTask(), Configurable {
         }
 
     @TaskAction
-    fun generate() = config?.let { generate(it) } ?: getConfiguration()?.let { generate(it) } ?: exitProcess(0)
+    fun generate() = config?.let { generate(it) } ?: generate(getConfiguration())
```
kamildoleglo, line 146: I think it would be best to leave it this way or take the exception into account in the `DokkaCollector` task, as `DokkaCollector` maps over dokka tasks and calls `getConfiguration()` on them. This way it will also fail when there are no source sets to document. Or maybe the fail is the desired behaviour there? cc @Kordyjan

PR #1007, opened 2020-06-23
Dokka core: remove runtime dependency to kotlin compiler

sellmair:


sellmair:
Closed because of #1008

PR #1014, opened 2020-06-25
Rendering information about extensions

Kordyjan:


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -181,9 +182,10 @@ open class DefaultPageCreator(
     }
 
     protected open fun contentForClasslike(c: DClasslike) = contentBuilder.contentFor(c) {
+        val sourceSets = c.sourceSets.toSet()
```
kamildoleglo, line 185: I think it's a Set now, so `toSet()` shouldn't be necessary but please check

PR #1040, opened 2020-06-29
Merger changes, javadoc minor fixes, constructor names change

pikinier20:
#979


plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
```diff
@@ -43,7 +45,16 @@ private fun <T> mergeExpectActual(
 
     fun analyzeExpectActual(sameDriElements: List<T>) = sameDriElements.reduce(reducer)
 
-    return elements.groupBy { it.dri }.values.map(::analyzeExpectActual)
+    fun T.isExpectActual() : Boolean =
+        this.safeAs<WithExtraProperties<T>>().let { it != null && it.extra[IsExpectActual] != null }
+
+    return elements.groupBy {
+        it.isExpectActual()
```
MarcinAman, line 52: Imho it won't work because of homo-hetero problem, we need to know if they are related in the expect-actual. This will merge cases when 2 classes are actual to different expect.

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
```diff
@@ -43,7 +45,16 @@ private fun <T> mergeExpectActual(
...
+    }.let {
+        it[false].orEmpty() +
```
MarcinAman, line 54: I love `it[true]` and `it[false]` but others won't appreciate its genius

I'd suggest using `partition` to split the list and then `map`

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -478,39 +478,42 @@ open class DefaultPageCreator(
...
-                            elements.map {
-                                instance(setOf(it.dri), it.sourceSets.toSet()) {
-                                    before {
-                                        contentForBrief(it)
-                                        contentForSinceKotlin(it)
-                                    }
-                                    divergent {
-                                        group {
-                                            +buildSignature(it)
+                    .map { (typeNamePair, elements) -> // This groupBy should probably use LocationProvider
```
MarcinAman, line 484: this looks awkward when there is no groupBy in the same line

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
pikinier20: i think when you define 2 expects in non-dependent sourcesets and then you have sourceset which depends on both of them, you cannot create actual because of name-clash, compiler won't let it go (how would he know to which expect you declare actual)

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
pikinier20: ohhh partition is what I wanted to have there, didnt know it exists... thanks

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
pikinier20: lul

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
pikinier20: Okey, I got it. You can have like: JSCommon -> JS, and JVMCommon -> JVM, and in this situation you can have expect/actual with the same name which you don't want to merge.

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
```diff
@@ -1,36 +1,78 @@
...
 import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.transformers.documentation.DocumentableMerger
 import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 
 internal object DefaultDocumentableMerger : DocumentableMerger {
+    var dependencyInfo: Map<DokkaConfiguration.DokkaSourceSet, List<DokkaConfiguration.DokkaSourceSet>>? = null //Sorry for that hack
 
     override fun invoke(modules: Collection<DModule>, context: DokkaContext): DModule {
+        this.dependencyInfo = context.getDependencyInfo()
```
pikinier20, line 16: This is hack that needs to be refactored, functions that are used to merge, are not in scope of object so you cannot easily reference to dokka context from these functions.

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
```diff
@@ -1,36 +1,78 @@
...
+private fun DokkaContext.getDependencyInfo()
+        : Map<DokkaConfiguration.DokkaSourceSet, List<DokkaConfiguration.DokkaSourceSet>> {
+
+        fun getDependencies(sourceSet: DokkaConfiguration.DokkaSourceSet) : List<DokkaConfiguration.DokkaSourceSet> =
+            listOf(sourceSet) + configuration.sourceSets.filter {
+                it.sourceSetID in sourceSet.dependentSourceSets
+            }.flatMap { getDependencies(it) }
+
+        return configuration.sourceSets.map { it to getDependencies(it) }.toMap()
+}
```
pikinier20, line 74: I don't know if this function should be here, or somewhere inside Dokka Configuration as it might be useful function in some cases. It gives you a map with dependencies (direct and transitive).

plugins/base/src/main/kotlin/transformers/documentables/DefaultDocumentableMerger.kt
```diff
@@ -41,9 +83,24 @@ private fun <T> mergeExpectActual(
     reducer: (T, T) -> T
 ): List<T> where T : Documentable, T : WithExpectActual {
 
-    fun analyzeExpectActual(sameDriElements: List<T>) = sameDriElements.reduce(reducer)
+    fun analyzeExpectActual(sameDriElements: List<T>) = sameDriElements
+        .partition { it.expectPresentInSet != null }
+        .let {
+            (expects, actuals) -> expects.map { expect ->
+                listOf(expect) + actuals.filter { actual -> DefaultDocumentableMerger.dependencyInfo?.get(actual.sourceSets.single())?.contains(expect.expectPresentInSet!!) ?: false }
```
pikinier20, line 90: It's use of hack that I commented on top of file. What is going here is just merging to list expect with all actuals that depend on sourceset in which expect is declared.

BarkingBad:
Now it is in #1177

PR #1041, opened 2020-06-29
Javadoc signature provider

### LINKED ISSUE #1025: Javadoc should have a separate signature provider
Javadoc should have a signature provider that outputs a signature that is just the same as in javadoc format (so it is split between the modifiers with return type and the rest).

To do that we should reuse kotlin as java converter.

MarcinAman:
closes #1025


MarcinAman:
Before merging in need to create a sanitizer for content, since some signatures are treated by korte as html tags ;-;

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -317,10 +317,19 @@ class DefaultPsiToDocumentableTranslator(
                     is PsiClassReferenceType -> {
                         val resolved: PsiClass = type.resolve()
                             ?: return UnresolvedBound(type.presentableText)
-                        if (resolved.qualifiedName == "java.lang.Object") {
-                            JavaObject
-                        } else {
-                            TypeConstructor(DRI.from(resolved), type.parameters.map { getProjection(it) })
+                        when {
+                            resolved.qualifiedName == "java.lang.Object" -> {
```
sellophane, line 321: We don't need brackets here, each time we return a simple instance

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
```diff
@@ -189,5 +176,11 @@ open class JavadocPageCreator(
         }
         return contents
     }
+
+    private fun DParameter.brief(sourceSet: DokkaSourceSet? = highestJvmSourceSet): List<ContentNode> =
+        briefFromContentNodes(paramsToContentNodes(sourceSet).dropWhile { it is ContentDRILink })
+
+    private fun ContentNode.asJavadocNode(): JavadocSignatureContentNode =
+        (this as ContentGroup).children.firstOrNull() as JavadocSignatureContentNode
```
sellophane, line 184: With `firstOrNull` we should return `JavadocSignatureContentNode?`

plugins/javadoc/src/main/kotlin/javadoc/JavadocPlugin.kt
```diff
@@ -31,13 +34,24 @@ class JavadocPlugin : DokkaPlugin() {
                 dokkaBasePlugin.querySingle { signatureProvider },
                 context.logger
             )
-        }
+        } applyIf { format == "javadoc" }
```
sellophane, line 37: please make desired format a const val String if we use it more than once

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -317,10 +317,19 @@ class DefaultPsiToDocumentableTranslator(
...
+                                JavaObject
+                            }
+                            resolved is PsiTypeParameter -> {
+                                OtherParameter(
+                                    declarationDRI = DRI.from(resolved.owner!!),
```
sellophane, line 326: Must we use `!!` here?

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
MarcinAman: ok

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
MarcinAman: Imho no because it will result in a lot of clutter in page creator and `JavadocPageCreator` will always be used with javadoc signature provider so the output will stay the same. Also if it isn't we will get an exception that something went really wrong. 

I've changed the exception to be more informative

plugins/javadoc/src/main/kotlin/javadoc/JavadocPlugin.kt
MarcinAman: ok

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
MarcinAman: i've included that null in the if statement, so it wont produce null pointer

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
```diff
@@ -43,20 +42,20 @@ open class JavadocPageCreator(
                 name = c.name.orEmpty(),
                 content = contentForClasslike(c),
                 dri = setOf(c.dri),
-                modifiers = listOfNotNull(c.visibility[jvm]?.name),
-                signature = signatureProvider.signature(c).nodeForJvm(jvm),
+                signature = signatureProvider.signature(c).nodeForJvm(jvm).asJavadocNode(),
```
kamildoleglo, line 45: Can we extract this call chain `signatureProvider.signature(c).nodeForJvm(jvm).asJavadocNode()` somewhere?

plugins/javadoc/src/main/kotlin/javadoc/renderer/JavadocContentToHtmlTranslator.kt
```diff
@@ -35,6 +35,19 @@ internal class JavadocContentToHtmlTranslator(
         }
     }.joinToString("<br>", """<span class="code">""", "</span>") { it }
 
+    private fun sanitize(content: String): String {
+        val replacements = listOf(">" to "&gt;", "<" to "&lt;")
```
kamildoleglo, line 39: We do have a helper in core that does exactly that I believe

plugins/javadoc/src/main/kotlin/javadoc/renderer/JavadocContentToTemplateMapTranslator.kt
```diff
@@ -73,45 +71,49 @@ internal class JavadocContentToTemplateMapTranslator(
...
-            val (modifiers, signature) = node.modifiersAndSignature
+        fun templateMapForFunctionNode(node: JavadocFunctionNode): TemplateMap {
             return mapOf(
-                "signature" to htmlForContentNode(node.signature, contextNode),
                 "brief" to htmlForContentNodes(node.brief, contextNode),
                 "parameters" to node.parameters.map { templateMapForParameterNode(it) },
-                "inlineParameters" to node.parameters.joinToString { "${it.type} ${it.name}" },
-                "modifiers" to htmlForContentNode(modifiers, contextNode),
-                "signatureWithoutModifiers" to htmlForContentNode(signature, contextNode),
+                "inlineParameters" to node.parameters.joinToString { "${htmlForContentNode(it.type, contextNode)} ${it.name}" },
```
kamildoleglo, line 84: I'd personally remove the string interpolation for the function call here as it looks ugly

plugins/kotlin-as-java/src/main/kotlin/KotlinAsJavaPlugin.kt
```diff
@@ -18,6 +18,6 @@ class KotlinAsJavaPlugin : DokkaPlugin() {
                     dokkaBasePlugin.commentsToContentConverter
                 ), ctx.logger
             )
-        }
+        } applyIf { format != "javadoc" }
```
kamildoleglo, line 21: I'd suggest `format == "kotlin-as-java"`

plugins/javadoc/src/main/kotlin/javadoc/renderer/JavadocContentToTemplateMapTranslator.kt
MarcinAman: done

plugins/javadoc/src/main/kotlin/javadoc/renderer/JavadocContentToHtmlTranslator.kt
MarcinAman: yup, removed

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
MarcinAman: extracted

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
```diff
@@ -189,5 +176,14 @@ open class JavadocPageCreator(
         }
         return contents
     }
+
+    private fun DParameter.brief(sourceSet: DokkaSourceSet? = highestJvmSourceSet): List<ContentNode> =
+        briefFromContentNodes(paramsToContentNodes(sourceSet).dropWhile { it is ContentDRILink })
+
+    private fun ContentNode.asJavadocNode(): JavadocSignatureContentNode =
+        (this as ContentGroup).children.firstOrNull() as? JavadocSignatureContentNode ?: throw IllegalStateException("JavadocPageCreator should be used only with JavadocSignatureProvider")
```
sellophane, line 184: Hint: I would rather go with sth like "No content for javadoc signature found"

Also, please, make `ContentComposite` implement `WithChildren<ContentNode>`, so that you can use 
`(this as ContentGroup).firstChildOfTypeOrNull< JavadocSignatureContentNode >()` here

plugins/javadoc/src/main/kotlin/javadoc/JavadocPlugin.kt
```diff
@@ -3,14 +3,18 @@ package org.jetbrains.dokka.javadoc
...
 import org.jetbrains.dokka.base.DokkaBase
+import org.jetbrains.dokka.base.resolvers.external.JavadocExternalLocationProviderFactory
+import org.jetbrains.dokka.kotlinAsJava.signatures.JavaSignatureProvider
 import org.jetbrains.dokka.plugability.DokkaPlugin
 import org.jetbrains.dokka.plugability.querySingle
 
 class JavadocPlugin : DokkaPlugin() {
 
     val dokkaBasePlugin by lazy { plugin<DokkaBase>() }
+    val javadocFormat = "javadoc"
```
sellophane, line 17: I would personally go with making it private companion object val, as this is an important static value, but thats just me

plugins/javadoc/src/main/kotlin/javadoc/renderer/JavadocContentToTemplateMapTranslator.kt
```diff
@@ -206,6 +203,9 @@ internal class JavadocContentToTemplateMapTranslator(
             )
         }
 
+        private fun renderInlineParameter(parameter: JavadocParameterNode): String =
+            "${htmlForContentNode(parameter.type, contextNode)} ${parameter.name}"
```
kamildoleglo, line 207: I was thinking more of a `htmlForContentNode(parameter.type, contextNode) + " ${parameter.name}"` instead but maybe that's just me

plugins/javadoc/src/main/kotlin/javadoc/JavadocPageCreator.kt
MarcinAman: done x2

plugins/javadoc/src/main/kotlin/javadoc/JavadocPlugin.kt
MarcinAman: makes sense :)

plugins/kotlin-as-java/src/main/kotlin/KotlinAsJavaPlugin.kt
MarcinAman: Paweł's changes were merged so i've used the override mechanism

PR #1046, opened 2020-06-30
Support dependent source sets in gradle by passing DokkaSourceSet object references instead of strings

BarkingBad:


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
```diff
@@ -193,12 +194,25 @@ open class DokkaTask : DefaultTask(), Configurable {
             return null
         }
 
+        val sourceSetIdCache: Map<String, String> = defaultModulesConfiguration.map {
```
sellmair, line 197: I think we can find a more straight-forward way here 🤔

BarkingBad: What do you mean?

sellmair:
Closed because of #1048

PR #1049, opened 2020-07-01
Enforce workspace unique SourceSetID

sellmair:


runners/cli/build.gradle.kts
```diff
@@ -10,7 +10,6 @@ repositories {
 
 dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-cli-jvm:0.2.1")
-    implementation("com.google.code.gson:gson:2.8.5")
```
BarkingBad, line 13: Why do we remove gson dependency for CLI?

BarkingBad: Oh never mind, I missed you actually made way to extract config from json

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/configurationImplementations.kt
```diff
@@ -25,64 +33,115 @@ class GradleSourceRootImpl : SourceRoot, Serializable {
...
+    override val sourceSetID: DokkaSourceSetID = DokkaSourceSetID(project, name)
+
     @Input
     override var sourceRoots: MutableList<SourceRoot> = mutableListOf()
+
     @Input
-    override var dependentSourceSets: MutableList<String> = mutableListOf()
+    override var dependentSourceSets: MutableSet<DokkaSourceSetID> = mutableSetOf()
+
+    // TODO NOW: support adding dependencies by name
```
kamildoleglo, line 60: Leftover?

sellmair: Yes! Fixed ✅

PR #1051, opened 2020-07-01
Refactor extension loading to allow for overloads and suppression

Kordyjan:


core/src/main/kotlin/plugability/DokkaContext.kt
```diff
@@ -86,14 +104,39 @@ private class DokkaContextConfigurationImpl(
...
+                .flatMap { buckets[it].orEmpty() }
+                .toSet()
+            mergedBucket.forEach { buckets[it] = mergedBucket }
         }
+        return buckets.values.distinct().associateBy(::findNotOverridden)
+    }
+
+    private fun findNotOverridden(bucket: Set<Extension<*, *, *>>): Extension<*, *, *> {
+        val filtered = bucket.filter { it !in suppressedExtensions }
+        return filtered.singleOrNull() ?: throw Error("Conflicting overrides: $filtered")
```
kamildoleglo, line 130: Maybe `IllegalStateException` would be better?

core/src/main/kotlin/plugability/DokkaContext.kt
```diff
@@ -67,14 +67,32 @@ private class DokkaContextConfigurationImpl(
...
+        }
+    }
 
-    private fun topologicalSort() {
+    private val rawExtensions = mutableListOf<Extension<*, *, *>>()
+    private val rawAdjacencyList = mutableMapOf<Extension<*, *, *>, MutableList<Extension<*, *, *>>>()
+    private val suppressedExtensions = mutableMapOf<Extension<*, *, *>, MutableList<Suppression>>()
 
-        val result: MutableList<Extension<*>> = mutableListOf()
+    fun topologicallySortAndPrune() {
```
kamildoleglo, line 84: I think we could make it `internal`

core/src/main/kotlin/plugability/DokkaContext.kt
Kordyjan: done

core/src/main/kotlin/plugability/DokkaContext.kt
Kordyjan: This function is defined in private class so there is no way to call it from outside of the module.

Kordyjan:
> Please update the developer guide
@MarcinAman done

PR #1055, opened 2020-07-02
Add GFM renderer tests

BarkingBad:


plugins/base/src/main/kotlin/renderers/DefaultRenderer.kt
```diff
@@ -170,6 +170,23 @@ abstract class DefaultRenderer<T>(
             renderPages(newRoot)
         }
     }
+
+    protected fun ContentDivergentGroup.groupDivergentInstances(
```
sellmair, line 174: Is there no option to make this simpler?
Maybe implement a custom type instead of using `Map<Pair<String, String>, List<Pair<ContentDivergentInstance, DokkaSourceSet>>>`

plugins/base/build.gradle.kts
```diff
@@ -4,6 +4,8 @@ plugins {
     id("com.jfrog.bintray")
 }
 
+val testUtils by configurations.creating
```
sellmair, line 7: What is the reason for this?

BarkingBad: It allows us to share `test` sources for other modules. It's the simpliest way to have access to abstract classes and `TestOutputWriter`, which are defined in `test` sources, in GFM without any major changes.  Checkout this line
https://github.com/Kotlin/dokka/blob/05db29f3150aa23cdf48b4cefcb3024e60c130d6/plugins/gfm/build.gradle.kts#L13

plugins/base/src/main/kotlin/renderers/DefaultRenderer.kt
BarkingBad: I can do, I just ripped it from here, hadn't ever thought about tough signature 
https://github.com/Kotlin/dokka/blob/dev-0.11.0/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt#L226

plugins/base/build.gradle.kts
sellmair: Why can't this be moved into `:test-tools` or `:testApi`? 🤔

BarkingBad: The problem is they are strongly coupled. There would be circular dependency between modules.

BarkingBad: The biggest problem is `TestOutputWriter`, as it is `base` EP, so it has to be aware of `base`.

PR #1059, opened 2020-07-03
Add tests for signatures rendering

BarkingBad:


plugins/base/src/test/kotlin/signatures/DivergentSignatureTest.kt
```diff
@@ -0,0 +1,122 @@
...
+
+        testFromData(
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
+                assert(writerPlugin.writer.contents.getValue("example/example/-clock/index.html")
+                    .let { Jsoup.parse(it) }.select("#content").single()
+                    .select("div.sourceset-depenent-content")
+                    .fold(0) { acc, elem ->
```
sellmair, line 56: Just use the `.count` function with predicate!

BarkingBad: Nice, I missed this util :)

BarkingBad:
This PR should be merged after GFM tests #1055, as there are some changes in API that need to be rebased and resolved

PR #1063, opened 2020-07-06
Show only first level of inheritance

MarcinAman:
Purpose of this PR is to have only first level of inheritance rendered. Also it includes a supertype type to reflect how it is represented in java via `extends/implements`. I took the level-based approach since it can be easily modified to render a inheritance tree in javadoc

### DIFF
--- a/core/src/main/kotlin/links/DRI.kt
+++ b/core/src/main/kotlin/links/DRI.kt
@@ -1,5 +1,9 @@
 package org.jetbrains.dokka.links
 
+import org.jetbrains.dokka.model.ClassKind
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.WithSupertypes
+
 /**
  * [DRI] stands for DokkaResourceIdentifier
  */
@@ -100,3 +104,4 @@ fun DriTarget.nextTarget(): DriTarget = when (this) {
     else -> this
 }
 
+data class DriWithKind(val dri: DRI, val kind: ClassKind)
--- a/core/src/main/kotlin/links/DRI.kt
+++ b/core/src/main/kotlin/links/DRI.kt
@@ -1,8 +1,6 @@
 package org.jetbrains.dokka.links
 
 import org.jetbrains.dokka.model.ClassKind
-import org.jetbrains.dokka.model.Documentable
-import org.jetbrains.dokka.model.WithSupertypes
 
 /**
  * [DRI] stands for DokkaResourceIdentifier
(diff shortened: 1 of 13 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -556,15 +554,30 @@ private class DokkaDescriptorVisitor(
         getDocumentation()?.toSourceSetDependent() ?: emptyMap()
 
     private fun ClassDescriptor.resolveClassDescriptionData(): ClassInfo {
-        val interfaces = hashSetOf<DRI>()
-        fun processSuperClasses(supers: List<ClassDescriptor>) {
-            supers.forEach {
-                if(it.kind == ClassKind.INTERFACE) interfaces.add(DRI.from(it))
-                processSuperClasses(it.getSuperInterfaces() + it.getAllSuperclassesWithoutAny())
-            }
+        tailrec fun processSuperClasses(
```
sellmair, line 557: I was a little bit confused by this name when starting to read this PR! 
Maybe something like `buildInheritenceTree` would be easier to understand?

Also: Could you explain to me why you are trying to model this inheritance **tree** with a flat data structure like a set?

sellmair: I also think you are creating a lot of information here. From the first reading: You only need the superclasses of inheritance level=0 and all implemented interfaces, right? If this is correct, then we should keep it simple and not do all the extra work.

MarcinAman: Okey, the name is not the best :)

i am building a whole structure since right now i need inheritance level=0 and all interfaces. In the nearest future i'll need a whole structure to render javadoc's inheritance tree (mby with some small adjustments to cover edge cases). I thought that it will be easier to just do it once and not rework the same piece of code few times

MarcinAman: I've changed name to `buildInheritanceInformation` since i wouldn't like to suggest that this structure is a tree

core/src/main/kotlin/links/DRI.kt
```diff
@@ -1,5 +1,9 @@
 package org.jetbrains.dokka.links
 
+import org.jetbrains.dokka.model.ClassKind
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.WithSupertypes
```
BarkingBad, line 5: You don't need those imports if I am correct

MarcinAman: Sure, removed

PR #1070, opened 2020-07-07
Separated search key and display label in search component

pikinier20:
image

### DIFF
--- a/plugins/base/frontend/src/main/components/search/search.tsx
+++ b/plugins/base/frontend/src/main/components/search/search.tsx
@@ -1,8 +1,9 @@
 import React, {useCallback, useState} from 'react';
 import {Select} from '@jetbrains/ring-ui';
+import {List} from '@jetbrains/ring-ui';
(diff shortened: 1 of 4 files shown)


plugins/base/frontend/src/main/components/search/search.tsx
```diff
@@ -43,15 +44,30 @@ const WithFuzzySearchFilterComponent: React.FC<Props> = ({data}: Props) => {
     )
 }
 
+const templateGenerator = (page:Page) => {
+    let classGenerator = (page:Page) => {
+        let classes = ""
+        if(page.level !== undefined) classes = classes + " indented"
+        if(page.disabled) classes = classes + "disabled"
```
MarcinAman, line 51: no space before `disabled`? This can lead to class like ` intenteddisabled`

pikinier20: Added space before disabled

Issue #1031, opened 2020-06-29
Sorting of tabs


Kordyjan:
The goal of this task is to provide a way for sorting of tabs of html pages and to create reasonable default for this sorting.

PR #1076, opened 2020-07-08
Implemented extension point for tab sorting strategy

pikinier20:


pikinier20:
#1031

plugins/base/src/main/kotlin/renderers/DefaultTabSortingStrategy.kt
```diff
@@ -0,0 +1,29 @@
...
+            ContentKind.Inheritors to mutableListOf(),
+            ContentKind.Source to mutableListOf(),
+            ContentKind.Sample to mutableListOf(),
+            ContentKind.Comment to mutableListOf()
+        )
+        val unrecognized: MutableList<T> = mutableListOf()
+        tabs.forEach {
+            tabMap[it.dci.kind]?.add(it) ?: unrecognized.add(it)
+        }
+        return tabMap.values.flatten() + unrecognized
```
Kordyjan, line 26: I like this implementation. Although it may be simpler if we create this map from readonly list of kinds.

PR #1079, opened 2020-07-08
Samples CSS fix

pikinier20:
Still TODO:

-  Code is shrinked before unfolding
- On Firefox div of sample is bigger than sample-container


kamildoleglo:
I think this was fixed. @MarcinAman?

MarcinAman:
Afaik yes, this is fixed. We can keep the PR for Filip to confirm that

PR #1082, opened 2020-07-09
Fix concurrency issue managing package request cache

SUPERCILEX:
The `getExternalLocation` is called on a bunch of different worker threads, causing the `cache` to effectively not be used due to concurrency issues. This causes a downstream issue where `openConnection()` throws a `SocketTimeoutException: connect timed out` regardless of the configured timeout settings, presumably due to the OS being overloaded with connections and immediately rejecting new requests at a certain point.

This CL fixes the issue by locking the network request and subsequent parsing.


sellmair:
LGTM. Thanks for the contribution ☺️
Triggered tests on TC 👍

PR #1083, opened 2020-07-09
Only apply KasJ plugin if using Javadoc format

SUPERCILEX:
Currently, you can't generate Javadoc and Kdoc from the same JAR. Since the Javadoc plugin applies the KasJ plugin, Kdoc will end up having snippets with Java code instead of Kotlin.


sellmair:
Hi @SUPERCILEX, thanks a lot for your contribution! Actually, your PR really sparked discussions internally and we think that the current outputFormat task configuration might be flawed. See #1087 for a one proposal on fixing this!

SUPERCILEX:
https://github.com/Kotlin/dokka/pull/1087#issuecomment-656493120 that SGTM, so closing this.

PR #1084, opened 2020-07-09
Improve JSON parsing for CLI

SUPERCILEX:
Currently one has to provide *every* field when using the JSON CLI option. This CL use the kotlinx/serialization library to provide defaults and improve parsing.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -3,12 +3,14 @@ import org.jetbrains.configureBintrayPublication
 plugins {
     `maven-publish`
     id("com.jfrog.bintray")
+    kotlin("plugin.serialization")
 }
 
 dependencies {
     api(project("dependencies", configuration = "shadow"))
 
     implementation(kotlin("reflect"))
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.20.0-1.4-M3")
     implementation("com.google.code.gson:gson:2.8.5")
     implementation("org.jsoup:jsoup:1.12.1")
 
--- a/core/src/main/kotlin/configuration.kt
+++ b/core/src/main/kotlin/configuration.kt
@@ -5,6 +5,7 @@ package org.jetbrains.dokka
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.UnstableDefault
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonConfiguration
 import java.io.File
 import java.net.URL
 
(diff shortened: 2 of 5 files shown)


sellmair:
Looks very interesting!
I'll try to review this today pretty soon! 👍

sellmair:
Running tests on Team City now 🧪

core/build.gradle.kts
```diff
@@ -3,12 +3,14 @@ import org.jetbrains.configureBintrayPublication
...
     id("com.jfrog.bintray")
+    kotlin("plugin.serialization")
 }
 
 dependencies {
     api(project("dependencies", configuration = "shadow"))
 
     implementation(kotlin("reflect"))
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.20.0-1.4-M3")
     implementation("com.google.code.gson:gson:2.8.5")
```
sellmair, line 14: Can't gson be removed now? 🤔

core/src/main/kotlin/configuration.kt
```diff
@@ -24,6 +25,9 @@ object DokkaDefaults {
     const val noJdkLink: Boolean = false
     val analysisPlatform: Platform = Platform.DEFAULT
     const val suppress: Boolean = false
+
+    const val displayName = "JVM"
```
sellmair, line 29: What about calling this property "sourceSetDisplayName". I know this is not the very name of the SourceSet's property, but I think this would clarify it a little bit better 🙄

core/build.gradle.kts
SUPERCILEX: No, it's used here: https://github.com/Kotlin/dokka/blob/e1f3417a22942d168ace994734add42cc01d3694/core/src/main/kotlin/plugability/DokkaPlugin.kt#L61 I don't really understand that code so I'd rather not touch it.

core/src/main/kotlin/configuration.kt
SUPERCILEX: Done

core/build.gradle.kts
sellmair: Okay, I will take care of replacing this with kotlinx-serialization before merging ☺️ 👍

SUPERCILEX: Sweet, that'd be great!

sellmair:
Merged your commits into #1159
Note: This commits won't be cherry picked into the 1.4.0-M3 release branch (which is almost identical to 0.11.0)

PR #1085, opened 2020-07-09
Run build in parallel with Gradle cache

SUPERCILEX:


gradle.properties
```diff
@@ -12,5 +12,7 @@ language_version=1.4
 # Code style
 kotlin.code.style=official
 
-# Memory settings
+# Gradle settings
 org.gradle.jvmargs=-Xmx4g
+org.gradle.parallel=true
+org.gradle.caching=true
```
sellmair, line 18: I think running Gradle in parallel by default inside the project is fine, but I still think caching should still stay a "opt in" for each developer 🤔

sellmair: @Kordyjan

SUPERCILEX: Any reason to *not* use the cache? Gradle will automatically clean it up, so storage space shouldn't really be an issue.

SUPERCILEX: Since there's another thumbs up, removed. Would still be curious to understand the reasoning.

sellmair: Unsatisfyingly, there is no objective reasoning. The build cache is theoretically a good idea. However, not everyone fully trusts a mechanism like this. I personally had some very strange issues with it. I am not saying that no one should use this build cache. You're proposing to enabled it _by default_ for every developer. I am just arguing that every developer should actively _choose_ to enable this for the project. Everyone should just be aware of it.

SUPERCILEX: I guess that's fair, thanks for sharing.

PR #1086, opened 2020-07-09
Minor style fixes for gfm signatures

BarkingBad:
What have been fixed:
- headers inside tables are now omitted
- styles are now properly rendered
- signatures in divergent now render only once
- some minor fixes

### DIFF
--- a/plugins/gfm/src/main/kotlin/GfmPlugin.kt
+++ b/plugins/gfm/src/main/kotlin/GfmPlugin.kt
@@ -98,7 +98,7 @@ open class CommonmarkRenderer(
                 buildList(it, pageContext)
             } else {
                 append("<li>")
-                it.build(this, pageContext)
+                append(buildString { it.build(this, pageContext, it.sourceSets) }.trim())
                 append("</li>")
             }
         }
@@ -138,7 +138,7 @@ open class CommonmarkRenderer(
         pageContext: ContentPage,
     ) {
         if (content is ContentGroup && content.children.firstOrNull { it is ContentTable } != null) {
-            buildContentNode(content, pageContext)
+            buildContentNode(content, pageContext, sourceSets)
         } else {
             val distinct = sourceSets.map {
                 it to buildString { buildContentNode(content, pageContext, setOf(it)) }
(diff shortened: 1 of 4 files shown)


plugins/gfm/src/main/kotlin/GfmPlugin.kt
```diff
@@ -211,10 +211,14 @@ open class CommonmarkRenderer(
     }
 
     override fun StringBuilder.buildText(textNode: ContentText) {
-        val decorators = decorators(textNode.style)
-        append(decorators)
-        append(textNode.text)
-        append(decorators.reversed())
+        if(textNode.text.isNotBlank()) {
+            val decorators = decorators(textNode.style)
+            append(textNode.text.takeWhile { it == ' ' } )
```
MarcinAman, line 216: what is the reason we need to parse the text node while rendering it?

plugins/gfm/src/main/kotlin/GfmPlugin.kt
```diff
@@ -251,33 +255,36 @@ open class CommonmarkRenderer(
         distinct.values.forEach { entry ->
             val (instance, sourceSets) = entry.getInstanceAndSourceSets()
 
-            append(sourceSets.joinToString(prefix = "#### [", postfix = "]") { "${it.moduleDisplayName}/${it.displayName}" })
+            append(sourceSets.joinToString(prefix = "[", postfix = "]") { it.displayName })
```
MarcinAman, line 258: it would be super nice to extract label generation to a fun. Right now it is in 3 places.

plugins/gfm/src/main/kotlin/GfmPlugin.kt
BarkingBad: There two bugs related to this:
- empty TextNodes with styles end up with e. g. "~~~~" representation, because it wants to surround empty text with styles
- there is some bug when styles characters (`*`, `**`, `~~`, etc) have white spaces at the beginning or at the end. If we provide text like `* there is some text *` vs ` *there is some text* `, the first one will end up rendered as it is with asterisks, the second one will be well formatted bold text (markserv tested)

plugins/gfm/src/main/kotlin/GfmPlugin.kt
BarkingBad: Yeah, you are right

PR #1087, opened 2020-07-09
Re-design format configuration

sellmair:


SUPERCILEX:
@sellmair Coming from #1083. Just to make sure I understand what that change is doing, you're creating different tasks with different plugin classpaths as a way of selectively applying the Dokka plugins? I think that could work for the Gradle plugin, but I'm trying to use the CLI. As I understand it, the way this is set up you would need a different executable binary per target format. Is that expected?

sellmair:
Yes, I am creating different tasks by default for all formats that we support out of the box. This would have been the intended way of supporting multiple formats even before this changes. 

Regarding CLI: This is even another argument for "why do I think the format setting is flawed". When using the CLI with *any* format, you would also need to link the plugin `.jar`. Just setting an output format using the setting would not have been enough.

I personally think, that it will not be possible to load a lot of plugins and keep them separate with configuration in the current design. I am convinced that the classpaths should be separated to ensure predictable behaviour and make it easy to use.

plugins/gfm/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ publishing {
 configureBintrayPublication("gfmPlugin")
 
 dependencies {
-    compileOnly(project(":plugins:base"))
+    implementation(project(":plugins:base"))
```
kamildoleglo, line 15: I'm not sure why this change is needed, could you please explain?

plugins/jekyll/src/main/kotlin/JekyllPlugin.kt
```diff
@@ -21,8 +22,8 @@ class JekyllPlugin : DokkaPlugin() {
     val renderer by extending {
         (CoreExtensions.renderer
                 providing { JekyllRenderer(it) }
-                applyIf { format == "jekyll" }
-                override plugin<DokkaBase>().htmlRenderer)
+                // TODO NOW: Is this correct?
```
kamildoleglo, line 25: Good question, I think it is, @Kordyjan?

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt
```diff
@@ -0,0 +1,50 @@
...
+
+
+abstract class AbstractDokkaTask : DefaultTask(), Configurable {
+    @Input
+    var outputDirectory: String = defaultDokkaOutputDirectory().absolutePath
+
+    @Suppress("unused")
+    @Deprecated("This setting is a noop and will be removed")
+    @get:Internal
+    var outputFormat: String = ""
```
kamildoleglo: I'd remove it entirely TBH. While it is a good practice to deprecate things, we have completely changed the DSL anyway, so one change more doesn't really matter

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaCollectorTask.kt
```diff
@@ -13,42 +11,44 @@ open class DokkaCollectorTask : DefaultTask() {
     var modules: List<String> = emptyList()
 
     @Input
-    var outputDirectory: String = ""
+    var outputDirectory: String = defaultDokkaOutputDirectory().absolutePath
 
     private lateinit var configuration: GradleDokkaConfigurationImpl
 
+    @Input
+    val dokkaTaskNames: MutableSet<String> = mutableSetOf()
```
kamildoleglo, line 19: Maybe `var` here would be better to allow replacing the list?

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaCollectorTask.kt
```diff
@@ -13,42 +11,44 @@ open class DokkaCollectorTask : DefaultTask() {
...
-        project.tasks.withType(DokkaTask::class.java).configureEach { it.config = configuration }
+        project.tasks.withType(DokkaTask::class.java).configureEach { it.enforcedConfiguration = configuration }
     }
 
     init {
-        finalizedBy(project.tasks.getByName(DOKKA_TASK_NAME))
+        // TODO: This this certainly not the ideal solution
+        dokkaTaskNames.forEach { dokkaTaskName ->
+            finalizedBy(dokkaTaskName)
+        }
```
kamildoleglo, line 50: Can we get those tasks and `dependOn` them instead? (I know that this was introduced in another commit, just wondering if we can fix it RN)

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/dokkaConfigurations.kt
```diff
@@ -0,0 +1,41 @@
+package org.jetbrains.dokka.gradle
+
+import org.gradle.api.Project
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.attributes.Usage
+
+internal fun Project.maybeCreateDokkaDefaultPluginConfiguration(): Configuration {
```
kamildoleglo, line 7: Can we `maybeRegister` them instead or is there a specific reason for creating them?

integration-tests/gradle/projects/it-basic-groovy/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
sellmair: Oh they are used! These integration projects are also designed in a way, that they can just be imported into the IDE, or being used to debug the current code-base. So the wrapper _is_ used. I can see the duplication, but I do not know if there is a good way around this.

plugins/gfm/build.gradle.kts
sellmair: Because it needs the base plugin at runtime and compile time. I am not sure why it was different before!

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt
sellmair: Removed ✅

plugins/gfm/build.gradle.kts
kamildoleglo: Not really, dokka gradle task adds dokka base in the runtime, so this way we end having the same dependency (dokka base) twice

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaCollectorTask.kt
sellmair: I am not sure if this double mutability is really a good idea, especially because of the fact that we are populating this with a default for default tasks. 
Making this a `var` might mislead people to think that it is "ok" to just replace this entirely. If they really want to do something custom, they could still clear it. 

I am not entirely against your proposal, but I am just a little bit afraid that it makes thinks more complicated and error prone and that the potential convenience might not be worth it 🙄

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaCollectorTask.kt
sellmair: I think the DokkaCollectorTask deserves to be entirely re-designed as well. Changing the configuration of another task during execution and then running those seems really wrong to me. 

I am to afraid to change this RN, but I want to tackle this as well in a later PR!

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/dokkaConfigurations.kt
sellmair: Unfortunately, I think there is no such thing as "maybeRegister", but I could just register them in the `main.kt`. Since the tasks are configured lazy, the maybeCreate is fine for them! 

Let me see what we can do there! ☺️

sellmair: Unfortunately, there is no such thing as "maybeRegister" in Gradle. 
Looking at the configuration code, I think there is no reason why this should be evaluated lazy:

```kotlin
        extendsFrom(maybeCreateDokkaDefaultRuntimeConfiguration())
        attributes.attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage::class.java, "java-runtime"))
        isCanBeConsumed = false
        defaultDependencies { dependencies ->
            dependencies.add(project.dokkaArtifacts.dokkaCore)
        }
```

I could use the register function in main.kt, but do you think we really gain something here?

plugins/jekyll/src/main/kotlin/JekyllPlugin.kt
Kordyjan: I don't think it is correct as long as `hmtlRenderer` is in the base plugin. I may be missing something but it looks like after applying jekyll plugin you would end up with two loaded renderers.

plugins/gfm/build.gradle.kts
sellmair: No, the dokka task just adds a "default" dependency (which is necessary, because otherwise you could not have a dokka task **without** base). Since this tasks depends on base, it should declare its dependency on base!

kamildoleglo: This way you won't be able to have a version difference between the plugin and the base, I'm now wondering if that's an issue though. 
You may want to have one version different than the other and while it should be possible to do with Gradle configuration I'm wondering if we should make it easier or not  🤔

PR #1090, opened 2020-07-10
Fix NPE

SUPERCILEX:


BarkingBad:
Can you provide what argument led to NPE?

SUPERCILEX:
TBH, I quite understand what's going on. Here's what I've figured out so far.

If I print a bunch of debug info like so:
```kotlin
println((type as PsiImmediateClassType).className)
println((type as PsiImmediateClassType).internalCanonicalText)
println((type as PsiImmediateClassType).presentableText)
println((type as PsiImmediateClassType).parameters.joinToString { it.canonicalText })
```

Here's what a good run looks like:
```
java.lang.Class<androidx.fragment.app.Fragment>
Class<Fragment>
androidx.fragment.app.Fragment
```

And the run that crashes:
```
java.lang.Class<Activity>
Class<Activity>
Activity
```

My understanding is that `Activity` isn't on the classpath which is why the resolve is failing.

PR #1100, opened 2020-07-10
Fix rendering of classlikes on packages page

BarkingBad:


plugins/javadoc/src/main/resources/views/components/indexTable.korte
```diff
@@ -1,4 +1,8 @@
-<table>
+{% if isTypeSummary %}
+    <table class="typeSummary">
```
MarcinAman, line 2: imho it would be a little bit cleaner like:
```
<table class="{{ styleClasses }}">
```
And then in indexPage:
```
{% set styleClasses = "typeSummary" %}
```
Since if classes are empty they are simply ignored in html

BarkingBad: Didn't know, thanks for the tip, unfortunately, it's too late :P

PR #1105, opened 2020-07-13
Single platform bubbles fix

BarkingBad:
obraz
obraz
obraz


mshishkina:
Fix is verified in dokka v.1.4.0-M3-dev-67

PR #1119, opened 2020-07-13
Fix javadoc rendering

BarkingBad:
Tested on https://github.com/Kotlin/kotlin-examples/tree/master/maven/dokka-maven-example

Remember to change version of `dokka.version` property to `0.11.0-SNAPSHOT` or whatever version of `0.11.0` you're using. Moreover, add following plugin repositories

```xml
<pluginRepository>
    <id>jcenter-dokka</id>
    <name>JCenterDokka</name>
    <url>https://dl.bintray.com/kotlin/kotlin-dev/</url>
    <releases>
        <enabled>true</enabled>
    </releases>
</pluginRepository>
<pluginRepository>
    <id>jcenter-md</id>
    <name>JCenterMd</name>
    <url>https://dl.bintray.com/jetbrains/markdown/</url>
    <releases>
        <enabled>true</enabled>
    </releases>
</pluginRepository>   
```

### DIFF
--- a/runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
+++ b/runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
@@ -166,7 +166,7 @@ abstract class AbstractDokkaMojo : AbstractMojo() {
     var failOnWarning: Boolean = DokkaDefaults.failOnWarning
 
     @Parameter
-    var dokkaPlugins: List<Dependency> = emptyList()
+    open var dokkaPlugins: List<Dependency> = emptyList()
 
     protected abstract fun getOutDir(): String
 
@@ -243,7 +243,7 @@ abstract class AbstractDokkaMojo : AbstractMojo() {
                 if (sourceSet.moduleDisplayName.isEmpty()) logger.warn("Not specified module name. It can result in unexpected behaviour while including documentation for module")
             },
             pluginsClasspath = getArtifactByAether("org.jetbrains.dokka", "dokka-base", dokkaVersion) +
-                    dokkaPlugins.map { getArtifactByAether(it.groupId, it.artifactId, it.version) }.flatten(),
+                    dokkaPlugins.map { getArtifactByAether(it.groupId, it.artifactId, it.version ?: dokkaVersion) }.flatten(),
             pluginsConfiguration = mutableMapOf(), //TODO implement as it is in Gradle
             modules = emptyList(),
             failOnWarning = failOnWarning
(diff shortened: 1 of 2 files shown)


runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -243,7 +243,7 @@ abstract class AbstractDokkaMojo : AbstractMojo() {
                 if (sourceSet.moduleDisplayName.isEmpty()) logger.warn("Not specified module name. It can result in unexpected behaviour while including documentation for module")
             },
             pluginsClasspath = getArtifactByAether("org.jetbrains.dokka", "dokka-base", dokkaVersion) +
-                    dokkaPlugins.map { getArtifactByAether(it.groupId, it.artifactId, it.version) }.flatten(),
+                    dokkaPlugins.map { getArtifactByAether(it.groupId, it.artifactId, it.version ?: dokkaVersion) }.flatten(),
```
sellmair, line 246: Do you think it makes sense for users that we automatically append **our** dokka version to artifacts? I can imagine that this leads to confusion and unwanted behaviour for users?

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -332,6 +332,9 @@ class DokkaMojo : AbstractDokkaMojo() {
 class DokkaJavadocMojo : AbstractDokkaMojo() {
```
sellmair, line 332: Could you please add a MavenIntegrationTest to cover this? I didn't even know that `DokkaJavadocMojo` exists 🙈

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -166,7 +166,7 @@ abstract class AbstractDokkaMojo : AbstractMojo() {
     var failOnWarning: Boolean = DokkaDefaults.failOnWarning
 
     @Parameter
-    var dokkaPlugins: List<Dependency> = emptyList()
+    open var dokkaPlugins: List<Dependency> = emptyList()
```
Kordyjan, line 169: Cannot we pass this list as constructor parameter for this abstract class?

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
BarkingBad: I can try, but I have never ever done integration tests in my life before, just know theory. Can you point me to example I could base on?

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
BarkingBad: It's not every artifact you can imagine. These are only dokka plugins. If you know certain version of plugin, you can specify it explicitly. However, if you don't know what plugin version should be linked, it will try to match plugin based on dokka-base version. It's useful for example for javadoc artifacts. I don't know what version user would like to use, I cannot pass it other way, but I can implicitly assume its version on dokka-base version

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
BarkingBad: I just missed that option, will change

PR #1124, opened 2020-07-14
Readme update for dokka 1.4-M3

kamildoleglo:


README.md
```diff
@@ -7,8 +7,57 @@ and can generate documentation in multiple formats including standard Javadoc, H
...
+Please see the usage instructions below for how to add plugins to dokka. 
+
+### Source sets 
+Dokka generates documentation based on source sets. 
+
+For single-platform projects, there is almost always only one source set - `main`.
+
+For multi-platform projects, source sets are the same as in Kotlin plugin - source set for each platform, eg. `jvmMain` or `jsMain` and for each
+common source set - eg. the default `commonMain` and custom ones, like `jsAndJvmMain`.
+When configuring multi-platform projects manually (eg. in the CLI), source sets must declare which other source sets they depend on.
```
sellmair, line 29: I thought this is still also necessary for the gradle plugin?

README.md
```diff
@@ -25,182 +74,203 @@ repositories {
...
 tasks.dokkaHtml {
     outputDirectory = "$buildDir/dokka"
+    dokkaSourceSets {
+        create("jvmMain")
+        create("jsMain") // or other names, identical to those in Kotlin-plugin
+    }
 }
 ```
-You may need to add an `import org.jetbrains.dokka.gradle.dokka` to the top of `build.gradle.kts` in this case.
+You may need to add an `import org.jetbrains.dokka.gradle.dokka` to the top of `build.gradle.kts`.
```
sellmair, line 103: We can remove this sentence: I removed the extension function in favour of gradle automated generated accessors

README.md
```diff
@@ -25,182 +74,203 @@ repositories {
...
 }
 ```
 
 #### Multiplatform
-Since version 0.10.0 dokka supports multiplatform projects. For a general understanding how a multiplatform documentation is generated, please consult the [FAQ](https://github.com/Kotlin/dokka/wiki/faq).
-In the multiplatform mode, instead of using the `configuration` block, you should use a `multiplatform` block with inner blocks for each platform. 
-The `configuration` block's parameters belong to those inner blocks, which can be named arbitrarly, however if you want to use source roots and classpath provided by Kotlin Multiplatform plugin, 
-they must have the same names as in the Kotlin Multiplatform plugin. See an example below:
+Dokka supports single-platform and multi-platform projects using source sets abstraction. For most mutli-platform projects,
+you should assume that `one platform == one source set`. Source sets can be named arbitrarily, however in order for autoconfiguration
```
sellmair, line 248: you should assume that `one platform == one source set`: Isn't this confusing when we talk about something like `jsAndJvmMain` above?

README.md
```diff
@@ -243,58 +313,43 @@ dokkaHtml {
...
+    dokkaHtmlPlugin("org.jetbrains.dokka:kotlin-as-java-plugin:1.4-M3")
+}
+``` 
+
+You can also create a custom dokka task and add plugins directly inside:
+
+```kotlin
+val customDokkaTask by creating(DokkaTask::class) {
+    dependencies {
+        plugins("org.jetbrains.dokka:kotlin-as-java-plugin:1.4-M3")
```
sellmair, line 332: Side note: Should we consider calling this porperty `plugin`? The plural looks strange here!

README.md
```diff
@@ -362,9 +446,9 @@ By default files will be generated in `target/dokka`.
 
 The following goals are provided by the plugin:
 
-  * `dokka:dokka` - generate HTML documentation in Dokka format (showing declarations in Kotlin syntax);
-  * `dokka:javadoc` - generate HTML documentation in JavaDoc format (showing declarations in Java syntax);
-  * `dokka:javadocJar` - generate a .jar file with JavaDoc format documentation.
+  * `dokka:dokka` - generate HTML documentation in Dokka format (showing declarations in Kotlin syntax)
+  * `dokka:javadoc` - generate HTML documentation in Javadoc format (showing declarations in Java syntax)
```
sellmair, line 450: This is not implemented yet, isn't it?

README.md
```diff
@@ -474,66 +558,115 @@ The available configuration options are shown below:
...
-  * `markdown` - markdown structured as `html`, Java classes are translated to Kotlin
-    * `gfm` - GitHub flavored markdown
-    * `jekyll` - Jekyll compatible markdown 
-  * `kotlin-website*` - internal format used for documentation on [kotlinlang.org](https://kotlinlang.org)
-
-### Platforms<a name="platforms"></a>
+  * `html` - HTML format used by default
+  * `javadoc` - looks like JDK's Javadoc, Kotlin classes are translated to Java
+  * `gfm` - GitHub flavored markdown
+  * `jakyll` - Jekyll compatible markdown
```
sellmair, line 654: typo!

core/migration_guide.md
```diff
@@ -4,72 +4,72 @@ There are two main changes between dokka 0.10.x and 0.11.0
 
 The first is the introduction of plugability - new documentation creating process is divided into several steps and each step provides extension points to be used. To learn more about new dokka pipeline and possible plugins, please read Developer's guide.
 
-Second difference comes with the change with the subject of dokka pass. Previously, separate dokka passes where set for every targeted platform, now every source set has its own pass.
+Second difference comes with the change with the subject of dokka pass. Previously, separate dokka passes where set for every targeted platform, now every source set has its own pass and the name itself changed to `sourceSet.
```
sellmair, line 7: Missing "`" at the end?

README.md
kamildoleglo: Yes, but only without autoconfiguration, I will clarify it

README.md
BarkingBad: They are, since today morning :P

README.md
kamildoleglo: It's still needed for custom tasks, removed though

README.md
kamildoleglo: Changed

README.md
kamildoleglo: We should, I agree

README.md
kamildoleglo: Thanks, fixed!

core/migration_guide.md
kamildoleglo: Fixed

README.md
```diff
@@ -7,8 +7,57 @@ and can generate documentation in multiple formats including standard Javadoc, H
 
 ## Using dokka
 
+### Plugins 
+Dokka can be customized with plugins. Each output format (apart from the default `HTML` one is internally a plugin).
```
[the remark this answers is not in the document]
kamildoleglo: Yes and no, IMHO it is also misleading without this information, as you have to have the dokka-base which provides the  HTML format if you want to generate anything, removed though

README.md
```diff
@@ -7,8 +7,57 @@ and can generate documentation in multiple formats including standard Javadoc, H
...
+* `kotlin-as-java-plugin` - translates Kotlin definitions to Java 
+
+Please see the usage instructions below for how to add plugins to dokka. 
+
+### Source sets 
+Dokka generates documentation based on source sets. 
+
+For single-platform projects, there is almost always only one source set - `main`.
+
+For multi-platform projects, source sets are the same as in Kotlin plugin - source set for each platform, eg. `jvmMain` or `jsMain` and for each
```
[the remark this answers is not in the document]
kamildoleglo: Changed

README.md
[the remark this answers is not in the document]
kamildoleglo: Added

README.md
```diff
@@ -7,8 +7,57 @@ and can generate documentation in multiple formats including standard Javadoc, H
...
+```groovy
+plugins {
+    id 'org.jetbrains.dokka' version '1.4-M3'
+}
+repositories {
+    jcenter() // or maven { url 'https://dl.bintray.com/kotlin/dokka' }
+}
+```
+
+or using the buildscript:
```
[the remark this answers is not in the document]
kamildoleglo: Thanks, changed manually

README.md
```diff
@@ -7,8 +7,57 @@ and can generate documentation in multiple formats including standard Javadoc, H
...
+        val jsMain by getting { dependsOn(jvmAndJsSecondCommonMain) }
+        val linuxX64Main by getting { dependsOn(commonMain) }
+    }
+}
+```
+Here `jsMain` and `jvmMain` both depend on `commonMain` (by default and transitively) and `jsAndJvmMain` while `linuxX64Main` only depends on `commonMain`. 
+
 ### Using the Gradle plugin
 
+```groovy
```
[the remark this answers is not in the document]
kamildoleglo: Changed

README.md
```diff
@@ -25,182 +74,203 @@ repositories {
...
+    }
+}
+```
 
-The available configuration options for single platform are shown below:
+Dokka documents single-platform as well as multi-platform projects. 
+Most of the configuration options are set per one source set.
+The available configuration options for are shown below:
 
 ```groovy
```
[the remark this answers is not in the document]
kamildoleglo: Changed

README.md
```diff
@@ -7,212 +7,254 @@ and can generate documentation in multiple formats including standard Javadoc, H
...
+    jcenter() // or maven { url 'https://dl.bintray.com/kotlin/dokka' }
+}
+```
+
+or using the buildscript block:
+
+```kotlin
 buildscript {
     repositories {
         jcenter()
```
sellmair, line 71: We might (for now) also show that you are supposed to add kotlin eap repository as well 🤔

README.md
```diff
@@ -7,212 +7,254 @@ and can generate documentation in multiple formats including standard Javadoc, H
...
     }
     dependencies {
-        classpath "org.jetbrains.dokka:dokka-gradle-plugin:${dokka_version}"
+        classpath("org.jetbrains.dokka:dokka-gradle-plugin:${dokka_version}")
```
mshishkina, line 74: Do we need such dependency, seems id("org.jetbrains.dokka") version "1.4-M3" is enough

sellmair: Depends on whether or not the `pluginManagement` block is used in settings.gradle.kts 
I would not show this approach, since it leads to a strange bug in gradle with the multimodule task aaaaand it seems not so common to me. People who want to use the `pluginManagement` approach will know how this works anyways 🙄 

Whats your opinion kamil?

kamildoleglo: I think you're right. I'm not a fan of the buildscript way either, I'll remove it

kamildoleglo: After further discussion decided to leave both ways, as one is easier and the other is better

PR #1126, opened 2020-07-14
Add a basic mathjax test

liutikas:
A simple tests that seems to show that MathjaxPlugin does not work. It does not seem insert the additional script resources.

### DIFF
--- a/plugins/mathjax/build.gradle.kts
+++ b/plugins/mathjax/build.gradle.kts
@@ -1,5 +1,14 @@
 import org.jetbrains.registerDokkaArtifactPublication
 
+dependencies {
+    testImplementation("org.jsoup:jsoup:1.12.1")
+    testImplementation(project(":plugins:base"))
+    testImplementation(project(":plugins:base:test-utils"))
+    testImplementation(project(":test-tools"))
+    testImplementation(kotlin("test-junit"))
+    testImplementation(project(":kotlin-analysis"))
+}
+
 registerDokkaArtifactPublication("mathjaxPlugin") {
     artifactId = "mathjax-plugin"
 }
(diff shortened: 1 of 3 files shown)


BarkingBad:
Hey, thank you for you contribution.

The reason why MathJax is not working is the fact you use the tag in the wrong way. It should not be called with any curly braces `{}`, just plain `@usesMathJax` anywhere, but __has to start from a new line__. For example, you can check it out how it works here: https://github.com/kamildoleglo/dokka-example/blob/master/src/jvmMain/kotlin/example/ParticularClock.kt#L28

(Note: In KDoc, I had to replace single dollar sign mathjax alias `$ ... $` with `\( ... \)` to work, single dollar signs ended up with plain text interpolation,  I need to investigate it in the near future)

And the output is:
obraz

However, your unit tests won't pass yet. There are few amendments that have to be done:
1. KDoc has to be above proper signature. For now you just left it floating somewhere. The easiest fix is to add some basic signature, for example:
```kotlin
        val source =
            """
            |/src/main/kotlin/test/Test.kt
            |package example
            | /**
            | * @usesMathJax
            | *
            | * \(\alpha_{out} = \alpha_{dst}\)
            | * \(C_{out} = C_{dst}\)
            | */
            | fun test(): String = ""
            """.trimIndent()
```
2. Secondly, you cannot use HTML tags in KDocs, I know javadoc accepts HTML, but KDoc interprets only markdown, so these paragraph tags will end up as a part of rendered text.
3. After adding proper signature, we need to read from a correct location. For us, it will be `root/example/<signature-name>.html`, so with provided example, you should point to `writerPlugin.writer.contents["root/example/test.html"]` 

Please fix these changes so we can merge it.

sellmair:
Running tests on Team City now ☺️ 👍

PR #1138, opened 2020-07-14
Fix page merging bug that prevented including Kotlin Playground

kamildoleglo:


sellmair:
Cherry picked into 1.4.0-M3

PR #1141, opened 2020-07-14
Fix presenting inline code in KDoc

BarkingBad:
I know that SinceKotlin is in wrong place. There is already PR fixing that
obraz
obraz


sellmair:
Cherry picked into 1.4.0-M3

PR #1145, opened 2020-07-14
Follow Javadoc file path convention

SUPERCILEX:
The current file path looks like `/androidx.fragment.app/package-summary.html` which is non-standard. This CL changes the path to look like `/androidx/fragment/app/package-summary.html`.


SUPERCILEX:
@sellophane Friendly ping

PR #1152, opened 2020-07-15
Gfm tables fix

BarkingBad:


sellmair:
Could you cover something like this with a test for the future? 🙄

sellmair:
Cherry picked into 1.4.0-M3

PR #1154, opened 2020-07-15
Use DokkaLogger instead of println and adjust log levels for less verbosity by default

sellmair:


sellmair:
Cherry picked into 1.4.0-M3

PR #1157, opened 2020-07-15
Update .idea/jarRepositories.xml to include maven local.

liutikas:
Importing dokka project into Intellij always results into this
diff. Checking it in, so it does not keep coming back.


sellmair:
Had to giggle a little:
image

Probably this is not necessary after that one? 🙄

PR #1159, opened 2020-07-16
Improve JSON parsing for CLI

sellmair:
Thanks to @SUPERCILEX

### DIFF
--- a/core/src/main/kotlin/configuration.kt
+++ b/core/src/main/kotlin/configuration.kt
@@ -5,6 +5,7 @@ package org.jetbrains.dokka
 import org.jetbrains.dokka.utilities.toJsonString
 import org.jetbrains.dokka.utilities.parseJson
 import java.io.File
+import java.io.Serializable
 import java.net.URL
 
 object DokkaDefaults {
@@ -53,7 +54,7 @@ enum class Platform(val key: String) {
 data class DokkaSourceSetID(
     val moduleName: String,
     val sourceSetName: String
-) {
+) : Serializable {
     override fun toString(): String {
         return "$moduleName/$sourceSetName"
     }
@@ -67,7 +68,7 @@ fun DokkaConfiguration.toJsonString(): String {
     return toJsonString(this)
 }
 
-interface DokkaConfiguration {
+interface DokkaConfiguration : Serializable {
     val outputDir: String
     val cacheRoot: String?
     val offlineMode: Boolean
@@ -77,7 +78,7 @@ interface DokkaConfiguration {
     val pluginsClasspath: List<File>
     val pluginsConfiguration: Map<String, String>
 
-    interface DokkaSourceSet {
+    interface DokkaSourceSet: Serializable {
         val sourceSetID: DokkaSourceSetID
         val displayName: String
         val moduleDisplayName: String
@@ -103,23 +104,23 @@ interface DokkaConfiguration {
         val analysisPlatform: Platform
     }
 
-    interface SourceRoot {
+    interface SourceRoot: Serializable {
         val path: String
     }
 
-    interface SourceLinkDefinition {
+    interface SourceLinkDefinition: Serializable {
         val path: String
         val url: String
         val lineSuffix: String?
     }
 
-    interface DokkaModuleDescription {
+    interface DokkaModuleDescription : Serializable {
         val name: String
         val path: String
         val docFile: String
     }
 
-    interface PackageOptions {
+    interface PackageOptions : Serializable{
         val prefix: String
         val includeNonPublic: Boolean
         val reportUndocumented: Boolean?
@@ -127,7 +128,7 @@ interface DokkaConfiguration {
         val suppress: Boolean
     }
 
-    interface ExternalDocumentationLink {
+    interface ExternalDocumentationLink : Serializable {
         val url: URL
         val packageListUrl: URL
 
--- a/core/src/main/kotlin/configuration.kt
+++ b/core/src/main/kotlin/configuration.kt
@@ -5,6 +5,7 @@ package org.jetbrains.dokka
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.UnstableDefault
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonConfiguration
 import java.io.File
 import java.net.URL
 
@@ -26,7 +27,7 @@ object DokkaDefaults {
     val analysisPlatform: Platform = Platform.DEFAULT
     const val suppress: Boolean = false
 
-    const val displayName = "JVM"
+    const val sourceSetDisplayName = "JVM"
     const val sourceSetName = "main"
 }
 
@@ -62,8 +63,9 @@ data class DokkaSourceSetID(
 }
 
 @OptIn(UnstableDefault::class)
-fun DokkaConfigurationImpl(json: String): DokkaConfigurationImpl {
-    return Json.parse(DokkaConfigurationImpl.serializer(), json)
+fun DokkaConfigurationImpl(input: String): DokkaConfigurationImpl {
+    val json = Json(JsonConfiguration.Default.copy(ignoreUnknownKeys = true))
+    return json.parse(DokkaConfigurationImpl.serializer(), input)
 }
 
 interface DokkaConfiguration {
--- a/core/src/main/kotlin/configuration.kt
+++ b/core/src/main/kotlin/configuration.kt
@@ -2,10 +2,8 @@
 
 package org.jetbrains.dokka
 
-import kotlinx.serialization.Serializable
-import kotlinx.serialization.UnstableDefault
-import kotlinx.serialization.json.Json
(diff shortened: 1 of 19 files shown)


sellmair:
@SUPERCILEX : May I ask for a review from your side now? Unfortunately, I had to switch to jackson for now 🙄

sellmair:
Note: This will not be cherry picked into 1.4.0-M3

core/src/main/kotlin/utilities/json.kt
```diff
@@ -0,0 +1,61 @@
...
+
+@PublishedApi
+internal fun <T : Any> parseJson(json: String, typeReference: TypeReference<T>): T {
+    return objectMapper.readValue(json, typeReference.jackson)
+}
+
+
+object FileSerializer : StdScalarSerializer<File>(File::class.java) {
+    override fun serialize(value: File, g: JsonGenerator, provider: SerializerProvider) {
+        g.writeString(value.path)
```
sellmair, line 50: I think this follows the proposal of @SUPERCILEX , but will break the existing format which uses something like 
`"path": "path/to/file"` as json format 🤔

asfalcone:
Can I ask why Jackson is required instead of kotlinx.serialization?

sellmair:
@asfalcone As of right now, we also do have to (de)-serialize types that are now known at compile time. We are planning to get rid of this requirement.

core/src/main/kotlin/configuration.kt
```diff
@@ -50,18 +53,18 @@ enum class Platform(val key: String) {
...
+) {
     override fun toString(): String {
         return "$moduleName/$sourceSetName"
     }
 }
 
 fun DokkaConfigurationImpl(json: String): DokkaConfigurationImpl {
-    return Gson().fromJson(json, DokkaConfigurationImpl::class.java)
+    return parseJson(json)
 }
```
kamildoleglo, line 65: Maybe make this oneline: `fun DokkaConfigurationImpl(json: String) = parseJson(json)`

core/src/main/kotlin/configuration.kt
```diff
@@ -50,18 +53,18 @@ enum class Platform(val key: String) {
...
 
-fun DokkaConfiguration.toJson(): String {
-    return Gson().toJson(this)
+fun DokkaConfiguration.toJsonString(): String {
+    return toJsonString(this)
 }
```
kamildoleglo, line 69: I'd suggest to also make it oneline with `=`

core/src/main/kotlin/utilities/json.kt
```diff
@@ -0,0 +1,69 @@
...
+    val module = SimpleModule().apply {
+        addSerializer(FileSerializer)
+        addSerializer(SourceRoot::class.java, SourceRootSerializer)
+        addDeserializer(SourceRootImpl::class.java, SourceRootImplDeserializer)
+        addDeserializer(SourceRoot::class.java, SourceRootImplDeserializer)
+    }
+    jacksonObjectMapper()
+        .registerModule(module)
+        .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+}
```
kamildoleglo, line 28: IMHO it would be more readable when wrote like this:
```kotlin
private val objectMapper = SimpleModule().apply {
... }.let { module ->
    jacksonObjectMapper()
        .registerModule(module)
        .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
}
```
but maybe that's only my preference

core/src/main/kotlin/utilities/json.kt
```diff
@@ -0,0 +1,69 @@
...
+
+@PublishedApi
+internal class TypeReference<T> private constructor(
+    internal val jackson: JacksonTypeReference<T>
+) {
+    companion object {
+        @OptIn(ExperimentalStdlibApi::class)
```
kamildoleglo, line 35: Probably an unnecessary annotation, I don't see any experimental API there

core/src/main/kotlin/utilities/json.kt
```diff
@@ -0,0 +1,69 @@
...
+
+@PublishedApi
+internal inline fun <reified T : Any> parseJson(json: String): T {
+    return parseJson(json, TypeReference())
+}
+
+@PublishedApi
+internal fun <T : Any> parseJson(json: String, typeReference: TypeReference<T>): T {
+    return objectMapper.readValue(json, typeReference.jackson)
+}
```
kamildoleglo, line 51: We can remove the `return` s here

core/src/main/kotlin/utilities/json.kt
```diff
@@ -0,0 +1,69 @@
...
+private object SourceRootSerializer : StdScalarSerializer<SourceRoot>(SourceRoot::class.java) {
+    override fun serialize(value: SourceRoot, g: JsonGenerator, provider: SerializerProvider) {
+        g.writeString(value.path)
+    }
+}
+
+private object SourceRootImplDeserializer : StdScalarDeserializer<SourceRootImpl>(SourceRootImpl::class.java) {
+    override fun deserialize(p: JsonParser, ctxt: DeserializationContext): SourceRootImpl {
+        return SourceRootImpl(p.text)
+    }
```
kamildoleglo, line 68: We can also omit the `return` there

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt
```diff
@@ -18,7 +18,7 @@ abstract class AbstractDokkaTask : DefaultTask(), Configurable {
     var outputDirectory: String = defaultDokkaOutputDirectory().absolutePath
 
     @Input
-    override val pluginsConfiguration: Map<String, String> = mutableMapOf()
+    override val pluginsConfiguration: MutableMap<String, String> = mutableMapOf()
```
kamildoleglo, line 21: (Comment to this property, not this PR)
 I'm not sure if Gradle DSL allows `val` s as it has to create a decorator for this property

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/configurationImplementations.kt
```diff
@@ -34,8 +34,8 @@ class GradleSourceRootImpl : SourceRoot, Serializable {
 }
 
 open class GradleDokkaSourceSet constructor(
-    @Transient @get:Input val name: String,
-    @Transient @get:Internal internal val project: Project
+    @get:JsonIgnore @Transient @get:Input val name: String,
+    @get:JsonIgnore @Transient @get:Internal internal val project: Project
```
kamildoleglo, line 38: `@Transient` was used to indicate that those properties shouldn't be serialized by GSON. If Jackson ignores it, we can remove them and only leave the `JsonIgnore` IMHO (and everywhere below)

core/src/main/kotlin/configuration.kt
sellmair: Done ✅

core/src/main/kotlin/configuration.kt
sellmair: Done ✅

core/src/main/kotlin/utilities/json.kt
sellmair: Hihi, I actually think the opposite: There is no reason for scoping, so I only see using a scoping function as making it even harder to parse (since scopes are expensive for me to parse in my head 🙄 )

core/src/main/kotlin/utilities/json.kt
sellmair: You're right. This was from a time that I tinkered with the `typeOf` function here. 
Removed it ✅

core/src/main/kotlin/utilities/json.kt
sellmair: Done ✅

core/src/main/kotlin/utilities/json.kt
sellmair: Done ✅

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt
sellmair: Added configuration for this to the BasicGroovyIntegrationTest and it seems to accept it 👍

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/configurationImplementations.kt
sellmair: Transient annotation is still necessary for Serializable to work (see the new test commitet 🎊 )

sellmair:
Cherry picked into 1.4.0-rc 🍒

PR #1160, opened 2020-07-16
Only register collector and multimodule tasks on projects that actually have children

sellmair:


sellmair:
Cherry picked into 1.4.0-M3 🍒

PR #1164, opened 2020-07-16
Fix signature duplication when filtering inherited functions

kamildoleglo:


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -62,9 +62,9 @@ open class DefaultPageCreator(
     open fun pageForTypeAlias(t: DTypeAlias) = MemberPageNode(t.name, contentForTypeAlias(t), setOf(t.dri), t)
 
     private val WithScope.filteredFunctions: List<DFunction>
-        get() = functions.flatMap { dFunction ->
-            dFunction.sourceSets.mapNotNull { sourceSetData ->
-                dFunction.takeIf { it.extra[InheritedFunction]?.isInherited(sourceSetData) != true }
+        get() = functions.mapNotNull { it ->
+            it.takeIf { function ->
+                function.sourceSets.any { function.extra[InheritedFunction]?.isInherited(it) != true }
```
sellmair, line 67: Using function and it on the same line might be rather confusing. Why not starting with 
```
functions.mapNotNull { function -> 
```

and use "function" from there

sellmair:
Cherry picked into 1.4.0-M3 🍒

PR #1174, opened 2020-07-17
Fixing javadoc comment parser for psi files

sellophane:


sellmair:
Please make sure to also push tests for this ☺️

sellmair:
@sellophane Can you investigate the test?S

plugins/base/src/main/kotlin/translators/psi/JavadocParser.kt
```diff
@@ -119,88 +131,118 @@ class JavadocParser(
...
+
+        private fun PsiElement.toDocumentationLinkString(
+            labelElement: PsiElement? = null
+        ): String? =
+            reference?.resolve()?.let {
+                if (it !is PsiParameter) {
+                    val dri = DRI.from(it)
+                    driMap[dri.toString()] = dri
+                    val label = labelElement ?: defaultLabel()
+                    """<a data-dri=$dri>${label.text}</a>"""
```
MarcinAman, line 166: I don't know if the assumption that the output format will always be able to render html is correct. This is basically coupling the JavadocParser with html-based output

PR #1176, opened 2020-07-17
Disable DokkaCollectorTask for all formats and DokkaMultimoduleTask for javadoc

sellmair:


Kordyjan:
`org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find org.jetbrains.kotlinx:kotlinx-html-jvm:0.6.10`🤔

PR #1177, opened 2020-07-17
Repair merger not to merge same name classlikes from different independent sources

BarkingBad:

PR #1180, opened 2020-07-17
Create sourceset documentable filter for javadoc

MarcinAman:
PR is waiting for #1177 to be finished due to merger changes

### DIFF
--- a/plugins/javadoc/src/main/kotlin/javadoc/transformers/documentables/JavadocDocumentableSourceSetFilterTransformer.kt
+++ b/plugins/javadoc/src/main/kotlin/javadoc/transformers/documentables/JavadocDocumentableSourceSetFilterTransformer.kt
@@ -1,0 +1,16 @@
+package javadoc.transformers.documentables
+
+import org.jetbrains.dokka.Platform
+import org.jetbrains.dokka.model.DModule
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.transformers.documentation.PreMergeDocumentableTransformer
+
+class JavadocDocumentableSourceSetFilterTransformer : PreMergeDocumentableTransformer {
+    private val List<Documentable>.jvmSourceSets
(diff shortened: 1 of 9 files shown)


plugins/javadoc/src/main/kotlin/javadoc/transformers/documentables/JavadocDocumentableSourceSetFilterTransformer.kt
```diff
@@ -0,0 +1,16 @@
+package javadoc.transformers.documentables
+
+import org.jetbrains.dokka.Platform
+import org.jetbrains.dokka.model.DModule
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.transformers.documentation.PreMergeDocumentableTransformer
+
+class JavadocDocumentableSourceSetFilterTransformer : PreMergeDocumentableTransformer {
+    private val List<Documentable>.jvmSourceSets
+        get() = flatMap { it.sourceSets }.filter { it.analysisPlatform == Platform.jvm }.distinct()
```
kamildoleglo, line 10: I'd personally also leave the `common` source sets that those jvm source sets depend upon

MarcinAman: uh ok, that means that i need to also figure out how to merge those sourcesets to ensure javadoc only works on 1 of them

kamildoleglo: Why, can't we just use the Documentable merger as always? 

Also, I'd log an info that some source sets were dropped

kamildoleglo: Bump

PR #1189, opened 2020-07-20
Re-introduce default DokkaCollectorTask, run on all submodules by def…

sellmair:
…ault, add Collector0IntegrationTest


sellmair:
Cherry picked into 1.4.0-rc 🍒

PR #1190, opened 2020-07-20
Implement dokka_integration_test_parallelism gradle property and set default to 2

sellmair:


kamildoleglo:
Maybe we can try something like 4 by default? 🤔

sellmair:
> Maybe we can try something like 4 by default? 🤔

I think every developer is free to increase this ☺️ 
But let's keep it low for people with low-end machines. 
(I think the projects default settings should work with almost all machines)

PR #1191, opened 2020-07-20
Wrap superclasses in Javadoc

MarcinAman:
This is quite sad since the bug is also present in javadoc (even for Java 14), so using custom styles in the only option here
image


sellmair:
Cherry-picked into 1.4.0-rc 🍒

PR #1193, opened 2020-07-21
Include information about local publish in README.md

MarcinAman:


sellmair:
Why are people not using our nightly dev builds? 🙄

MarcinAman:
> Why are people not using our nightly dev builds? 🙄

Maybe they use but imho it is worth to at least inform that there is an option to easily build new versions of dokka. This is also important if somebody would like to contribute and see how the changes affect the documentation

sellmair:
I see: This might be good informations for developers with little Gradle experience. You're right!
The issue is, that the Readme is already pretty long and I am not sure if including this information is really worth it. 
Would be interested in your opinion @kamildoleglo !

kamildoleglo:
I'd put it in the CONTRIBUTING.md

MarcinAman:
Ok, done

PR #1195, opened 2020-07-21
Temporarily fix coroutines publication

kamildoleglo:


sellmair:
I cherry picked this into 1.4.0-rc 🍒 
There is no potential harm for users, but this will enable us to compare coroutines across branches.

PR #1216, opened 2020-07-23
Navigation pane improvements

MarcinAman:
Small PR for low-hanging fruits like sorting in navigation pane on left and improving performance of CSS. 
I'd CSS to make it to RC.

You can check if samples still work on stdlib: `/kotlin-stdlib/kotlin-stdlib/kotlin.sequences/as-sequence.html`
I havn't found any sample that is in tab so here is a screen for our project:

Jul-23-2020 10-10-50


sellmair:
Cherry-picked into 1.4.0-rc 🍒

PR #1221, opened 2020-07-24
Bugfix for bad package of TestLogger

pikinier20:


kamildoleglo:
Could you please also move the `testApi.utils` in this PR

PR #1225, opened 2020-07-27
Fix error message when no dokka source sets are registered

sellmair:


sellmair:
Cherry picked into 1.4.0-rc 🍒

PR #1233, opened 2020-07-30
Fix for creating searchbar pages data - escaping apostrophes

pikinier20:


plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -107,3 +113,94 @@ class SourcesetDependencyAppender(val context: DokkaContext) : PageTransformer {
         )
     }
 }
+
+class SearchbarDataInstaller(val context: DokkaContext): PageTransformer {
+    private val pageList = mutableMapOf<String, Pair<String, String>>()
+    lateinit var locationProvider : LocationProvider
+
+        override fun invoke(input: RootPageNode): RootPageNode {
+            locationProvider = context.plugin<DokkaBase>().querySingle { locationProviderFactory }
```
MarcinAman, line 122: I am not so sure about fetching location provider like that. It requires a root page that can be changed in other `PageTransformer`, so unless you are entirely sure that this page transformer is last i'd just create this object in HtmlRenderer and pass location provider via constructor

plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -107,3 +113,94 @@ class SourcesetDependencyAppender(val context: DokkaContext) : PageTransformer {
...
+            )
+
+            return input.modified(
+                children = input.children + data
+            )
+    }
+
+    private fun traverse(page:PageNode): Unit {
+        if(page is ContentPage && page !is ModulePageNode && page !is PackagePageNode) processPage(page)
+        page.children.forEach(::traverse)
```
MarcinAman, line 138: [Minor] Imho a tailrec function with accumulator would look better than side effect

plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
pikinier20: I've changed usage of SearchbarDataInstaller - I don't declare it as PageTransformer and use it in HtmlRenderer to register and finally to generate page list.

plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
pikinier20: I deleted this recursion as it was failing with strange exceptions when tried to run it after rendering

PR #1234, opened 2020-07-30
Cache redirector

sellmair:


IgnatBeresnev:
The PR seems stale and outdated, we'll re-implement it sometime in the future if needed

PR #1237, opened 2020-07-30
Javadoc package-list

MarcinAman:


Kordyjan:
I understand opening classes for page nodes, but please keep in mind two things:
- methods that are an implementation of some abstract method and are not supposed to be overridable down the line should be marked as `final override`.
- Do not declare properties in inheriting class constructor if they are already declared in base class. It should look like in the following example:
```kotlin
interface A {
    val a: Int
}

class B(
    final override val a: Int 
): A 

class C(a: Int): B(a) // here `a` is ordinary parameter, not property
```
This should result in less bloated bytecode and easier debugging.

MarcinAman:
@Kordyjan something like in latest commit?: https://github.com/Kotlin/dokka/pull/1237/commits/a849b476fd2c1160c9638fb2ee9a82515a6dae0a

Kordyjan (approved):

sellmair:
This is not for the 1.4.0 release? 🙄

sellmair:
Cherry picked into 1.4.0 🍒

PR #1244, opened 2020-08-03
Index page for javadoc

sellophane:


gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Project Settings
-dokka_version_base=1.4.0-rc
+dokka_version_base=0.11.0
```
sellmair, line 2: I think this was not an intended change, right?

sellophane: yes, definitely not

sellmair:
Cherry picked into 1.4.0 🍒

PR #1259, opened 2020-08-04
Highlight anchor

MarcinAman:


MarcinAman:
Oh XD
I guess another question is do we want to do it this way? Tbh that was the first thought that i had on this issue. Maybe you have something better. 
It is fairly visible but quite coarse imho, maybe making it blue would help to some degree

Kordyjan:
My idea was to change the border in similar fashion how Github does that when displaying anchored comments. Click: https://github.com/Kotlin/dokka/pull/1259#issuecomment-669281261 to see what I mean on your comment.

MarcinAman:
ok, thats a nice idea:
image

PR #1260, opened 2020-08-04
Fix broken link in README.md

drewhannay:
"Dokka Gradle example project" was pointing to a path that no longer exists


sellmair:
Thanks for your contribution ☺️

PR #1271, opened 2020-08-05
Fix broken link in README.md

vladimirshefer:
Without `https://` prefix the url leads to `https://github.com/Kotlin/dokka/blob/master/github.com/Kotlin/dokka/releases/download/v1.4.0-rc/dokka-cli-1.4.0-rc.jar` as a relative url.


Kordyjan:
Whoops! I was too quick with merging the previous pr. Thanks for another contribution.

PR #1273, opened 2020-08-05
Create navigation search component

MarcinAman:


core/src/main/kotlin/pages/RendererSpecificPage.kt
```diff
@@ -1,8 +1,12 @@
 package org.jetbrains.dokka.pages
 
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.renderers.Renderer
 import kotlin.reflect.KClass
 
+typealias LocationResolver = (DRI, Set<DokkaConfiguration.DokkaSourceSet>) -> String
```
MarcinAman, line 8: Imho this is the most questionable line of this PR. The reason behind it is that `LocationProvider` is introduced on the `base` plugin level and here we are not even in plugin

core/src/main/kotlin/pages/RendererSpecificPage.kt
```diff
@@ -1,8 +1,12 @@
 package org.jetbrains.dokka.pages
 
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.DisplaySourceSet
 import org.jetbrains.dokka.renderers.Renderer
 import kotlin.reflect.KClass
 
+typealias LocationResolver = (DRI, Set<DisplaySourceSet>) -> String
```
Kordyjan, line 8: I really like this idea. If you want you can even change it from `typealias` to `fun interface`.

plugins/base/frontend/src/main/components/navigationPaneSearch/navigationPaneSearch.scss
```diff
@@ -0,0 +1,37 @@
...
+        background: $white;
+        border: 1px solid $grey-border;
+        box-sizing: border-box;
+        border-radius: 4px;
+        padding: 8px;
+        height: $defaultHeight;
+    }
+    
+    .navigation-pane-search {
+        width: 100% !important;
```
Kordyjan, line 20: Is this `!important` really necessary?

plugins/base/frontend/src/main/components/navigationPaneSearch/navigationPaneSearch.tsx
```diff
@@ -0,0 +1,73 @@
...
+        (element: Option) => {
+            window.location.replace(`${(window as IWindow).pathToRoot}${element.location}`)
+            onSelected(element);
+        },
+        [selected]
+    );
+
+    const onFilter = (filterValue: string, filteredRecords?: Option[]) => {
+        if(filteredRecords){
+            const requiredWidth = Math.max(...filteredRecords.map(e => e.label.length*9), defaultWidth)
```
Kordyjan, line 26: Could you please extract this magic `9` into constant. It is not obvious why is it here and it can quickly become unmaintainable.

plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -63,6 +87,11 @@ object NavigationPageInstaller : PageTransformer {
                 children.filter { it is ContentPage && it.documentable is DEnumEntry }.map { visit(it as ContentPage) }
             else -> emptyList()
         }.sortedBy { it.name.toLowerCase() }
+
+    private tailrec fun flattenNavigationNodes(nodes: List<NavigationNode>, acc: List<NavigationNode> = emptyList()): List<NavigationNode> {
+        if(nodes.isEmpty()) return acc
+        return flattenNavigationNodes(nodes.flatMap { it.children }, nodes.map { it.copy(children = emptyList()) } + acc)
+    }
```
Kordyjan, line 94: This recursive function is really fancy but making `NavigationNode` implement `WithChildren` and then calling `.withDescendants` is simpler, more consistent and is guaranteed to be lazily evaluated.

Kordyjan (changes requested):

plugins/base/frontend/src/main/components/navigationPaneSearch/navigationPaneSearch.scss
MarcinAman: Tbh i don't see another solution. The issue is that this component requires me to provide a min-width value in pixels (by default 240px), so i can make it exactly 248px (280px - paddings) but i'd much rather tell it to be 100% of the container it is in. The `!important` part is caused by the fact, that this value in pixels is provided to this component via styles in html so it is impossible to provide a more specific selector.

PR #1295, opened 2020-08-10
Adding missing subclasses pages

sellophane:


MarcinAman (approved):
LGTM, we can think about adding a dot in between the outer class and the inner one on the package view as is in javadoc

Kordyjan:
@sellophane Please rebase and merge it.

PR #1303, opened 2020-08-12
Fix generic supertypes to hold TypeParameters

BarkingBad:


BarkingBad:
obraz
obraz
From Java sources
obraz

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -182,9 +183,14 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
                 }
             }
             if (c is WithSupertypes) {
-                c.supertypes.filter { it.key == sourceSet }.map { (s, dris) ->
-                    list(dris, prefix = " : ", sourceSets = setOf(s)) {
-                        link(it.dri.sureClassNames, it.dri, sourceSets = setOf(s))
+                c.supertypes.filter { it.key == sourceSet }.map { (s, typeConstructors) ->
+                    list(typeConstructors, prefix = " : ", sourceSets = setOf(s)) {
+                        link(it.typeConstructor.dri.sureClassNames, it.typeConstructor.dri, sourceSets = setOf(s))
+                        if ( it.typeConstructor.projections.isNotEmpty() ) {
```
MarcinAman, line 189: I think this if is already included in list function

BarkingBad: Oh, yes, thanks :D

BarkingBad:
obraz

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -128,22 +125,27 @@ class DefaultPsiToDocumentableTranslator(
...
             fun parseSupertypes(superTypes: Array<PsiClassType>, level: Int = 0) {
                 if(superTypes.isEmpty()) return
-                val parsedClasses = superTypes.filter { !it.shouldBeIgnored }.mapNotNull {
-                    it.resolve()?.let {
+                val parsedClasses = superTypes.filter { !it.shouldBeIgnored }.mapNotNull { psi ->
+                    psi.resolve()?.let {
                         when {
                             it.isInterface -> DRI.from(it) to JavaClassKindTypes.INTERFACE
                             else -> DRI.from(it) to JavaClassKindTypes.CLASS
+                        }?.let {
```
sellmair, line 139: IMO this gets a little bit convoluted again here. Maybe you could give some `it` a name or just use assignments. This scoping is not the most difficult to parse, that I have seen in the project, but its still not easy either.

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -555,28 +554,34 @@ private class DokkaDescriptorVisitor(
...
 
     private fun ClassDescriptor.resolveClassDescriptionData(): ClassInfo {
-        tailrec fun buildInheritanceInformation(
-            inheritorClass: ClassDescriptor?,
-            interfaces: List<ClassDescriptor>,
+
+        fun toTypeConstructor(kt: KotlinType) =
+            TypeConstructor(DRI.from(kt.constructor.declarationDescriptor as DeclarationDescriptor), kt.arguments.map { it.toProjection() })
+
+        tailrec fun buildAncestryInformation(
```
sellmair, line 561: I know this was not introduced with this PR,; It still seems a little bit strange to me, why you are trying to represent a tree like structure within a flat `Set`? 🤔

BarkingBad: As far as I know it was introduced by @MarcinAman. I guess the clue is we don't need to preserve the relations between two given ancestors, we focus only on distance from given node. Therefore, the Set with `level` property is much more convenient, since you don't have to write your own abstraction and add utils for e. g. traversing the structure and collecting certain nodes. It can be rewritten, however I don't know whether it is the crucial task for now.

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
BarkingBad: Done

Issue #1258, opened 2020-08-04
Anchors are not unique


MarcinAman:
In some situations anchors are not unique, which means that linking does not work correctly

Example of it would be `example/example/main.html` in https://github.com/kamildoleglo/dokka-example where `Sources` all have the same anchor: `example//main/#/PointingToDeclaration/`. Same thing is visible on `example/example/-clock/index.html` where anchor copied from inheritors redirects to samples.

PR #1307, opened 2020-08-13
Anchors hint

MarcinAman:
Basically anchors should right now work in the same way they did before


plugins/base/src/main/resources/dokka/scripts/platformContentHandler.js
```diff
@@ -42,15 +42,14 @@ function handleAnchor() {
...
-                let found = document.querySelector('.tabs-section > .section-tab[data-togglable="' + tab.getAttribute("data-togglable") + '"]')
                 toggleSections(tab)
-                const content = element.nextElementSibling
-                if(content){
-                    content.classList.add('anchor-highlight')
-                    highlightedAnchor = content
-                }
-                element.scrollIntoView({behavior: "smooth"})
             }
+            const content = element.nextElementSibling
```
MarcinAman, line 47: This change is unrelated to anchors rendering, but I saw that anchors are not highlighted correctly when there is no tab (eg. on module page)

Kordyjan (changes requested):
In my opinion this pr should add potential anchors for all table rows. Of course, I'm open for discussion.

MarcinAman:
> In my opinion this pr should add potential anchors for all table rows. Of course, I'm open for discussion.

I thought about it and i don't see a reason why we should preemptively generate anchors. Also this is super bad in current state since anchors are not unique (see https://github.com/Kotlin/dokka/issues/1258 and anchors don't have a sourceset in them so every sourceset tab would be broken). Using an extra is a decent idea since it gives us control over what we want to link to and introduces a straight forward logic how to do it.

BarkingBad (commented):
I think currently we currently generate anchors for every element possible (I think I added that to handle typealiases and properties (which will have their own pages since #1494)) so we should check that PR for any possible inconvenient behaviour.

kamildoleglo (approved):

kamildoleglo (approved):

PR #1308, opened 2020-08-13
DokkaCollectorTask: Merge source set bubbles

### LINKED ISSUE #1294: DokkaCollectorTask showing confusing and seemingly unnecessary source set bubbles
E.g. running `dokkaHtmlCollector` on spring framework.
image

sellmair:
Closes #1294


core/src/main/kotlin/pages/ContentNodes.kt
```diff
@@ -182,11 +245,17 @@ data class ContentDivergentGroup(
...
+    override val sourceSets: Set<ContentSourceSet>
         get() = children.flatMap { it.sourceSets }.distinct().toSet()
 
     override fun withNewExtras(newExtras: PropertyContainer<ContentNode>): ContentDivergentGroup =
         copy(extra = newExtras)
+
+    override fun transformChildren(transformer: (ContentNode) -> ContentNode): ContentDivergentGroup =
+        copy(children = children.map(transformer).filterIsInstance<ContentDivergentInstance>())
+
+    // TODO NOW?
```
sellmair, line 257: There is no intuitive behaviour for me here. I think this should be tackled when we think about ContentNode<->SourceSet associations in the future!

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/JavadocContentNodes.kt
```diff
@@ -12,12 +11,16 @@ enum class JavadocContentKind : Kind {
...
     kind: Kind,
-    override val sourceSets: Set<DokkaSourceSet>
+    override val sourceSets: Set<ContentSourceSet>
 ) : ContentNode {
     override val dci: DCI = DCI(dri, kind)
     override val style: Set<Style> = emptySet()
     override val extra: PropertyContainer<ContentNode> = PropertyContainer.empty()
     override fun withNewExtras(newExtras: PropertyContainer<ContentNode>): ContentNode = this
+
+    // TODO: Support needed?
```
sellmair, line 21: I think we should implement support for this, but data classes would really help here 🙄

core/src/main/kotlin/model/ContentSourceSet.kt
```diff
@@ -0,0 +1,32 @@
+package org.jetbrains.dokka.model
+
+import org.jetbrains.dokka.DokkaConfiguration.DokkaSourceSet
+import org.jetbrains.dokka.DokkaSourceSetID
+import org.jetbrains.dokka.Platform
+
+data class ContentSourceSet(
```
sellmair, line 7: Naming alternatives welcome. I am not 100% happy with this name :/

core/src/main/kotlin/model/CompositeSourceSetID.kt
```diff
@@ -0,0 +1,30 @@
+package org.jetbrains.dokka.model
+
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.DokkaSourceSetID
+
+data class CompositeSourceSetID(
```
sellmair, line 6: Naming alternatives: 
- `MultiSourceSetID` 
- `SourceSetIDComposite`

core/src/main/kotlin/model/ContentSourceSet.kt
sellmair: Alternaives: 
- `DisplaySourceSet`  -> "Because this is what is displayed
- `DisplayableSourceSet` ->".."
- `SourceSetBubble`-> "Because in the end, it represents those bubbles"
- `MergedSourceSet` -> Because it can represent multiple original source sets
- `CompositeSourceSet`-> ...
... 

Please, help 🙈

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/JavadocContentNodes.kt
sellmair: Done ✅

core/src/main/kotlin/model/CompositeSourceSetID.kt
Kordyjan: `CompositeSourceSetID` is the best IMO.

core/src/main/kotlin/model/ContentSourceSet.kt
Kordyjan: `CompositeSourceSet` would be most consistent with `ContentSourceSetId`. But on the other hand the most important thing in this class is not that it is any composite of source sets, but composite that can be displayed in the final documentation. I would go with `DisplayableSourceSet` / `PresentationSourceSet`.

plugins/base/src/main/kotlin/transformers/pages/merger/SourceSetMergingPageTransformer.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class SourceSetMergingPageTransformer(context: DokkaContext) : PageTransformer {
+
+    private val mergedSourceSets = context.configuration.sourceSets.toContentSourceSets()
+        .associateBy { sourceSet -> sourceSet.key }
+
+    override fun invoke(input: RootPageNode): RootPageNode {
+        return input.transformContentPagesTree { contentPage ->
+            val content: ContentNode = contentPage.content
+            val newContent = transformWithMergedSourceSets(content)
```
MarcinAman, line 18: nit: imho you can inline it to: `val newContent = transformWithMergedSourceSets(contentPage.content)`

sellmair: Good point ☺️ 
Done ✅

PR #1310, opened 2020-08-14
Add generic typealiases proper parsing

BarkingBad:
obraz

### DIFF
--- a/core/src/main/kotlin/model/Documentable.kt
+++ b/core/src/main/kotlin/model/Documentable.kt
@@ -332,8 +332,9 @@ data class DTypeAlias(
     override val documentation: SourceSetDependent<DocumentationNode>,
     override val expectPresentInSet: DokkaSourceSet?,
     override val sourceSets: Set<DokkaSourceSet>,
-    override val extra: PropertyContainer<DTypeAlias> = PropertyContainer.empty()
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/model/Documentable.kt
```diff
@@ -332,8 +332,9 @@ data class DTypeAlias(
     override val documentation: SourceSetDependent<DocumentationNode>,
     override val expectPresentInSet: DokkaSourceSet?,
     override val sourceSets: Set<DokkaSourceSet>,
-    override val extra: PropertyContainer<DTypeAlias> = PropertyContainer.empty()
-) : Documentable(), WithType, WithVisibility, WithExtraProperties<DTypeAlias> {
+    override val extra: PropertyContainer<DTypeAlias> = PropertyContainer.empty(),
+    override val generics: List<DTypeParameter>
```
sellmair, line 336: @Kordyjan Do you agree with this solution? The rest of the PR looks good to me ☺️

Kordyjan: Yup, we've forgotten that typealiase can have generic parameters, so it definitely should be here.
Nitpick: by convention and since `generics` is a parameter without default value it should go before the `extra`.

BarkingBad: Updated

PR #1317, opened 2020-08-17
Javadoc deprecated page

sellophane:


plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/JavadocPageNodes.kt
```diff
@@ -262,9 +272,60 @@ class AllClassesPage(val classes: List<JavadocClasslikePageNode>) : JavadocPageN
 
 }
 
+class DeprecatedPage(
+    val elements: IdentityHashMap<DeprecatedPageSection, MutableSet<DeprecatedNode>>,
```
Kordyjan, line 276: I really don't like mutability in page model. Why cannot it be expressed immutably?

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/htmlPreprocessors.kt
```diff
@@ -90,4 +96,61 @@ object IndexGenerator: PageTransformer {
             IndexPage(i + 1, set.sortedBy { it.getId().toLowerCase() }, keys, input.sourceSets())
         })
     }
+}
+
+object DeprecatedPageCreator : PageTransformer {
+    override fun invoke(input: RootPageNode): RootPageNode {
+        val elements = IdentityHashMap<DeprecatedPageSection, MutableSet<DeprecatedNode>>().apply {
```
Kordyjan, line 103: If `DeprecatedPageSection` is made into enum this identity map can be just an ordinary map.

Kordyjan (changes requested):

PR #1321, opened 2020-08-17
Move plugin version management to buildSrc/build.gradle.kts

floscher:
While researching a similar issue, I recently came across https://github.com/gradle/gradle/issues/11090#issuecomment-666249417 (cc @sellmair ) about how the `pluginManagement{}` block in `settings.gradle.kts` can't access logic defined in `buildSrc`.

So I thought I'd share how I'm trying to solve that problem in some of my projects. Just in case this could be useful for this project, too.

There are two parts to this solution:

---

There is a second `buildSrc` project inside the existing `buildSrc` project. It only contains a short `build.gradle.kts` script and just compiles the sources of the first `buildSrc` project one additional time, so those classes are on the classpath of `buildSrc/build.gradle.kts`.

At the moment this is not utilized in this pull request, but you could then add build logic to `buildSrc/src/main/kotlin` that can be reused in the buildscript of `buildSrc` for e.g. plugin version management.

This may sound like a hack, but in my experience this works pretty reliably.

---

The plugin versions are defined in `buildSrc/build.gradle.kts` as `dependencies.implementation("‹…›")` instead of using the `pluginManagement{}` block in `settings.gradle.kts`.

This gets rid of version inconsistencies like https://github.com/Kotlin/dokka/blob/b19ce3c7622666bd638baf2f840e898bb00efd90/settings.gradle.kts#L31 vs https://github.com/Kotlin/dokka/blob/b19ce3c7622666bd638baf2f840e898bb00efd90/buildSrc/build.gradle.kts#L11 Both are now defined in one location.

Also subprojects can then no longer declare conflicting plugin versions like https://github.com/Kotlin/dokka/blob/b19ce3c7622666bd638baf2f840e898bb00efd90/settings.gradle.kts#L32 and https://github.com/Kotlin/dokka/blob/b19ce3c7622666bd638baf2f840e898bb00efd90/runners/gradle-plugin/build.gradle.kts#L6 because Gradle will issue a warning: `Plugin request for plugin already on the classpath must not include a version`.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -2,12 +2,36 @@ plugins {
     `kotlin-dsl`
 }
 
-repositories {
-    mavenCentral()
-    jcenter()
+/**
+ * These plugins are added to the buildscript classpath, so subprojects can refer to the plugin in the plugins block
+ * by using the [PluginDependenciesSpec.id] function, without the need to also call [PluginDependencySpec.version].
+ */
+val pluginVersions = setOf(
+  "com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.5",
+  "com.github.jengelman.gradle.plugins:shadow:5.2.0",
+  "com.gradle.publish:plugin-publish-plugin:0.12.0"
+).map {
+    dependencies.implementation(it) ?: throw NullPointerException("Plugin dependency must not be null!")
 }
 
-dependencies {
-    implementation("com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4")
-    implementation("com.github.jengelman.gradle.plugins:shadow:2.0.4")
+repositories {
+    exclusiveContent {
+        forRepository {
+            gradlePluginPortal()
+        }
+        filter {
+            pluginVersions.forEach { includeVersion(it.group!!, it.name, it.version!!) }
+        }
+    }
+    jcenter()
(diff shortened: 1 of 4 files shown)


buildSrc/build.gradle.kts
```diff
@@ -2,12 +2,36 @@ plugins {
...
+        content {
+            includeGroupByRegex("^org\\.jetbrains\\.(anko|dokka|kotlin|kotlinx)(\\..+)?$")
+        }
+    }
+    maven("https://dl.bintray.com/kotlin/kotlin-dev/") {
+        content {
+            includeGroupByRegex("^org\\.jetbrains\\.(anko|dokka|intellij|kotlin|kotlinx)(\\..+)?$")
+        }
+    }
 }
```
floscher, line 37: Here, I added an `exclusiveContent` rule for the plugins to only be taken from `gradlePluginPortal()` and some regex rules for `kotlin-eap` and `kotlin-dev` to avoid unnecessary requests to those repos.

This could be replaced by some more elaborate logic in `buildSrc/src/main/kotlin` that could then be reused in any other Gradle buildscript in the project (except `buildSrc/buildSrc/build.gradle.kts`).

IgnatBeresnev:
Thank you for the effort! Sorry it hasn't been addressed properly. 

I'll close this PR in favour of #2704, there's quite a bit of refactoring, including the move of plugin version management to buildSrc.

PR #1332, opened 2020-08-17
Hide unused generic type variables in constructor

BarkingBad:


BarkingBad:
obraz

plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocClasslikeTemplateMapTest.kt
```diff
@@ -312,7 +312,7 @@ internal class JavadocClasslikeTemplateMapTest : AbstractJavadocTemplateMapTest(
             val ownMethods = assertIsInstance<List<*>>(methods["own"]).first()
             val sampleFunction = assertIsInstance<Map<String, Any?>>(ownMethods)
 
-            assertEquals("final &lt;D extends <a href=Generic.html>T</a>&gt; <a href=Generic.html#sampleFunction()>D</a> <a href=Generic.html#sampleFunction()>sampleFunction</a>()", sampleFunction.signatureWithModifiers())
+            assertEquals("final &lt;D extends <a href=Generic.html#Generic>T</a>&gt; <a href=Generic.html#sampleFunction()>D</a> <a href=Generic.html#sampleFunction()>sampleFunction</a>()", sampleFunction.signatureWithModifiers())
```
MarcinAman, line 315: I am not so sure about that. In original javadoc the link is going to `Generic.html` and afaik there is no `Generic` anchor on that page. Why is this change needed?

BarkingBad: After applying changes, it was value that was calculated. Moreover, Kotlin and Javadoc link providers were producing given link, which seems reasonable, so I dedided to change the test expected value. If we want it to be without anchor, we need to consider why location provider skips the usual link. I will debug it soon so we will know.

plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
```diff
@@ -133,6 +135,10 @@ interface JvmSignatureUtils {
                 it.dri == DRI("kotlin", "Deprecated")
                         || it.dri == DRI("java.lang", "Deprecated")
             } == true) setOf(TextStyle.Strikethrough) else emptySet()
+
+    infix fun DFunction.uses(t: DTypeParameter): Boolean =
+        t.dri in (listOfNotNull(receiver?.type?.drisOfAllNestedBounds, receiver?.dri) +
+                parameters.flatMap { listOf(it.dri) + it.type.drisOfAllNestedBounds })
```
Kordyjan, line 141: This is wrong. The list you are looking at has inferred type `List<Any>` since it contains both `DRI`s and `List<DRI>`s.

Kordyjan (changes requested):

plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
BarkingBad: Done

Kordyjan (approved):

sellmair:
Cherry picked into 1.4.0 🍒

PR #1355, opened 2020-08-20
Make MarkdownParser output text instead of broken links

### LINKED ISSUE #1345: S3 Coroutines unresolved links
Related to markdown on the module page

kamildoleglo:
MarkdownParser should also output info about the unresolved link and its location. This however should be done properly when refactoring the parser 

Closes #1345

### DIFF
--- a/plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
+++ b/plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
@@ -1,7 +1,6 @@
 package org.jetbrains.dokka.base.parsers
 
 import com.intellij.psi.PsiElement
-import org.jetbrains.dokka.model.doc.*
 import org.intellij.markdown.MarkdownElementTypes
 import org.intellij.markdown.MarkdownTokenTypes
 import org.intellij.markdown.ast.ASTNode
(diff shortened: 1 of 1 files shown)


sellmair:
Do you want this being merged into 1.4.0?

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -159,17 +161,37 @@ class MarkdownParser(
...
+                    MarkdownElementTypes.INLINE_LINK,
+                    params = params,
+                    children = linkText.children.drop(1).dropLast(1).evaluateChildren(),
+                    dri = dri
+                )
+            }
+        }
+
+        private fun String.isRemoteLink() = try {
+            URL(this)
```
sellmair, line 191: I was looking for some more straight-forward way of checking this. 
@Kordyjan Do you know a better method? 🙄

kamildoleglo:
> Do you want this being merged into 1.4.0?

Not sure 🤔 We can do that I think

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
sellmair: I think this is the way many people do this. It's not the nicest API, but seems to work for others as well 🤷

PR #1367, opened 2020-08-21
Module page feedback

MarcinAman:


MarcinAman:
For the reference, now it looks like this:
image

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -308,22 +310,18 @@ open class DefaultPageCreator(
     protected open fun contentForComments(
         d: Documentable
     ): List<ContentNode> {
-        val tags: GroupedTags = d.documentation.flatMap { (pd, doc) ->
-            doc.children.asSequence().map { pd to it }.toList()
-        }.groupBy { it.second::class }
-
```
Kordyjan, line 314: I don't get this change. Now you are evaluating `groupedTags` every time you access it, so 7 times inside of this function. This is bad.

plugins/base/src/main/kotlin/translators/documentables/briefFromContentNodes.kt
```diff
@@ -0,0 +1,26 @@
+package org.jetbrains.dokka.base.translators.documentables
+
+import org.jetbrains.dokka.pages.ContentGroup
+import org.jetbrains.dokka.pages.ContentNode
+import org.jetbrains.dokka.pages.ContentText
+import org.jetbrains.dokka.pages.TextStyle
+
+fun briefFromContentNodes(description: List<ContentNode>): List<ContentNode> {
+    val firstSentenceRegex = Regex("^((?:[^.?!]|[.!?](?!\\s))*[.!?])")
```
Kordyjan, line 9:
```suggestion
    val firstSentenceRegex = """^((?:[^.?!]|[.!?](?!\s))*[.!?])""".toRegex()
```

Kordyjan: I know it is only moved from another place, but this is not the idiomatic way of doing that.

Kordyjan (changes requested):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
MarcinAman: Uh ok, my bad
changed to `val tags = d.groupedTags`

sellmair:
Cherry picked into 1.4.0 🍒

PR #1370, opened 2020-08-23
Update README.md

tom5079:
File.resolve can't be used in Groovy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ tasks.dokkaHtml.configure {
 
 Groovy
 ```kotlin
-tasks.named("dokkaHtml") {
-    outputDirectory.set(buildDir.resolve("dokka"))
+tasks.dokkaHtml {
+    outputDirectory = "$buildDir/dokka"
 }
 ```
 


README.md
```diff
@@ -115,8 +115,8 @@ tasks.dokkaHtml.configure {
 
 Groovy
 ```kotlin
-tasks.named("dokkaHtml") {
-    outputDirectory.set(buildDir.resolve("dokka"))
+tasks.dokkaHtml {
+    outputDirectory = "$buildDir/dokka"
```
sellmair, line 119: The `.set` is still necessary in Groovy also. We are trying to investigate this, soon.

README.md
```diff
@@ -115,8 +115,8 @@ tasks.dokkaHtml.configure {
 
 Groovy
 ```kotlin
```
sellmair, line 117: We might also want to fix this here also, right? 🙄

sellmair (commented):
Thanks for the PR ☺️ 
Some minor things would need to be adapted!

README.md
tom5079: I didn't test it so I have no idea how it will work out in Kotlin DSL
I could test it on Kotlin DSL as well if you want me to do so

sellmair: Oh I think we can just replace ```kotlin with ```groovy ☺️

README.md
tom5079: Somehow it doesn't work if I do it with `.set`

```
dokkaHtml {
    outputDirectory.set("$buildDir/dokka")
}
```
results in error `No signature of method: build_2u87vgqvma4krbutdh5r4jh3w.dokkaHtml() is applicable for argument types: (build_2u87vgqvma4krbutdh5r4jh3w$_run_closure8) values: [build_2u87vgqvma4krbutdh5r4jh3w$_run_closure8@3cbe711b]`

but
```
dokkaHtml {
    outputDirectory = "$buildDir/dokka"
}
```
works just fine.

Oh, and the `tasks.` in front of the `dokkaHtml` seems to be redundant

tom5079: I found the reason why it's not working

the `outputDirectory` property was type `Property<File>` in [`master`](https://github.com/Kotlin/dokka/blob/master/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt) branch
but it was changed to `String` in [`1.4.0-RC`](https://github.com/Kotlin/dokka/blob/1.4.0-rc/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt) branch

README.md
tom5079: I think you don't need to specify a language there?

tom5079: I think you should replace `kotlin` with `gradle`  
  
source:  
[Github official guide](https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks#syntax-highlighting)
[Supported Language List](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1891)

README.md
tom5079: I think you have to change the Kotlin DSL part too...?

IgnatBeresnev:
This section has been moved from README.md to [user guides](https://kotlin.github.io/dokka/1.6.10/user_guide/gradle/usage/) a while ago and has been fixed. Thanks!

PR #1372, opened 2020-08-24
Relative paths fix

BarkingBad:
I don't know why, the previous solution was correctly resolving the file, but during serialization it used to lose sometimes correct location (while debuggin I observed `File` serializes to its path, which in case it was provided as relative path, it is serialized to one, which caused trouble while deserialization)

Of course I could change it to be `File(value).toAbsoulteFile()`, but I have more trust in `Paths` since it cooperates well on different OSes

### DIFF
--- a/runners/cli/src/main/kotlin/cli/main.kt
+++ b/runners/cli/src/main/kotlin/cli/main.kt
@@ -246,7 +246,7 @@ private fun parseSourceSet(args: Array<String>): DokkaConfiguration.DokkaSourceS
 }
 
 object ArgTypeFile : ArgType<File>(true) {
-    override fun convert(value: kotlin.String, name: kotlin.String): File = File(value)
+    override fun convert(value: kotlin.String, name: kotlin.String): File = Paths.get(value).toRealPath().toFile()
     override val description: kotlin.String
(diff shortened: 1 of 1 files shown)


sellmair (approved):
Maybe one could even use `toCanonicalPath`

BarkingBad:
Good point, will change

BarkingBad:
Basing on documentation I beliver for `Path` there is `getRealPath` equivalent of `File.getCanonicalPath()` https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toRealPath-java.nio.file.LinkOption...-

sellmair:
Can you give further details on why we would not want tu just use java.io.File here? 🙄

BarkingBad:
It actually is `java.io.File` at the end. Just construction of the `File` object is done via `java.nio` which is newer and more flexible apporach eliminating risk of not resolving path on certain OS

sellmair:
Can you elaborate on the risks more? What risks can happen here?

sellmair (approved):

sellmair:
Cherry picked into 1.4.0 🍒

PR #1373, opened 2020-08-24
CLI runner clarification in README

BarkingBad:


README.md
```diff
@@ -579,6 +579,21 @@ You can also use a JSON file with dokka configuration:
  java -jar <dokka_cli.jar> <path_to_config.json>
  ```
 
+Using CLI runner to generate default documentation requires providing all dependencies manually on classpath.
```
sellmair, line 582: Let's put CLI and maybe even maven plugin readmes into the subfolders 🙄 
This is just not relevant to most users 🙄

BarkingBad: This PR is just clarification for future useres about using CLI, I think moving these parts to subfolders we can discuss

sellmair: You could still link from the main readme to runners/cli/README.md 🙄 
I think the current readme is just toooo big now 🤖

BarkingBad:
https://github.com/kamildoleglo/dokka/pull/2 moved changes here

PR #1376, opened 2020-08-24
Fix losing tokens while parsing markdown italic text

BarkingBad:


kamildoleglo (approved):

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -48,9 +48,18 @@ class MarkdownParser(
...
             )
 
+        private fun strongHandler(node: ASTNode): DocTag =
+            DocTagsFromIElementFactory.getInstance(
+                node.type,
+                children = node.children.evaluateChildrenWithDroppedEnclosingTokens(2)
+            )
+
+        private fun List<ASTNode>.evaluateChildrenWithDroppedEnclosingTokens(count: Int) =
+            drop(count).dropLast(count).evaluateChildren()
```
kamildoleglo, line 61: I'd personally just create a method only for dropping *count* first and last tokens and call `evaluateChildren()` explicitly (I mean like: `node.children.dropSurrounding(1).evaluateChildren()`) but this is also OK I think

PR #1378, opened 2020-08-24
Drop Android SDK installation requirement

### LINKED ISSUE #1362: Failing test when ANDROID_SDK_ROOT is not set
Fully-qualified name of test:
```kotlin
org.jetbrains.dokka.gradle.AndroidAutoConfigurationTest.`source sets have non-empty classpath`
```

sellmair:
Closes #1362 

This moves
`:testApi` -> `:core:test-api`
`:test-tools` -> `:core:content-matcher-test-utils`

This creates new
`:test-utils` -> basic APIs for testing (assertIsInstance, assumeAndroidSdkInstalled, ...)

Package coordinates for `test-api` is still untouched, but up for discussion ☺️


Kordyjan:
Thanks for changing it to `the-correct-naming-convention`. It was bothering me for a while but recently I haven't encountered any task that would justify that change, because it is small but deeply reaching.

sellmair:
I am still a little bit unsatisfied with this current proposal. It feels like moving a lot just for such a simple use case. 
Also I feel like the new test-utils can be confusing and might not be justified just for such little functions. 
I just feel this is not the best solution. But probably a lot of my dis-satisfaction comes from the fact that I (dumb me) introduced some tests with kotlin-test and now we have a little chaos in our organisation of tests 😢 

I'll hopefully tackle this after 1.4.0

Kordyjan:
My two cents:
- I don't like how `:plugins:base:base-test-utils` looks like. I think that it still will be unambiguous if we drop the second `base`.
- We are not publishing any of our test-utils-like modules. We should change that. I don't know if we want to have that before 1.4.0. If so it would be nice if you add that to the current pr.

sellmair:
I do not like :plugins:base:base-test-utils either, but I think there is a bug in Gradle/Kotlin Compiler that prevents us from naming it test-utils 😢 

I tried several times, but It would not just not compile with this name :(

PR #1379, opened 2020-08-24
Fix external links to Android and Androix documentations

### LINKED ISSUE #1230: Anchors to androidx are broken
For an example project see the `it-android-0` in the `integration-tests`

Dokka creates a link to: https://developer.android.com/reference/kotlin/androidx/appcompat/app/AppCompatActivity.html#findViewById-kotlin.Int-
while the correct one looks like this:
https://developer.android.com/reference/kotlin/androidx/appcompat/app/AppCompatActivity.html#findviewbyid

kamildoleglo:
Closes #1230

### DIFF
--- a/plugins/base/src/main/kotlin/resolvers/external/javadoc/JavadocExternalLocationProviderFactory.kt
+++ b/plugins/base/src/main/kotlin/resolvers/external/javadoc/JavadocExternalLocationProviderFactory.kt
@@ -1,5 +1,8 @@
 package org.jetbrains.dokka.base.resolvers.external.javadoc
 
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.androidSdk
+import org.jetbrains.dokka.androidX
 import org.jetbrains.dokka.base.resolvers.external.ExternalLocationProvider
 import org.jetbrains.dokka.base.resolvers.external.ExternalLocationProviderFactory
 import org.jetbrains.dokka.base.resolvers.external.ExternalLocationProviderFactoryWithCache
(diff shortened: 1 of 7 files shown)


plugins/base/src/main/kotlin/resolvers/external/javadoc/JavadocExternalLocationProviderFactory.kt
```diff
@@ -11,15 +14,21 @@ class JavadocExternalLocationProviderFactory(val context: DokkaContext) :
...
-                    RecognizedLinkFormat.Javadoc1 ->
-                        JavadocExternalLocationProvider(doc, "()", ", ", context) // Covers JDK 1 - 7
-                    RecognizedLinkFormat.Javadoc8 ->
-                        JavadocExternalLocationProvider(doc, "--", "-", context) // Covers JDK 8 - 9
-                    RecognizedLinkFormat.Javadoc10,
-                    RecognizedLinkFormat.DokkaJavadoc ->
-                        JavadocExternalLocationProvider(doc, "()", ",", context) // Covers JDK 10
-                    else -> null
+                when (doc.packageList.url) {
+                    DokkaConfiguration.ExternalDocumentationLink.Companion.androidX().packageListUrl,
```
sellmair, line 18: NIT: Don't write out the Companion here 🙄

plugins/base/src/test/resources/locationProvider/android-package-list
```diff
@@ -0,0 +1,228 @@
+$dokka.format:java-layout-html
```
sellmair, line 1: Thanks for pushing a test! 
I think we could remove some potential danger here. 
Having this list here might be a little bit dangerous, because we might want to maintain it, or are confused why it is not maintained. 

What about having just the subset of this file that is actually necessary for your test being writting out by your test before running? It would also reduce the test scope!

kamildoleglo: Yeah, but then it isn't the `android-package-list` anymore 🤔 
I just thought that it may be a good idea to have an entire file as someone might want to use it someday to their tests and then will be surprised when this file isn't complete
Anyway, left only the necessary lines 👍

plugins/base/src/main/kotlin/resolvers/external/javadoc/JavadocExternalLocationProviderFactory.kt
kamildoleglo: Oops, haven't noticed it when IntelliJ imported it by itself. Fixed 👍

plugins/base/src/test/resources/locationProvider/android-package-list
sellmair: Yes, thats the whole point! 
I would warn you against offering this to others as resource, because then you should maintain it (which is uncomfortable). It might just be easer to materialize this file during the test (can be from constant string value e.g.)

kamildoleglo: I think this won't change much so maintaining it should not be a problem but the idea of materializing those files during the tests is great. Changed to a similar thing 👍

sellmair (approved):

plugins/base/src/test/resources/locationProvider/android-package-list
sellmair: Awesome! Your solution is even much better 😍

sellmair:
Cherry picked into 1.4.0 🍒

PR #1383, opened 2020-08-24
Changing dokka logo into link to main page

sellophane:


Kordyjan (approved):

Kordyjan:
@sellmair Could you check if everything is ok in gradle runner?

core/src/main/kotlin/configuration.kt
```diff
@@ -80,6 +81,7 @@ interface DokkaConfiguration : Serializable {
     val modules: List<DokkaModuleDescription>
     val pluginsClasspath: List<File>
     val pluginsConfiguration: Map<String, String>
+    val globalMainPage: String
```
kamildoleglo, line 84: IMHO this should be a `Path` or better yet `File` to be more consistent with our current configuration

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -644,6 +641,13 @@ open class HtmlRenderer(
 
     private fun resolveLink(link: String, page: PageNode): String = if (URI(link).isAbsolute) link else page.root(link)
 
+    private fun pathToRootPage(currentPage: PageNode): String {
+        return context.configuration.globalMainPage.takeIf { it.isNotBlank() }
```
kamildoleglo, line 645: Maybe a `=` instead?

runners/cli/src/main/kotlin/cli/main.kt
```diff
@@ -52,6 +52,11 @@ class GlobalArguments(args: Array<String>) : DokkaConfiguration {
         "Throw an exception if the generation exited with warnings"
     ).default(DokkaDefaults.failOnWarning)
 
+    override val globalMainPage by parser.option(
+        ArgType.String,
+        description = "Entry point for dokka multimodule documentation"
```
kamildoleglo, line 57: IMHO this explains nothing, I'd write something like "Path to the global main page for multimodule documentation"

kamildoleglo (changes requested):

core/src/main/kotlin/configuration.kt
```diff
@@ -14,6 +14,7 @@ object DokkaDefaults {
     val cacheRoot: File? = null
     const val offlineMode: Boolean = false
     const val failOnWarning: Boolean = false
+    const val globalMainPage: String = ""
```
sellmair, line 17: What should an empty String represent here? 
I would much rather go with something like optional file. If it is not specified, do not link anywhere 🤔

core/src/main/kotlin/configuration.kt
sellmair: I am also not 100% happy with the name. Is there a "non-global main page" that somebody who configures this should care? 

Ideas: 
- `mainPage: File?`
- `mainPageFile: File?`

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt
```diff
@@ -36,6 +36,10 @@ abstract class AbstractDokkaTask(
     val failOnWarning: Property<Boolean> = project.objects.safeProperty<Boolean>()
         .safeConvention(DokkaDefaults.failOnWarning)
 
+    @Input
+    val globalMainPage: Property<String> = project.objects.safeProperty<String>()
```
sellmair, line 40: Hmm, maybe we should set this up: We should know this information and I would wait until somebody comes up with a good use case for making this configurable.

core/src/main/kotlin/configuration.kt
sellmair: Also: I think there should be a way of figuring the "main page" out during generation. I think this can be tackled by me during this bigger ticket for "multi-module documentation". Therefore this configuration _might_ just be temporary 🙄

sellmair (changes requested):

core/src/main/kotlin/configuration.kt
Kordyjan: Even if this can be figured out by runner, it still should be configurable. I can imagine cases in which the "main page" is something one level higher than dokka's main page.

sellmair: Can you give me an example, just so I have a better image of it?
In which scenario would we want to configure this different?

Kordyjan: Library documentation can be part of some bigger website and creators may want this logo to lead to the homepage of the library that can have a high-level explanation with examples and from that links to interesting dokka pages.
I can also think about a case in which there is collection of libraries that have separate documentations but the common homepage.

sellophane: I changed it to URI due to both `File` and `Path` limitations regarding possible schemes

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -659,6 +656,12 @@ open class HtmlRenderer(
 
     private fun resolveLink(link: String, page: PageNode): String = if (URI(link).isAbsolute) link else page.root(link)
 
+    private fun pathToRootPage(currentPage: PageNode): String =
+        context.configuration.homePage?.toASCIIString()
```
kamildoleglo, line 660: Do we need `toASCIIString()` ? IMHO just `toString()` would be sufficient, I see no reason not to support all Unicode characters, but maybe I don't see something

kamildoleglo (approved):

sellmair (changes requested):
Previous Review comments are not all addressed.

sellophane:
> addressed

sellophane:
With new approach to multimodule, with page serialization and rendering of whole instead of separate modules and multimodule page on top, gradle-based name supplying will be not needed, but so far this is what we can achieve.

sellmair:
We will unfortunately not merge this for now. This will be done later after #1402

PR #1389, opened 2020-08-25
Add basic handling for generic params variance

BarkingBad:


plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/signatures/JavadocSignatureProvider.kt
```diff
@@ -150,7 +150,7 @@ class JavadocSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLo
                 annotationsBlock(t)
             }
             signatureWithoutModifiers {
-                text(t.name)
+                text(t.name) // Investigate if java classes can be somehow variant
```
BarkingBad, line 153: Then it should be something like that `? extends T extends SomeInterface`

BarkingBad: On the other hand we should somehow indicate that when the class comes from Kotlin sources

plugins/kotlin-as-java/src/main/kotlin/signatures/JavaSignatureProvider.kt
```diff
@@ -141,7 +141,7 @@ class JavaSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLogge
     private fun signature(t: DTypeParameter) =
         t.sourceSets.map {
             contentBuilder.contentFor(t, styles = t.stylesIfDeprecated(it), sourceSets = setOf(it)) {
-                text(t.name.substringAfterLast("."))
+                text(t.name.substringAfterLast(".")) // Investigate if java classes can be somehow variant
```
BarkingBad, line 144: Same as above

plugins/base/src/test/kotlin/signatures/SignatureUtils.kt
```diff
@@ -0,0 +1,12 @@
...
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import utils.TestOutputWriter
+
+fun TestOutputWriter.renderedContent(path: String = "root/example.html") =
+    contents.getValue(path).let { Jsoup.parse(it) }.select("#content")
+        .single()
+
+fun Element.signature() = select("div.symbol.monospace")
+fun Element.firstSignature() = signature().first()
```
sellmair, line 12: Hey small tip: 
You might want to enable the intelliJ setting: 
"New line on save" to ensure that you always end files with a newline ☺️

core/src/main/kotlin/model/Documentable.kt
```diff
@@ -366,6 +389,12 @@ enum class FunctionModifiers {
     NONE, FUNCTION, EXTENSION
 }
 
+fun Variance<TypeParameter>.withNewDri(dri: DRI) = when(this) {
```
sellmair, line 392: NIT: I think you can easily drop the `new` part. `withDRI` or `withDri` will be perfectly understandable IMHO

sellmair, line 392: NIT: I think you can easily drop the new part. withDRI or withDri will be perfectly understandable IMHO

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/signatures/JavadocSignatureProvider.kt
sellmair: I think `? extends SomeInterface` or `? super SomeInterface`  or am I wrong?  🤔

sellmair (commented):
Very, nice ☺️ 👍

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/signatures/JavadocSignatureProvider.kt
BarkingBad: Yeah, but that works for properties/parameters. There is no such mechanism for class declaration, which we here discuss.

BarkingBad: I mean, in Kotlin we can have `class Klass<out T : SomeInterface>`. To accomplish that in Java, you should have something like `class Klass<? extends T extends SomeInterface>()` which is of course not a correct syntax.

plugins/base/src/test/kotlin/signatures/SignatureUtils.kt
BarkingBad: Ok, will change, thanks

core/src/main/kotlin/model/Documentable.kt
BarkingBad: Will change

PR #1394, opened 2020-08-27
Fix unstable ordering of methods

### LINKED ISSUE #1333: Unstable sort order for overloaded member functions
**Describe the bug**
Hi, on CI we verify if the documentation gets generated correctly and is uptodate when creating PRs.
The code in the `Screenshots` section, generates files which are unstable for version control because the order changes on each dokka invocation.

**Expected behaviour**
Documentation generation should be stable for version control.

**Screenshots**

It is the following code:

https://github.com/detekt/detekt/blob/c6ff7112b0b33afdf62e043075d67508c9e1e5b9/detekt-api/src/main/kotlin/io/gitlab/arturbosch/detekt/api/Context.kt#L13-L45

2020-08-17-202736_1913x435_scrot

**To Reproduce**
1. `git clone https://github.com/detekt/detekt`
2. `git checkout -b "try-dokka-1.4" "origin/try-dokka-1.4"`
3. `gradle generateDocumentation --rerun-tasks` 2x

**Dokka configuration**

```kotlin
    outputDirectory = "$rootDir/docs/pages/kdoc"
}
```

**Installation**
- Operating system: Linux
- Build tool: Gradle v6.6.0
- Dokka version: 1.4.0-rc

**Additional context**
We are upgrading to Kotlin 1.4 and want to upgrade dokka.
Old dokka does not allow us to upgrade Gradle because it throws an OOM exception for Gradle > 6.5 on Java 11.

kamildoleglo:
Closes #1333
This PR also fixes a bug in `SearchbarDataInstaller` which had a race condition when used asynchronously from `HtmlRenderer`

### DIFF
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -88,10 +88,11 @@ open class DefaultPageCreator(
             val documentations = it.sourceSets.map { platform ->
                 it.descriptions[platform]?.also { it.root }
             }
-            val haveSameContent = documentations.all { it?.root == documentations.firstOrNull()?.root && it?.root != null }
+            val haveSameContent =
+                documentations.all { it?.root == documentations.firstOrNull()?.root && it?.root != null }
 
             link(it.name, it.dri)
-            if(it.sourceSets.size == 1 || (documentations.isNotEmpty() && haveSameContent)){
+            if (it.sourceSets.size == 1 || (documentations.isNotEmpty() && haveSameContent)) {
                 documentations.first()?.let { firstSentenceComment(kind = ContentKind.Comment, content = it) }
             }
         }
@@ -129,7 +130,7 @@ open class DefaultPageCreator(
         divergentBlock("Types", types, ContentKind.Classlikes, extra = mainExtra + SimpleAttr.header("Types"))
         divergentBlock(
             "Functions",
-            s.functions,
+            s.functions.sorted(),
             ContentKind.Functions,
             extra = mainExtra + SimpleAttr.header("Functions")
         )
@@ -153,8 +154,8 @@ open class DefaultPageCreator(
             if (map.values.any()) {
                 header(2, "Inheritors") { }
                 +ContentTable(
-                    listOf(contentBuilder.contentFor(mainDRI, mainSourcesetData){
-                            text("Name")
+                    listOf(contentBuilder.contentFor(mainDRI, mainSourcesetData) {
+                        text("Name")
                     }),
                     map.entries.flatMap { entry -> entry.value.map { Pair(entry.key, it) } }
                         .groupBy({ it.second }, { it.first }).map { (classlike, platforms) ->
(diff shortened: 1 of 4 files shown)


BarkingBad (approved):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -129,7 +130,7 @@ open class DefaultPageCreator(
         divergentBlock("Types", types, ContentKind.Classlikes, extra = mainExtra + SimpleAttr.header("Types"))
         divergentBlock(
             "Functions",
-            s.functions,
+            s.functions.sort(),
```
sellmair, line 133: Just a question: 
Would we maybe benefit from (whoever is containing this functions) already having the ordered the same way?
Maybe we are using the functions somewhere else and have unexpected unintuitive order somewhere? 🤔

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -174,6 +175,9 @@ open class DefaultPageCreator(
         }
     }
 
+    private fun Collection<DFunction>.sort() =
```
sellmair, line 178: Extension might maybe be better on `Iterable` than `Collection`?
Would it make sense to let `DFunction` impelement `Comaprable` or is this order to specific to the current use case?

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
kamildoleglo: I agree and that was my first idea, but then after a discussion with @Kordyjan we decided to do it this way, as the presentation side is responsible for the correct displaying of declarations (and TBH it's faster this way as we don't have to modify the data structure)

sellmair: Thats okay ☺️ Good argument!

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
kamildoleglo: > Extension might maybe be better on Iterable than Collection?

Sure, changed 👍 

> Would it make sense to let DFunction impelement Comaprable or is this order to specific to the current use case?

We can do that, I have two issues with this approach though:
1) It would be kind of inconsistent with other Documentables that don't implement `Comparable`
2) The order here is introduced only for sensible presentation (eg. we prefer methods alphabetically with less arguments) and not for some sort of actual order. When implementing `Comparable` I feel like one can assume that one method is *bigger* or *smaller* than other (as it would be then possible to do `fun1 >= fun2`) and it may lead to some wrong impressions based on different assumptions what *bigger* means. 

Maybe we should rename `sort` to something like `orderLexicographically()` ?

sellmair: Very good arguments presented ☺️ 
I would be fine with calling it sort. Will merge this if you are either ☺️

sellmair (approved):

sellmair:
Cherry picked into 1.4.0 🍒

Issue #1335, opened 2020-08-18
Feedback: All modules page


sellmair:
Transcript: 

1) The standalone `/` seems off. Expected project name instead
2) Modules submenu should actually list all modules for fast navigation and to be consistent with the packages (e.g. compare it with http://dokka-snapshots.s3.eu-central-1.amazonaws.com/navigation-pane-overflow/coroutines/latest/kotlinx-coroutines-core/build/dokka/html/kotlinx-coroutines-core/index.html)
3) This big empty white indent seems to be redundant 

![image](https://user-images.githubusercontent.com/34319766/90503994-a16b0480-e150-11ea-86c5-c933239ecc72.png)

PR #1398, opened 2020-08-27
All modules page #1335

MarcinAman:
For the kotlinx.serialization it looks like this:
[dokka.zip](https://github.com/Kotlin/dokka/files/5137082/dokka.zip)
image

### DIFF
--- a/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/MultiModule0IntegrationTest.kt
+++ b/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/MultiModule0IntegrationTest.kt
@@ -55,7 +55,8 @@ class MultiModule0IntegrationTest(override val versions: BuildVersions) : Abstra
         outputDir.allHtmlFiles().forEach { file ->
             assertContainsNoErrorClass(file)
             assertNoUnresolvedLinks(file)
-            assertNoHrefToMissingLocalFileOrDirectory(file)
+            //Navigation file for multimodule will contain paths that are incorrect since they are resolved finally on frontend
+            if(!file.toString().contains("htmlMultiModule/navigation")) assertNoHrefToMissingLocalFileOrDirectory(file)
             assertNoEmptyLinks(file)
         }
 
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -395,7 +395,7 @@ open class HtmlRenderer(
                         }
 
                         it.filter { it !is ContentLink }.takeIf { it.isNotEmpty() }?.let {
-                            if(pageContext is ModulePage){
+                            if(pageContext is ModulePage || pageContext is MultimoduleRootPage){
                                 it.forEach {
                                     span(classes = if(it.dci.kind == ContentKind.Comment) "brief-comment" else "") {
                                         it.build(this, pageContext, sourceSetRestriction)
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -525,9 +525,8 @@ open class HtmlRenderer(
         div(classes = "breadcrumbs") {
             val path = locationProvider.ancestors(page).filterNot { it is RendererSpecificPage }.asReversed()
 
-            text(projectName)
+            if(page is MultimoduleRootPage) text(projectName)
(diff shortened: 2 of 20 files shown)


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleFileLayout.kt
```diff
@@ -79,5 +79,27 @@ internal fun DokkaMultiModuleTask.copyChildOutputDirectory(child: AbstractDokkaT
     }
 
     sourceChildOutputDirectory.copyRecursively(targetChildOutputDirectory, overwrite = true)
+    copyNavigationForMultiModule(sourceChildOutputDirectory, outputDirectory.getSafe(), child)
+}
+
+internal fun copyNavigationForMultiModule(
```
sellmair, line 85: Hmm. I see. This is the workaround you were talking about! 
I am not quite sure, if we should really go that way 🙄

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/MultiModule0IntegrationTest.kt
```diff
@@ -55,7 +55,8 @@ class MultiModule0IntegrationTest(override val versions: BuildVersions) : Abstra
         outputDir.allHtmlFiles().forEach { file ->
             assertContainsNoErrorClass(file)
             assertNoUnresolvedLinks(file)
-            assertNoHrefToMissingLocalFileOrDirectory(file)
+            //Navigation file for multimodule will contain paths that are incorrect since they are resolved finally on frontend
+            if(!file.toString().contains("htmlMultiModule/navigation")) assertNoHrefToMissingLocalFileOrDirectory(file)
```
sellmair, line 59: NIT: Maybe be more explicit and use `path` instead of  `toString`?

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleFileLayout.kt
```diff
@@ -79,5 +79,27 @@ internal fun DokkaMultiModuleTask.copyChildOutputDirectory(child: AbstractDokkaT
...
+    child: AbstractDokkaTask
+) {
+    val childProjectName = child.projectName.getSafe()
+    val currentToDesiredDestinationMapping = listOf(
+        "/scripts/navigation-pane.json" to "/scripts/navigation-pane-${childProjectName}.json",
+        "/navigation.html" to "/navigation-${childProjectName}.html"
+    )
+
+    currentToDesiredDestinationMapping.forEach { (currentLocation, desiredLocation) ->
+        run {
```
sellmair, line 97: What is the point of opening a scope here?

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleFileLayout.kt
```diff
@@ -79,5 +79,27 @@ internal fun DokkaMultiModuleTask.copyChildOutputDirectory(child: AbstractDokkaT
...
+            File(sourceChildOutputDirectory.absolutePath + currentLocation)
```
sellmair, line 98: Use `.resolve` instead of adding paths might be nicer here?

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -680,6 +685,9 @@ open class HtmlRenderer(
                     }
                 }
                 script { unsafe { +"""var pathToRoot = "${locationProvider.pathToRoot(page)}";""" } }
+                if(page is MultimoduleRootPage){
+                    script { unsafe { + """var modules = ${context.configuration.modules.joinToString(prefix = "[", postfix = "]"){ "\"${it.path}\"" }}"""}}
```
sellmair, line 689: NIT: Liiiittle bit long line

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/MultiModule0IntegrationTest.kt
sellmair: NIT: I do not know why I would prefer the following
```kotlin
 if("htmlMultiModule/navigation" !in file.path)
     assertNoHrefToMissingLocalFileOrDirectory(file)
```

sellmair:
Closing this for now. We will tackle this again after something like this #1402

PR #1416, opened 2020-08-31
Render react resources as separate css to allow overriding

MarcinAman:
Motivation for that PR was the meeting with scala team, that faced an issue with not being able to override all styles. Also i thought it is weird that in order to override styles you need to modify 2 extension points so i've merged them into categories (styles, scripts and assets).

### DIFF
--- a/plugins/base/build.gradle.kts
+++ b/plugins/base/build.gradle.kts
@@ -18,10 +18,28 @@ dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-html-jvm:$kotlinx_html_version")
 }
 
-task("copy_frontend", Copy::class) {
-    from(File(project(":plugins:base:frontend").projectDir, "dist/"))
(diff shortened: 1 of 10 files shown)


MarcinAman:
Given our current dsl is there any way to test it?

plugins/base/build.gradle.kts
```diff
@@ -18,10 +18,28 @@ dependencies {
...
+}
+
+val copyCssFiles by tasks.registering(Copy::class){
+    from(projectDistDir){
+        include("*.css")
+    }
+    destinationDir = File(sourceSets.main.get().resources.sourceDirectories.singleFile, "dokka/styles")
+}
+
+task("copy_frontend") {
```
Kordyjan, line 37: Why not `by registering`?

Kordyjan (changes requested):

MarcinAman:
Rebased

BarkingBad (approved):

PR #1428, opened 2020-09-02
Null pointer on images

MarcinAman:
image
image


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -346,8 +347,7 @@ open class HtmlRenderer(
         node: ContentEmbeddedResource,
         pageContext: ContentPage
     ) { // TODO: extension point there
-        val imageExtensions = setOf("png", "jpg", "jpeg", "gif", "bmp", "tif", "webp", "svg")
-        return if (File(node.address).extension.toLowerCase() in imageExtensions) {
+        return if (node.isImage()) {
```
kamildoleglo, line 350: We can probably use `fun(...) = (...)` instead of `fun(...) { return (...) }`

MarcinAman: sure, done 👍

PR #1431, opened 2020-09-03
Adding validity check at gradle runner level

sellophane:


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
```diff
@@ -35,6 +36,9 @@ abstract class DokkaTask : AbstractDokkaTask(DokkaBootstrapImpl::class) {
             .filterNot { it.suppress.getSafe() }
 
     override fun buildDokkaConfiguration(): DokkaConfigurationImpl {
+        val validityIssues = preConfigureValidityCheck.mapNotNull { (check, message) -> message.takeUnless{ check() } }
+        if (validityIssues.isNotEmpty())
+            throw IllegalStateException(validityIssues.joinToString("; "))
```
kamildoleglo, line 41: Maybe `DokkaException` would be better?

PR #1445, opened 2020-09-08
Adding configuration mpp check for javadoc

sellophane:


plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/validity/MultiplatformConfiguredChecker.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import org.jetbrains.dokka.Platform
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.validity.PreGenerationChecker
+import org.jetbrains.dokka.validity.PreGenerationCheckerOutput
+
+class MultiplatformConfiguredChecker(val context: DokkaContext) : PreGenerationChecker {
+
+    override fun invoke(): PreGenerationCheckerOutput {
+        val isSinglePlatform = context.configuration.sourceSets.fold(true) { acc, sourceSet ->
```
MarcinAman, line 11: would `.all { sourceset -> ...` work?

core/src/main/kotlin/validity/PreGenerationChecker.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.dokka.validity
+
+interface PreGenerationChecker : () -> PreGenerationCheckerOutput {
+
+    override fun invoke(): PreGenerationCheckerOutput
+}
+
+data class PreGenerationCheckerOutput(val result: Boolean, val messages: List<String>) {
+
+    operator fun plus(pair: Pair<Boolean, List<String>>) =
```
MarcinAman, line 10: Imho this should have a generic list type parameter and be extracted so it can be used somewhere else

MarcinAman (approved):

core/src/main/kotlin/validity/PreGenerationChecker.kt
BarkingBad: Or we can even consider using [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) which brings some cool error handling functions

BarkingBad (approved):

core/src/main/kotlin/validity/PreGenerationChecker.kt
sellophane: Yes, we can consider.
After considering, I come to conclusion, that `Result` would be great here IF  ONLY Kotlin would allow it to be return type for functions. So I am sticking to what is.

sellophane: And I don't expect it to be used anywhere else.

kamildoleglo (approved):

PR #1459, opened 2020-09-11
Hide derived members that are not public

BarkingBad:
```kotlin
open class A {
    private val shouldBeHidden: Int = 1
    val shouldBeVisible: Int = 2

    private fun privateFun(): Int = 3
    fun publicFun(): Int = 4
}

open class B : A()

open class C : B()
```

obraz
obraz
obraz

obraz
obraz
obraz

### DIFF
--- a/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
@@ -388,12 +388,13 @@ private class DokkaDescriptorVisitor(
         }
     }
 
-    private suspend fun visitPropertyDescriptor(descriptor: PropertyDescriptor, parent: DRIWithPlatformInfo): DProperty {
-        val dri = parent.dri.copy(callable = Callable.from(descriptor))
+    private suspend fun visitPropertyDescriptor(originalDescriptor: PropertyDescriptor, parent: DRIWithPlatformInfo): DProperty {
+        val dri = parent.dri.copy(callable = Callable.from(originalDescriptor))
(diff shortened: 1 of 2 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -351,14 +354,15 @@ private class DokkaDescriptorVisitor(
         )
     }
 
-    fun CallableMemberDescriptor.createDRI(wasOverridenBy: DRI? = null): Pair<DRI, DRI?> =
-        if (kind == CallableMemberDescriptor.Kind.DECLARATION || overriddenDescriptors.isEmpty())
-            Pair(DRI.from(this), wasOverridenBy)
-        else
-            overriddenDescriptors.first().createDRI(DRI.from(this))
```
BarkingBad, line 358: This was wrong as it was passing only neighbouring DRI to the next level of recursion. We want the most upper and lower DRIs. Moreover it was wrongly interpreted then as an `inheritedFrom` insted of `overridenBy` in 361 line

BarkingBad: Anyway, I think it is nice to use this function for `visitPropertyDescriptor` as well. I would also change the extra `InheritedFunction` to `InheritedCallable` and then change presentation of property/function row to something like this

obraz

@Kordyjan @sellmair @kamildoleglo what do you think on that?

BarkingBad:
I know it's 528 line, but ~500 are just one test. I would keep it more concise though autoformatter had a different opinion about that

kamildoleglo (approved):

MarcinAman (approved):

PR #1464, opened 2020-09-12
Always show protected members

### LINKED ISSUE #434: Allow including docs for internal/protected members but not private members
Currently Dokka only exposes the boolean configuration option `includeNonPublic`, which forces you to choose between including only public members; or private, internal, protected and public members with no in-between.

It would be nice to be able to be able to include documentation for internal/protected members without having a plethora of warnings about not documenting private members.

SUPERCILEX:
They are part of the public API and should always be shown.

Fixes #434


BarkingBad:
I think what we want here (accrodgin to #434) is to change binary option `public` vs `includeNonPublic` to possibility of specifying certain level of privacy of the API as it is in Javadoc. So there should be greater changes in configuration rather than assuming `protected` is `public`
obraz

SUPERCILEX:
Closing since I won't have time to fix the PR. Yup, I agree that you should be able to choose between different levels, but the default should still include protected members.

mukeshsolanki:
Has this been resolved? as im facing issues with protected methods

IgnatBeresnev:
@mukeshsolanki it has, hopefully will be included in 1.6.20, see https://github.com/Kotlin/dokka/issues/434

PR #1472, opened 2020-09-14
Add MD parser

mkondratek:


BarkingBad:
Changes look good. However, you should fix it so it builds and passes all CI tests. You possibly will have merge conflicts if #1473 gets merged earlier.

BarkingBad (approved):

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -10,503 +10,480 @@ import org.intellij.markdown.ast.impl.ListItemCompositeNode
...
 
-        private fun blockquotesHandler(node: ASTNode): DocTag =
-            DocTagsFromIElementFactory.getInstance(
-                node.type, children = node.children
-                    .filterIsInstance<CompositeASTNode>()
-                    .evaluateChildren()
-            )
+    override fun parse(text: String): DocumentationNode {
+        return DocumentationNode(listOf(Description(parseStringToDocNode(text))))
+    }
```
BarkingBad, line 44: You should not override default behaviour from base class, especially not in that way

BarkingBad: Just delete it, so the base class `parse` will be dispatched

BarkingBad (changes requested):

BarkingBad (approved):

PR #1473, opened 2020-09-14
Fix nested paragraphs

BarkingBad:


BarkingBad:
Comparing to latest master on stdlib, git diff found exactly 6220 different lines, all have one paragraph less nested (if there were at least two). I assume it did not change any behaviour.

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -465,46 +466,46 @@ class MarkdownParser(
             DocumentationNode(
                 (listOf(kDocTag) + getAllKDocTags(findParent(kDocTag))).map {
                     when (it.knownTag) {
-                        null -> if (it.name == null) Description(parseStringToDocNode(it.getContent())) else CustomTagWrapper(
-                            parseStringToDocNode(it.getContent()),
+                        null -> if (it.name == null) Description(parseStringToDocNode(it.getContent()) as RootDocTag) else CustomTagWrapper(
```
Kordyjan, line 469: Is it even possible for `parseStringToDocNode` to return something other than `RootDocTag`. If no move cast into that function. If this is possible consider slightly modifying tha api.

Kordyjan (changes requested):

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
BarkingBad: Yeah, I shouldn't have left it that way, will refactor soon

sellophane:
1) Please squash commits before merge
2) `Fix tests` breaks compilation

plugins/base/src/main/kotlin/transformers/pages/comments/DocTagToContentConverter.kt
```diff
@@ -175,7 +176,18 @@ object DocTagToContentConverter : CommentsToContentConverter {
...
+            else -> when {
+                docTag is CustomDocTag && docTag.name == MarkdownElementTypes.MARKDOWN_FILE.name -> if (docTag.children.size > 1)
+                    listOf(ContentGroup(
+                        buildChildren(docTag),
+                        dci,
+                        sourceSets.toDisplaySourceSets(),
+                        styles,
+                        extra = extra
+                    )) else buildChildren(docTag)
+                else -> buildChildren(docTag)
```
kamildoleglo, line 189: Extract to separate method

runners/gradle-plugin/src/test/kotlin/org/jetbrains/dokka/gradle/AndroidAutoConfigurationTest.kt
```diff
@@ -61,6 +61,7 @@ class AndroidAutoConfigurationTest {
         }
     }
 
+    @Ignore
```
kamildoleglo, line 64: Please remove

kamildoleglo (changes requested):

PR #1477, opened 2020-09-16
Fix scrolling to relevant records in search, filter all that doesnt m…

### LINKED ISSUE #1474: New HTML format: Search shows the most unrelevant results first
Dokka version: 1.4.10-dev-72
I use 1.4.10/stdlib.
When typing 'compare' in search, the list of results is scrolled down and shows the most unrelevant results:
I'm not sure it worths showing such results in the list at all (coincidence is 1 letter in each word).
image

MarcinAman:
…atch required amout of letters

closes #1474

### DIFF
--- a/plugins/base/frontend/src/main/components/search/dokkaFuzzyFilter.tsx
+++ b/plugins/base/frontend/src/main/components/search/dokkaFuzzyFilter.tsx
@@ -2,7 +2,8 @@ import {Select} from "@jetbrains/ring-ui";
 import {Option, OptionWithHighlightComponent, OptionWithSearchResult, SearchRank} from "./types";
 import fuzzyHighlight from '@jetbrains/ring-ui/components/global/fuzzy-highlight.js'
 import React from "react";
-import {SearchResultRow} from "./searchResultRow";
+import {SearchResultRow, signatureFromSearchResult} from "./searchResultRow";
+import _ from "lodash";
 
 const orderRecords = (records: OptionWithSearchResult[], searchPhrase: string): OptionWithSearchResult[] => {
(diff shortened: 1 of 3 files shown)


plugins/base/frontend/src/main/components/search/dokkaFuzzyFilter.tsx
```diff
@@ -66,7 +72,7 @@ export class DokkaFuzzyFilterComponent extends Select {
                     rank: SearchRank.NameMatch
                 }
             })
-            .filter((record: OptionWithSearchResult) => record.matched)
+            .filter((record: OptionWithSearchResult) => record.matched && (hasApproperateLenght(record, 3) || filterPhrase === ''))
```
MarcinAman, line 75: Is a sequence of 3 a good idea? So basically it will discard every record that doesn't have at least 3 characters in sequence matched

plugins/base/frontend/src/main/components/search/dokkaFuzzyFilter.tsx
```diff
@@ -38,6 +39,11 @@ const highlightMatchedPhrases = (records: OptionWithSearchResult[]): OptionWithH
     })
 }
 
+const hasApproperateLenght = (searchResult: OptionWithSearchResult, length: number): boolean => {
```
dolik-rce, line 42: "Approperate" is not a word. I think you meant "appropriate".

MarcinAman: Fair point, as i read this PR again i think this name was flawed from the beginning. Changed to something i found more descriptive: `hasAnyMatchedPhraseLongerThan`

Kordyjan (approved):

PR #1480, opened 2020-09-16
Fix merging documentations of modules and packages

BarkingBad:
This PR will not build until #1473 will be merged into master

### DIFF
--- a/plugins/base/src/test/kotlin/linkableContent/LinkableContentTest.kt
+++ b/plugins/base/src/test/kotlin/linkableContent/LinkableContentTest.kt
@@ -312,4 +312,66 @@ class LinkableContentTest : AbstractCoreTest() {
         }
 
     }
+
+    @Test
+    fun `Include module with description parted in two files`() {
+
+        val testDataDir = getTestDataDir("multiplatform/basicMultiplatformTest").toAbsolutePath()
+        val includesDir = getTestDataDir("linkable/includes").toAbsolutePath()
+
+        val configuration = dokkaConfiguration {
+            moduleName = "example"
+            sourceSets {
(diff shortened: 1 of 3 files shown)


plugins/base/src/test/kotlin/linkableContent/LinkableContentTest.kt
```diff
@@ -312,4 +312,66 @@ class LinkableContentTest : AbstractCoreTest() {
         }
 
     }
+
+    @Test
+    fun `Include module with description parted in two files`() {
```
sellmair, line 317: This IMHO is the wrong kind of test. You should really write a unit test for everything that is unit testable. 
I am not proposing to remove this test, but adding a proper unit test as well!

sellophane: But with this engine we are doing mainly functional tests and this one does not look much different than the others

sellophane (approved):

sellophane (changes requested):
Something here is not compiling...

BarkingBad:
> Something here is not compiling...

Yeah, is should be rebased on top of #1473 as it uses its new API, so it's better to wait until former gets merged

plugins/base/src/test/kotlin/linkableContent/LinkableContentTest.kt
sellmair: Yes the project is almost entirely relying on functional tests and this is not ideal! This particular component was designed to support unit tests. 

Maybe @Kordyjan can give further directions on this issue :)

kamildoleglo (approved):

PR #1482, opened 2020-09-17
Remove dokka context from creating analysis

MarcinAman:
If someone uses just the dokka-analysis should not be forced into creating a dokka context only to pass configuration, sourcesets and logger

### DIFF
--- a/kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/KotlinAnalysis.kt
+++ b/kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/KotlinAnalysis.kt
@@ -6,18 +6,24 @@ import org.jetbrains.dokka.DokkaConfiguration.DokkaSourceSet
 import org.jetbrains.dokka.DokkaSourceSetID
 import org.jetbrains.dokka.model.SourceSetDependent
 import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.utilities.DokkaLogger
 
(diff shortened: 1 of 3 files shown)


BarkingBad (approved):
When speaking of completely decoupling KotlinAnalysis, I think you can chang `DokkaConfiguration` to `List<DokkaSourceSet>` and `DokkaLogger` to `DokkaLogger?` so the user can string everything to absolute minimum

MarcinAman:
> When speaking of completely decoupling KotlinAnalysis, I think you can chang `DokkaConfiguration` to `List<DokkaSourceSet>` and `DokkaLogger` to `DokkaLogger?` so the user can string everything to absolute minimum

Fair point, changed

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/KotlinAnalysis.kt
```diff
@@ -6,19 +6,25 @@ import org.jetbrains.dokka.DokkaConfiguration.DokkaSourceSet
...
+            logger = logger,
+            sourceSets = sourceSets,
             sourceSet = sourceSet
         )
     }
 
     return KotlinAnalysisImpl(environments)
 }
 
+@Deprecated(message = "Construct using DokkaConfiguration and logger",
```
kamildoleglo, line 23: The message should rather be: "Construct using list of DokkaSourceSets and logger" indicating what has to be used to not contradict the `replaceWith` message

kamildoleglo (changes requested):

PR #1483, opened 2020-09-17
Add better handling of functional types in rendered output

BarkingBad:
260 lines are just tests

```kotlin
val n1: Function1<Int, String> = { _ -> "" }
val n2: (Int) -> String = { _ -> "" }
val n3: Boolean.(Int) -> String = { _ -> "" }
val n4: (param: Int) -> String = { _ -> "" }
val n5: suspend (Int) -> String = { _ -> "" }
val n6: suspend Boolean.(Int) -> String = { _ -> "" }
val n7: suspend (param: Int) -> String = { _ -> "" }
val n8: suspend (param1: suspend Boolean.(param2: List<Int>) -> Boolean) -> String = { _ -> "" }
```

obraz

### DIFF
--- a/core/src/main/kotlin/model/Documentable.kt
+++ b/core/src/main/kotlin/model/Documentable.kt
@@ -340,12 +340,13 @@ data class DTypeParameter(
     constructor(
         dri: DRI,
         name: String,
+        presentableName: String?,
         documentation: SourceSetDependent<DocumentationNode>,
         expectPresentInSet: DokkaSourceSet?,
         bounds: List<Bound>,
         sourceSets: Set<DokkaSourceSet>,
         extra: PropertyContainer<DTypeParameter> = PropertyContainer.empty()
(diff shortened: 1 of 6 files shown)


BarkingBad:
Also works for Java sources

```java
public class JavaClass {
    public Function<String, Integer> javaFunction = null;
    public Function1<String, Integer> kotlinFunction = null;
}
```

obraz

MarcinAman:
why `val n1: Function1<Int, String> = { _ -> "" }` is represented the same as `(Int) -> String = { _ -> "" }`? shouldn't we preserve information about the type exactly?

BarkingBad:
@MarcinAman because `(Int) -> String` and `Function1<Int, String>` are exactly the same type, former is just syntactic sugar

MarcinAman:
Oh ok, was just wondering if this was intentional. If so then ok

Kordyjan (approved):

PR #1484, opened 2020-09-17
Add processing of annotations to typealiases

BarkingBad:
By mistake, we forgot to process annotations for `typealiases` completetly.

Also fixes reported bug with not deprecating them.

```kotlin
@MustBeDocumented
@Target(AnnotationTarget.TYPEALIAS)
annotation class SomeAnnotation

@SomeAnnotation
@Deprecated(
    message = "Replace with ImageRequest.",
    replaceWith = ReplaceWith("ImageRequest", "coil.request.ImageRequest")
)
typealias GetRequest = Int
```

obraz

### DIFF
--- a/plugins/base/src/main/kotlin/transformers/documentables/DeprecatedDocumentableFilterTransformer.kt
+++ b/plugins/base/src/main/kotlin/transformers/documentables/DeprecatedDocumentableFilterTransformer.kt
@@ -57,6 +57,10 @@ class DeprecatedDocumentableFilterTransformer(val context: DokkaContext) : PreMe
                     modified = modified || listModified
                     list
                 }
+                val typeAliases = filterTypeAliases(pckg.typealiases).let { (listModified, list) ->
+                    modified = modified || listModified
+                    list
+                }
                 when {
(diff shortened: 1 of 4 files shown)


MarcinAman (approved):
Would be nice if we could have another test for the deprecated case

BarkingBad:
Yeah, I was just lazy. If I have spare time, I will add one :P

plugins/base/src/main/kotlin/transformers/documentables/DeprecatedDocumentableFilterTransformer.kt
```diff
@@ -57,14 +57,19 @@ class DeprecatedDocumentableFilterTransformer(val context: DokkaContext) : PreMe
                     modified = modified || listModified
                     list
                 }
+                val typeAliases = filterTypeAliases(pckg.typealiases).let { (listModified, list) ->
```
sellophane, line 60: We can possibly extract this function, as we are using it 4 times now

sellophane (approved):

Issue #1339, opened 2020-08-18
`K docs` logo should be easily replaceable


sellmair:
It should be rather easy to provide a custom `K doc` logo
![image](https://user-images.githubusercontent.com/34319766/90505502-04f63180-e153-11ea-8c30-eb4683c03e44.png)

romanowski:
This is how we did it in scala3doc: https://github.com/lampepfl/scala3doc/pull/9

PR #1488, opened 2020-09-18
Make logo replaceable #1339

MarcinAman:
In order to change a logo you need to override a `logo-styles.css` file with your own and in this file add the `background-image` property on element `id='logo'` like this:
```
#logo {
    background-image: url('https://upload.wikimedia.org/wikipedia/commons/9/9d/Ubuntu_logo.svg');
}
```

This would result in your page looking like this:
image

As not all images are created equal you may want to make some css adjustments (padding, height, width etc) to make it look nice. You should do it in `logo-styles.css` file.

With this PR dokka allows to add custom styles (in `customStyles`) and resources (in `customAssets`). Styles file will be added to **every page** as embedded resource. Styles will be added to the `styles` directory and assets to `images`.

Sample configuration can look like this:
```
    customStyleSheets.set(listOf(file("customResources/logo-styles.css"), file("customResources/custom-style-to-add.css")))
    customAssets.set(listOf(file("customResources/custom-resource.svg")))
```


BarkingBad (approved):
Maybe we should update user manual with info about extending custom styles and assets

runners/cli/src/main/kotlin/cli/main.kt
```diff
@@ -254,7 +264,7 @@ object ArgTypeFile : ArgType<File>(true) {
 object ArgTypePlatform : ArgType<Platform>(true) {
     override fun convert(value: kotlin.String, name: kotlin.String): Platform = Platform.fromString(value)
     override val description: kotlin.String
-        get() = "{ String thar represents paltform }"
+        get() = "{ String thar represents platform }"
```
kamildoleglo, line 267: `thar` -> `that`

kamildoleglo (changes requested):
Also you forgot to update Maven configuration

kamildoleglo (approved):

PR #1504, opened 2020-09-23
On this page

MarcinAman:
Component is available on pages with tabs and anchors as for other pages it doesn't make sense. 
It is made to be on hover, just to quickly navigate to desired section.
What do you think about it?


plugins/base/frontend/src/main/components/root.tsx
```diff
@@ -1,39 +1,59 @@
...
+      }
+    }
+  })
 }
 
 let renderApp = () => {
   render(
-      <App/>,
-      rootEl
+    <App />,
```
sellophane, line 43: Is this white-space-agnostic?

plugins/base/src/main/kotlin/allModulePage/MultimodulePageCreator.kt
```diff
@@ -48,18 +50,18 @@ class MultimodulePageCreator(
...
+                    val header = linkNode(module.name, dri, DCI(setOf(dri), ContentKind.Main), extra = extraWithAnchor)
                     val content = ContentGroup(
                         children =
                         if (displayedModuleDocumentation != null)
                             DocTagToContentConverter.buildContent(displayedModuleDocumentation, dci, emptySet())
                         else emptyList(),
                         dci = dci,
                         sourceSets = emptySet(),
-                        style = emptySet()
+                        style = emptySet(),
```
sellophane, line 62: Am I not seeing here, or is this coma completely unwanted here?

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -103,6 +106,7 @@ open class HtmlRenderer(
             }
             node.hasStyle(TextStyle.Paragraph) -> p(additionalClasses) { childrenCallback() }
             node.hasStyle(TextStyle.Block) -> div(additionalClasses) { childrenCallback() }
+            node.isAnchorable -> buildAnchor(node.anchor, node.anchorLabel!!, node.sourceSetsFilters) { childrenCallback() }
```
sellophane, line 109: Must we have a `!!` here?

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -365,51 +369,51 @@ open class HtmlRenderer(
...
-                            attributes["data-filterable-current"] = node.sourceSets.joinToString(" ") {
-                                it.sourceSetIDs.merged.toString()
-                            }
-                            attributes["data-filterable-set"] = node.sourceSets.joinToString(" ") {
-                                it.sourceSetIDs.merged.toString()
-                            }
+                buildAnchor(node)
+                div(classes = "table-row") {
+                    if (!style.contains(MultimoduleTable)) {
+                        attributes["data-filterable-current"] = node.sourceSets.joinToString(" ") {
```
sellophane, line 375: You are generating same value twice.

plugins/base/src/main/kotlin/resolvers/anchors/AnchorsHint.kt
```diff
@@ -1,9 +1,13 @@
...
 import org.jetbrains.dokka.model.properties.ExtraProperty
 import org.jetbrains.dokka.pages.ContentNode
 
-// TODO IMPORTANT: https://github.com/Kotlin/dokka/issues/1054
-object SymbolAnchorHint: ExtraProperty<ContentNode>, ExtraProperty.Key<ContentNode, SymbolAnchorHint> {
-    override val key = this
+data class SymbolAnchorHint(val anchorName: String): ExtraProperty<ContentNode> {
+    object SymbolAnchorHintKey : ExtraProperty.Key<ContentNode, SymbolAnchorHint>
+    override val key: ExtraProperty.Key<ContentNode, SymbolAnchorHint> = SymbolAnchorHintKey
+    companion object: ExtraProperty.Key<ContentNode, SymbolAnchorHint> {
```
Kordyjan, line 10: What is the reason of this object implementing `ExtraProperty.Key` if it is not used as a key?

plugins/base/frontend/src/main/components/root.tsx
MarcinAman: Yes

plugins/base/src/main/kotlin/resolvers/anchors/AnchorsHint.kt
MarcinAman: I don't understand. It is since when i'd like to get values from extras i use it

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
MarcinAman: fixed

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
MarcinAman: yes, because without it this doesn't make sense. 
Also this wont produce NPE since `isAnchorable` checks if `anchorLabel` != null

MarcinAman:
I dont know if you have realized but this PR depends on: https://github.com/Kotlin/dokka/pull/1307, so this one also needs a review ;D

plugins/base/src/main/kotlin/allModulePage/MultimodulePageCreator.kt
MarcinAman: rmed

plugins/base/src/main/kotlin/resolvers/anchors/AnchorsHint.kt
kamildoleglo: Fixed in the second PR

PR #1515, opened 2020-09-30
Fix typealiases filtering when private

BarkingBad:


plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
```diff
@@ -290,5 +294,22 @@ class DocumentableVisibilityFilterTransformer(val context: DokkaContext) : PreMe
             }
             return Pair(classlikesListChanged, filteredClasslikes)
         }
+
+        private fun filterTypeAliases(
+            functions: List<DTypeAlias>,
```
Kordyjan, line 299: `functions`?

BarkingBad: Copy-paste leftover Will fix

BarkingBad: Fixed

PR #1527, opened 2020-10-01
Dri index plugin

kamildoleglo:


elizarov:
It would be great to have some tests. They will help to see the sample of the output format of this plugin.

PR #1545, opened 2020-10-09
Update urls to fonts

MarcinAman:


kamildoleglo (approved):

abyssal:
@MarcinAman This needs to be updated again, the URLs changed. I think the Mono team replaced 'fonts' with 'webfonts' for these ones.

PR #1555, opened 2020-10-13
Distinguish types from functions on navigation

MarcinAman:


plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -69,6 +70,13 @@ object NavigationPageInstaller : PageTransformer {
                 children.filter { it is ContentPage && it.documentable is DEnumEntry }.map { visit(it as ContentPage) }
             else -> emptyList()
         }.sortedBy { it.name.toLowerCase() }
+
+    private val ContentPage.displayableName: String
+        get() = if(documentable is DFunction){
```
kamildoleglo, line 75: Please reformat

kamildoleglo (changes requested):

kamildoleglo (approved):

PR #1578, opened 2020-10-18
Cleaning querying from plugins where possible

sellophane:


core/src/main/kotlin/plugability/DokkaPlugin.kt
```diff
@@ -60,7 +60,10 @@ inline fun <reified P : DokkaPlugin, reified E : Any> P.query(extension: P.() ->
     context?.let { it[extension()] } ?: throwIllegalQuery()
 
 inline fun <reified P : DokkaPlugin, reified E : Any> P.querySingle(extension: P.() -> ExtensionPoint<E>): E =
-    context?.single(extension()) ?: throwIllegalQuery()
+    querySingleOrNull(extension) ?: throwIllegalQuery()
+
+inline fun <reified P : DokkaPlugin, reified E : Any> P.querySingleOrNull(extension: P.() -> ExtensionPoint<E>): E? =
+    context?.single(extension())
```
Kordyjan, line 66: This is not working as intended. `context.single` should never return null. What can be null here is the context itself.

Kordyjan (changes requested):

core/src/main/kotlin/plugability/DokkaPlugin.kt
MarcinAman: I am not a fan of this, since it gives you a false hope that whenever you want to resolve something that isn't there null will be returned. Instead we will throw an exception. 

A possible solution would be to add a `singleOrNull` to `DokkaContext`, but i am not entirely sure it is a good idea because of stability reasons. Also a single nullable value doesn't reflect 3 states that could occur: no element, 1 element, multiple elements

plugins/base/src/test/kotlin/transformers/ContextModuleAndPackageDocumentationReaderTest1.kt
```diff
@@ -110,13 +112,11 @@ class ContextModuleAndPackageDocumentationReaderTest1 : AbstractContextModuleAnd
 
     @Test
     fun `assert moduleA with unknown source set`() {
-        val documentation = reader[
```
MarcinAman, line 113: why this will now throw exception rather than return an empty map?

core/src/main/kotlin/plugability/DokkaPlugin.kt
sellophane: Rewerted.

plugins/base/src/test/kotlin/transformers/ContextModuleAndPackageDocumentationReaderTest1.kt
sellophane: Due to changes in initialization - previously we were passing nulled kotlinAnalyse directly, now it is being retrieved from context and is initialize. And when we try to query it for invalid source set, we receive exception instead.

MarcinAman (approved):

Kordyjan (approved):

PR #1590, opened 2020-10-24
New breadcrumbs

MarcinAman:


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -99,8 +99,7 @@ open class HtmlRenderer(
             node.hasStyle(TextStyle.Span) -> span() { childrenCallback() }
             node.dci.kind == ContentKind.Symbol -> div("symbol $additionalClasses") { childrenCallback() }
             node.dci.kind == ContentKind.BriefComment -> div("brief $additionalClasses") { childrenCallback() }
-            node.dci.kind == ContentKind.Cover -> div("cover $additionalClasses") {
-                filterButtons(pageContext)
+            node.dci.kind == ContentKind.Cover -> div("cover $additionalClasses") { //TODO this can be removed
```
sellophane, line 102: If it can, then remove

sellophane (commented):
Apart from the one comment, looks good

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
kamildoleglo: For now it is needed as it appends the `cover` class. I'll fix that in a following PR with refactored HtmlRenderer

kamildoleglo (approved):

PR #1606, opened 2020-11-06
Add javadocJar task for publishing to MavenCentral

kamildoleglo:


MarcinAman:
TODO: we need to sign those jars: https://central.sonatype.org/pages/gradle.html

MarcinAman (approved):

PR #1615, opened 2020-11-11
Make searchbar an extension point

MarcinAman:


plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -13,22 +13,24 @@ import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.plugability.configuration
 import org.jetbrains.dokka.transformers.pages.PageTransformer
 
-object NavigationPageInstaller : PageTransformer {
+open class NavigationPageInstaller : PageTransformer {
```
MarcinAman, line 16: I have some doubts it this should have a separate interface

sellophane: Yes.

plugins/base/src/main/kotlin/DokkaBase.kt
```diff
@@ -207,4 +208,8 @@ class DokkaBase : DokkaPlugin() {
             MultimodulePageCreator(it)
         }
     }
+
+    val baseSearchbarDataInstaller by extending {
```
sellophane, line 212: It's acceptable solution, but my personal preference would be to have this as yet another htmlPreprocessor, which adds to pages tree a new, special page, let's call it `PlainJsPage`, with contents from `searchbarDataInstaller.generatePagesList()` and renders it in HtmlRenderer as `outputWriter.write("scripts/pages", (page as PlainJsPage).content, ".js")`

This omits the need of having another extension point for this particular reason and is more consistent with how we generate similar pages (Index, Navigation, etc)

plugins/base/src/main/kotlin/renderers/html/SearchbarDataInstaller.kt
```diff
@@ -4,74 +4,107 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
...
+    data class PageWithId(val id: PageId, val page: ContentPage) {
+        val displayableSignature = getSymbolSignature(page)?.let { flattenToText(it) } ?: page.name
+    }
 
-    private val pageList = ConcurrentHashMap<String, Pair<String, String>>()
+    private val mapper = jacksonObjectMapper()
 
-    open fun generatePagesList(): String {
-        val pages = pageList.entries
+    open fun generatePagesList(pages: Map<PageId, PageWithId>, locationResolver: PageResolver): Json =
```
pikinier20, line 32: I think returning there Json implies that you need to use jackson. String was more flexible

pikinier20 (commented):
I think it's OK, just take a look at my previous comment there

plugins/base/src/main/kotlin/renderers/html/SearchbarDataInstaller.kt
MarcinAman: I don't think it does since JSON is just an alias for String

kamildoleglo (approved):

PR #1622, opened 2020-11-13
Remove suspend from SourceToDocumentableTranslator

kamildoleglo:


Kordyjan (changes requested):
There is a way to not abandon completely the benefits of an asynchronous translator and make it still usable from other jvm languages. We can extract sub-interface `AsynchronousSourceToDocumentableTranslator` that will have suspending version of athe invoke method and the default implementation for the non-suspending invoke that would just wrap suspending one in `runBlocking`. Dokka can check if given translator is asynchronous or not and call corresponding method.

Kordyjan (approved):

PR #1630, opened 2020-11-17
Fixes from scala3doc

romanowski:


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -143,6 +143,11 @@ open class HtmlRenderer(
             }
         }
 
+    fun FlowContent.withHtml(content: String): Unit = when (this){
```
MarcinAman, line 146: Why do we need it? It is not used anywhere

romanowski: We do need this in scala3doc. Also it will be super usefull in any other non-Kotlin usages.

MarcinAman (approved):

PR #1636, opened 2020-11-18
Exclude top-level const val from generated getters

### LINKED ISSUE #1624: dokkaJavadoc incorrectly renders top-level const val
**Describe the bug**
Javadoc output displays these as private final fields on the generated `{Filename}Kt` class, with final getters.

**Expected behaviour**
Javadoc output displays these as public static final fields on the generated `{Filename}Kt` class, without getters, reflecting the generated bytecode.

**To Reproduce**
Create Kt file with the following content:
```kotlin
const val FIRST = "String"
const val SECOND = 2
```

**Dokka configuration**

```groovy
dependencies {
    classpath "org.jetbrains.dokka:dokka-gradle-plugin:1.4.10.2"
}
...
apply plugin: 'org.jetbrains.dokka'
```

**Installation**
- Operating system: Windows
- Build tool: Gradle 6.3
- Dokka version: 1.4.10.2

bishiboosh:
This fixes #1624

### DIFF
--- a/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocClasslikeTemplateMapTest.kt
+++ b/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocClasslikeTemplateMapTest.kt
@@ -316,6 +316,33 @@ internal class JavadocClasslikeTemplateMapTest : AbstractJavadocTemplateMapTest(
         }
     }
 
+    @Test
+    fun `class with top-level const`() {
+        dualTestTemplateMapInline(
+            kotlin =
+                """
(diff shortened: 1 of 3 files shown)


bishiboosh:
I've just realized I've not tested this with Javadoc tests, and so I can't assume this is working right now, I'll look into it

MarcinAman:
Hi, is this PR ready for review?

bishiboosh:
@MarcinAman yup !

bishiboosh:
@MarcinAman done, let me know if there are any issues

kamildoleglo (approved):

MarcinAman:
Thanks ☺️

PR #1639, opened 2020-11-21
fix grammar error

HollowMan6:
an -> a
referring to https://www.quora.com/Which-is-the-correct-grammar-usage-a-unique-or-an-unique


MarcinAman (approved):

MarcinAman:
Thanks ☺️

PR #1644, opened 2020-11-27
Use maven directly to download additional dependencies

### LINKED ISSUE #1408: Does mvn's dokka:javadoc download a bunch of stuff?
Screenshot from 2020-08-30 19-08-40

This is my resources during a build.  You can see that around 30 seconds ago, it switched from CPU intensive, to a solid download.  This was the exact point where the `dokka-maven-plugin:1.4.0-rc:javadoc` task was being executed by maven.  It didn't do this in gradle.

```text
$ mvn --version
Apache Maven 3.6.0
Maven home: /usr/share/maven
Java version: 11.0.8, vendor: Ubuntu, runtime: /usr/lib/jvm/java-11-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.15.0-112-generic", arch: "amd64", family: "unix"
```

```xml
<plugin>
	<groupId>org.jetbrains.dokka</groupId>
	<artifactId>dokka-maven-plugin</artifactId>
	<version>1.4.0-rc</version>
	<executions>
		<execution>
			<phase>pre-site</phase>
			<goals>
				<goal>dokka</goal>
			</goals>
		</execution>
	</executions>
</plugin>
```

Huh.  I eventually waited maybe 4 or 5 minutes and it completed.  Rebuilding again, it only downloaded for 10 or 20 seconds.  What's it doing?  Why?  You might want to add some maven output like, "Downloading the flux capacator...  23% complete" or something.  This is a mixed Java/Kotlin/Maven project.

### LINKED ISSUE #1625: dokka-maven-plugin ignores non-standard maven local repo
**Describe the bug**
When you run your maven build and use a local maven repo at a non-standard location (not `~/.m2/repository`), then dokka-maven-plugin still downloads dependencies there. If you want to keep your dependencies separate, that's not possible then.

**Expected behaviour**
When Maven is configured or instructed to use a specific location for the local maven repository, then the default local repo location should not be modified.

**To Reproduce**
* A maven project with the dokka-maven-plugin added
* clear your default local maven repo location (e.g. `mv ~/.m2/repository ~/.m2/repository-old`)
* run `mvn clean verify -Dmaven.repo.local=$(pwd)/m2repo`

Now all dependencies should be placed into `./m2repo` and `~/.m2/repository` should not exist.

**Dokka configuration**

```xml
            <plugin>
                <groupId>org.jetbrains.dokka</groupId>
                <artifactId>dokka-maven-plugin</artifactId>
                <version>${dokka.version}</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>javadocJar</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

**Installation**
- Operating system: Linux
- Build tool: Maven
- Dokka version: 1.4.10.2

**Additional context**
The path `~/.m2/repository` seems to be hard coded here: https://github.com/Kotlin/dokka/blob/eba25082251b5cc9b6e398e780b8092f2bb9fb9c/runners/maven-plugin/src/main/kotlin/DokkaMojo.kt#L268

### LINKED ISSUE #1626: dokka-maven-plugin ignores maven.wagon.httpconnectionManager.ttlSeconds
**Describe the bug**
When using Maven in Azure (e.g. via github actions), one must configure the TTL of the pooled http connections, otherwise downloading of dependencies might fail (see actions/virtual-environments#1499).

This can be done by running Maven with these additional options: `-Dmaven.wagon.httpconnectionManager.ttlSeconds=180`. Alternatively pooling can be disabled completely via `-Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false`.

This works until the module is built, that contains dokka-maven-plugin. Then this setting is ignored. It seems, that dokka-maven-plugin downloads additional dependencies in a different way (see #1408 and #1625).

**Expected behaviour**
The http connection settings should be applied for dokka-maven-plugin as well. Ideally, dokka-maven-plugin does not need to do anything, it should just inherit/use the Maven settings.

**To Reproduce**
To make the problem evident, enable logging for http client. Start Maven with these properties:

```
-Dmaven.wagon.httpconnectionManager.ttlSeconds=180 
-Dorg.slf4j.simpleLogger.log.org.apache.maven.wagon.providers.http.httpclient=DEBUG 
-Dorg.slf4j.simpleLogger.log.org.apache.maven.wagon.providers.http.httpclient.wire=ERROR 
-Dorg.slf4j.simpleLogger.log.org.apache.maven.wagon.providers.http.httpclient.impl.conn=DEBUG 
-Dorg.slf4j.simpleLogger.log.org.apache.maven.wagon.providers.http.httpclient.impl.client=DEBUG 
-Dorg.slf4j.simpleLogger.log.org.apache.maven.wagon.providers.http.httpclient.client=DEBUG
```

Also use a fresh local maven repository, so that the dependencies are downloaded again - otherwise you might not see http client loggings at all.

You should see the normal Maven logs like `Downloading ....` and `Downloaded ... ` and in between DEBUG logs like `[DEBUG] http-outgoing-8 >> GET /maven2/....` and so on. These logs disappear as soon as dokka-maven-plugin is initialized. The dependencies for the following maven modules are downloaded without the logs of the http client.

It seems, that dokka-maven-plugin brings in an own version of the http client and replaces this within the current Maven session? 

Note: the maven http client is shaded, hence the different packages: `org.apache.maven.wagon.providers.http.httpclient` vs. `org.apache.http.impl.client`. The log is enabled only on the shaded packages, that support the http configuration from Maven.

**Dokka configuration**

```xml
            <plugin>
                <groupId>org.jetbrains.dokka</groupId>
                <artifactId>dokka-maven-plugin</artifactId>
                <version>${dokka.version}</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>javadocJar</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

**Installation**
- Operating system: Linux
- Build tool: Maven
- Dokka version: 1.4.10.2

**Additional context**
The problem seems to stem from the way how the additional plugins are downloaded, see #1408. While the preloading or loading standard plugins as default is nice, this is maybe a complexity, that is not needed?
One alternative is, to declare these additional plugin dependencies as "normal" plugin dependencies. Then these additional plugins are on the classpath and you can detect them, e.g. via [ServiceLoader](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html. If these dependencies are declared as normal dependencies, then the normal Maven mechanics to download dependencies are inherited/used.

adangel:
Use maven instead of aether to resolve additional dependencies. Aether is still used, but now indirectly by maven. Using maven directly allows to reuse all the configurations like repositories, mirrors, proxies and the local repository.

This fixes #1625 . A local test showed, that the additional dependencies like dokka-base, javadoc-plugin (and their dependencies) are downloaded correctly. If they are in the local repository already, now download is done again. This probably fixes #1408 as well.

**Note**: I've changed the code to only consider remote plugin repositories for downloading these artifacts, as this seems more correct - dokka-base/javadoc-plugin are in fact plugin dependencies and not project dependencies. This could be easily changed back, see comment.

I then removed the now unneeded aether dependencies. This seems to fix #1626. At least, if running maven with wagon-logging enabled, the httpclient logs are showing up now. I didn't test whether the TTL config is actually applied though. I assume, that one of these aether dependencies provided another plexus component for "org.sonatype.aether.RepositorySystem" which was picked up by maven in general and replaced the already existing component - thus maven didn't use anymore its bundled httpclient but the httpclient provided through transitive dependencies of aether.

~**Note:** I was not able to do a full test of my use case: dokka-maven-plugin by default fetches e.g. dokka-base with the same version - on master 1.4.10.2-SNAPSHOT. That would mean, I need to publish these already into my local repo - thus the download would not be tried again, because it's already in my local repository. So I temporarily changed this to download the stable version 1.4.10.2 for dokka-base/javadoc-plugin instead. With that setting, I could test and see, that these additional artifacts are downloaded into the correct local maven repo. However, these stable versions don't seem to be compatible anymore with the current master (java.lang.NoSuchMethodError: 'org.jetbrains.dokka.plugability.ExtensionPoint org.jetbrains.dokka.CoreExtensions.getPreMergeDocumentableTransformer()'), so I actually never fully executed dokka-maven-plugin...~ [Updated: see comment below]

### DIFF
--- a/runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
+++ b/runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
@@ -2,6 +2,10 @@ package org.jetbrains.dokka.maven
 
 import org.apache.maven.archiver.MavenArchiveConfiguration
 import org.apache.maven.archiver.MavenArchiver
+import org.apache.maven.artifact.DefaultArtifact
+import org.apache.maven.artifact.handler.DefaultArtifactHandler
+import org.apache.maven.artifact.resolver.ArtifactResolutionRequest
+import org.apache.maven.artifact.resolver.ArtifactResolutionResult
 import org.apache.maven.execution.MavenSession
 import org.apache.maven.model.Dependency
 import org.apache.maven.plugin.AbstractMojo
@@ -9,30 +13,15 @@ import org.apache.maven.plugin.MojoExecutionException
 import org.apache.maven.plugins.annotations.*
 import org.apache.maven.project.MavenProject
 import org.apache.maven.project.MavenProjectHelper
-import org.apache.maven.repository.internal.MavenRepositorySystemUtils
+import org.apache.maven.repository.RepositorySystem
 import org.codehaus.plexus.archiver.Archiver
 import org.codehaus.plexus.archiver.jar.JarArchiver
 import org.codehaus.plexus.util.xml.Xpp3Dom
-import org.eclipse.aether.DefaultRepositorySystemSession
-import org.eclipse.aether.RepositorySystem
-import org.eclipse.aether.RepositorySystemSession
-import org.eclipse.aether.artifact.DefaultArtifact
-import org.eclipse.aether.collection.CollectRequest
-import org.eclipse.aether.connector.basic.BasicRepositoryConnectorFactory
-import org.eclipse.aether.graph.DependencyNode
-import org.eclipse.aether.impl.DefaultServiceLocator
-import org.eclipse.aether.repository.LocalRepository
-import org.eclipse.aether.repository.RemoteRepository
-import org.eclipse.aether.resolution.DependencyRequest
-import org.eclipse.aether.spi.connector.RepositoryConnectorFactory
-import org.eclipse.aether.spi.connector.transport.TransporterFactory
-import org.eclipse.aether.transport.file.FileTransporterFactory
-import org.eclipse.aether.transport.http.HttpTransporterFactory
-import org.eclipse.aether.util.graph.visitor.PreorderNodeListGenerator
 import org.jetbrains.dokka.*
 import org.jetbrains.dokka.DokkaConfiguration.ExternalDocumentationLink
 import java.io.File
 import java.net.URL
+import java.util.stream.Collectors
 
 class SourceLinkMapItem {
     @Parameter(name = "path", required = true)
@@ -58,11 +47,18 @@ class ExternalDocumentationLinkBuilder {
 
 abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependency>) : AbstractMojo() {
 
-    @Parameter(defaultValue = "\${project}", readonly = true)
-    private var mavenProject: MavenProject? = null
+    @Parameter(defaultValue = "\${project}", readonly = true, required = true)
+    protected var mavenProject: MavenProject? = null
 
-    @Parameter()
-    private var session: RepositorySystemSession? = null
+    /**
+     * The current build session instance. This is used for
+     * for dependency resolver API calls via repositorySystem.
+     */
+    @Parameter(defaultValue = "\${session}", required = true, readonly = true)
+    protected var session: MavenSession? = null
+
+    @Component
+    private var repositorySystem: RepositorySystem? = null;
 
     class PackageOptions : DokkaConfiguration.PackageOptions {
         @Parameter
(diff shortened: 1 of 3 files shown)


runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -254,52 +250,27 @@ abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependenc
...
-        repoSystem.resolveDependencies(session, dependencyRequest)
-        val nlg = PreorderNodeListGenerator()
-        node.accept(nlg)
-        return nlg.files
+
+        val request: ArtifactResolutionRequest = ArtifactResolutionRequest()
+        request.isResolveRoot = true
+        request.isResolveTransitively = true
+        request.localRepository = session!!.localRepository
+        request.remoteRepositories = mavenProject!!.pluginArtifactRepositories
```
adangel, line 263: Only the configured plugin repositories are considered. This could be easily changed to consider also the project repositories:

```
request.remoteRepositories = mavenProject!!.remoteArtifactRepositories.plus(mavenProject!!.pluginArtifactRepositories)
```

adangel:
I could now test the complete use case and it works. The plugins are downloaded via maven's own wagon transport, allowing to set TTL and other http parameters.

First I didn't have all snapshots in my test snapshot repo, so javadoc-plugin was not found. I've added a resolutionErrorHandler to break the build if any artifact could not be resolved. Otherwise dokka still executes, but just without the missing plugins giving undesired results.

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -254,52 +254,30 @@ abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependenc
...
+        request.isOffline = session!!.isOffline
+        request.isForceUpdate = session!!.request.isUpdateSnapshots
+        request.servers = session!!.request.servers
+        request.mirrors = session!!.request.mirrors
+        request.proxies = session!!.request.proxies
+        request.artifact = DefaultArtifact(groupId, artifactId, version, "compile", "jar", null,
```
kamildoleglo, line 273: Please use the `apply` block here

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
kamildoleglo: That should not be necessary IMHO

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -254,52 +254,30 @@ abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependenc
...
+                DefaultArtifactHandler("jar"))
+
+        log.debug("Resolving $groupId:$artifactId:$version ...")
+
+        val result: ArtifactResolutionResult = repositorySystem!!.resolve(request)
+        resolutionErrorHandler!!.throwErrors(request, result);
+        return result.artifacts.stream().map { it.file }.collect(Collectors.toList())
```
kamildoleglo, line 280:
```suggestion
        return result.artifacts.map { it.file }
```
should suffice

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -254,52 +254,30 @@ abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependenc
...
+        request.servers = session!!.request.servers
+        request.mirrors = session!!.request.mirrors
+        request.proxies = session!!.request.proxies
+        request.artifact = DefaultArtifact(groupId, artifactId, version, "compile", "jar", null,
+                DefaultArtifactHandler("jar"))
+
+        log.debug("Resolving $groupId:$artifactId:$version ...")
+
+        val result: ArtifactResolutionResult = repositorySystem!!.resolve(request)
+        resolutionErrorHandler!!.throwErrors(request, result);
```
kamildoleglo, line 279:
```suggestion
        resolutionErrorHandler!!.throwErrors(request, result)
```

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -58,11 +48,21 @@ class ExternalDocumentationLinkBuilder {
...
 abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependency>) : AbstractMojo() {
 
-    @Parameter(defaultValue = "\${project}", readonly = true)
-    private var mavenProject: MavenProject? = null
+    @Parameter(defaultValue = "\${project}", readonly = true, required = true)
+    protected var mavenProject: MavenProject? = null
+
+    /**
+     * The current build session instance. This is used for
+     * for dependency resolver API calls via repositorySystem.
```
kamildoleglo, line 56: Typo: "for" two times

kamildoleglo (changes requested):
Thanks for this contribution! I only noticed a few minor issues, if you have time to fix them, please do ☺️

adangel:
@kamildoleglo Thanks for the review! I've pushed a fix.

kamildoleglo:
Thanks for this PR ☺️

PR #1655, opened 2020-12-02
Resolve DRI-s from valueOf and values in java enum

MarcinAman:


plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -336,9 +336,13 @@ class DefaultPsiToDocumentableTranslator(
         private fun parseFunction(
             psi: PsiMethod,
             isConstructor: Boolean = false,
-            inheritedFrom: DRI? = null
+            inheritedFrom: DRI? = null,
+            parentDRI: DRI? = null
         ): DFunction {
-            val dri = DRI.from(psi)
+            val currentDri = DRI.from(psi)
```
kamildoleglo, line 342: `currentDri` is only used once in the line below, so it would be better to inline it there

kamildoleglo (changes requested):

PR #1661, opened 2020-12-04
Add dokka to GitHub link

StefMa:
I just discovered a dokka website in a friends pet project.
At the bottom right I saw this nice `Sponsored and developed by dokka ->` text.
I wanted to press at `dokka` but realized that only the arrow is clickable.
Geist_1607084524165

I guess it is a common pattern that "a text" is also linkend in such cases.
Therefore I moved the "dokka" text into the anchor as well. At least I hope so 😁 
I have never used `kotlinx.html` nor found a good documentation :( 
So I basically copied the `span` `text` thingy into the anchor tag. I hope this is working 😁 .
If not, sorry. Please guide me how to do it correctly 😇

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -777,8 +777,9 @@ open class HtmlRenderer(
                             }
                             span { text("© 2020 Copyright") }
                             span("pull-right") {
-                                span { text("Sponsored and developed by dokka") }
+                                span { text("Sponsored and developed by ") }
                                 a(href = "https://github.com/Kotlin/dokka") {
+                                    span { text("dokka") }
                                     span(classes = "padded-icon")
                                 }
                             }
--- a/plugins/base/src/main/resources/dokka/styles/style.css
+++ b/plugins/base/src/main/resources/dokka/styles/style.css
@@ -1109,6 +1109,10 @@ td.content {
     padding: 0 16px;
 }
 
+.footer a {
+    color: var(--breadcrumb-font-color);
+}
+
 .footer .padded-icon {
     padding-left: 0.5em;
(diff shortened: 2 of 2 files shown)


MarcinAman:
Thanks ☺️ 
Could you also update the font color to match the others?
image

StefMa:
You're welcome.

Can you please say which color do you mean and how change this in code?
Thank you ☺

MarcinAman:
i'd like the link to have the same color that the text has (i think it is `var(--breadcrumb-font-color)` in css). Css file is here: https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/resources/dokka/styles/style.css

CSS selector can be something like `.footer a`

StefMa:
Done ✅ 
Hopefully correctly 😁 
Thanks for guidance!

Btw. Is there a test which outputs a sample Dokka page for each PR? 🤔 
In case something changed at the frontend (like in this PR) it can be directly seen a reviewer.
If not, does it make sense to create something like that?

MarcinAman:
we have integration tests for some libraries that are built on teamcity and then uploaded to S3. They mostly use html

Unfortunately right now there is no way for external contributor to see the results, i'll try do make it better since we have more and more of those 🥇

MarcinAman (approved):

MarcinAman:
Thanks for the contribution 🥇

StefMa:
A simple solution would be to copy [this Action](https://github.com/Kotlin/dokka/blob/e2be33f25da62ac25ea95b160540e30cc7afca1f/.github/workflows/s3-snapshots.yml#L44-L75) but instead of pushing it to S3 you could save it as an [artifact](https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts).
The output are simply static files. There is no need to upload them anywhere for PRs at least 🙂

PR #1675, opened 2020-12-15
Parsing of JvmName

MarcinAman:
First part of the annotation story, in this PR i handle `@JvmName` annotation both on file and property/function level.


bishiboosh:
(in the end I'm really happy you took this task yourself, the code is way cleaner than what I've tried to do, perks of knowing the codebase well I guess :D)

plugins/kotlin-as-java/src/main/kotlin/KotlinAsJavaPlugin.kt
```diff
@@ -13,13 +14,20 @@ class KotlinAsJavaPlugin : DokkaPlugin() {
     val kotlinAsJavaDocumentableTransformer by extending {
         CoreExtensions.documentableTransformer with KotlinAsJavaDocumentableTransformer()
     }
+
+    val jvmNameTransformer by extending {
```
kamildoleglo, line 18: IMHO this should run after the KotlinToJavaConverter.kt

kamildoleglo (changes requested):

PR #1688, opened 2021-01-07
Add maven central publication

kamildoleglo:


runners/gradle-plugin/build.gradle.kts
```diff
@@ -81,8 +70,10 @@ publishing {
     }
 }
 
+tasks.withType<PublishToMavenRepository>().configureEach {
```
MarcinAman, line 73: what it does?

MarcinAman (approved):

PR #1691, opened 2021-01-08
Footer customisation

MarcinAman:


kamildoleglo (approved):

gnarea:
Thanks for this! Any idea when the next release that will ship this change will be?

MarcinAman:
I am preparing release notes so it should be released this week

FWDekker:
Thanks! I'm a bit confused, though. Where do I add the `footerMessage` property? The documentation talks about the "base plugin configuration" but I have no idea where that is. I can't seem to add it to `dokkaHtml.configure` or to my source set's configuration or anywhere else where I'd expect to have to add it.

MarcinAman:
Here is an example of base plugin configuration: https://github.com/Kotlin/kotlin-examples/blob/25040fe92994ab46ebac6a6450d3ad1d3e870af1/gradle/dokka/dokka-customFormat-example/build.gradle.kts#L21

but instead of passing customAssets/stylesheets you need to add `footerMessage` there. 

Also it is possible to do it without adding base plugin to buildScript:
```
pluginsMapConfiguration.set(mapOf("org.jetbrains.dokka.base.DokkaBase" to """{ "footerMessage": "<div style=\"color: red\">custom message</div>" }"""))
```

You can read more on those in our gradle docs:
https://kotlin.github.io/dokka/1.4.32/user_guide/gradle/usage/#applying-plugins

FWDekker:
Oh I see, i didn't get that the base plugin was a separate plugin. I thought it was the base of the plugin, like a shared module or something.

Either way, the snippet you provided works perfectly, thank you!

bogdan-richard:
I might be missing something, but I cannot add my own footer for **dokkaJavadoc** tasks.  Tried with @MarcinAman!s example, the only change is this is in Groovy:
```
pluginsMapConfiguration.set(["org.jetbrains.dokka.plugability.DokkaBase" : """{ "footerMessage": "<div style=\"color: red\">custom message</div>" }"""])
```
Could anyone please give some hints?

MarcinAman:
@bogdan-richard aforementioned customisation option only works for html output format. Currently your are not able to customise footer in javadoc

bogdan-richard:
Thanks @MarcinAman for the quick response!

PR #1702, opened 2021-01-15
Bring back displaying jvm name when KaJ is not present

MarcinAman:


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -931,21 +928,18 @@ private class DokkaDescriptorVisitor(
...
+            mustBeDocumented(),
             scope
         )
-    }
 
-    private fun AnnotationDescriptor.mustBeDocumented(dri: DRI): Boolean =
-        if (dri.isJvmName()) false
-        else annotationClass!!.annotations.hasAnnotation(FqName("kotlin.annotation.MustBeDocumented"))
+    private fun AnnotationDescriptor.mustBeDocumented(): Boolean =
+        annotationClass!!.annotations.hasAnnotation(FqName("kotlin.annotation.MustBeDocumented"))
```
kamildoleglo, line 942: Please change to false by default

plugins/kotlin-as-java/src/main/kotlin/transformers/JvmNameDocumentableTransformer.kt
```diff
@@ -42,6 +44,17 @@ class JvmNameDocumentableTransformer : DocumentableTransformer {
             }
         } as T
 
+    private fun PropertyContainer<DFunction>.withoutJvmName(): PropertyContainer<DFunction> {
+        val annotationsWithoutJvmName = get(Annotations)?.let { annotations ->
+            annotations.copy((annotations.directAnnotations + annotations.fileLevelAnnotations).map { (sourceset, annotations) ->
+                sourceset to annotations.filterNot { it.isJvmName() }
```
kamildoleglo, line 50: File-level name won't be processed

kamildoleglo (changes requested):

kamildoleglo (approved):

PR #1712, opened 2021-01-27
JvmOverloads

bishiboosh:
This feature adds `JvmOverloads` support by generating the required functions.

As `DFunction.asJava` was modified, I took the liberty of also fixing the fact that top-level functions weren't marked as static.


bishiboosh:
(cc @MarcinAman, this is the first PR, I'll try to do the `JvmSynthetic` one tomorrow)

plugins/kotlin-as-java/src/main/kotlin/jvmOverloads.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.dokka.kotlinAsJava
+
+import org.jetbrains.dokka.model.Annotations
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.properties.WithExtraProperties
+import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
+
+internal fun WithExtraProperties<out Documentable>.jvmOverloads(): Annotations.Annotation? =
```
kamildoleglo, line 8: I'd personally call it something like `hasJvmOverloads` just to be more specific

plugins/kotlin-as-java/src/main/kotlin/jvmOverloads.kt
```diff
@@ -0,0 +1,12 @@
...
+    extra[Annotations]?.directAnnotations?.entries?.firstNotNullResult { (_, annotations) -> annotations.jvmOverloadsAnnotation() }
+
+internal fun List<Annotations.Annotation>.jvmOverloadsAnnotation(): Annotations.Annotation? =
```
kamildoleglo, line 11: Since this is not used in any other place than the line before, I'd make it private, make it return `Boolean` and call it `hasJvmOverloads`; or alternatively just create a method `isJvmOverloads` that would check if an annotation is a `@JvmOverloads` one and use `annotations.any { it.isJvmOverloads() }` in line 9. 

It is functionally equivalent but just looks better IMHO

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -184,8 +185,36 @@ internal fun DFunction.asJava(containingClassName: String): DFunction {
...
+    val baseFunction = asJava(containingClassName, newName, parameters, isTopLevel)
+    return if (jvmOverloads() == null) {
+        listOf(baseFunction)
+    } else {
+        val functions = mutableListOf(baseFunction)
+        val parameters = this.parameters.toMutableList()
+        while (parameters.removeLastIf { it.extra[DefaultValue] != null }) {
+            functions.add(asJava(containingClassName, newName, parameters, isTopLevel))
+        }
+        functions
```
kamildoleglo, line 216: IMHO this could be written like this and then we can remove the `removeLastIf` entirely
```suggestion
        val endParamsWithDefaults = parameters.takeLastWhile { it.extra[DefaultValue] != null }
        val startParamsWithoutDefaults = parameters - endParamsWithDefaults

        endParamsWithDefaults.runningFold(startParamsWithoutDefaults) { acc, param -> acc + param }.map { params -> asJava(containingClassName, newName, params, isTopLevel) }

```

kamildoleglo (changes requested):
Nice PR, thanks! Just a few minor comments

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
bishiboosh: I took your advice on this and extracted the logic to another method but changed a little from what you gave, both because your code didn't take the case where there was a non-default param between default params and to ensure the order of generated functions respects what's in the [`JvmOverloads` documentation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/-init-.html)

MarcinAman (approved):

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
kamildoleglo: Nice catch! Thanks ☺️

kamildoleglo (approved):

MarcinAman:
Thanks ☺️

PR #1738, opened 2021-02-09
Fix displaying default java visibility

MarcinAman:


plugins/base/src/test/kotlin/model/ClassesTest.kt
```diff
@@ -309,15 +309,6 @@ class ClassesTest : AbstractModelTest("/src/main/kotlin/classes/Test.kt", "class
...
 
-//    @Test fun companionObjectExtension() {
-//        checkSourceExistsAndVerifyModel("testdata/classes/companionObjectExtension.kt", defaultModelConfig) { model ->
-//            val pkg = model.members.single()
-//            val cls = pkg.members.single { it.name == "Foo" }
-//            val extensions = cls.extensions.filter { it.kind == NodeKind.CompanionObjectProperty }
-//            assertEquals(1, extensions.size)
-//        }
-//    }
-
```
kamildoleglo, line 320: Can we reenable it?

MarcinAman: I guess there is no point since the same test is done line above. I've changed it so i actually does something rather than being just a TODO

PR #1789, opened 2021-03-17
Fix configuration for suppressing obvious functions

MarcinAman:
I am wondering whether or not this should be in sourceset configuration. Technically you could have some sourcesets that you would like to show those functions....

@kamildoleglo ?


runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -159,6 +156,9 @@ abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependenc
     @Parameter
     var failOnWarning: Boolean = DokkaDefaults.failOnWarning
 
+    @Parameter
```
kamildoleglo, line 159: IMHO it would be better to annotate it with `@Parameter(defaultValue = "${DokkaDefaults.suppressObviousFunctions}")` I'm not sure how Kotlin default values play with Maven

kamildoleglo (changes requested):

PR #1804, opened 2021-03-24
Reintroduce multimodule documentation

MarcinAman:


core/src/main/kotlin/configuration.kt
```diff
@@ -100,6 +100,7 @@ interface DokkaConfiguration : Serializable {
     val pluginsConfiguration: List<PluginConfiguration>
     val delayTemplateSubstitution: Boolean
     val suppressObviousFunctions: Boolean
+    val includes: Set<File>
```
kamildoleglo, line 103: I'm not sure if this is the best place for this property. This configuration is AFAIK shared with single- and multi-module tasks and single-module tasks already have a configuration option with this name in the source sets. We can still add it here though, but we should support it in the single-module then

kamildoleglo (changes requested):

kamildoleglo (approved):

PR #1805, opened 2021-03-25
Handle html in kdoc

MarcinAman:


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -707,6 +707,9 @@ open class HtmlRenderer(
 
     override fun FlowContent.buildText(textNode: ContentText) =
         when {
+            textNode.hasStyle(ContentStyle.Html) -> {
+                consumer.onTagContentUnsafe { raw(textNode.text) }
+            }
```
kamildoleglo, line 712: Can we add it to `extra` instead of `styles` ?

kamildoleglo (changes requested):

kamildoleglo (approved):

PR #1808, opened 2021-03-26
Replace jcenter with mavenCentral in usage doc

LouisCAD:
I think dokka is fully published to Maven Central now, so it makes sense to update the usage doc.

### DIFF
--- a/docs/src/doc/docs/user_guide/gradle/usage.md
+++ b/docs/src/doc/docs/user_guide/gradle/usage.md
@@ -14,7 +14,7 @@ plugins {
 }
 
 repositories {
-    jcenter() // or maven(url="https://dl.bintray.com/kotlin/dokka")
+    mavenCentral()
 }
 ```
 
@@ -23,7 +23,7 @@ settings.gradle.kts:
 pluginManagement {
     repositories {
         gradlePluginPortal()
-        jcenter()
+        mavenCentral()
     }
 }
 ```


MarcinAman:
Hi, this is not yet possible since we rely on kotlinx.html that is not present on maven central. So the minimal set of repositories would be something like this:
```
    maven("https://kotlin.bintray.com/kotlinx")
    mavenCentral()
```

asarkar:
> we rely on kotlinx.html 
@MarcinAman Can this be documented? Removing jcenter causes the build to fail:
```
Execution failed for task ':dokkaHtml'.
> Could not resolve all files for configuration ':dokkaHtmlPlugin'.
   > Could not find org.jetbrains.kotlinx:kotlinx-html-jvm:0.7.2.
     Required by:
         project : > org.jetbrains.dokka:dokka-base:1.4.30
```

kamildoleglo:
@asarkar it is documented in the README

MarcinAman:
@LouisCAD Information about jcenter still holds true. AFAIK you can still just use jcenter (without maven central) if you want to, thats why i don't know why we should update it. 

Ofc dokka is present on maven central but we are still waiting for kotlinx.html to be published. Thats why you can use mavenCentral + jcenter or mavenCentral + `maven { url "https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven" }` (https://github.com/kotlin/kotlinx.html/wiki/Getting-started)

MarcinAman:
i think this PR can be closed as it was resolved in: https://github.com/Kotlin/dokka/commit/8c7ab58f9e2adcf86a7da4697c99dde38db91ac8

PR #1812, opened 2021-03-29
Fix some compiler warnings

kamildoleglo:


plugins/base/src/main/kotlin/transformers/documentables/ActualTypealiasAdder.kt
```diff
@@ -8,8 +8,8 @@ import org.jetbrains.dokka.transformers.documentation.DocumentableTransformer
 
 class ActualTypealiasAdder : DocumentableTransformer {
 
-    override fun invoke(modules: DModule, context: DokkaContext) = modules.generateTypealiasesMap().let { aliases ->
-        modules.copy(packages = modules.packages.map { it.copy(classlikes = addActualTypeAliasToClasslikes(it.classlikes, aliases)) })
```
MarcinAman, line 12: why?

MarcinAman (approved):

plugins/base/src/main/kotlin/transformers/documentables/ActualTypealiasAdder.kt
kamildoleglo: There was a warning that in the `DocumentableTransformer` interface the argument is called `original` instead of `modules` and this may introduce some bugs, so I renamed it

PR #1831, opened 2021-04-09
Fix several bugs in configuration examples in gradle user guide

AzimMuradov:
Fixes
- `".*\.internal.*"` -> `""".*\.internal.*"""` in "Configuration options" example (**in regular kotlin strings `\` is not allowed on its own**)
- `sourceRoot.from(...)` -> `sourceRoots.from(...)` in "Configuration options" example (**typo**)
- `url = ...` -> `url.set(...)` in "Configuration options" example (**works in groovy, but not in kotlin**)
- indention in "Kotlin Multiplatform plugin configuration" example (**typo**)


kamildoleglo (approved):

MarcinAman (approved):
Thanks!

PR #1858, opened 2021-04-16
Remove unused repositories

MarcinAman:


buildSrc/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
 
 repositories {
     mavenCentral()
-    jcenter()
+    gradlePluginPortal()
 }
```
kamildoleglo, line 8: This can probably removed altogether

MarcinAman: gradlePluginPortal() has to stay because otherwise we get unresolved dependencies

kamildoleglo (approved):

PR #1886, opened 2021-04-23
Option to Suppress annotations

### LINKED ISSUE #1861: Option to skip things marked with an annotation
I have a few `@RequiresOptIn` annotations that I'm using in my project.  One of them is for internal APIs that I need to have public for other modules.  I would like to not generate documentation for anything with this annotation.

A way to skip declarations that have certain annotations.

Skip declarations for any `@RequiresOptIn` annotation: this is bad because `@RequiresOptIn` can be used for experimental annotations as well as internal, or things like a plugin API that will be used in some cases but not all.

A selector in the HTML docs (and potentially others) to filter by `@RequiresOptIn` annotations that are registered in the configuration (w/ configurable defaults): I would like to see this as well, but it's not exclusive with skipping some altogether.

rnett:
Fixes #1861.

I did this in the base plugin rather than a separate plugin since it fits with things like `supressedFiles` and `skipDeprecated`, and it seemed like overkill to make a new plugin just for this.

I can't run tests locally due to `Node Sass does not yet support your current environment: Windows 64-bit with Unsupported runtime (72)`, any idea how to fix?  I'm assuming the CI will catch anything.

### DIFF
--- a/core/src/main/kotlin/configuration.kt
+++ b/core/src/main/kotlin/configuration.kt
@@ -135,6 +135,7 @@ interface DokkaConfiguration : Serializable {
         val noStdlibLink: Boolean
         val noJdkLink: Boolean
         val suppressedFiles: Set<File>
+        val suppressedAnnotations: Set<String>
         val analysisPlatform: Platform
     }
 
--- a/core/src/main/kotlin/defaultConfiguration.kt
+++ b/core/src/main/kotlin/defaultConfiguration.kt
@@ -49,6 +49,7 @@ data class DokkaSourceSetImpl(
     override val noStdlibLink: Boolean = DokkaDefaults.noStdlibLink,
     override val noJdkLink: Boolean = DokkaDefaults.noJdkLink,
     override val suppressedFiles: Set<File> = emptySet(),
+    override val suppressedAnnotations: Set<String> = emptySet(),
     override val analysisPlatform: Platform = DokkaDefaults.analysisPlatform,
 ) : DokkaSourceSet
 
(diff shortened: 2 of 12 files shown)


MarcinAman:
I think your issue is related to the java version. Which one are you using?

As for the PR: I don't see an appealing reason to have this functionality because use case for that is insanely limited because most people treat those values as a part of a public API, which it is in fact since it is available publicly. To my mind it is a way of fixing issues related to repository structure in docs. As an extension, this should be your own plugin. You can create one using our template: https://github.com/Kotlin/dokka-plugin-template and publish it to maven central or other repository of choice

rnett:
I'm using Java 14.

You may have a better idea of the target use for this than I do, but as I understand it the `@RequiresOptIn` annotation is there specifically to support APIs that need to have public visibility but aren't "public" semantically ([KTIJ-933](https://youtrack.jetbrains.com/issue/KTIJ-933), [KT-37649](https://youtrack.jetbrains.com/issue/KT-37649), [KT-40231](https://youtrack.jetbrains.com/issue/KT-40231)).  Perhaps hiding them isn't the right way to go about this, but I do think it would be very useful for dokka to have an equivalent to the "lower resolution in IDE unless opted-in" functionality (tags maybe?).

MarcinAman:
I think they are public despite annotation. They may have lower priority in completion but they are still preset there. And it makes sense because you still can use them, it just require a compilation flag or an annotation. 

To my mind filtering by annotation doesn't solve a problem popular enough to include in main product. That said, dokka is pluggable so everyone is free to create any plugin he wants for their documentation. If you are interested in having it in your docs please create a one and share it with the world ☺️

PR #1887, opened 2021-04-24
Fix line separator issues

### LINKED ISSUE #1883: Module/package description linebreaks have no space
**Describe the bug**
Markdown linebreaks in module/package description (i.e. two spaces followed by a a line break) cause a line break, but no space in the generated HTML.

This may be an issue with JetBrains/markdown instead.

**Expected behaviour**
Line breaks are added with some spacing, like in the IDE markdown preview (shown) or GitHub:
image

**Screenshots**
Rendered docs:
image

**To Reproduce**
Use this file:
https://github.com/rnett/krosstalk/blob/8877632b5d124742bc903c285e67612cbf7011a9/core/krosstalk-client/README.md

**Dokka configuration**
https://github.com/rnett/krosstalk/blob/8877632b5d124742bc903c285e67612cbf7011a9/build.gradle.kts

**Installation**
- Operating system: Windows
- Build tool: Gradle v6.8.3
- Dokka version: 1.4.32

### LINKED ISSUE #1884: HTML: Code block in module/package doc isn't closed in render
**Describe the bug**
Seems to be the same as #433.  I have a code block in my module description:
```md
```kotlin
actual object MyKrosstalk : Krosstalk(), KtorKrosstalkClient {
    actual override val serialization = KotlinxBinarySerializationHandler(Cbor { })

    override val client: KtorClient = KtorClient(HttpClient(Apache))
    override val serverUrl: String = "http://localhost:8080"

    actual object Auth : Scope, KtorClientBasicAuth()
}
\```
```
(the \\``` is for GitHub)

It is rendered as:
image

**Expected behaviour**
For the code block to work correctly.

**To Reproduce**
https://github.com/rnett/krosstalk/blob/34325f775480b9a65ef730628873355659e7069e/core/krosstalk-client/README.md

**Dokka configuration**
https://github.com/rnett/krosstalk/blob/34325f775480b9a65ef730628873355659e7069e/build.gradle.kts

**Installation**
- Operating system: Windows
- Build tool: Gradle v6.8.3
- Dokka version: 1.4.32

rnett:
Simple fix for #1884 and #1883, just replaces CRLF line separators with LF before markdown parsing.  Ideally this should be fixed in the parser, but this works for now.

Also adds a test for this.

For GitHub:
Fixes #1883
Fixes #1884

### DIFF
--- a/plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
+++ b/plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
@@ -12,8 +12,24 @@ import org.intellij.markdown.flavours.gfm.GFMFlavourDescriptor
 import org.intellij.markdown.flavours.gfm.GFMTokenTypes
 import org.jetbrains.dokka.base.parsers.factories.DocTagsFromIElementFactory
 import org.jetbrains.dokka.links.DRI
-import org.jetbrains.dokka.model.doc.*
+import org.jetbrains.dokka.model.doc.Author
+import org.jetbrains.dokka.model.doc.Constructor
+import org.jetbrains.dokka.model.doc.CustomTagWrapper
+import org.jetbrains.dokka.model.doc.Description
+import org.jetbrains.dokka.model.doc.DocTag
+import org.jetbrains.dokka.model.doc.DocumentationLink
+import org.jetbrains.dokka.model.doc.DocumentationNode
+import org.jetbrains.dokka.model.doc.Param
+import org.jetbrains.dokka.model.doc.Property
+import org.jetbrains.dokka.model.doc.Receiver
+import org.jetbrains.dokka.model.doc.Return
+import org.jetbrains.dokka.model.doc.Sample
+import org.jetbrains.dokka.model.doc.See
+import org.jetbrains.dokka.model.doc.Since
 import org.jetbrains.dokka.model.doc.Suppress
+import org.jetbrains.dokka.model.doc.TagWrapper
+import org.jetbrains.dokka.model.doc.Text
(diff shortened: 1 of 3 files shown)


rnett:
Tests are failing because they were in the wrong spot, `parseModuleAndPackageDocFragment` does not use the markdown parser.  `parseModuleAndPackageDocFragment` supports CRLF fine, but leaves them in the output.  Is that intentional/a good idea, or should I switch those to LF as well (it won't line up with the markdown parser unless I do)?

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -37,7 +53,7 @@ open class MarkdownParser(
         return visitNode(markdownAstRoot)
     }
 
-    override fun preparse(text: String) = text
+    override fun preparse(text: String) = text.replace("\r\n", "\n").replace("\r", "\n")
```
MarcinAman, line 56: I have a bad feeling about that. What would happen if somebody would add `\r\n` in their documentation? Or any other escaped version of those characters?

MarcinAman (changes requested):

plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
rnett: I assume (unless you're doing some kind of preprocessing somewhere else) that they won't be unescaped, i.e. a "\n" in the markdown document will be parsed as `['\', 'n']` i.e. `"\\n"`, not `['\n']`.  I'll add this to the tests.

PR #1890, opened 2021-04-25
Add ability to specify older versions as a list of files

### LINKED ISSUE #1889: Versioning: option to specify list of files or glob instead of directory.
I'm publishing a few different sets of documentation, all of which use the versioning plugin.  The structure looks something like:
```
${version}
    core
    plugins
    plugin-development
```

Currently, the old versions for versioning must be in the same directory, which does not work with the structure above.

I would like to be able to pass a list of files (well, directories) or a glob to the versioning plugin instead of being limited to a directory,

Probably.

rnett:
Fixes #1889.  I'm not aware of a good Java globbing implementation (the JDK one doesn't use base directories) unless you already have a dependency for one.  If you do I can add a setter method.

The older version directory's subfolders and the older versions list are added together, so either or both can be specified.  I don't love the `olderVersions` name but I couldn't think of a better one.  `additionalOlderVersions` maybe?

### DIFF
--- a/plugins/versioning/src/main/kotlin/versioning/VersioningConfiguration.kt
+++ b/plugins/versioning/src/main/kotlin/versioning/VersioningConfiguration.kt
@@ -5,6 +5,7 @@ import org.jetbrains.dokka.plugability.DokkaContext
 import java.io.File
 
 data class VersioningConfiguration(
+    var olderVersionsDir: File? = defaultOlderVersionsDir,
     var olderVersions: List<File>? = defaultOlderVersions,
     var versionsOrdering: List<String>? = defaultVersionsOrdering,
     var version: String? = defaultVersion,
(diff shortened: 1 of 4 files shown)


rnett:
I managed to fix my build issues by updating `sass-loader` to `^10.0.5` so that it supports `node-sass` 5+.  Can I commit that here or should I open another PR?

plugins/versioning/src/main/kotlin/versioning/VersioningConfiguration.kt
```diff
@@ -6,14 +6,23 @@ import java.io.File
...
+    var olderVersions: List<File>? = defaultOlderVersions,
     var versionsOrdering: List<String>? = defaultVersionsOrdering,
     var version: String? = defaultVersion,
 ) : ConfigurableBlock {
     internal fun versionFromConfigurationOrModule(dokkaContext: DokkaContext): String =
         version ?: dokkaContext.configuration.moduleVersion ?: "1.0"
 
+    internal fun allOlderVersions(): List<File> {
+        if (olderVersionsDir != null)
+            assert(olderVersionsDir!!.isDirectory) { "Supplied previous version $olderVersionsDir is not a directory!" }
```
MarcinAman, line 18: Are kotlin's smartcasts not working here?

MarcinAman (approved):
Please update the docs about versioning. An example of this style of configuration also would be nice to put there

MarcinAman:
>I managed to fix my build issues by updating sass-loader to ^10.0.5 so that it supports node-sass 5+. Can I commit that here or should I open another PR?

Please do a PR. Thanks!

plugins/versioning/src/main/kotlin/versioning/VersioningConfiguration.kt
rnett: It's since it's a `var`.  I could make it a `?.let` if it's an issue.

MarcinAman:
Thanks!

Issue #1867, opened 2021-04-21
Provide DSL to configure partial and non-partial tasks at the same time.


rnett:
**Is your feature request related to a problem? Please describe**
I have gotten tripped up several times now by configuring `dokkaHtml` and then using the multi module task, or configuring `org.jetbrains.dokka.gradle.DokkaTask` when I mean to use `org.jetbrains.dokka.gradle.DokkaTaskPartial`.  I'm not actually trying to generate multi-module and single-module docs, but that is a possible use case as well.

**Describe the solution you'd like**
A way to configure format-specific tasks for partial (multi-module) and non-partial.  Ideally something like a `dokka.html{` block (not named that, probably), but at minimum lift `dokkaSourceSets` to `AbstractDokkaTask` since it's the same type in both subclasses.  Ideally, though, the groupings of formats and partial/non-partial would be orthogonal, so I could configure groups like "all tasks", "all HTML tasks", "all partial tasks", "all (well, the) partial HTML task", etc.

Extensions for `tasks.withType<org.jetbrains.dokka.gradle.DokkaTaskPartial>()`, `tasks.withType<org.jetbrains.dokka.gradle.DokkaTask>() `, and `tasks.withType<org.jetbrains.dokka.gradle.AbstractDokkaTask>() ` would also be nice if this route is taken.

Personally, I'd move away from the task based configuration as the default and use something more like  plugin:
```kotlin
dokka{
    html{
        partial{
            // partial only
        }
        // all html
    }
    // all
    partial{
        // all partial
    }
}
```
but that's just personal preference.

**Describe alternatives you've considered**
See above, I'm not suggesting something specific.

MarcinAman:
I don't see a use case behind it except prototyping. The thing is projects are either single or multi module, there is no way that this changes dynamically. So you either configure partials for the multimodule or dokka tasks for single module. Any mix of those is invalid. 

Could you elaborate more on your use case? Why do you switch between single and multimodule docs?

rnett:
The change over was when I added modules during development.  I'm publishing multiple sets of docs to work around #1888 and #1862, and one of those sets went from being a single module to multi module when I restructured the project.  I have also been using the single module tasks to inspect a single module's docs without re-building the whole multi-module project.  It's possible to use multi-module and single module at the same time if you aren't generating a single set of docs for the entire project.

[...]

PR #1891, opened 2021-04-25
Adds a supertype for leaf dokka tasks that defines dokkaSourceSets

rnett:
Helps with #1867.  I'm not familiar enough with Gradle to write a proper extension to configure things outside of the tasks.

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
@@ -6,10 +6,10 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Nested
 import org.jetbrains.dokka.*
 
-abstract class DokkaTask : AbstractDokkaTask() {
+abstract class DokkaTask : AbstractDokkaLeafTask() {
 
     @get:Internal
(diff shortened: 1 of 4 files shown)


MarcinAman (approved):

lastthyme (approved):

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaTask.kt
```diff
@@ -6,10 +6,10 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Nested
 import org.jetbrains.dokka.*
 
-abstract class DokkaTask : AbstractDokkaTask() {
+abstract class DokkaTask : AbstractDokkaLeafTask() {
 
     @get:Internal
-    val dokkaSourceSets: NamedDomainObjectContainer<GradleDokkaSourceSetBuilder> =
+    override val dokkaSourceSets: NamedDomainObjectContainer<GradleDokkaSourceSetBuilder> =
```
kamildoleglo, line 12: Could you extract the implementation of this property and the second one in the `DokkaTaskPartial` class to the `AbstractDokkaLeafTask` instead of overriding it please, since the implementations are the same AFAIK

kamildoleglo (changes requested):

MarcinAman:
Thanks!

PR #1905, opened 2021-05-05
Cachable task

MarcinAman:


integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleCachingIntegrationTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        val customResourcesDir = File(templateProjectDir, "customResources")
+        if(customResourcesDir.exists() && customResourcesDir.isDirectory) {
+            val destination = File(project.parentFile, "customResources")
+            destination.mkdirs()
+            destination.deleteRecursively()
+            customResourcesDir.copyRecursively(destination)
+        }
+
+        // clean local cache for each test
+        if (versions.gradleVersion.version == "7.0") {
```
kamildoleglo, line 24: If Gradle 7.0 removed the old syntax then this `if` should be for versions `>= 7.0`

kamildoleglo (changes requested):

kamildoleglo (approved):

Issue #1679, opened 2020-12-19
Throws section does not render properly in HTML


Egorand:
**Describe the bug**

If a function has a `@throws` tag in its docs, Dokka will generate a "Throws" section in the function's documentation, listing the exception type and the message side by side. If the exception type name is too long, or if the screen width is too small, the message and the type name will be displayed on top of each other.

**Expected behaviour**

The exception type name and the message should not be displayed on top of each other. If the screen width is too small they should perhaps be stacked vertically.

**Screenshots**

![image](https://user-images.githubusercontent.com/1186186/102695150-876f2a80-41f3-11eb-991b-62ea2be42dd1.png)

**To Reproduce**

1. Clone the following project: https://github.com/Egorand/dokka-throws-html-bug
2. Build the documentation: `./gradlew dokkaHtml`
3. Open the following file (path is relative to the project's directory): `build/dokka/html/dokka-throws-html-bug/com.squareup.egor.dokka/-test/test.html`
4. If the "Throws" section renders properly, try resizing the browser window to a smaller width.

**Dokka configuration**

No custom configuration, the plugin is applied as follows:

```kotlin
plugins {
    kotlin("jvm") version "1.4.21"
    id("org.jetbrains.dokka") version "1.4.20"
}
```

**Installation**
- Operating system: macOS Big Sur 11.1
- Build tool: Gradle v6.7
- Dokka version: 1.4.20

RegisMelgaco:
I think this solves the problem > https://github.com/Kotlin/dokka/pull/1912

PR #1912, opened 2021-05-11
Fix throws section does not render properly in HTML, solving issue 1679.

RegisMelgaco:
Fixes [issue 1679](https://github.com/Kotlin/dokka/issues/1679)

### DIFF
--- a/plugins/base/src/main/resources/dokka/styles/style.css
+++ b/plugins/base/src/main/resources/dokka/styles/style.css
@@ -796,6 +796,9 @@ td.content {
 .main-subrow > div > span {
     display: flex;
     position: relative;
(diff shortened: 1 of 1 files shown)


MarcinAman:
Hi, your solution works but i think this can be done in a better way since it is not breaking the text semantically but rather "as is". Imho you could use something like [this](https://github.com/Kotlin/dokka/blob/76c26f3325ebe0c06779fca09bbef0eefac5aa5a/plugins/base/src/main/kotlin/renderers/html/htmlFormatingUtils.kt#L41)

RegisMelgaco:
You are right, it is a much pleasant solution breaking semantically. I will give it a shot!

PR #1946, opened 2021-05-29
GFM renderer: cleanup paragraphs and tables

### LINKED ISSUE #1853: Trim sections in GFM
image

msink:
ready for review and merge

Fixes #1853

### DIFF
--- a/plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
+++ b/plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
@@ -36,6 +36,11 @@ open class CommonmarkRenderer(
                 childrenCallback()
                 buildParagraph()
             }
+            node.dci.kind == ContentKind.BriefComment -> {
+                buildParagraph()
+                childrenCallback()
+                buildParagraph()
+            }
             else -> childrenCallback()
         }
     }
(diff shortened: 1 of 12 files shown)


MarcinAman:
Hi, could you update the tests?

msink:
> update the tests

Tests has been updated from the beginning, did I miss some?

msink:
fixed

MarcinAman:
Hi, is the PR ready for review?

msink:
I hope yes, this time all tests passed and two extra tests added

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
```diff
@@ -35,6 +36,11 @@ open class CommonmarkRenderer(
                 childrenCallback()
                 buildParagraph()
             }
+            node.dci.kind == ContentKind.BriefComment -> {
```
kamildoleglo, line 39: We shouldn't switch by the kind here. I know it's done this way in several places, but it's not a good solution. If the brief comment requires this sequence, it should have a `TextStyle.Block` or some similar one added, possibly by a format-specific transformer

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
```diff
@@ -259,25 +262,26 @@ open class CommonmarkRenderer(
...
+            buildParagraph()
             buildSourceSetTags(sourceSets)
+            buildNewLine()
 
             instance.before?.let {
-                buildNewLine()
-                append("Brief description")
-                buildNewLine()
+                ///append("Brief description")
+                ///buildNewLine()
```
kamildoleglo, line 271: Please remove the comments here and below, if they are not needed

kamildoleglo (changes requested):

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
msink: Removed.

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
msink: Transformer - too complex for me, sorry :) 
Maximum - can add a comment with TODO.

And what is wrong with that, in HtmlRenderer it's wrapped in `div`, and `div` is block-level element anyway
```kotlin
node.dci.kind == ContentKind.BriefComment -> div("brief $additionalClasses") { childrenCallback() }
```

It's needed there because without it properties table 
> | Name | Summary |
> |---|---|
> | [padded](padded.md) | `var padded: Boolean`<br>If `true`, the container insert some space between children. |
> | [stretchy](stretchy.md) | `var stretchy: Boolean = false`<br>Next added child should expand to use all available size. |
```
| [padded](padded.md) | `var padded: Boolean`<br>If `true`, the container insert some space between children. |
```

renders without \<br> and even spaces:
> | Name | Summary |
> |---|---|
> | [padded](padded.md) | `var padded: Boolean`If `true`, the container insert some space between children. |
> | [stretchy](stretchy.md) | `var stretchy: Boolean = false`Next added child should expand to use all available size. |
```
| [padded](padded.md) | `var padded: Boolean`If `true`, the container insert some space between children. |
```

kamildoleglo: It's unmaintainable, as HtmlRenderer shows, and wasn't meant to be used there. Kinds should never be used at this point, they should only be used in transformers for identifying the actual element (eg. finding all signatures in the ContentNodes tree). That's my fault that they are accessible here though. 
I'll fix this part then, thanks!

msink:
So - if only the `ContentKind` part is pending - maybe merge it "as is" and do cleanup refactoring later?
In more global context than in this PR only.

MarcinAman (approved):

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
msink: Thanks @kamildoleglo,  9895b2e works as expected.
So this is resolved?

kamildoleglo: Yup, it is. Let me get some performance measurements and we'll probably get that merged

kamildoleglo (approved):
Sorry for the force-push, but had to rename the method

kamildoleglo:
Thanks!

PR #1956, opened 2021-06-08
Line wrap

### LINKED ISSUE #1932: Soft NewLine model
Currently NewLine modelling is very limited - only `ContentBreakLine` exist, meaning `HardLineBreak` -\
`<br>` in HTML and `\\\n` or `  \n` (two spaces here) in Markdown.
As a result - line separators in source KDoc are lost.

Adding something like `ContentSoftNewLine`, meaning  single `\n` in both HTML and in Markdown.
When parsing KDoc, insert it in appropriate places and call
```
   override fun T.buildSoftNewLine() = append("\n")
```
in renderer.

Maybe also add something like
```
   override fun T.buildParagraphEnd()
``` 
but in this part i'm unsure, is it really needed or not.

msink:
Attempt to fix #1932

RFC, not sure I'm not going in totally wrong direction.


MarcinAman (changes requested):
I am not a fan of this because this implies that formatting will be proxied to the output. In some cases (if the author of documentation wanted to do it) it is desirable, but there are so many cases where the line break was placed because someone run out of screen real estate or even it was opposed by a formatter like this: https://github.com/facebookincubator/ktfmt/blob/cd09a6c887391301c8f3a80f5ba1ae2ae615cbe7/core/src/test/java/com/facebook/ktfmt/FormatterKtTest.kt#L3978

I case someone wanted to have an enter he can just make another one and create a paragraph that will be rendered in separate line

core/src/main/kotlin/pages/ContentNodes.kt
```diff
@@ -25,12 +25,13 @@ interface ContentNode : WithExtraProperties<ContentNode>, WithChildren<ContentNo
...
 /** Simple text */
 data class ContentText(
-    val text: String,
+    val wholeText: String,
     override val dci: DCI,
     override val sourceSets: Set<DisplaySourceSet>,
     override val style: Set<Style> = emptySet(),
     override val extra: PropertyContainer<ContentNode> = PropertyContainer.empty()
 ) : ContentNode {
+    val text: String get() = wholeText.replace('\n', ' ')
```
msink, line 34: Tried it this way - but now one test failed:
```
LinkableContentTest > Samples multiplatform documentation() FAILED
    org.opentest4j.AssertionFailedError: expected: <import p2.JsClass
    fun main() {
       //sampleStart
       JsClass().printWithExclamation("Hi, Js")
       //sampleEnd
    }> but was: <import p2.JsClass fun main() {     //sampleStart     JsClass().printWithExclamation("Hi, Js")     //sampleEnd }>
```

msink:
> I am not a fan of this because this implies that formatting will be proxied to the output. In some cases (if the author of documentation wanted to do it) it is desirable, but there are so many cases where the line break was placed because someone run out of screen real estate or even it was opposed by a formatter like this: https://github.com/facebookincubator/ktfmt/blob/cd09a6c887391301c8f3a80f5ba1ae2ae615cbe7/core/src/test/java/com/facebook/ktfmt/FormatterKtTest.kt#L3978
> 
> I case someone wanted to have an enter he can just make another one and create a paragraph that will be rendered in separate line

Idea there was that nothing changes for anyone using already existing property "text", but if in some (possibly rare) cases original soft-line-ends are needed - they are preserved in new property "wholeText".
Maybe bad idea, but it's worth experimenting anyway.

PR #1973, opened 2021-06-21
Implement url to multiple sources

qurbonzoda:
* For signatures declared in a single location only `(source)` hyperlink is added.
* For builtins (common, jvm, js take from `kotlin/core/builtins` while k/n has separate implementation) `(Common source) (Native source)` hyperlinks are added.
* For multiplatform declarations with `expect/actual` `(Common source) (JVM source) (JS source) (Native source)` hyperlinks are added.
* For signatures declared in multiple platform, e.g. JVM and Native, `(JVM source) (Native source)` hyperlinks are added.

The hyperlinks follow the signature, thus often make the signature line very long. Maybe adding the hyperlinks in a separate line or even in a separate section might be better. Another approach is to add hyperlinks to signature platform tags.

### DIFF
--- a/core/src/main/kotlin/Formats/KotlinWebsiteHtmlFormatService.kt
+++ b/core/src/main/kotlin/Formats/KotlinWebsiteHtmlFormatService.kt
@@ -179,6 +179,33 @@ open class KotlinWebsiteHtmlOutputBuilder(
         }
     }
 
+    override fun appendSourceLinks(platforms: PlatformsData, appendSourceLink: DocumentationNode.(platform: String?) -> Unit) {
+        // For each platform find a source url.
+        // If several platforms use the same source url, use the source url only for the first platform.
+        // If only one platform has a source url, drop platform information.
+        val usedSourceUrls = mutableSetOf<String>()
+        val platformToSourceUrl = mutableMapOf<String, DocumentationNode>()
+
+        val nonJRE6 = platforms.filterNot { it.key.isJREVersion() && it.key.endsWith("6") }
+        nonJRE6.forEach { (platform, nodes) ->
+            nodes.firstOrNull { node ->
+                node.details(NodeKind.SourceUrl).firstOrNull()?.takeUnless { usedSourceUrls.contains(it.name) }?.also {
+                    usedSourceUrls.add(it.name)
+                } != null
+            }?.let { nodeWithUnusedSourceUrl ->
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/Formats/KotlinWebsiteHtmlFormatService.kt
```diff
@@ -179,6 +179,33 @@ open class KotlinWebsiteHtmlOutputBuilder(
...
+        // If only one platform has a source url, drop platform information.
+        val usedSourceUrls = mutableSetOf<String>()
+        val platformToSourceUrl = mutableMapOf<String, DocumentationNode>()
+
+        val nonJRE6 = platforms.filterNot { it.key.isJREVersion() && it.key.endsWith("6") }
+        nonJRE6.forEach { (platform, nodes) ->
+            nodes.firstOrNull { node ->
+                node.details(NodeKind.SourceUrl).firstOrNull()?.takeUnless { usedSourceUrls.contains(it.name) }?.also {
+                    usedSourceUrls.add(it.name)
+                } != null
```
ilya-g, line 194: Minor: perhaps can be simplified to
```suggestion
                node.details(NodeKind.SourceUrl).firstOrNull()?.takeIf { usedSourceUrls.add(it.name) } != null
```

ilya-g (approved):
In general I think any sensible implementation of source link merging will do. Let's see how it works out for stdlib in this branch  (stdlib-link-fix-1.4) and if it fits us, we could request implementing the similar logic in the new Dokka.
 

> For builtins .... (Common source) (Native source) hyperlinks are added.

JS-IR also has its own builtins sources. After it stabilizes it would make sense to add links to them too.

> The hyperlinks follow the signature, thus often make the signature line very long. Maybe adding the hyperlinks in a separate line or even in a separate section might be better. Another approach is to add hyperlinks to signature platform tags.

This is not formatted, but indeed it can look confusing:

image

Let's try placing it on a separate line.

core/src/main/kotlin/Formats/KotlinWebsiteHtmlFormatService.kt
qurbonzoda: done

qurbonzoda:
The source urls are now on the next line (`<br/>`) after the signature code but inside the signature `<div>`.
How can I see formatted documentation pages?

PR #1978, opened 2021-06-22
Remove limitation on minimal matched length

MarcinAman:


MarcinAman:
@kamildoleglo could you check it out? This introduces more search results that have matched phrases distributed all over the place

PR #2007, opened 2021-07-07
Add missing text styles in html

MarcinAman:


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -703,18 +703,37 @@ open class HtmlRenderer(
...
-            textNode.hasStyle(TextStyle.Cover) -> buildBreakableText(textNode.text)
             else -> text(textNode.text)
         }
+    }
+
+    private inline fun FlowContent.applyStyle(styleToApply: Style, crossinline body: FlowContent.() -> Unit){
+        when(styleToApply){
+            TextStyle.Bold -> b { body() }
+            TextStyle.Italic -> i { body() }
+            TextStyle.Strikethrough -> samp { body() }
```
kamildoleglo, line 732: Strikethrough is `<strike>`

kamildoleglo (changes requested):

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
MarcinAman: Fixed:
image

I had to make my own tag, since i am not gonna wait for kotlinx.html's release cycle: https://github.com/Kotlin/kotlinx.html/pull/184

PR #2026, opened 2021-07-15
Missing receiver docs

MarcinAman:
image

In fact they were not "missing" but just were displayed with parameters. Caveat was that when function had no parameters then the receiver also was not displayed. I've changed it to be present as normal tag since i don't think that presenting it as in the compiled code is the way to go


vmishenev (approved):
I agree doc for receiver should has not place into parameters table.

PR #2030, opened 2021-07-20
Migrate Dokka examples to Dokka repo (KT-47798)

vmishenev:
https://youtrack.jetbrains.com/issue/KT-47798


.github/workflows/gh-pages-examples.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        if: steps.filter.outputs.examples_changed == 'true'
+        with:
+          name: ${{ matrix.projects }}
+          path: examples/gradle/${{ matrix.projects }}/build/dokka
+
+
+  deploy-examples:
```
MarcinAman, line 47: Will this action run every time we push to master? Or only when `build-examples` job will detect changes?

plugins/all-modules-page/out/index.md
```diff
@@ -0,0 +1,10 @@
+/
```
MarcinAman, line 1: I think this is some sort of garbage. Please revert it

MarcinAman (approved):

.github/workflows/gh-pages-examples.yml
vmishenev: The job runs every time we push to master. But if there are no changes the second step ```peaceiris/actions-gh-pages``` says ```nothing to push```.

Issue #1171, opened 2020-07-16
SameMethodNamePageMergerStrategy does not update parents' source sets 


kamildoleglo:
When squashing DivergentGroups from different source sets the parents' source sets should all be updated up to the root

PR #2038, opened 2021-07-26
Fix broken link in kotlinx.serialization

MarcinAman:
hacked out untill https://github.com/Kotlin/dokka/issues/1171

### DIFF
--- a/plugins/base/src/main/kotlin/resolvers/local/DokkaLocationProvider.kt
+++ b/plugins/base/src/main/kotlin/resolvers/local/DokkaLocationProvider.kt
@@ -35,15 +35,15 @@ open class DokkaLocationProvider(
         pageGraphRoot.children.forEach { registerPath(it, emptyList()) }
     }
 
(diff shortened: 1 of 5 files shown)


plugins/base/src/main/kotlin/resolvers/local/DokkaLocationProvider.kt
```diff
@@ -130,7 +139,7 @@ open class DokkaLocationProvider(
     private val PageNode.pathName: String
         get() = if (this is PackagePageNode) name else identifierToFilename(name)
 
-    protected data class DRIWithSourceSet(val dri: DRI, val sourceSet: DisplaySourceSet?)
+    protected data class DRIWithSourceSet(val dri: DRI, val sourceSet: Set<DisplaySourceSet>)
```
vmishenev, line 142: Do `DRIWithSourceSet` rename to `DRIWithSourceSet`**s** similarly `withSourceSets` function?

Issue #2015, opened 2021-07-08
Improved rendering on annotations


qwwdfsad:
Consider the following signature from `kotlinx.coroutines` with Dokka 1.5.0: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-android/kotlinx.coroutines.android/as-coroutine-dispatcher.html

The signature has two annotations:
1) `@JvmOverloads()`. For this one, it would be nice to omit parentheses -- they are not present in the original declaration in the source code and also add additional confusion because parentheses are rendered with a colour different from the annotation itself

2) `@JvmName(name = from)`.  For this one, it would be nice to properly distinguish string literals from the parameter names. E.g. the `@JvmName(name = "from")`.
Without the clear disambiguation, it can be interpreted differently -- as the reference to the previous parameter name or as the constant in code. Additionally, it would be nice to use the colour scheme consistent with the IDE where integers, nulls, and string constants have different colour

PR #2045, opened 2021-07-28
Improve annotations rendering

vmishenev:
According to  #2015:

1. omit parentheses if there are no parameters 
2. add quotes for string literals into annotations

### DIFF
--- a/plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
+++ b/plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
@@ -69,20 +69,19 @@ interface JvmSignatureUtils {
             is Never -> Unit
         }
         link(a.dri.classNames!!, a.dri)
-        text("(")
-        a.params.entries.forEachIndexed { i, it ->
-            group(styles = setOf(TextStyle.BreakableAfter)) {
-                text(it.key + " = ")
-                when (renderAtStrategy) {
-                    is All -> All
-                    is Never, is OnlyOnce -> Never
-                }.let { strategy ->
-                    valueToSignature(it.value, strategy, listBrackets, classExtension)
-                }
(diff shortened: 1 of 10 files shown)


plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
```diff
@@ -93,20 +92,29 @@ interface JvmSignatureUtils {
...
+    private fun<T> PageContentBuilder.DocumentableContentBuilder.listParams(
+        params: Collection<T>,
+        listBrackets: Pair<Char, Char>?,
+        outFn: PageContentBuilder.DocumentableContentBuilder.(T) -> Unit
+    ) {
+        listBrackets?.let{ text(it.first.toString()) }
+        params.forEachIndexed { i, it ->
+            group(styles = setOf(TextStyle.BreakableAfter)) {
+                this.outFn(it)
+                if (i != params.size - 1) text(", ")
```
MarcinAman, line 112: you can use `lastIndex`

MarcinAman (approved):
Could you write a test with `NullValue` being used?

plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
vmishenev: It is only for List or Array, not Collection.

vmishenev:
> Could you write a test with `NullValue` being used?

I already wrote one in  ContentForAnnotationsTest.kt ( `rich documented annotation` ). Need another test for `NullValue`?

MarcinAman:
> > Could you write a test with `NullValue` being used?
> 
> I already wrote one in ContentForAnnotationsTest.kt ( `rich documented annotation` ). Need another test for `NullValue`?

Nah, thats fine. I've missed it ☺️

PR #2048, opened 2021-07-29
Make briefs contain first paragraph as in spec, resolve issues with i.e.

MarcinAman:


vmishenev (changes requested):
I changed one of your tests a bit (added `<!-- -->`). It's failed now. Is the test correct?
```kotlin
    @Test
    fun `second brief for functions should work with html comment for Java`() {
        testInline(
            """
            |/src/main/java/test/Example.java
            |package test;
            |
            |public class Example {
            |   /**
            |    * This is a simulation of Prof.<!-- --> Knuth's MIX computer. This is definitely not a brief in java <!-- -->
            |    */
            |   public static String test() {
            |       return "TODO";
            |   }
            |}
        """.trimIndent(),
            testConfiguration
        ) {
           pagesTransformationStage = { module ->
                val functionBriefDocs = module.singleFunctionDescription("Example")

                assertEquals(
                    "This is a simulation of Prof.<!-- --> Knuth's MIX computer.",
                    functionBriefDocs.children.joinToString("") { (it as ContentText).text })
            }
        }
    }
```

PR #2049, opened 2021-07-30
Property annotation for getter and setter

vmishenev:


core/src/main/kotlin/model/additionalExtras.kt
```diff
@@ -61,7 +61,7 @@ data class Annotations(
         }.toMap()
 
     enum class AnnotationScope {
-        DIRECT, FILE
+        DIRECT, FILE, GETTER, SETTER
```
MarcinAman, line 64: why do you need those?

MarcinAman (changes requested):
This PR resolves half of the problem. User still doesn't know about annotations placed on getters/setters. For example:

```kotlin
@get:JvmName("xdDep")
@set:JvmName("asdDep")
var propertyDep: String
    @SinceKotlin("1.3")
    @Deprecated("Dont use")
    get() = ""
    @SinceKotlin("1.4")
    set(value) {}
```

Shows only annotations from `@get`, `@set`. They are placed above the property which is questionable but ok for now. We are missing annotations placed directly above getter/setter

vmishenev:
> This PR resolves half of the problem. User still doesn't know about annotations placed on getters/setters. For example:
> 
> ```kotlin
> @get:JvmName("xdDep")
> @set:JvmName("asdDep")
> var propertyDep: String
>     @SinceKotlin("1.3")
>     @Deprecated("Dont use")
>     get() = ""
>     @SinceKotlin("1.4")
>     set(value) {}
> ```
> 
> Shows only annotations from `@get`, `@set`. They are placed above the property which is questionable but ok for now. We are missing annotations placed directly above getter/setter

Also we are displaying annotations placed directly above getter/setter.

```kotlin
@MustBeDocumented()
@Target(AnnotationTarget.FUNCTION)
annotation class Marking
            
@get:JvmName("xdDep")
@set:JvmName("asdDep")
var propertyDep: String
  @Marking()
  get() = ""
  @Marking()
  set(value) {}
```

It renders to
image

In your case dokka ignores annotations  ```SinceKotlin``` and `Deprecated` because of https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/kotlin/signatures/KotlinSignatureUtils.kt#L16

core/src/main/kotlin/model/additionalExtras.kt
vmishenev: I use these scopes to render in https://github.com/Kotlin/dokka/pull/2049/files#diff-8cd255845347168a7599ddf3b534ad987e31b9d92fe4e16303279cd3b6172ca9R74
Or I can introduce an extra parameter of that function to define the annotation target?

MarcinAman (approved):

Issue #2041, opened 2021-07-26
DefaultPsiToDocumentableTranslator translates PsiArrays to unexpected type


fsladkey:
### `DefaultPsiToDocumentableTranslator` translates `PsiArrays` to unexpected type

I noticed that the `DefaultPsiToDocumentableTranslator` translates most types as I expect, but for `PsiArray`s, the resulting DRI will always be `kotlin.Array` regardless of the source language. For example, Java source `int[] numbers` will be assigned a type of `kotlin.Array`. I'm not 100% sure what type would be expected here, but it would be great to distinguish it from actual Kotlin arrays.

Here's a test case I wrote in `JavaTest.kt` which illustrates the unexpected type.

```kotlin
@Test
fun javaArrayType() {
    inlineModelTest(
        """
        |public class Foo {
        |  public void bar(int[] baz) {}
        |}
        """, configuration = configuration
    ) {
        with((this / "java" / "Foo").cast<DClass>()) {
            with((this / "bar").cast<DFunction>()) {
                val paramType = this.parameters.first().type as GenericTypeConstructor
                assertNotEquals("kotlin", paramType.dri.packageName)
            }
        }
    }
}
```

If it's useful, my particular use case is to identify arrays of Java primitives and convert them to Kotlin `IntArray`s (or equivalent for other types) when displaying Kotlin docs for Java sources. Let me know if I've misinterpreted anything or if you have any suggestions. Thanks!

[Relevant source](https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt#L463)

**Installation**
- Operating system: macOS/Linux
- Build tool: Gradle
- Dokka version: 1.5.0

MarcinAman:
I think this is partially correct because there are 2 cases present:
- type is `byte[]`, `char[]`, `short[]`, `int[]`, `long[]`, `boolean[]` they are converted in compiler into their equivalents from `Arrays.kt`
- other types are compiled into `Array<out T>`

So our approach is only partially correct. We should add those 'special' cases from point 1

PR #2051, opened 2021-08-04
Replace arrays of primitive types on std arrays

vmishenev:
According to https://github.com/Kotlin/dokka/issues/2041#issuecomment-887444062

### DIFF
--- a/plugins/base/src/main/kotlin/transformers/documentables/KotlinArrayDocumentableReplacerTransformer.kt
+++ b/plugins/base/src/main/kotlin/transformers/documentables/KotlinArrayDocumentableReplacerTransformer.kt
@@ -1,0 +1,63 @@
+package org.jetbrains.dokka.base.transformers.documentables
+
+import org.jetbrains.dokka.Platform
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.plugability.DokkaContext
+
+class KotlinArrayDocumentableReplacerTransformer(context: DokkaContext):
+    DocumentableReplacerTransformer(context) {
+
+    private fun Documentable.isJVM() =
+        sourceSets.any{ it.analysisPlatform == Platform.jvm }
+
+    override fun processGenericTypeConstructor(genericTypeConstructor: GenericTypeConstructor): AnyWithChanges<GenericTypeConstructor> =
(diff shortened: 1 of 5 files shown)


vmishenev:
I don't replace parameter types into `Callable` of `DRI`

plugins/base/src/main/kotlin/transformers/documentables/KotlinArrayDocumentableReplacerTransformer.kt
```diff
@@ -0,0 +1,63 @@
...
+                                    true)
+                            else -> null
+                        }
+                    }
+                }
+            }
+            ?: super.processGenericTypeConstructor(genericTypeConstructor)
+
+    override fun processModule(module: DModule): AnyWithChanges<DModule>  =
+        if(module.isJVM())
```
MarcinAman, line 60: I think that we can narrow down the scope of searching only to java classes, but watch for mixed packages (where java and kotlin classes are in 1 package).

The easiest way to check language would be to look at the source like i did here: https://github.com/Kotlin/dokka/pull/2048/files

plugins/base/src/test/kotlin/filter/KotlinArrayDocumentableReplacerTest.kt
```diff
@@ -0,0 +1,199 @@
...
+        """.trimMargin(),
+            configuration
+        ) {
+            preMergeDocumentablesTransformationStage = {
+                val params = it.firstOrNull()?.packages?.firstOrNull()?.functions?.firstOrNull()?.parameters
+
+                val typeArrayNames = listOf("IntArray", "BooleanArray", "FloatArray", "DoubleArray", "LongArray", "ShortArray",
+                    "CharArray", "ByteArray")
+
+                Assertions.assertEquals(typeArrayNames.size, params?.size)
```
MarcinAman, line 42: You don't need to prefix those with `Assertions`

MarcinAman (approved):
I think it looks ok, some minor comments. 

One thing that i am concerned about is what is the runtime penalty of such transformer. Could you check it out on some bigger repository like `ktor`?

vmishenev:
I ran dokka on `ktor` without- and with such transformer 4 times. I don't see a significant difference. But my  measurements are very scattered. 
image

PR #2058, opened 2021-08-07
Align border with logo button border

hfhbd:
Before:
Before
After:
After


MarcinAman:
This PR is made in quite an unfortunate time as we are currently redesigning the frontend to be more in line with that kotlin website and intellij docs will look like. In the newer version this issue wont be present

PR #2060, opened 2021-08-09
Keep formatting within code inline

vmishenev:


MarcinAman (approved):

MarcinAman:
Could you try it out on google's repo to see if their test also passes?

PR #2079, opened 2021-08-17
Manual highlighting webhelp

vmishenev:
compile-time highlighting  is used  for signatures.
 runtime one (with PrismJs) is used for code blocks.


plugins/base/src/main/kotlin/signatures/JvmSignatureUtils.kt
```diff
@@ -116,23 +117,23 @@ interface JvmSignatureUtils {
         }
         is EnumValue -> link(a.enumName, a.enumDri)
         is ClassValue -> link(a.className + classExtension, a.classDRI)
-        is StringValue -> group(styles = setOf(TextStyle.Breakable)) { text( "\"${a.text()}\"") }
-        is LiteralValue -> group(styles = setOf(TextStyle.Breakable)) { text(a.text()) }
+        is StringValue -> group(styles = setOf(TextStyle.Breakable)) { text( "\"${a.text()}\"", styles = mainStyles + TokenStyle.String) }
+        is LiteralValue -> group(styles = setOf(TextStyle.Breakable)) { text(a.text(), styles = mainStyles + TokenStyle.Constant) }
```
MarcinAman, line 121: pls make helper functions for that like `constant(...)` or any other notation

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -128,34 +129,36 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
...
-                    text("annotation class ")
+                    text("annotation class ", styles = mainStyles + TokenStyle.Keyword)
                 }
             }
             link(c.name!!, c.dri)
             if (c is WithGenerics) {
-                list(c.generics, prefix = "<", suffix = ">") {
+                list(c.generics, prefix = "<", suffix = ">",
+                    separatorStyles = mainStyles + TokenStyle.Punctuation,
+                    borderStyles = mainStyles + TokenStyle.Operator) {
```
MarcinAman, line 161: i'd name it `surroundingCharactersStyle` or any other name related to `surrounding`

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -220,14 +229,23 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
...
                 p.extra[DefaultValue]?.run {
-                    text(" = $value")
+                    text(" = ", styles = mainStyles + TokenStyle.Operator)
+                    highlightValue(value)
                 }
             }
         }
 
+    private fun PageContentBuilder.DocumentableContentBuilder.highlightValue(v: String) {
+        if(v.matches(HighlightingRegex.isNumber)) text(v, styles = mainStyles + TokenStyle.Number)
```
MarcinAman, line 242: I think this information probably can be obtained from analysis, for me current regex-based solution is unacceptable.

MarcinAman (changes requested):
It would be nice if it could work with enums constructor values like:
image

Code for it is:
```kotlin
enum class WithDefaultValues(val inner: String) {
    FIRST("ASD"), SECOND("asd2")
}
```

MarcinAman:
Also: please write a test or two for some simple signatures

plugins/base/base-test-utils/src/main/kotlin/renderers/JsoupUtils.kt
```diff
@@ -4,12 +4,22 @@ import org.jsoup.nodes.Element
 import org.jsoup.nodes.Node
 import org.jsoup.nodes.TextNode
 
-fun Element.match(vararg matchers: Any): Unit =
+fun Element.match(vararg matchers: Any, ignoreSpan:Boolean = false): Unit =
```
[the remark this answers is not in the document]
vmishenev: It simplifies the testing of signatures.
_I've renamed `ignoreSpan`  to fit better_

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
vmishenev: This information is absent in Documentable Model for default values because they are stored in string format.

MarcinAman: Yes, but nothing prevents you from adding types to it. For example a default value for:
```kotlin
fun simpleFun(int: Int = 1, string: String = "string"): String = ""
```
is stored as `KtConstantExpression`. I am not sure how to properly get the type from that as the only place i see is a debug name but i'll try to find it out tomorrow / on monday

vmishenev: Yes, of course, I was afraid to change the model that the plugin developers use.
As it's stored as `KtConstantExpression` (I mean an expression), the highlighting becomes harder. 
```kotlin
 fun simpleFun(string: String = "string" + "string2" )
```
It needs not only the type.
Or we can highlight the default values as `token constant`.

MarcinAman: Okey, i've debugged a couple of cases and i think that for now the best solution would be to recognise only constants (like in my first snippet). If default value has an expression then we will render it as text and not parse its arguments. 

This should cover most of the cases and we can add a proper parser later as this is not that trivial.

vmishenev:
Possible issue: we don't keep format of long literals, e.g. `0xFF` or '0b01001' (and skipping underscore). The same is for annotation parameter value.

PR #2083, opened 2021-08-19
Handle `@link` tag

vmishenev:


plugins/base/src/main/kotlin/translators/psi/parsers/JavadocParser.kt
```diff
@@ -334,7 +333,8 @@ class JavadocParser(
         ) =
             when (tag.name) {
                 "link", "linkplain" -> tag.referenceElement()
-                    ?.toDocumentationLinkString(tag.dataElements.filterIsInstance<PsiDocToken>())
+                    ?.toDocumentationLinkString(tag.dataElements.filterIsInstance<PsiDocToken>().joinToString("") { it.stringifyElementAsText(keepFormatting = true)
+                        .toString() })
```
MarcinAman, line 337: use `.orEmpty` or something like this since if you use `toString` on `null` it will be displayed as null

plugins/base/src/test/kotlin/content/params/ContentForParamsTest.kt
```diff
@@ -256,14 +256,15 @@ class ContentForParamsTest : BaseAbstractTest() {
...
+            | * Return the target fragment set by {@link #setTargetFragment} or {@link
+            | * #setTargetFragment}.
             | *
             | * @deprecated Instead of using a target fragment to pass results, the fragment requesting a
             | *              result should use
             | * {@link java.util.HashMap#containsKey(java.lang.Object) FragmentManager#setFragmentResult(String, Bundle)} to deliver results to
             | * {@link java.util.HashMap#containsKey(java.lang.Object) FragmentResultListener} instances registered by other fragments via
             | * {@link java.util.HashMap#containsKey(java.lang.Object) FragmentManager#setFragmentResultListener(String, LifecycleOwner,
-            | * FragmentResultListener)}.
+            | *  FragmentResultListener)}.
```
MarcinAman, line 267: please check google's plugin after those changes

MarcinAman (approved):

plugins/base/src/test/kotlin/content/params/ContentForParamsTest.kt
vmishenev: I've checked. I had failed tests, but all they are related with `code` tag.

PR #2089, opened 2021-08-23
Don't force @literal in javadoc to be surrounded by code tag

### LINKED ISSUE #2088: Javadoc @literal tag is always surrounded by `<code>` and unnecessary newlines.
### Javadoc @literal tag is always surrounded by `<code>` and unnecessary newlines.

Given an input javadoc such as

```
/**
* An example of using the literal tag
* <pre>
* {@literal @}Entity
* public class User {}
* </pre>
*/
```

I expect the output to look like

dokka-PATCH

However, the literal `@` is enclosed in a `code` tag and newlines, so it looks like this instead.

dokka-HEAD

**Installation**
- Operating system: macOS//Linux
- Dokka version: 1.5.21-dev-113

After some experimentation I may have found a fix for this, so I will open a PR. However, I wanted to have a bug open to track this in case my solution does not make sense.

fsladkey:
### Don't force @literal in javadoc to be surrounded by code tag

Fixes #2088

After experimenting while looking into #2088, I found what appears to be a solution (at least for my use case), so I figured I'd make a PR just in case this is a reasonable general solution as well. The change is small, we simple do not wrap the value of `{@literal @}` in a code tag, as I believe the semantics of it do not require that (it can itself be wrapped in `pre` or `code`) and among other things, it makes formatting examples of annotations in doc strings difficult.

I've run my simple example from the last test in the PR with a few different versions of Dokka for comparison.

#### Dokka 1.5.0

dokka-1 5 0

#### Dokka 1.5.21-dev-113

dokka-HEAD

#### Dokka patch (this PR)

dokka-PATCH

### DIFF
--- a/plugins/base/src/main/kotlin/translators/psi/parsers/JavadocParser.kt
+++ b/plugins/base/src/main/kotlin/translators/psi/parsers/JavadocParser.kt
@@ -333,8 +333,8 @@ class JavadocParser(
             when (tag.name) {
                 "link", "linkplain" -> tag.referenceElement()
                     ?.toDocumentationLinkString(tag.dataElements.filterIsInstance<PsiDocToken>())
-                "code", "literal" -> "<code data-inline>${tag.dataElements.joinToString("") { it.stringifyElementAsText(keepFormatting = true)
-                    .toString() }.htmlEscape()}</code>"
+                "code" -> "<code data-inline>${dataElementsAsText(tag)}</code>"
+                "literal" -> dataElementsAsText(tag)
                 "index" -> "<index>${tag.children.filterIsInstance<PsiDocTagValue>().joinToString { it.text }}</index>"
                 "inheritDoc" -> inheritDocResolver.resolveFromContext(context)
                     ?.fold(ParsingResult(javadocTag)) { result, e ->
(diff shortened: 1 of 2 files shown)


MarcinAman:
It works well for `@` 👏 

One small issue that actually @vmishenev found is that it doesn't handle `<` or `>` well. Lets take a look at the example from [oracle's website](https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/whatsnew-1.5.0.html):

```
{@literal a<B>c}
```

We this PR parses it as:
image

whereas before it it was:
image

Ofc, previous version was not ideal, but i think that it would be nice to have the literal grouped in 1 place rather than scattered across multiple nodes. 


I was experimenting with your PR and figured out that if we add a custom tag for literal in JavadocParser then we are quite close to a decent solution: `"literal" -> "<literal>${dataElementsAsText(tag)}</literal>"`. This may not be the only way to fix it but at least it is a starting point. Also: please add the example from javadoc to the test ☺️

fsladkey:
Thanks for the quick review and feedback @MarcinAman!

> I was experimenting with your PR and figured out that if we add a custom tag for literal in JavadocParser then we are quite close to a decent solution: "literal" -> "<literal>${dataElementsAsText(tag)}</literal>". This may not be the only way to fix it but at least it is a starting point.

That makes sense, and if it seems like a reasonable solution to you it's good enough for me! Cleans things up nicely by grouping. I also tested it against our sources it and as expected there too. I've just pushed a new commit with that change + the new test for `{@literal a<B>c}`. Let me know what you think!

MarcinAman (approved):

vmishenev (approved):

Issue #2057, opened 2021-08-06
Markdown syntax within list items is changed to literal text by GFM plugin


sgilson:
**Describe the bug**
The GFM plugin uses HTML tags to display lists. Since Markdown syntax is not allowed inside block-level tags, markdown inside list items is interpreted as literal text. This means that bold text, italic text, and links in Markdown syntax are not interpreted correctly.

Reference: https://www.markdownguide.org/basic-syntax/#html-best-practices 

**Expected behaviour**
The semantics of the original markdown should be maintained.

**Screenshots**
n/a

**To Reproduce**
Write the following KDoc:

```kotlin
/**
 * - an **unordered list**
 */
```

Run `mvn pre-site` and open the generated markdown file. The content will contain this:

```markdown
<ul><li>an **unordered list**</li></ul>
```

I've pasted that content below to demonstrate how this will be rendered:

<ul><li>an **unordered list**</li></ul>

In KDoc, we wrote that we should have an unordered list where the list item contains bold text. With the addition of the HTML tags, the text is not in bold as we would expect.

**Dokka configuration**
Configuration of dokka used to reproduce the bug

```xml
  <plugin>
    <groupId>org.jetbrains.dokka</groupId>
    <artifactId>dokka-maven-plugin</artifactId>
    <version>1.5.0</version>
    <executions>
      <execution>
        <phase>pre-site</phase>
        <goals>
          <goal>dokka</goal>
        </goals>
      </execution>
    </executions>
    <configuration>
      <dokkaPlugins>
        <plugin>
          <groupId>org.jetbrains.dokka</groupId>
          <artifactId>gfm-plugin</artifactId>
          <version>1.5.0</version>
        </plugin>
      </dokkaPlugins>
    </configuration>
  </plugin>
```

**Installation**
- Operating system: macOS
- Build tool: Maven 3.6.0
- Dokka version: 1.5.0

**Additional context**

**Are you willing to provide a PR?**
Yes. Will look through the source over the weekend to see if I can assist

PR #2009, opened 2021-07-08
Add sitemaps plugin

MarcinAman:

PR #2098, opened 2021-08-29
GFM: Use Markdown syntax to render lists

sgilson:
Addresses #2057

- Switch list representation in CommonmarkRenderer.kt to use Markdown syntax instead of HTML
- Add ListBuilder to PageContentBuilder.kt to make testing feasible through builder DSL
- Switch to non-deprecated JUnit Assert in SimpleElementsTest.kt


sgilson:
Can this be run again? The errors in the previous build seem related to a lack of memory, but I don't see how this could be related to the changes.

sgilson:
Updated API file to fix the failing check

sgilson:
I've tried reproducing the Java 11 and 16 tests locally on my windows machine with the same gradle arguments with no luck. I'm seeing a similar paging/connection reset issue on other PRs, but I'm not exactly sure what could be causing this:

- https://github.com/Kotlin/dokka/pull/2104/checks?check_run_id=3498303510
- https://github.com/Kotlin/dokka/pull/2009/checks?check_run_id=3161456723
- https://github.com/Kotlin/dokka/pull/2112/checks?check_run_id=3505373157

Any thoughts on how to debug this?

MarcinAman:
> Can this be run again? The errors in the previous build seem related to a lack of memory, but I don't see how this could be related to the changes.

JDK 16 on windows sometimes failes as i think they use a weaker machine for that, don't worry.

As for the api check: weird, but i see that it passes now so it is ok

PR #2104, opened 2021-08-31
Improve versioning plugin

vmishenev:
- support for single module projects
- version navigator is on all pages
- dropdown arrow for version navigator

**NOTE**
I removed `versioning plugin` dependency  from `multimodule plugin`. 

image


plugins/versioning/src/main/kotlin/versioning/VersionsNavigationCreator.kt
```diff
@@ -1,53 +1,86 @@
...
+import org.jetbrains.dokka.plugability.configuration
 import org.jetbrains.dokka.plugability.plugin
 import org.jetbrains.dokka.plugability.querySingle
 import java.io.File
-import java.nio.file.Files.isDirectory
-import java.nio.file.Path
 
 interface VersionsNavigationCreator {
-    operator fun invoke(): String
+    operator fun invoke(location: String): String
```
MarcinAman, line 15: i don't like this as it is too general for what you are doing therefore doesn't let the reader know that this is in fact a relative location from the current version. 
If you really want to pass the location always (either in string or file format) why don't we just stick to files?

plugins/versioning/src/main/kotlin/versioning/VersionsNavigationCreator.kt
```diff
@@ -1,53 +1,86 @@
...
+                    var relativePosition: String? = null
+                    var activeVersion = ""
+                    div(classes = "versions-dropdown-button") {
+                        versions.minByOrNull { (_, versionLocation) ->
+                            versionLocation?.let { position.toRelativeString(it).length } ?: 0xffffff
+                        }?.let { (version, versionRoot) ->
+                            relativePosition = versionRoot?.let { output.toRelativeString(it) }
+                            activeVersion = version
+                            text(version)
+                        }
```
MarcinAman, line 53: i have no idea what this piece of code does

MarcinAman (changes requested):

plugins/versioning/src/main/kotlin/versioning/PreviousDocumentationCopier.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.dokka.versioning.VersioningPlugin
+import java.io.File
+
+class DefaultPreviousDocumentationCopier(val context: DokkaContext) : PostAction {
+    private val versioningHandler by lazy { context.plugin<VersioningPlugin>().querySingle { versioningHandler } }
+    private val processingStrategies: List<TemplateProcessingStrategy> =
+        context.plugin<TemplatingPlugin>().query { templateProcessingStrategy }
+
+    override fun invoke() {
+        versioningHandler.previousVersions.onEach { (_, dirs) -> copyVersion(dirs.src, dirs.dst) }
```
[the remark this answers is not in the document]
vmishenev: I use the version source to resolve a link in `VersionsNavigationCreator `.  I check is there a needed file in the corresponding  version source folder to avoid generating the invalid links (In the navigator such version will be strikethrough). At the copying (also creating navigator) moment the destination folder can't be used since not all version folder are copied.

plugins/versioning/src/main/kotlin/versioning/VersioningConfiguration.kt
```diff
@@ -9,6 +9,7 @@ data class VersioningConfiguration(
     var olderVersions: List<File>? = defaultOlderVersions,
     var versionsOrdering: List<String>? = defaultVersionsOrdering,
     var version: String? = defaultVersion,
+    var isOnlyOnRootPage: Boolean? = defaultIsOnlyOnRootPage
```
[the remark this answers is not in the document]
vmishenev: By default it supports the old behavior The navigator was only on root page in the old plugin.

plugins/versioning/src/main/kotlin/versioning/VersionsNavigationCreator.kt
```diff
@@ -1,53 +1,84 @@
...
+                                            ?.exists() == true
+
+                                val absolutePath =
+                                    if(isExistsFile) path?.resolve(relativePosition) else path?.resolve("index.html")
+
+                                a(href = absolutePath?.toRelativeString(position)) {
+                                    if (isExistsFile)
+                                        text(version)
+                                    else
+                                        strike { text(version) }
```
MarcinAman, line 71: not sure if everyone will know why some versions are crossed out. Can we have some tooltip or any other form of informing documentation user that this is not present in certain version. Maybe Eugene / Mike will have a better idea for that

MarcinAman (approved):

vmishenev:
I've set `renderVersionsNavigationOnAllPages` to `true` by default.
But important -- the plugin should be enabled in all submodules to render the navigator on all pages since it adds needed stylesheets in submodules pages.

MarcinAman:
> I've set `renderVersionsNavigationOnAllPages` to `true` by default.
> But important -- the plugin should be enabled in all submodules to render the navigator on all pages since it adds needed stylesheet in submodules pages. It can require to reconfigure existed multimodule projects.

Please write it in the docs

plugins/versioning/src/main/resources/dokka/not-found-version.html
```diff
@@ -0,0 +1,190 @@
...
+                </svg>
+                <svg class="page-404__beam">
+                    <use xlink:href="#page-404-beam"></use>
+                </svg>
+            </div>
+        </div>
+        <div class="wt-col-5">
+            <h1 class="_big">uh-oh!</h1>
+            <div class="sub-title margin-top-vertical-unit-half">You are requesting a page that not
+                available in documentation <span id="version">1.0</span> version
```
MarcinAman, line 184: i'd say that `not available in version 1.0` is sufficient

MarcinAman (approved):
Minor comments, LGTM!

PR #2112, opened 2021-09-03
Fix links of inherited properties, do not generate pages for them

vmishenev:


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -51,7 +51,7 @@ open class DefaultPageCreator(
             e.nameAfterClash(), contentForEnumEntry(e), setOf(e.dri), e,
             e.classlikes.renameClashingDocumentable().map(::pageForClasslike) +
                     e.filteredFunctions.renameClashingDocumentable().map(::pageForFunction) +
-                    e.properties.renameClashingDocumentable().mapNotNull(::pageForProperty)
+                    e.filteredProperties.renameClashingDocumentable().mapNotNull(::pageForProperty)
```
MarcinAman, line 54: It think this filters out the entries even when we have `suppressObviousMembers = true` and / or `suppressInheritedMembers = true`, shouldn't this be filtered out in `InheritedEntriesDocumentableFilterTransformer`?

plugins/base/src/test/kotlin/model/PropertyTest.kt
```diff
@@ -138,6 +138,29 @@ class PropertyTest : AbstractModelTest("/src/main/kotlin/property/Test.kt", "pro
         }
     }
 
+    @Test
```
MarcinAman, line 141: Could you also update the previous test (`propertyOverride`) to include assertion about DRI?

MarcinAman (changes requested):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: We don't have `isObviousMembers` for properties. 
It should be filtered out in `InheritedEntriesDocumentableFilterTransformer` as well if  `suppressInheritedMembers` is set (by default it's `false`). 
This filter only allows to not generate the pages for inherited properties, but the properties  are kept.

MarcinAman: >We don't have isObviousMembers for properties.
This is not good overall but ok, for this PR

MarcinAman (approved):

PR #2125, opened 2021-09-07
Copy button theming

MarcinAman:


plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -387,7 +390,7 @@ code.paragraph {
     left: -15em;
 }
 
-.copy-popup-wrapper.active-popup {
+.symbol:hover .copy-popup-wrapper.active-popup {
```
MarcinAman, line 393: a little unrelated change but this makes the copy popup hidden when user "unhovers" the symbol box

vmishenev (approved):

PR #2130, opened 2021-09-09
Fix wrong path to root

vmishenev:


plugins/base/src/main/kotlin/resolvers/local/DokkaLocationProvider.kt
```diff
@@ -118,22 +118,25 @@ open class DokkaLocationProvider(
             "${page::class.simpleName}(${page.name}) does not belong to the current page graph so it is impossible to compute its path"
         )
 
-        val contextNode =
-            if (context !is ClasslikePageNode && context?.children?.isEmpty() == true && context.parent() != null) context.parent() else context
         val nodePath = pathFor(node)
-        val contextPath = contextNode?.let { pathFor(it) }.orEmpty()
+        val contextPath = context?.let { pathFor(it) }.orEmpty()
+        val endedContextPath = if (context?.isIndexPage() == false)
+            contextPath.toMutableList().also { it.removeLastOrNull() }
```
MarcinAman, line 124: what are we exactly removing here?

vmishenev: We remove a last component of path from a root to context. 
Before we got a path from the root to a parent of context (it was incorrect if the parent was module page, see line 29).

MarcinAman: Therefore the bug shouldn't be fixed here but in `pathsIndex`

vmishenev: The problem in the path of module page. 
`pathsIndex` from  a  module page should be empty, shouldn't it?
Otherwise I can move checking (whether a page is modulePage -- line 29) to here (line 131).

MarcinAman (approved):

PR #2131, opened 2021-09-10
Fix link to javadoc enum entity

vmishenev:


plugins/base/src/main/kotlin/resolvers/external/javadoc/JavadocExternalLocationProvider.kt
```diff
@@ -30,9 +30,13 @@ open class JavadocExternalLocationProvider(
         if (classNames == null) {
             return "$docWithModule$packageLink/package-summary$extension".htmlEscape()
         }
+        val classSplit = classNames?.split(".").orEmpty()
+        val enumEntityAnchor = classSplit.takeIf { classSplit.size > 1 }?.get(1)
```
MarcinAman, line 34: Not only enums can have dot in class names but also inner classes:
https://docs.oracle.com/javase/8/docs/api/java/rmi/activation/ActivationGroupDesc.CommandEnvironment.html

MarcinAman (changes requested):

plugins/base/src/main/kotlin/resolvers/external/javadoc/JavadocExternalLocationProvider.kt
vmishenev: Then I suggest adding a some `extra` in DRI for enum entry, e.g. "enum entry" (DRI extra is a text, it's weird).

MarcinAman: From where would you like to get this information?

vmishenev: in `DRIFactory.kt`

MarcinAman: i wasn't talking about a place in code where to do it, i was more concerned about the information that we have available for example in `@see` tag: `@see java.nio.file.StandardCopyOption.REPLACE_EXISTING`

vmishenev: If I got it correctly, a link to entities can be in `@see`, `@throws`, `@exception` and MD links. But the information we get from "kotlin analysis" descriptors.

MarcinAman: So i've checked out this branch:
- it is quite bad that we have different DRI-s for the same link depending on the language that we run over. For example in java the link syntax is: `@see java.nio.file.StandardCopyOption#REPLACE_EXISTING` whereas in kotlin `java.nio.file.StandardCopyOption.REPLACE_EXISTING`. This results in different dri-s being created (kotlin as a dot in classnames but java has REPLACE_EXISTING as callable). 

- I think we are forced to include something in the extras of DRI as i realy dont want to change it's structure for compatibility reasons. Imho kotlin the way it is represented in kotlin is ok. I've debugged there and we have a type of the entity in analysis so adding this extra should be easy. 

Format of this extra is up to you

vmishenev:
Also Dokka generated a wrong link `https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#CREATE--` instead of `https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#CREATE`

core/src/main/kotlin/links/DRI.kt
```diff
@@ -23,6 +23,10 @@ data class DRI(
     }
 }
 
+enum class DRIExtra(val value: String) {
```
MarcinAman, line 26: I am not a fan of this design. It has a fundamental flaw of not being extensible to plugin creators. Imho a good extra should be extensible, type safe and (possibly) easily serialisable

plugins/base/src/test/kotlin/locationProvider/JavadocExternalLocationProviderTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test
+    fun `link to enum entity of javadoc`() {
+        val locationProvider = getTestLocationProvider()
+        val ktDri = DRI(
+            "java.nio.file",
+            "StandardOpenOption.CREATE",
+            extra = DRIExtra.EnumEntry.value
+        )
+        val javaDri = DRI(
```
MarcinAman, line 49: I think that this deserves a github issue if you dont want to do it in this PR

MarcinAman (changes requested):

core/src/main/kotlin/links/DRI.kt
```diff
@@ -23,6 +26,26 @@ data class DRI(
     }
 }
 
+object EnumEntryDRIExtra: DRIExtraProperty<EnumEntryDRIExtra> {
+    override val key = "EnumEntry"
+}
+
+interface DRIExtraProperty<T> {
```
MarcinAman, line 33: i think it should also have: `@JsonTypeInfo(use = CLASS)` as this will avoid confusion between different entries with the same structure.

MarcinAman (approved):

PR #2156, opened 2021-09-20
Fix whitespaces inside `@link` tag

vmishenev:
JDK Javadoc replaces breaklines inside `@link` tag on whitespaces.

### DIFF
--- a/plugins/base/src/main/kotlin/translators/psi/parsers/JavadocParser.kt
+++ b/plugins/base/src/main/kotlin/translators/psi/parsers/JavadocParser.kt
@@ -332,8 +332,8 @@ class JavadocParser(
         ) =
             when (tag.name) {
                 "link", "linkplain" -> tag.referenceElement()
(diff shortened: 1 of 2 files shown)


MarcinAman (approved):

MarcinAman:
Please also check it out on google's repo

vmishenev:
There is a wrong test on their side. 
```
    -      <li>Database: This annotation marks a class as a database. It should be an abstract class that extends RoomDatabase. At runtime, you can acquire an instance of it via Room.databaseBuilder or  Room.inMemoryDatabaseBuilder. 
    +      <li>Database: This annotation marks a class as a database. It should be an abstract class that extends RoomDatabase. At runtime, you can acquire an instance of it via Room.databaseBuilder or Room.inMemoryDatabaseBuilder. 
```

MarcinAman:
Google's repo may contain broken tests as they sometimes adjust them to match our bugs

The output after your changes looks better so i think it is ok. You can merge if you are done with this PR

PR #2159, opened 2021-09-22
Support JS KLib

vmishenev:
The implementation is like Native KLib. 
I duplicated JS Resolver from [Intellij](https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/analysis/src/org/jetbrains/kotlin/caches/resolve/JsResolverForModuleFactory.kt) because of [KTIJ-17809.](https://youtrack.jetbrains.com/issue/KTIJ-17809)

Project reproducer: https://github.com/vmishenev/sample-js-project


MarcinAman (approved):
Could you add this project to integration tests and assert that no page contains ERROR CLASS. That would be super handy when migrating kotlins

MarcinAman (approved):

PR #2168, opened 2021-09-24
Fix adding new custom stylesheets in submodules

vmishenev:
Or this can be implemented through a new template command.  It might be easier to configure. What's better?
On the other hand, somebody can want to use different stylesheets in submodules

### DIFF
--- a/core/test-api/api/test-api.api
+++ b/core/test-api/api/test-api.api
@@ -152,6 +152,7 @@ public final class testApi/testRunner/TestDokkaConfigurationBuilder {
 	public fun <init> ()V
 	public final fun build ()Lorg/jetbrains/dokka/DokkaConfigurationImpl;
 	public final fun getCacheRoot ()Ljava/lang/String;
+	public final fun getDelayTemplateSubstitution ()Z
 	public final fun getFailOnWarning ()Z
 	public final fun getFormat ()Ljava/lang/String;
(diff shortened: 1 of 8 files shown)


MarcinAman (approved):
I don't think we should care about this 1 person that would like to have each module with different stylesheet. 

I think this way is ok. Imho it is worth adding an assertion for that in our multimodule tests. Please also check if those custom stylesheets only occur once in the output (so they are not present in module but only on the root level)

vmishenev:
I have a problem with `SearchbarDataInstaller`. It keeps `RootPageNode` . But other html-preprocessors change `RootPageNode `, e.g. an asset installer.
When I add a new installer (e.g. for version plugin) after `SearchbarDataInstaller`, it causes an exception in `LocationProvider` since  `RootPageNode`  in `SearchbarDataInstaller`  is old. 
Order of html-preprocessors  is non-deterministic.  It changes whenever I changes the preprocessors.
It was broken here. An integration test is failed.

MarcinAman:
This is an issue that we had for a while and up until now I was mostly changing it to use DRIResolvableWrite rather than page one. DRI will always be resolved given the same DRI and sourcesets. 

The root cause of this issue is that we have a map in Location provider therefore if the page is the same but some of it contents are different (for example has different custom resources) then we can't link to it. This is... maybe not terrible but something that could be fixed, don't think it has a lot of priority tho

PR #2181, opened 2021-10-08
Fix for incorrect rendering of deprecated code

### LINKED ISSUE #2177: Gfm adding consecutive ~~ for deprecated code
**Describe the bug**
When running Gfm against code marked with `@Deprecated`, it adds consecutive `~~` in certain cases.  Take this example:

```
data class Test(
    val var1: Int? = null,
    val var2: String? = null
) {
    /**
     *
     * @param var3 A parameter
     */
    @Deprecated("This is deprecated")
    fun test(var3: Int) {
    }
}
```

This produces a test.md file (among other md files) with the following contents:

```
# test

[androidJvm]\
~~fun~~ [~~test~~](test.md)~~(~~~~var3~~~~:~~ [Int](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-int/index.html)~~)~~

## Parameters

androidJvm

| | |
|---|---|
| var3 | A parameter |
```

**Expected behaviour**
I think ideally there would only be `~~` at the beginning and end of this entire line:  `fun test(var3: Int)`.  It looks like however it is being parsed, it is passing in each of these when being sent to buildText, so each is surrounded with ~~:
- fun
- test
- (
- var3
- :
- )

**Screenshots**
N/A

**To Reproduce**
You can use that data class example above and generate Gfm docs from it.

**Dokka configuration**

```Kotlin
        tasks.named("dokkaGfmPartial") {
            dokkaSourceSets {
                configureEach {
                    reportUndocumented.set(false)
                    skipEmptyPackages.set(false)
                    skipDeprecated.set(false)
                }
            }
        }
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v6.8.0
- Dokka version:  1.5.32

**Additional context**
N/A

I can try if someone points me in the right direction.  I've tried setting this up locally to debug against an Android project, but couldn't do anything other than get it to log.

stevenzeck:
Closes #2177.  This PR fixes an issue where consecutive `ContentText` nodes have strikethrough styling.  This isn't the best fix, but it's the cleanest as it only changes CommonmarkRenderer.  Instead of surrounding `ContentText` with `~~`, it surrounds it with `<s>` and `</s>`.  I think ideally the `ContentGroup` should be surrounding with strikethrough instead of each text, but that requires more changes.

### DIFF
--- a/plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
+++ b/plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
@@ -213,10 +213,11 @@ open class CommonmarkRenderer(
             append(textNode.text)
         } else if (textNode.text.isNotBlank()) {
             val decorators = decorators(textNode.style)
+            val containsStrikethrough = textNode.style.contains(TextStyle.Strikethrough)
             append(textNode.text.takeWhile { it == ' ' })
-            append(decorators)
+            if (containsStrikethrough) append("<s>") else append(decorators)
             append(textNode.text.trim().htmlEscape())
-            append(decorators.reversed())
+            if (containsStrikethrough) append("</s>") else append(decorators.reversed())
             append(textNode.text.takeLastWhile { it == ' ' })
         }
     }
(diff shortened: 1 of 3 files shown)


vmishenev:
https://github.com/Kotlin/dokka/pull/2622 Now we have strikethrough only for declaration names and not for the whole signature as before.

PR #2210, opened 2021-10-27
Make PrimitiveJavaObject WithExtraProperties

owengray-google:
Issue: 2207. This is to be able to represent @NonNull String[]


vmishenev:
1. Can you rebase the PR after my force pushing? I am sorry for that.
2. Would you like to make `UnresolvedBound`, `TypeAliased` with ExtraProperties as well?
3. Can you use the `extra` property [here](https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt#L473) and [here for `UnresolvedBound`](https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt#L467), [`TypeAliased`](https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt#L818) to annotate `UnresolvedBound`, `PrimitiveJavaObject`, `TypeAliased`? Adding `extra = annotations()` is enough.  
4. If you add some tests for annotations it will be great.

vmishenev (changes requested):
See above

owengray-google:
Obsoleted by https://github.com/Kotlin/dokka/pull/2414

PR #2220, opened 2021-11-08
Allow filtering of property setter

vmishenev:


plugins/base/src/test/kotlin/transformers/SuppressTagFilterTest.kt
```diff
@@ -88,6 +89,26 @@ class SuppressTagFilterTest : BaseAbstractTest() {
...
+            """
+            |/src/suppressed/Suppressed.kt
+            |var property: Int
+            |/** @suppress */
+            |private set
+            """.trimIndent(), configuration
+        ) {
+            preMergeDocumentablesTransformationStage = { modules ->
+                val prop = modules.flatMap { it.packages }.flatMap { it.properties }
+                    .firstOrNull { it.name == "property" } as? DProperty
```
IgnatBeresnev, line 104: Seems like you can get rid of `as?` if you use `find` (which uses the same `firstOrNull`)
```suggestion
                    .find { it.name == "property" }
```

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
```diff
@@ -142,9 +148,37 @@ class DocumentableVisibilityFilterTransformer(val context: DokkaContext) : PreMe
...
+                    Pair(modified,
+                        if (modified)
+                            with(original) {
+                                copy(
+                                    setter = setter?.second?.firstOrNull(),
+                                    getter = getter?.second?.firstOrNull()
+                                )
+                            }
+                        else original
+                    )
```
IgnatBeresnev, line 171: Seems like `with(original) {}` does not make any difference, `original.copy` looks just fine and imho requires less brain power to understand :)

And I'd extract the result of if-else into a local variable and add indentation like below. Otherwise it looks a bit difficult to understand what is used where and how
```suggestion
                    val property =
                        if (modified)
                            original.copy(
                                setter = setter?.second?.firstOrNull(),
                                getter = getter?.second?.firstOrNull()
                            )
                        else original

                    modified to property
```

IgnatBeresnev (approved):
lgtm, left minor style-related comments that are up to you

PR #2258, opened 2021-12-05
Fix setup js deps

vmishenev:


kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/AnalysisEnvironment.kt
```diff
@@ -289,23 +290,32 @@ class AnalysisEnvironment(val messageCollector: MessageCollector, val analysisPl
...
+                                    kotlinLibrary,
+                                    analyzerServices,
+                                    dependencyResolver
+                                )
+                                else DokkaJsKlibLibraryInfo(kotlinLibrary, analyzerServices, dependencyResolver)
+                            )
+                        }
+                    } catch (e: Throwable) {
+                        configuration.getNotNull(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY)
+                            .report(CompilerMessageSeverity.WARNING, "Can not resolve KLIB. " + e.message)
```
IgnatBeresnev, line 316: What type of exception are we expecting to catch here? Or is this try-catch just in case? 

Throwable is rather wide, we can end up catching all sorts of things, including `StackOverflowError`/`OutOfMemoryError`, which can hide/mask a problem down the road. As stated in `java.lang.Error`: 
```
 * An {@code Error} is a subclass of {@code Throwable}
 * that indicates serious problems that a reasonable application
 * should not try to catch. 
```

Is it possible to narrow it down to a particular exception or maybe a group? Or preferably write a validation for input to make sure we don't cause any exceptions ourselves and if something does happen, it's completely unrelated?

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/AnalysisEnvironment.kt
```diff
@@ -496,8 +507,9 @@ class AnalysisEnvironment(val messageCollector: MessageCollector, val analysisPl
     fun addClasspath(paths: List<File>) {
         if (analysisPlatform == Platform.js) {
             configuration.addAll(JSConfigurationKeys.LIBRARIES, paths.map { it.absolutePath })
+        } else {
+            configuration.addJvmClasspathRoots(paths)
```
IgnatBeresnev, line 511: Is it possible that a similar problem will arise if `paths` contains files from `Platform.native` or `Platform.common`?

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/AnalysisEnvironment.kt
vmishenev: To be honest we do not except a exception here, i would say this is just in case. I have already written a validation.

I will not want to terminate Dokka even if `Error` happens. Anyway a user will get warning and Dokka will be able to generate the documentation.
  
P.S. Also you can take a look at [the validation](https://github.com/JetBrains/kotlin/blob/3f22321643f5dcc8a23cf0f28875790c4588f1b4/compiler/util-klib/src/org/jetbrains/kotlin/library/impl/KotlinLibraryImpl.kt#L349) using in Kotlin Gradle.

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/AnalysisEnvironment.kt
vmishenev: Possible, but it will depend on runners. I do not expect the similar problem here.
This PR solves only issues which are related with JS deps.

Issue #2213, opened 2021-10-28
Description List not handled in docs


asfalcone:
**Describe the bug**
Description List (`<dl>`, `<dt>`, `<dd>`) isn't parsed properly.

In java, the description items are removed. In Kotlin they appear as body text instead of structured DocTag elements

**Expected behaviour**
Dl, Dt, DD appear as structured `DocTag` elements

**Screenshots**
From Kotlin source:
![image](https://user-images.githubusercontent.com/1848683/139303575-26554d11-149c-4198-9e59-a2c89e67d9f3.png)

From Java source:
![image](https://user-images.githubusercontent.com/1848683/139303746-8ea6b809-1b92-4d71-811b-43715abb7027.png)

**To Reproduce**
```
/** 
 * <dl>
 *     <dt>
 *         <code>name="<i>name</i>"</code>
 *     </dt>
 *     <dd>
 *         A URI path segment. To enforce security, this value hides the name of the subdirectory
 *         you're sharing. The subdirectory name for this value is contained in the
 *         <code>path</code> attribute.
 *     </dd>
 *     <dt>
 *         <code>path="<i>path</i>"</code>
 *     </dt>
 *     <dd>
 *         The subdirectory you're sharing. While the <code>name</code> attribute is a URI path
 *         segment, the <code>path</code> value is an actual subdirectory name. Notice that the
 *         value refers to a <b>subdirectory</b>, not an individual file or files. You can't
 *         share a single file by its file name, nor can you specify a subset of files using
 *         wildcards.
 *     </dd>
 * </dl>
 */
 * class Foo
 ```

**Installation**
- Dokka version: 1.6.0-dev-122

IgnatBeresnev:
@asfalcone Hi! I can't tell from `Foo` the class whether it's a kotlin or a java class :) What's the use case? Would you like to be able to make description lists in JavaDocs only or in KDocs as well?

The reason I'm asking is it's unclear to me at the moment whether we should support parsing html tags in KDoc at all.

IgnatBeresnev:
Will be included in the upcoming 1.6.10. The dev version is out.

PR #2259, opened 2021-12-06
Description list support for JavaDocs (#2213)

IgnatBeresnev:
#2213


core/src/main/kotlin/pages/ContentNodes.kt
```diff
@@ -350,6 +350,28 @@ enum class ContentStyle : Style {
...
+     * or definition list, usually followed by [DescriptionDetails] for one or more
+     * terms. Formatting example: see `<dt>` html tag
+     */
+    DescriptionTerm,
+
+    /**
+     * If used within [DescriptionList] context, provides the definition or other
+     * related text associated with [DescriptionTerm]. Formatting example: see `<dd>` html tag
+     */
+    DescriptionDetails
```
IgnatBeresnev, line 372: I thought about extracting it to a separate enum `ListContentStyle`, but it seems unlikely that we'll ever add anything else there, so probably not worth it?

plugins/base/src/main/kotlin/transformers/pages/comments/DocTagToContentConverter.kt
```diff
@@ -39,14 +39,14 @@ open class DocTagToContentConverter : CommentsToContentConverter {
...
-        fun buildList(ordered: Boolean, start: Int = 1) =
+        fun buildList(ordered: Boolean, newStyles: Set<Style> = emptySet(), start: Int = 1) =
             listOf(
                 ContentList(
                     buildChildren(docTag),
                     ordered,
                     dci,
                     sourceSets.toDisplaySourceSets(),
-                    styles,
+                    if (newStyles.isEmpty()) styles else styles + newStyles,
```
vmishenev, line 49: Why!? It is not essential optimization.

vmishenev (changes requested):
* This does not work with:
```
<dl>
    <div>
        <dt>Term1</dt>
        <dd>Desc2</dd>
    </div>
    <div>
        <dt>Term2</dt>
        <dd>Desc2</dd>
    </div>
</dl> 
```
* What is about another tags, e.g. `dfn`?

IgnatBeresnev:
> This does not work with:
> ```
> <dl>
>   <div>
>        <dt>Term1</dt>
>        <dd>Desc2</dd>
>    </div>
>    <div>
>        <dt>Term2</dt>
>        <dd>Desc2</dd>
>    </div>
> </dl> 
> ```

Unfortunately, it seems to be the limitation of the content model in general since there's no separate content type for list items, which are now handled as `ContentGroup`. For instance, for `ul/ol`, we consider all children to be `li`, even if they're not. In javadocs it results in this
```
    /**
     * <ul>
     *     <div>one</div>
     *     <div>two</div>
     * </ul>
     */
```
being rendered as
* one
* two

__________________

As for description lists, `dl` is `ContentList` and `dt/dd` are `ContentGroup` with a list style, and just like for `ol/ul`, I only look for list items among the top level children. If we want to be able to render nested `dd/dt` using the current content model, I see the following options:
* Map `dt/dd` into `ContentList` (but this won't work for `ol/ul` due to nested lists)
* Add an `if (node.hasStyle(DescriptionTag))` into `buildGroup`

And I don't like either solution :)

We could extend the content model and introduce `ContentListItem` and render it separately, but I would expect you not to like it ;)

All in all, I don't see an easy way out, let me know if you do. Hopefully your example is a rare case and the solution can be postponed

IgnatBeresnev:
> What is about another tags, e.g. dfn?

It seems like `dfn` can indeed be used within description lists, but it's not exclusive to that, so I think it falls more under "support more javadoc html tags", which I indeed didn't do. We can plan it as a separate issue/pr

plugins/base/src/main/kotlin/transformers/pages/comments/DocTagToContentConverter.kt
vmishenev: Anyway this  is not a good place for optimization. It is an issue of implementation of `plus`. 

This make code more complex to read and can consider as boilerplate.

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -295,8 +294,31 @@ open class HtmlRenderer(
...
+        dl { buildListItems(node.children, pageContext, sourceSetRestriction) }
+    } else {
+        ul { buildListItems(node.children, pageContext, sourceSetRestriction) }
+    }
+
+    open fun DL.buildListItems(
+        items: List<ContentNode>,
+        pageContext: ContentPage,
+        sourceSetRestriction: Set<DisplaySourceSet>? = null
+    ) {
```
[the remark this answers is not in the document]
vmishenev: Please make the all three. Unfortunately code is not ideal. Our goal is to make it better.

vmishenev:
> It seems like dfn can indeed be used within description lists, but it's not exclusive to that
It is a good case to think about the design of styles beforehand.

> We can plan it as a separate issue/pr
Can you open an issue for it?

vmishenev:
```
    /**
     * <ul>
     *     <div>one</div>
     *     <div>two</div>
     * </ul>
     */
```
It is impossible for HTML scheme.
See https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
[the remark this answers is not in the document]
IgnatBeresnev: > Unfortunately code is not ideal. Our goal is to make it better.

Making everything into a single expression function is not making the code better, it's subjective and is purely about projects style guide :)

I tried making the other two into single-expression, but the compiler's not able to infer types properly and you have you specify it explicitly. I'm sorry, but this looks worse:
```
    open fun DL.buildListItems(
        items: List<ContentNode>,
        pageContext: ContentPage,
        sourceSetRestriction: Set<DisplaySourceSet>? = null
    ): Unit = items.forEach {
``` 
Let's leave it as it is

vmishenev (approved):
Please add my case to tests

PR #2270, opened 2021-12-13
Introduce documentedVisibilities setting

IgnatBeresnev:


vmishenev (changes requested):
* Why public? In this case I expect I can make public entities undocumented. 
* Please add the option in the integration tests

IgnatBeresnev:
> Why public? In this case I expect I can make public entities undocumented.

I don't see any sensible and valid general use-cases where I as a user would want to document ONLY private/protected/internal code and exclude public (within the same module/package). I can see reasons why you need control over protected/internal/private, but I don't see any to exclude publics, maybe you can provide some examples.

I can, however, see the user forget to include public into
```
documentedVisibilities.set(setOf(Visibility.PROTECTED))
```
and then wonder why it doesn't generate documentation for public code (and maybe even not notice it at first). That's the primary reason I decided to make it the default

If you want to exclude a package/module altogether, I believe there's other mechanisms for that in Dokka and it should be used instead of excluding public visibility via `documentedVisibilities`.

> Please add the option in the integration tests

Can you please tell me which tests you have in mind? I think I've added it to all the places I could find that deal with something similar.

vmishenev:
I mean the existence of `Visibility.PUBLIC` is ambiguous. Then, on the other hand, `documentedVisibilities.set(setOf(Visibility.PROTECTED))` can  be considered as only for protected members.
In my opinion it would be more obvious if another name, e.g. `documentedExtraVisibilities`, was used.

>  Can you please tell me which tests you have in mind? 
For example
https://github.com/Kotlin/dokka/blob/master/integration-tests/gradle/projects/it-basic/build.gradle.kts

IgnatBeresnev:
> I mean the existence of Visibility.PUBLIC is ambiguous.

Yeah, I thought about that and I agree it raises some questions, but at the same time I think it would allow for a clearer configuration where the author needs to show/stress their intentions by using the default value, like in the example below
```kotlin
named("main") {
    documentedVisibilities.set(
        setOf(
            Visibility.PUBLIC,
            Visibility.PRIVATE,
            Visibility.PROTECTED,
            Visibility.INTERNAL
        )
    )
    perPackageOption {
        matchingRegex.set(".*api.*")
        documentedVisibilities.set(
            setOf(
                Visibility.PUBLIC
            )
        )
    }
}
```

I thought about it more and out of all the options I see (all of them have trade offs, including renaming), I least dislike the idea of removing hardcoded generation for publics and generating documentation for whatever is set in `documentedVisibilities`. The default will be `setOf(Visibility.PUBLIC)`. 

I guess the worst that can happen in this case is the user can forge to include public if they override defaults, but at least the API will be concise. 

What do you think about that?

vmishenev:
Compare two configurations:
```
 ...
documentedVisibilities.set(
        setOf(
            Visibility.PRIVATE,
        )
    ) 
```
```
 ...
documentedVisibilities.set(
        setOf(
            Visibility.PUBLIC,
            Visibility.PRIVATE
        )
    ) 
```
Is it easy to tell the difference between them without looking in our documentation?
The configuration must be self-documenting and obvious. It is difficult to guess the default inclusion `Visibility.PUBLIC` into an user-defined set by the only configuration.
 
Why do you dislike the idea of renaming to `documentedExtraVisibilities` (for example)?

IgnatBeresnev:
> It is difficult to guess the default inclusion Visibility.PUBLIC into an user-filled set by the only configuration.

As I said earlier, I agree and that's why I proposed to remove the hardcoded inclusion of `public` up above. 

> Why do you dislike the idea of renaming to documentedExtraVisibilities (for example)?

Renaming doesn't really solve it completely. It gives a hint something's off, but it raises the same question - what's the default? 
* If something's extra, then something is base. It's safe to assume `public` would in in the base documented visibilities, but what about `protected`? You'd have to look at enum values or read documentation for the definition of `extra` anyway
* `*extraVisibilities` to me sounds like there are more visibilities than the ones defined by stdlib (like `protected internal`). `extraDocumentedVisibilities` is better, but one might think that these visibilities would be documented more than others (like extra documentation). I initially had `includeVisibilities` as the name, but it still doesn't solve the issue that you brought up
* If we go by `extra` / `additional` then having `Visibility.PUBLIC` would make little sense and we'd have to remove it (since it wouldn't be 'additional', it'd be base), which limits the use of this enum (it won't be complete) if we or plugin creators need similar settings. We could rename the enum, of course, but `ExtraVisibilities.PRIVATE` looks strange

As I mentioned, I think the option with least questions would be the straightforward approach:

If you write this, you only get `private` documented.
```
documentedVisibilities.set(
        setOf(
            Visibility.PRIVATE,
        )
    ) 
```
If you write this, then it's `public` + `private` that are documented.
```
documentedVisibilities.set(
        setOf(
            Visibility.PUBLIC,
            Visibility.PRIVATE
        )
    ) 
```

If you don't set anything, default is `documentedVisibilities = setOf(Visibility.PUBLIC)`. Although I would consider adding `protected` to the default as well. It's part of public API and `javadoc` task generates documentation for protected members.

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
```diff
@@ -21,9 +21,25 @@ class DocumentableVisibilityFilterTransformer(val context: DokkaContext) : PreMe
         fun Visibility.isAllowedInPackage(packageName: String?) = when (this) {
             is JavaVisibility.Public,
             is KotlinVisibility.Public -> true
```
vmishenev, line 23: By your description you have a mistake here.

vmishenev (changes requested):

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
```diff
@@ -152,6 +155,12 @@ abstract class AbstractDokkaMojo(private val defaultDokkaPlugins: List<Dependenc
     @Parameter
     var includeNonPublic: Boolean = DokkaDefaults.includeNonPublic
 
+    @Parameter(property = "visibility")
+    var documentedVisibilities: Set<DokkaConfiguration.Visibility> = DokkaDefaults.documentedVisibilities
+        // hack to set the default value for lists, didn't find any other safe way
```
vmishenev, line 160: Why you do not use `defaultValue` in Parameter?

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
```diff
@@ -20,10 +21,31 @@ class DocumentableVisibilityFilterTransformer(val context: DokkaContext) : PreMe
...
+                packageOptions.firstOrNull { Regex(it.matchingRegex).matches(name) }
+            }
+
+            val (documentedVisibilities, includeNonPublic) = when {
+                packageOpts != null -> packageOpts.documentedVisibilities to packageOpts.includeNonPublic
+                else -> globalOptions.documentedVisibilities to globalOptions.includeNonPublic
+            }
+
+            val isDocumentedVisibilitiesOverridden = documentedVisibilities !== DokkaDefaults.documentedVisibilities
+            return if (isDocumentedVisibilitiesOverridden) {
```
vmishenev, line 44: Is `isDocumentedVisibilitiesOverridden` extra?

runners/maven-plugin/src/main/kotlin/DokkaMojo.kt
IgnatBeresnev: I want to take the default value from `DokkaDefaults.documentedVisibilities` and you can't use it in annotations since it's a list and can't be a `const`. `defaultValue` can do some expression evaluation, but I didn't find a way to reference a field. The only thing I could write in there is `PUBLIC`, which works, but if defaults change / refactoring occurs - we could have a problem here

vmishenev: In this case you need to manually set the default value in the POM.

vmishenev: although it can be inconsistent with `DokkaDefaults.documentedVisibilities`

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
IgnatBeresnev: What do you mean by extra? If you're proposing to inline it, then no, it's not extra - `isOverridden` in the name is supposed to explain the reason for reference equality check

vmishenev: For example `documentedVisibilities.ifEmpty{ DokkaDefaults.documentedVisibilities }.contains(visibility) || includeNonPublic`?

IgnatBeresnev: In this case we're imposing the default value on documented visibilities even if the user chose to override it with an `emptySet()`. The idea was that if the user has explicitly overridden it, we always prefer it and ignore deprecated `includeNonPublic`. If we try to mix them, it can lead to unexpected behaviour

I noticed an inconsistency with `visibility == PUBLIC` and pushed a fix, perhaps you'll like it more. Also left a comment with the explanation in case it wasn't obvious enough

vmishenev (approved):

PR #2273, opened 2021-12-14
Update kotlin and dokka versions to 1.6.10

IgnatBeresnev:


IgnatBeresnev:
Yeah, I understand now that updating dokka and updating documentation and examples should be done separately, otherwise examples can't be built :) Will create another PR

Issue #1446, opened 2020-09-08
Missing some constructors when using Kotlin-as-Java


BarkingBad:
Occurs for subtypes from other classes than default `Any`

IgnatBeresnev:
A comment in #2277 mentioned that the PR fixed this issue, so I assume we just forgot to close it.

Do let us know if you encounter any issues related to that though.

PR #2277, opened 2021-12-16
Add filtering stategies extension point and fix constructors in Kotlin-as-Java

BarkingBad:


BarkingBad:
Rebased (unfortunately, I lost access to push onto dokka branches, so I open a separate PR)

BarkingBad:
Fixes of course https://github.com/Kotlin/dokka/issues/1446

asfalcone:
@vmishenev @IgnatBeresnev Would you please review this PR that is necessary for development of our plugin? Thank you! 🙏

plugins/base/src/main/kotlin/translators/documentables/DefaultDocumentableFilteringStrategies.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.dokka.base.translators.documentables
+
+import org.jetbrains.dokka.model.DFunction
+import org.jetbrains.dokka.model.PrimaryConstructorExtra
+
+object DefaultDocumentableFilteringStrategies : DocumentableFilteringStrategies {
+    override fun shouldConstructorBeInPage(constructor: DFunction): Boolean =
+        constructor.extra[PrimaryConstructorExtra] == null || constructor.documentation.isNotEmpty()
```
IgnatBeresnev, line 8: I'd leave a comment explaining the check. Not too obvious for someone encountering this for the first time
```suggestion
        // primary kotlin constructors are rendered as part of the class signature
        // and are not included in any separate constructor blocks
        constructor.extra[PrimaryConstructorExtra] == null || constructor.documentation.isNotEmpty()
```

plugins/base/src/main/kotlin/translators/documentables/DocumentableFilteringStrategies.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.dokka.base.translators.documentables
+
+import org.jetbrains.dokka.model.DFunction
+
+interface DocumentableFilteringStrategies {
+    fun shouldConstructorBeInPage(constructor: DFunction): Boolean
```
IgnatBeresnev, line 6: Can you leave a KDoc that will allow plugin developers to understand what this extension method is for? Without context, it's not obvious which page this method is referring to.

Something along the lines of 

> Specifies whether [constructor] should appear in a separate constructors block/page. You may want to omit a constructor if it should not be visible to the user or is documented elsewhere (like Kotlin's primary constructor, which is rendered as part of the class signature)

Or, ideally, try to think of a better name, but I didn't manage to come up with anything short :(

IgnatBeresnev:
@BarkingBad we're having some internal discussions on why primary constructors were not included in Constructors block in the first place. Can you please hold off on fixing my review comments? We'll write up takeaways shortly

vmishenev:
We have discussed this issue. We propose removing of the filter `c.constructors.filter { it.extra[PrimaryConstructorExtra] == null || it.documentation.isNotEmpty() },`.  
There are no reasons to hide a non-documented primary constructor in the constructor block.
This filter `it.documentation.isNotEmpty()` does not work properly for a class having a KDoc comment. It is a bad filter that does not depend on a source set.
Also we carry on generating a html pages for all constructors.

What do you think about our proposal?
If you agree, you will be able just to remove the filter in this PR.

BarkingBad:
Hi, sorry for all inconsistency with the current dokka architecture, it is actually rebase of an old PR from September 2020 (over 14 months old, so I guess some things have changed over time). I will try to address all of your concerns so we can move forward with the issue :)

plugins/kotlin-as-java/src/test/kotlin/KotlinAsJavaPluginTest.kt
```diff
@@ -87,7 +88,10 @@ class KotlinAsJavaPluginTest : BaseAbstractTest() {
                     .map { it.content }
 
                 val children = contentList.flatMap { content ->
-                    content.mainContents.first().cast<ContentGroup>().children
+                    content.mainContents.single { it is ContentGroup }.children // TODO: Fix nesting of groups in PageCreator,
+                        // becuase now there is different level of nesting constructors and properties/functions
+                        // `contentForScope` is called on same level as constructors are
+                        // best solution would be possibility to merge these groups
```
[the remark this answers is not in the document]
BarkingBad: This TODO was written 14 months ago, I don't know if it is still valid

BarkingBad:
I've updated the PR, let me know what you think about that

vmishenev:
Thank you. Could you fix the tests?
Also it wolud be great to fix another [one](https://github.com/Kotlin/dokka/blob/master/plugins/base/src/test/kotlin/content/functions/ContentForBriefTest.kt#L107 )

IgnatBeresnev (approved):

vmishenev (approved):

PR #2287, opened 2021-12-22
Use common gradle+kotlin versions across gradle integration tests

IgnatBeresnev:


integration-tests/gradle/projects/it-js-ir-0/src/main/kotlin/RootPackageClass.kt
```diff
@@ -12,7 +11,8 @@ class RootPackageClass {
     val description = "I do live in the root package!"
 }
 
-fun RComponent<*, *>.params() = URLSearchParams()
+// sample method that uses classes from dom and react, should compile
+fun URLSearchParams.react(props: Props, state: State) {}
```
IgnatBeresnev, line 15: RComponent existed in 1.4, but seems to have been removed in 1.5 or 1.6. Moreover, `Props` and `State` were also moved from `kotlin-react-router-dom` to `kotlin-react` in 1.5. This makes it difficult to support testing of 1.4. 

Since 1.4 is already deprectaed, I decided to drop it and make sure it works for 1.5 and 1.6. Hopefully, we'll drop 1.4 generally when 1.7 comes out, so not a big deal in my opinion.

integration-tests/gradle/projects/it-android-0/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
```
Goooler, line 3:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.2-bin.zip
```

Bump all these `6.9` to `6.9.2`?

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/JsIRGradleIntegrationTest.kt
```diff
@@ -10,16 +10,23 @@ class JsIRGradleIntegrationTest(override val versions: BuildVersions) : Abstract
...
-        )
+        val versions = TestedVersions.BASE
     }
 
-    val reactVersionArg = mapOf(
-        "1.5.0" to "-Pdokka_it_react_kotlin_version=5.2.0-pre.202-kotlin-1.5.0",
-        "1.5.30" to "-Pdokka_it_react_kotlin_version=5.2.0-pre.246-kotlin-1.5.30",
-        "1.5.31" to "-Pdokka_it_react_kotlin_version=5.2.0-pre.251-kotlin-1.5.31"
+    // https://mvnrepository.com/artifact/org.jetbrains.kotlin-wrappers/kotlin-react
+    private val reactVersionArg = mapOf(
```
vmishenev, line 17: This can be in `TestedVersions`.

vmishenev (approved):

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/JsIRGradleIntegrationTest.kt
IgnatBeresnev: Indeed :+1:

Issue #2223, opened 2021-11-10
Add documentation for json CLI


owengray-google:
Right now, dokka's primary CLI documentation is [here](https://kotlin.github.io/dokka/1.6.0/user_guide/cli/usage/).

We have been forced to use the json option instead of the CLI as described on that page, because of 
```
Execution failed for task ':docs-tip-of-tree:dackkaDocs'.
> A failure occurred while executing androidx.build.dackka.DackkaWorkAction
   > Process 'command '$CHECKOUT/prebuilts/jdk/jdk11/linux-x86/bin/java'' could not be started because the command line exceed operating system limits.
```

Basically, the `-classpath` argument for a `sourceSet` is getting passed ~750 androidx and kotlin artifacts, classes.jar and dependencies, so that we can resolve references. With a ~180 character path each, this is getting near the 2^17 character limit (130k) that many terminals have for commands.

We are now using the json option, and it appears to be working fine, but it has some clear API differences from the normal CLI method, such as the `-src` CLI option being the equivalent of `sourceRoots`, or the `packageListUrl`s requiring `file://` be prepended if they are local.

It would be valuable to have documentation for the json option for the CLI.
(If there is any, I can't find it. I think it deserializes to a [DokkaConfiguration](https://github.com/Kotlin/dokka/blob/master/core/src/main/kotlin/configuration.kt) but that source file doesn't contain information about e.g. the `file://` difference)

vmishenev:
You are right. JSON configuration is deserialization to a [DokkaConfiguration](https://github.com/Kotlin/dokka/blob/master/core/src/main/kotlin/configuration.kt).
I have added the sample configuration. The values of particular parameters are the same as in the CLI options.
All `*Url` parameters must correspond to the URI scheme (`file://` for local), the same for CLI.

PR #2292, opened 2021-12-27
Add global settings to JSON dokka cli input

### LINKED ISSUE #2261: Inconsistency with CLI params and json/gradle
**Describe the bug**
Inconsistency with CLI params and json/gradle. 
`-globalLinks` for example exists as a command line argument but doesn't exist in Gradle or config.json 

Further, Gradle params link from: https://kotlin.github.io/dokka/1.6.0/user_guide/cli/usage/#example-using-json is broken

**Expected behaviour**
These are 1:1 consistent

BarkingBad:
Fixes #2261


runners/cli/src/main/kotlin/cli/main.kt
```diff
@@ -406,14 +404,35 @@ fun parseLinks(links: List<String>): List<ExternalDocumentationLink> {
...
+            }
+
+            sourceSets.forEach {
+                it.sourceLinks.cast<MutableSet<SourceLinkDefinitionImpl>>().addAll(globals.sourceLinks ?: emptyList())
+            }
+
+            sourceSets.forEach {
+                it.externalDocumentationLinks.cast<MutableSet<ExternalDocumentationLink>>().addAll(defaultLinks(it))
+            }
+        }
```
IgnatBeresnev, line 427: it seems that very similar logic is present in [GlobalArguments's init](https://github.com/Kotlin/dokka/blob/0ae0222561d86a44325e17c0296cfc87f93f380e/runners/cli/src/main/kotlin/cli/main.kt#L133), what do you think about moving the contents of this method there for consistency?

runners/cli/src/test/kotlin/cli/CliTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import junit.framework.Assert.assertTrue
+import org.junit.Test
+import java.lang.IllegalStateException
+import kotlin.test.assertEquals
+
+class CliIntegrationTest {
+
+    @Test
+    fun testGlobalArgs() {
```
IgnatBeresnev, line 11: Would be nice to correct the naming to represent expected behaviour, since new developers might not know what contracts `global` configs have. something like `should apply global settings to all source sets`

IgnatBeresnev (commented):
Can you please update usage docs? `docs/src/doc/docs/user_guide/cli/usage.md` and fix tests for windows? There seems to be a problem with paths
```
org.jetbrains.dokka.CliIntegrationTest > testGlobalArgs FAILED
    java.nio.file.InvalidPathException at CliTest.kt:15
```

Overall, lgtm, left some minor comments/questions

runners/cli/src/main/kotlin/cli/main.kt
```diff
@@ -406,14 +404,35 @@ fun parseLinks(links: List<String>): List<ExternalDocumentationLink> {
             }
 }
 
+fun initializeConfiguration(globalArguments: GlobalArguments): DokkaConfiguration = if (globalArguments.json != null) {
+        val jsonContent = Paths.get(checkNotNull(globalArguments.json)).toFile().readText()
+        val globals: GlobalDokkaConfiguration = parseJson(jsonContent)
+        val dokkaConfigurationImpl = DokkaConfigurationImpl(jsonContent)
+        dokkaConfigurationImpl.run {
```
vmishenev, line 411: 'apply' is better.

vmishenev (approved):
Ok.
In the future I would like to avoid double parsing of JSON file. For example, for Google [size of the file can exceed 130k]( https://github.com/Kotlin/dokka/issues/2223).

plugins/all-modules-page/out/index.md
```diff
@@ -0,0 +1,10 @@
+/
+
+# Sample project
+
+Sample documentation with [external link](https://www.google.pl)
+
+## All modules:
+
+| Name |
+|---|
```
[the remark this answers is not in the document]
BarkingBad: Removed

runners/cli/src/main/kotlin/cli/JsonMapperForCLI.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.SerializerProvider
+import com.fasterxml.jackson.databind.module.SimpleModule
+import com.fasterxml.jackson.databind.ser.std.StdScalarSerializer
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.jacksonTypeRef
+import java.io.File
+
+// THIS IS COPIED FROM BASE SINCE IT NEEDS TO BE INSTANTIATED ON THE SAME CLASS LOADER AS PLUGINS
+
```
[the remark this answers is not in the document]
BarkingBad: It is a copy left-over though I think I can do it another way. I will come back to this issue in a separate comment section

runners/cli/src/main/kotlin/cli/JsonMapperForCLI.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun serialize(value: File, g: JsonGenerator, provider: SerializerProvider) {
+        g.writeString(value.path)
+    }
+}
+
+data class GlobalDokkaConfiguration(
+    val perPackageOptions: List<PackageOptionsImpl>?,
+    val externalDocumentationLinks: List<ExternalDocumentationLinkImpl>?,
+    val sourceLinks: List<SourceLinkDefinitionImpl>?
+)
```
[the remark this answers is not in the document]
BarkingBad: Done

BarkingBad:
> In the future I would like to avoid double parsing of JSON file.

The problem is that we would have to copy the whole `DokkaConfigurationImpl` to be a new data class having these properties. Maybe there is an easier way to achieve that by some custom Jackson mapper.

runners/cli/src/main/kotlin/cli/main.kt
BarkingBad: Done

runners/cli/src/main/kotlin/cli/main.kt
BarkingBad: `init` method is used for `cli` parameters while these settings are obtained from json file, which is processed as a completely different path of execution (it always has been since this check: 
```
    val configuration = if (globalArguments.json != null)
        DokkaConfigurationImpl(
            Paths.get(checkNotNull(globalArguments.json)).toFile().readText()
        )
    else
        globalArguments
```
)
I assume you cannot both have some settings setup as `cli` params and some passed via JSON, so I think it's not worth it trying to unify these two completely paths of execution, but I am open to your opinions

core/src/main/kotlin/configuration.kt
```diff
@@ -86,6 +86,20 @@ data class DokkaSourceSetID(
...
+ * Otherwise, the user would be enforced to repeat all these options per each sourceset.
+ */
+data class GlobalDokkaConfiguration(
+    val perPackageOptions: List<PackageOptionsImpl>?,
+    val externalDocumentationLinks: List<ExternalDocumentationLinkImpl>?,
+    val sourceLinks: List<SourceLinkDefinitionImpl>?
+)
+
+fun GlobalDokkaConfiguration(json: String): GlobalDokkaConfiguration = parseJson(json)
+
```
BarkingBad, line 102: @IgnatBeresnev In order to not copy-paset whole json utilities, I had to move these to `core` since I cannot access `parseJson` from `core` being in `cli` module as it is `internal`. On the other hand, this `GlobalDokkaConfiguration` is not so `core` feature, rather `cli` specific, so it should not be there. Do you have any idea how we should deal with this problem? The third solution could be "opening" `parseJson` function in `core` module by removing `internal` keyword.

IgnatBeresnev: To me, `GlobalDokkaConfiguration` seems to be a `json` feature rather than a `cli` one (although the only place we use json configuration is indeed in the cli), and since we already provide the ability to create dokka configuration from json up above, conceptually I'm fine with having this here

The only downside is that I, as a plugin developer, am unable to apply `GlobalDokkaConfiguration` to `DokkaConfigurationImpl` since this logic resides in `initializeConfiguration` (which is in CLI module):
```kotlin
        val globals = GlobalDokkaConfiguration(jsonContent)
        ...
        dokkaConfigurationImpl.apply {
            sourceSets.forEach {
                ...
            }
            ...
        }
```
What do you think about moving this to an extension function in core, say `fun DokkaConfiguration.apply(globalConfiguration: GlobalDokkaConfiguration)`?

If I'm writing a plugin, I then could use json configs like
```
val configuration = DokkaConfigurationImpl(mainJson)
val globalConfiguration = GlobalDokkaConfiguration(globalJson)

configuration.apply(globalConfiguration)
```

BarkingBad: I don't know if I got you clear, but do you propose to have two separate jsons?

IgnatBeresnev: Not necessarily, could be one, could be separate, could be no json at all - that's up to the plugin writer.

My main proposed point was to add the ability to apply `GlobalDokkaConfiguration` to `DokkaConfiguration` (extract this code from `initializeConfiguration` in cli), then having this class in `core` is fine by me, since it can be reused by anyone

BarkingBad: Oh, ok, I got you wrong the first time I read your comment. Now it makes sense, will do fixes today.

IgnatBeresnev: Nice, thanks!

integration-tests/cli/build.gradle.kts
```diff
@@ -12,6 +12,8 @@ evaluationDependsOn(":plugins:base")
 dependencies {
     implementation(kotlin("stdlib"))
     implementation(kotlin("test-junit"))
+    val jackson_version: String by project
+    implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jackson_version")
```
IgnatBeresnev, line 16: after json parsing was moved back to core, is this needed anymore?

runners/cli/build.gradle.kts
```diff
@@ -10,6 +10,10 @@ dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-cli-jvm:0.3.3")
     implementation(project(":core"))
     implementation(kotlin("stdlib"))
+    val jackson_version: String by project
+    implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jackson_version")
```
IgnatBeresnev, line 14: Same here, is it still needed?

IgnatBeresnev (approved):
lgtm, thanks for addressing so many comments!

Let's fix `apiCheck` check (new method is missing) and I think we're ready to merge

integration-tests/cli/build.gradle.kts
BarkingBad: My fault, you're right

runners/cli/build.gradle.kts
BarkingBad: Same here

PR #2302, opened 2022-01-11
Fix wrong Groovy DSL example

chkpnt:
File#resolve(String) doen't exist in Groovy


IgnatBeresnev (approved):
Thanks! 

From my experience, groovy DSL is indeed poorly documented, we'll re-visit it before stable release, but if you decide to contribute more - I'll gladly review the PRs :)

PR #2307, opened 2022-01-13
Add external documentable provider

Kordyjan:
There are some cases when there is a need to process documentables of classlikes that were not defined in the project itself but are supertypes of classlikes defined in the documented project. This pr adds a simple service that can be accessed via
```kotlin
context.plugin<DokkaBase>().querySingle { externalDocumentablesProvider }
```
and then queried with DRI and source set to obtain a documentable.

Currently, it supports obtaining only DClasslikes, as I cannot see any use case for wanting to obtain documentables for external members or packages. Nevertheless, if the need arise one can add a new, more general interface without breaking backward compatibility.

### DIFF
--- a/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
@@ -2,6 +2,7 @@ package org.jetbrains.dokka.base.translators.descriptors
 
 import com.intellij.psi.PsiNamedElement
 import com.intellij.psi.util.PsiLiteralUtil.*
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.runBlocking
@@ -55,6 +56,7 @@ import org.jetbrains.kotlin.resolve.constants.KClassValue.Value.LocalClass
 import org.jetbrains.kotlin.resolve.constants.KClassValue.Value.NormalClass
 import org.jetbrains.kotlin.resolve.descriptorUtil.annotationClass
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
+import org.jetbrains.kotlin.resolve.descriptorUtil.parents
 import org.jetbrains.kotlin.resolve.scopes.MemberScope
 import org.jetbrains.kotlin.resolve.source.KotlinSourceElement
 import org.jetbrains.kotlin.resolve.source.PsiSourceElement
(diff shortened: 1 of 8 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -162,7 +173,7 @@ private class DokkaDescriptorVisitor(
         }
     }
 
-    private suspend fun visitClassDescriptor(descriptor: ClassDescriptor, parent: DRIWithPlatformInfo): DClasslike =
+    suspend fun visitClassDescriptor(descriptor: ClassDescriptor, parent: DRIWithPlatformInfo): DClasslike =
```
IgnatBeresnev, line 176: This doesn't seem to leak to public API now (I don't see it in any of the `*.api` files), but I'd prefer `internal` here if possible

plugins/base/src/main/kotlin/translators/descriptors/ExternalDocumentablesProvider.kt
```diff
@@ -0,0 +1,10 @@
+package org.jetbrains.dokka.base.translators.descriptors
+
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.DClasslike
+
+
+interface ExternalDocumentablesProvider {
+    fun findClasslike(dri: DRI, sourceSet: DokkaConfiguration.DokkaSourceSet): DClasslike?
+}
```
IgnatBeresnev, line 10: I really liked the explanation for this extension point in the PR description, can you please leave a similar KDoc here or somplace else?  Preferably with an example/usecase (about supertypes from external libraries or something)? 

So that new plugin developers/maintainers can understand when to use it and why it exists, thanks!

IgnatBeresnev (commented):
Don't know much about extension points or DRI side of things yet, but all in all LGTM

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -109,6 +111,15 @@ class DefaultDescriptorToDocumentableTranslator(
             )
         }
     }
+
+    override fun translateDescriptor(descriptor: ClassDescriptor, sourceSet: DokkaSourceSet): DClasslike {
```
vmishenev, line 115: `translateClassDescriptor`? Likewise there is the same naming convention  for `visit`-methods.

vmishenev (approved):
Thia PR allows to get an arbitrary `Classlike` from classpath. It is not necessary one is a supertype in the documented project.

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
Kordyjan: `DokkaDescriptorVisitor` is private in `DefaultDescriptorToDocumentableTranslator` so `visitClassDescriptor(...)` cannot be used from outside of enclosing class.

IgnatBeresnev: oh, I see, thanks

IgnatBeresnev (approved):
:+1:

PR #2309, opened 2022-01-13
KT-50292 - Implement vertical alignment of parameter

IgnatBeresnev:
Do tell me if you'd like to see screenshots of before/after

### DIFF
--- a/core/src/main/kotlin/pages/ContentNodes.kt
+++ b/core/src/main/kotlin/pages/ContentNodes.kt
@@ -318,7 +318,7 @@ interface Kind
 
 enum class ContentKind : Kind {
 
-    Comment, Constructors, Functions, Parameters, Properties, Classlikes, Packages, Symbol, Sample, Main, BriefComment,
+    Comment, Constructors, Functions, Parameters, Parameter, Properties, Classlikes, Packages, Symbol, Sample, Main, BriefComment,
     Empty, Source, TypeAliases, Cover, Inheritors, SourceSetDependentHint, Extensions, Annotations;
 
     companion object {
@@ -347,7 +347,8 @@ enum class TextStyle : Style {
 }
 
 enum class ContentStyle : Style {
-    RowTitle, TabbedContent, WithExtraAttributes, RunnableSample, InDocumentationAnchor, Caption
+    RowTitle, TabbedContent, WithExtraAttributes, RunnableSample, InDocumentationAnchor, Caption,
+    Wrapped, Indented
 }
 
 enum class ListStyle : Style {
--- a/core/src/main/kotlin/pages/ContentNodes.kt
+++ b/core/src/main/kotlin/pages/ContentNodes.kt
@@ -316,9 +316,25 @@ data class PlatformHintedContent(
 interface Style
 interface Kind
 
+/**
+ * [ContentKind] represents a grouping of content of one kind. This can be rendered
+ * as either a part of a composite page (one tab/block within a class's page, for instance)
+ * or as a separate page altogether.
+ */
 enum class ContentKind : Kind {
 
-    Comment, Constructors, Functions, Parameters, Parameter, Properties, Classlikes, Packages, Symbol, Sample, Main, BriefComment,
+    /**
+     * Marks all sorts of signatures. Can contain sub-kinds marked as [SymbolContentKind]
+     *
+     * Some examples:
+     * - primary constructor: `data class CoroutineName(name: String) : AbstractCoroutineContextElement`
+     * - constructor: `fun CoroutineName(name: String)`
+     * - function: `open override fun toString(): String`
+     * - property: `val name: String`
+     */
+    Symbol,
+
+    Comment, Constructors, Functions, Parameters, Properties, Classlikes, Packages, Sample, Main, BriefComment,
     Empty, Source, TypeAliases, Cover, Inheritors, SourceSetDependentHint, Extensions, Annotations;
 
     companion object {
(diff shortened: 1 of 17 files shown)


core/src/main/kotlin/pages/ContentNodes.kt
```diff
@@ -318,7 +318,7 @@ interface Kind
 
 enum class ContentKind : Kind {
 
-    Comment, Constructors, Functions, Parameters, Properties, Classlikes, Packages, Symbol, Sample, Main, BriefComment,
+    Comment, Constructors, Functions, Parameters, Parameter, Properties, Classlikes, Packages, Symbol, Sample, Main, BriefComment,
```
vmishenev, line 321: It is really confusing with `Parameters` that has another application.
You can create another `Kind` or style.

plugins/base/base-test-utils/src/main/kotlin/renderers/SignatureUtils.kt
```diff
@@ -2,11 +2,15 @@ package signatures
...
 fun TestOutputWriter.renderedContent(path: String = "root/example.html") =
     contents.getValue(path).let { Jsoup.parse(it) }.select("#content")
         .single()
 
 fun Element.signature() = select("div.symbol.monospace")
-fun Element.firstSignature() = signature().first()
\ No newline at end of file
+fun Element.firstSignature() = signature().first()
+
+class Parameters(vararg matchers: Any) : Tag("span", *matchers, expectedClasses = listOf("parameters"))
```
vmishenev, line 15: the same

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -89,6 +89,21 @@ open class HtmlRenderer(
...
             }
+            node.dci.kind == ContentKind.Parameters -> {
+                span("parameters $additionalClasses") {
+                    childrenCallback()
+                }
+            }
+            node.dci.kind == ContentKind.Parameter -> {
+                span("parameter $additionalClasses") {
+                    if (node.hasStyle(ContentStyle.Indented)) {
+                        // could've been done with CSS (padding-left), but the indent needs to
```
vmishenev, line 100: I don't like having extra spaces in copied text. 

Have you tried this!?
```css
.intented::before {
    content: '    '
}
```

vmishenev (changes requested):
Have you checked this for GFM format?

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
IgnatBeresnev: Me neither, but unfortunately both padding and content don't work with select or copy. I added `content: '    '`, here's the result (it's now 8 spaces, 4 physical and 4 from content)

indented-ff

IgnatBeresnev: Also, I'm not sure what you mean by "I don't like having extra spaces in copied text."

Would you like the constructor to be copied like this?

```
class PrimaryConstructorWithVals(
val valString: String = "val def", 
var varString: String = "var def", 
var valStringWithoutDefaultValue: String, 
paramString: String
)
```

If so, content will do just fine then, but I don't think it's right

IgnatBeresnev: Discussed in person. In chrome everything works as expected, whereas Firefox does copy text without line breaks, so it results in
```
class PrimaryConstructorWithVals(    val valString: String = "val def",     var varString: String = "var def",     var valStringWithoutDefaultValue: String,     paramString: String)
```

core/src/main/kotlin/pages/ContentNodes.kt
IgnatBeresnev: Discussed in person. `ContentKind` seems to be about big blocks of content. For instance, `ContentKind.Parameter` is this:
parameterscontent

I thought `ContentKind` was more of a universal way to mark any content anywhere, so it is indeed strange to use it for function parameters. Will re-do to `SymbolStyle` and write KDocs

IgnatBeresnev:
GFM works as expected (without any changes). Here's an example:

_________________

//[Dokka Debug KTS](../../../index.md)/[me.beresnev.dokka.debug](../index.md)/[PrimaryConstructorWithVals](index.md)

# PrimaryConstructorWithVals

[jvm]\
class [PrimaryConstructorWithVals](index.md)&lt;[T](index.md)&gt;(valString: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html), varString: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html), valStringWithoutDefaultValue: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html), paramString: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html))

## Constructors

| | |
|---|---|
| [PrimaryConstructorWithVals](-primary-constructor-with-vals.md) | [jvm]<br>fun [PrimaryConstructorWithVals](-primary-constructor-with-vals.md)(valString: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html) = "val def", varString: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html) = "var def", valStringWithoutDefaultValue: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html), paramString: [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html)) |

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
IgnatBeresnev: It seems to be a firefox-specific bug. Created an issue for this: https://github.com/Kotlin/dokka/issues/2314

plugins/base/base-test-utils/src/main/kotlin/renderers/SignatureUtils.kt
IgnatBeresnev: I think it's ok since it's within `SignatureUtils` class

core/src/main/kotlin/pages/ContentNodes.kt
IgnatBeresnev: Introduced `SymbolContentKind`, added KDocs to avoid such misunderstandings in the future

PR #2312, opened 2022-01-17
Close input stream in `FileWriter`

vmishenev:


plugins/base/src/main/kotlin/renderers/FileWriter.kt
```diff
@@ -79,7 +79,8 @@ class FileWriter(val context: DokkaContext): OutputWriter {
                 val filePath = file.toAbsolutePath().toString()
                 withContext(Dispatchers.IO) {
                     Paths.get(root.path, rebase(filePath)).toFile().writeBytes(
-                        this@FileWriter.javaClass.getResourceAsStream(filePath).readBytes()
+                        this@FileWriter.javaClass.getResourceAsStream(filePath).use { it?.readBytes() } ?:
+                        throw IllegalStateException("Can not get a resource from $filePath")
```
IgnatBeresnev, line 83: I usually write it like this, but I guess it's a matter of preference

```suggestion
this@FileWriter.javaClass.getResourceAsStream(filePath).use { it?.readBytes() } 
    ?: throw IllegalStateException("Can not get a resource from $filePath")
```

IgnatBeresnev (approved):

PR #2315, opened 2022-01-18
Reuse `ObjectMapper`

IgnatBeresnev:
Accidentally came across this. I don't think it's very performance critical, but it's recommended and usually is reused (common practice)

As per [documentation](https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#basics-things-you-should-do-anyway),

> There are some basic ground rules to follow to ensure that Jackson processes things efficiently (close to optimal level). These are things that you should "do anyway", even if you do not have actual performance problems ...
> 1. Reuse heavy-weight objects: ObjectMapper (data-binding) ...
> ...

### DIFF
--- a/core/src/main/kotlin/links/DRI.kt
+++ b/core/src/main/kotlin/links/DRI.kt
@@ -33,17 +33,20 @@ abstract class DRIExtraProperty<T> {
         ?: (this.javaClass.let { it.`package`.name + "." + it.simpleName.ifEmpty { "anonymous" } })
 }
 
-private val DRI_EXTRA_OBJECT_MAPPER = ObjectMapper()
 
 class DRIExtraContainer(val data: String? = null) {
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/links/DRI.kt
```diff
@@ -33,15 +33,17 @@ abstract class DRIExtraProperty<T> {
         ?: (this.javaClass.let { it.`package`.name + "." + it.simpleName.ifEmpty { "anonymous" } })
 }
 
+private val DRI_EXTRA_OBJECT_MAPPER = ObjectMapper()
```
vmishenev, line 36: What is about `DRIExtraContainer.companion`?

vmishenev (changes requested):

core/src/main/kotlin/links/DRI.kt
IgnatBeresnev: Yeah, for some reason I forgot that exists :sweat_smile: Fixed

vmishenev (approved):

Issue #1227, opened 2020-07-29
Create a way to display indirect inheritors


MarcinAman:
possible solutions involve a text-based structure like in javadoc or even a graph using some JS library. To be discussed

DanySK:
Hi, I'm chipping in just to say that it would be very useful for some of my projects.

PR #2326, opened 2022-01-25
Refactor Ancestry Graphs

BarkingBad:
Progresses https://github.com/Kotlin/dokka/issues/1227

Actually saves the output of the ancestry graphs via extra PropertyContainer. Can be handy for example for this issue:
https://issuetracker.google.com/issues/170124934

Note it is still in WIP, I need to add some tests covering the graph creation.


plugins/base/src/main/kotlin/translators/psi/parsers/PsiCommentsUtils.kt
```diff
@@ -98,7 +98,7 @@ internal fun findClosestDocComment(element: PsiNamedElement, logger: DokkaLogger
             return findClosestDocComment(superMethods.single(), logger)
         }
 
-        val superMethodDocumentation = superMethods.map { method -> findClosestDocComment(method, logger) }
+        val superMethodDocumentation = superMethods.map { method -> findClosestDocComment(method, logger) }.distinct()
```
BarkingBad, line 101: Without `distinct`, the below `if` condition was pointless since the arity of the list won't change while `mapping` and we are testing 3 lines earlier that list is not a single value.

IgnatBeresnev:
Hi! Overall, the changes seem fine, although it's not very clear when this will come in handy. Was this refactoring "just-in-case" or are you preparing ground for another PR that will come soon? :)

Could the google issue that you [linked](https://issuetracker.google.com/issues/170124934) be solved with #2307 ? It seems like the same use case.

BarkingBad:
Maybe you're right that we could reuse #2307 to get achieve the same thing. Nonetheless, I think these changes are improving the model since the previous implementation was keeping all ancestors in a plain list, so there were redundant interfaces in some nodes and you were not able to recreate the whole ancestry tree. The one thing we should consider is whether we can keep the ancestry tree as an extra (I'm talking about memory performance issues). If we cannot keep these as an extra, I will reuse #2307, but I feel like all the other changes are valid and this PR could be merged, don't you think?

One thing I probably should remove is `level` property, which is now pointless since we can calculate the level finally recursively because it is not longer list.

vmishenev (changes requested):
This might be handy for Dokka. But right now we do not plan to implement displaying indirect inheritors https://github.com/Kotlin/dokka/issues/1227 in our format.
I am afraid this PR can be memory consuming.
1. Can you estimate the memory consumption in large projects (Dokka with the PR vs Dokka without one) like [aws-sdk-kotlin](https://github.com/awslabs/aws-sdk-kotlin)? You can use https://github.com/gradle/gradle-profiler
It is not necessary to merge this. 
2. Can you introduce a special flag to save ancestry graph into annotations? By default it is disabled.
3. I agree `level` is redundant. It made sense when we had `Set<AncestryLevel>.`

plugins/base/src/test/kotlin/ancestry/AncestryTest.kt
```diff
@@ -0,0 +1,599 @@
...
...rclass=null, interfaces=[])])]), AncestryNode(level=4, typeConstructor=GenericTypeConstructor(dri=kotlin.collections/MutableCollection///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={})), superclass=null, interfaces=[AncestryNode(level=5, typeConstructor=GenericTypeConstructor(dri=kotlin.collections/Collection///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={})), superclass=null, interfaces=[AncestryNode(level=6, typeConstructor=GenericTypeConstructor(dri=kotlin.collections/Iterable///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={})), superclass=null, interfaces=[])]), AncestryNode(level=5, typeConstructor=GenericTypeConstructor(dri=kotlin.collections/MutableIterable///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={})), superclass=null, interfaces=[AncestryNode(level=6, typeConstructor=GenericTypeConstructor(dri=kotlin.collections/Iterable///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={})), superclass=null, interfaces=[])])])])]), interfaces=[]), interfaces=[])>
```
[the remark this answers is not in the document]
BarkingBad: Oh, my bad. I have been testing on that :P Will remove

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -344,16 +352,16 @@ class DefaultPsiToDocumentableTranslator(
...
                         )
                     )
                 }
             }
         }
 
-        private fun isExceptionExtra(ancestryTree: List<AncestryLevel>): ExceptionInSupertypes? =
-            ancestryTree.mapNotNull { it.superclass }.filter { it.dri.isDirectlyAnException() }
-                .takeIf { it.isNotEmpty() }?.let { ExceptionInSupertypes(it.toSourceSetDependent()) }
+        private fun AncestryNode.isExceptionExtra(): ExceptionInSupertypes? =
```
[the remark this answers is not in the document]
BarkingBad: I know, I was also confused, but I was already named like that, I just changed parameters signature, so I left it that way. I will change it.

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -1093,13 +1084,8 @@ private class DokkaDescriptorVisitor(
...
                 containingDeclaration.fqNameOrNull()?.asString()
                     ?.let { it == "kotlin.Any" || it == "kotlin.Enum" || it == "java.lang.Enum" || it == "java.lang.Object" } == true
-}
 
-private data class AncestryLevel(
-    val level: Int,
-    val superclass: TypeConstructor?,
-    val interfaces: List<TypeConstructor>
-)
+    private fun AncestryNode.isExceptionExtra(): ExceptionInSupertypes? =
```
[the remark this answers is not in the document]
BarkingBad: As above

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -206,11 +193,35 @@ class DefaultPsiToDocumentableTranslator(
...
+                            val javaClassKind = when {
+                                supertypePsiClass.isInterface -> JavaClassKindTypes.INTERFACE
+                                else -> JavaClassKindTypes.CLASS
+                            }
+                            supertypePsiClass to javaClassKind
+                        }
+                    }
+                }
+
+                fun parsePsiClass(psiClass: PsiClass, level: Int = 0): AncestryNode {
```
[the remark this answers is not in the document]
BarkingBad: I can come up with more self explanatory name.

BarkingBad:
> Can you introduce a special flag to save ancestry graph into an extra? By default it is disabled. Although If we do not need this extra, I will prefer to not keep this.

Can do that.

> Can you estimate the memory consumption in large projects (Dokka with the PR vs Dokka without one) like [aws-sdk-kotlin](https://github.com/awslabs/aws-sdk-kotlin)? You can use https://github.com/gradle/gradle-profiler
    It is not necessary to merge this.
    
Ok, I will provide the assessments.

BarkingBad:
I decided to drop extra property for now as It was consuming an additional 124% of memory (too much even for togglable utility). However, I keep model refactors since they are quite handy

vmishenev (approved):

core/src/main/kotlin/model/additionalExtras.kt
```diff
@@ -123,4 +123,17 @@ data class ConstructorValues(val values: SourceSetDependent<List<Expression>>) :
...
+    val superclass: AncestryNode?,
+    val interfaces: List<AncestryNode>,
+) {
+    fun allImplementedInterfaces(): List<TypeConstructor> {
+        fun traverseInterfaces(ancestry: AncestryNode): List<TypeConstructor> =
+            ancestry.interfaces.flatMap { listOf(it.typeConstructor) + traverseInterfaces(it) } +
+                (ancestry.superclass?.let(::traverseInterfaces) ?: emptyList())
+        return traverseInterfaces(this).distinct()
+    }
+}
```
IgnatBeresnev, line 139: Should this be in `additionalExtras.kt` if it's not an extra?

plugins/base/src/main/kotlin/translators/isException.kt
```diff
@@ -1,6 +1,15 @@
 package org.jetbrains.dokka.base.translators
 
 import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.AncestryNode
+import org.jetbrains.dokka.model.TypeConstructor
+
+internal fun AncestryNode.typeConstructorsBeingExceptionsOrNull(): List<TypeConstructor>? {
```
IgnatBeresnev, line 7: I think it's widely accepted that returning an empty list instead of a null is a good practice

Is there maybe a good reason to not do that here? If not, what do you think about returning an empty list instead?

BarkingBad: It is laterly chained in functions that return Extra or null.
```
        private fun AncestryNode.exceptionInSupertypesOrNull(): ExceptionInSupertypes? =
            typeConstructorsBeingExceptionsOrNull()?.let { ExceptionInSupertypes(it.toSourceSetDependent()) }
```
I can make it emptyList and test for that using `takeIf { it.isNotEmpty() }` if it is worth it,

```
        private fun AncestryNode.exceptionInSupertypesOrNull(): ExceptionInSupertypes? =
            typeConstructorsBeingExceptions().takeIf { it.isNotEmpty() }?.let { ExceptionInSupertypes(it.toSourceSetDependent()) }
```

core/src/main/kotlin/model/additionalExtras.kt
BarkingBad: Hard to tell, I can move it to another file, though I felt like the file is for containing addition model elements (I know that all are extra related). Should I move it to some new file like `AncestryNode.kt`?

plugins/base/src/main/kotlin/translators/isException.kt
BarkingBad: Ok, will change

IgnatBeresnev (approved):

PR #2340, opened 2022-01-28
Handle name changes for JvmName and JvmField annotations

BarkingBad:
Fixes https://issuetracker.google.com/issues/168340963

Most of the things were already covered in #1678  and #1675


BarkingBad:
Actually making @JvmField with @JvmName is impossible (JvmName has no target to the `field`, which I forgot) so it is impossible to rename it this way.

PR #2342, opened 2022-02-03
Fix MathJax rendering bug (#2175)

### LINKED ISSUE #2175: MathJax does not produce results
**Describe the bug**
I am attempting to use the MathJax plugin to add math support for my documentation. Whenever I add the `@usesMathJax` annotation to a KDoc, the documentation fails to render. 

**Expected behaviour**
The documentation to appear with correct math rendering.

**Screenshots**
Given the following class and method,
```kotlin
/**
 * An example class.
 */
class Example {
    /**
     * @usesMathjax
     * An example method.
     * \(a^2 + b^2 = c^2\)
     */
    fun foo() = println("Hello, world.")
}
```
this is what's rendered:
image

A similar render appears with classic Javadoc (nothing appears for the documentation of the method).

**To Reproduce**
* Create a piece of documentation and apply the `@usesMathJax` annotation
* Include valid MathJax in the documentation
* Run `dokkaHtml` (or `dokkaJavadoc`)
* Observe the output

[**Here's an example repo to test.**](https://github.com/wyskoj/TestDokkaMathJax)

**Dokka configuration**
```kotlin
plugins {
    id("org.jetbrains.kotlin.jvm") version "1.5.31"
    id("org.jetbrains.dokka") version "1.5.30"
    java
    idea
}

repositories {
    mavenCentral()
}

dependencies {
    implementation("org.jetbrains.kotlin:kotlin-stdlib")
    implementation("org.jetbrains.dokka:mathjax-plugin:1.5.31")
}
```

**Installation**
- Operating system: Same results on both Windows 11 and Ubuntu 20.04
- Build tool: Gradle v7.1
- Dokka version: 1.5.30

**Additional context**
I found little documentation on how to use this plugin, so it could be simply I'm using it all wrong. If that's the case, how can I appropriately use this plugin?

Sure.

wyskoj:
Fix a bug where using the `@usesMathJax` custom tag would cause the
documentation to fail to render. I fixed this by making all custom tags
render to the screen, which were being ignored.

Along with this, I updated the existing test case to check for the
presence of content in the output.

Fixes #2175

### DIFF
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -340,6 +340,18 @@ open class DefaultPageCreator(
                     }
                 }
             }
+            val customTags = d.customTags
+            if (customTags.isNotEmpty()) {
+                group(styles = setOf(TextStyle.Block)) {
+                    customTags.forEach {
+                        group(styles = setOf(TextStyle.Block)) {
+                            it.value.values.forEach {
+                                comment(it.root)
+                            }
+                        }
+                    }
+                }
+            }
 
             val unnamedTags = tags.filterNot { (k, _) -> k.isSubclassOf(NamedTagWrapper::class) || k in specialTags }
                 .values.flatten().groupBy { it.first }.mapValues { it.value.map { it.second } }
@@ -705,6 +717,9 @@ open class DefaultPageCreator(
 
     private val Documentable.descriptions: SourceSetDependent<Description>
         get() = groupedTags.withTypeUnnamed<Description>()
+
+    private val Documentable.customTags: Map<String, SourceSetDependent<CustomTagWrapper>>
+        get() = groupedTags.withTypeNamed()
 
     private val Documentable.hasSeparatePage: Boolean
         get() = this !is DTypeAlias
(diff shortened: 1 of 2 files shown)


IgnatBeresnev:
Hi! Sorry for such a long wait with this PR

Changes you're proposing as is would add a few bugs: right now, all custom doctags are ignored, but there are some that have special handling. For instance, `@hide` tag that android codebase uses to hide specific members (just a marker), or `@SinceKotlin` annotation that gets converted to a `Since Kotlin` custom tag to be custom rendered on the page. So we can't just dump its content on the page as is.

We could add a check like `if (it.name == "usesMathJax")`, but it'd be strange to have mathjax internals leak into `dokkaBase` (all of the mathjax logic is in a separate plugin module) and it would only solve this one specific problem, leaving the rest be.

You've done a great job of finding the exact place where this needs to be fixed, which is the most difficult part. So, inspired by this PR, we've added a custom extension point in #2343 that would allow plugins to add special rendering of custom doc tags such as `Since Kotlin` and `@usesMathJax`. If no tag content provider is present for a custom tag, it gets ignored just as before.

I tested it for mathjax, but removed it just before merging as I'd like it to be included in this PR instead (you have some proper tests here and I think credits for fixing this should go to you :)).

Would you be able to add changes from [this commit](https://github.com/Kotlin/dokka/pull/2343/commits/0ef0fb63531fe3f23ae29f9902242b9c97b42336) to the PR? It should work then.

Also, can you rebase onto latest master? There seems to a bug with dependencies that cause CI test failures

Looking forward to merging this, thanks!

wyskoj:
@IgnatBeresnev I think this should be it... let me know if this is good.

IgnatBeresnev (approved):

IgnatBeresnev:
yeah, lgtm, thanks!

vmishenev (approved):

PR #2348, opened 2022-02-07
Fix filtering suppresed extensions

BarkingBad:
A potential fix for a problem that suppressing extensions was breaking the dokka run. I hope I will manage to cover it with some tests.

### DIFF
--- a/core/src/main/kotlin/plugability/DokkaContext.kt
+++ b/core/src/main/kotlin/plugability/DokkaContext.kt
@@ -124,7 +124,11 @@ private class DokkaContextConfigurationImpl(
     }
 
     private fun findNotOverridden(bucket: Set<Extension<*, *, *>>): Extension<*, *, *> {
-        val filtered = bucket.filter { it !in suppressedExtensions }
+        // Let's filter out all suppressedExtensions that are not only overrides.
+        // suppressedExtensions can be polluted by suppressions that completely disables the extension, and would break dokka behaviour
+        // if not filtered out
+        val suppressedExtensionsByOverrides = suppressedExtensions.filterNot { it.value.any { it !is Suppression.ByExtension } }
+        val filtered = bucket.filterNot { it in suppressedExtensionsByOverrides }
         return filtered.singleOrNull()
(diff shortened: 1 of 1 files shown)


vmishenev (changes requested):

core/src/main/kotlin/plugability/DokkaContext.kt
```diff
@@ -124,7 +124,7 @@ private class DokkaContextConfigurationImpl(
     }
 
     private fun findNotOverridden(bucket: Set<Extension<*, *, *>>): Extension<*, *, *> {
-        val filtered = bucket.filter { it !in suppressedExtensions }
+        val filtered = bucket.filter { it !in suppressedExtensions.filter { it.value.all { it is Suppression.ByExtension } } }
```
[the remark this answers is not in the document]
BarkingBad: > Is bucket.filterNot(..) better?

filterNot with any and `!is` will be more efficient since it will abort on the first occurrence. 

> all or any?

It has to be filter + all or filetrNot + any + `!is` since we want only suppressedExtensions by overrides. If there is any suppression other than overrides we handle it in a separate method. It was breaking the previous implementation when we had only `Suppresion.ByPlugin` by manually disabling the extensions.

vmishenev, line 127: I got it, thank you for the explanation.
It would be great to leave a code comment here.

```suggestion
        val suppressedExtensionsByOverrides = suppressedExtensions.filter { it.value.any { it is Suppression.ByExtension }
        val filtered = bucket.filterNot { it in suppressedExtensionsByOverrides } }
```

BarkingBad: Fixed

vmishenev (approved):

BarkingBad:
I guess there is some flaky issue with links, because windows CIs and locally at my PC with Java tests are passing

vmishenev:
I have re-run tests. Also I have opened https://github.com/Kotlin/dokka/issues/2361

IgnatBeresnev (approved):
:+1:

PR #2351, opened 2022-02-08
Expose Kotlin-as-Java methods that could be used not only as part of transformers

BarkingBad:


BarkingBad:
CI failed on windows 17, though I cannot rerun it :(

plugins/kotlin-as-java/src/main/kotlin/transformers/JvmNameDocumentableTransformer.kt
```diff
@@ -55,33 +53,34 @@ class JvmNameDocumentableTransformer : DocumentableTransformer {
         return extraWithoutAnnotations.addAll(listOfNotNull(annotationsWithoutJvmName))
     }
 
-    private fun transformClassLike(documentable: DClasslike): DClasslike =
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun transformClassLike(documentable: DClasslike, context: DokkaContext): DClasslike =
```
vmishenev, line 57: Is it not better to make `transform` public?

vmishenev (changes requested):
I would rather have a special interface for `KotlinAsJavaDocumentableTransformer` and  `JvmNameDocumentableTransformer` to transform  `Documentable`.

plugins/kotlin-as-java/src/main/kotlin/transformers/JvmNameDocumentableTransformer.kt
BarkingBad: Yes, it will be more accurate, will change

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -241,7 +241,7 @@ internal fun DFunction.asJava(containingClassName: String, isTopLevel: Boolean =
     }
 }
 
-internal fun DClasslike.asJava(): DClasslike = when (this) {
+fun DClasslike.asJava(): DClasslike = when (this) {
```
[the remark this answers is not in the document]
BarkingBad: I can expose them too

BarkingBad:
>I would rather have a special interface for KotlinAsJavaDocumentableTransformer and JvmNameDocumentableTransformer to transform Documentable.


Do you want some artificial interfaces with 1 method in each of them, or do you think about something more specific?

plugins/kotlin-as-java/src/main/kotlin/transformDocumentableToJava.kt
```diff
@@ -0,0 +1,10 @@
+package org.jetbrains.dokka.kotlinAsJava
+
+import org.jetbrains.dokka.kotlinAsJava.converters.asJava
+import org.jetbrains.dokka.kotlinAsJava.transformers.JvmNameDocumentableTransformer
+import org.jetbrains.dokka.model.DClasslike
+import org.jetbrains.dokka.plugability.DokkaContext
+
+fun DClasslike.transformClasslikeToJava(context: DokkaContext): DClasslike {
+    return JvmNameDocumentableTransformer().transform(this.asJava(), context)
+}
```
BarkingBad, line 10: I have done it only for classlikes, since making a general function for all intermediate elements would be cumbersome (additional parameters for asJava functions), so they would have to be at least overloaded. WDYT @vmishenev?

vmishenev: To my mind, this is ok.

plugins/kotlin-as-java/src/main/kotlin/transformToJava.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.jetbrains.dokka.plugability.DokkaContext
+
+
+fun DPackage.transformToJava(context: DokkaContext): DPackage {
+    return JvmNameDocumentableTransformer().transform(this.asJava(), context)
+}
+
+fun DClasslike.transformToJava(context: DokkaContext): DClasslike {
+    return JvmNameDocumentableTransformer().transform(this.asJava(), context)
+}
```
IgnatBeresnev, line 18: I also wonder if it makes sense to extract the creation of `JvmNameDocumentableTransformer` into a private lazy field, like
```
private val JVM_NAME_DOCUMENTABLE_TRANSFORMER by lazy { JvmNameDocumentableTransformer() }
```

I'd assume `transformToJava` methods will be called recursively and probably more than once, so it'd help reducing allocation rate and ease gc work.

IgnatBeresnev: Or, better yet, I think `KotlinAsJavaPlugin#jvmNameTransformer` can be reused. This should also help avoid potential configuration problems in the future if the transformer implementation changes

BarkingBad: Recursive calls are not a problem, since these are `asJava` that go recursively and then `JvmNameDocumentableTransformer` is called only once. 

Resuing `jvmNameTransformer` as a property is also problematic since I cannot access `transform` method, I have access only to `invoke` method that is designed for `DModule` only.

The only solution here is to have lazy singleton of that transformer.

IgnatBeresnev (approved):

PR #2364, opened 2022-02-15
Remove bintray publish plugins

Goooler:
Seems jcenter supports could be removed.


IgnatBeresnev (approved):

buildSrc/src/main/kotlin/org/jetbrains/binaryCompatibility.kt
```diff
@@ -15,15 +15,13 @@ internal object BinaryCompatibilityConfig {
     )
 }
 
-internal fun Project.registerBinaryCompatibilityCheck(publicationName: String) {
-    publicationName.takeIf {
-        it !in BinaryCompatibilityConfig.ignoredPublications
-    }?.let {
+internal fun Project.registerApiValidation(publicationName: String) {
```
IgnatBeresnev, line 18: Hi! Wanted to merge this, but thought about this.

Can you actually revert renaming? I'm fine with the `if` part, but I feel like `BinaryCompatibility` is better because it's a widely used plugin within kotlin ecosystem. 

Both dependency and the config are `BinaryCompatibility`, not `ApiValidation`, so it's more consistent, too

IgnatBeresnev (changes requested):

IgnatBeresnev (approved):

buildSrc/src/main/kotlin/org/jetbrains/binaryCompatibility.kt
IgnatBeresnev: Fastest gun in the west :) thanks

PR #2365, opened 2022-02-15
Fix HTML head and favicon in multi-module projects

vmishenev:
`Jsoup` can not process custom html tags inside head block so template commands are kept as html comments

### DIFF
--- a/plugins/templating/src/main/kotlin/templates/CommandHandler.kt
+++ b/plugins/templating/src/main/kotlin/templates/CommandHandler.kt
@@ -2,10 +2,17 @@ package org.jetbrains.dokka.templates
 
 import org.jetbrains.dokka.base.templating.Command
 import org.jsoup.nodes.Element
+import org.jsoup.nodes.Node
 import java.io.File
 
 interface CommandHandler {
     fun handleCommand(element: Element, command: Command, input: File, output: File)
+    fun canHandle(command: Command): Boolean
+    fun finish(output: File) {}
+}
+
+interface CommentCommandHandler {
+    fun handleCommand(nodes: List<Node>, command: Command, input: File, output: File)
     fun canHandle(command: Command): Boolean
     fun finish(output: File) {}
 }
--- a/plugins/templating/src/main/kotlin/templates/CommandHandler.kt
+++ b/plugins/templating/src/main/kotlin/templates/CommandHandler.kt
@@ -8,9 +8,11 @@ import java.io.File
 
 interface CommandHandler  {
     @Deprecated("This was renamed to handleCommandAsTag", ReplaceWith("handleCommandAsTag(command, element, input, output)"))
-    fun handleCommand(element: Element, command: Command, input: File, output: File) =
-        handleCommandAsTag(command, element, input, output)
-    fun handleCommandAsTag(command: Command, body: Element, input: File, output: File)
+    fun handleCommand(element: Element, command: Command, input: File, output: File) { }
+
+    @Suppress("DEPRECATION")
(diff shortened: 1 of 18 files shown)


plugins/templating/src/main/kotlin/templates/DirectiveBasedTemplateProcessing.kt
```diff
@@ -20,9 +25,17 @@ class DirectiveBasedHtmlTemplateProcessingStrategy(private val context: DokkaCon
...
             val document = Jsoup.parse(input, "UTF-8")
             document.outputSettings().indentAmount(0).prettyPrint(false)
+
             document.select("dokka-template-command").forEach {
                 handleCommand(it, parseJson(it.attr("data")), input, output)
             }
+            extractCommandsFromComments(document) { nodes, command ->
+                val nodesTrimed =
+                    nodes.dropWhile { (it is TextNode && it.isBlank).also { res -> if (res) it.remove() } }
+                        .dropLastWhile { (it is TextNode && it.isBlank).also { res -> if (res) it.remove() } }
```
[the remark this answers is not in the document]
vmishenev: > Also, am I correct to understand that we remove all blank text nodes in the beginning and in the end, but leave empty nodes in the middle? Is it just in case or was there something specific?


Because of formatting (indenting) of tags.

```
    <!--[+]cmd:-->
    <link href="###styles/main.css" rel="Stylesheet">
    <link href="###styles/main.css" rel="Stylesheet">
    <!--[-]cmd-->
```
----->
 ```
    <link href="###styles/main.css" rel="Stylesheet">
    <link href="###styles/main.css" rel="Stylesheet">
```

File: plugins/templating/src/main/kotlin/templates/CommandHandler.kt [hunk 2]
```diff
@@ -2,10 +2,16 @@ package org.jetbrains.dokka.templates
...
 
-interface CommandHandler {
-    fun handleCommand(element: Element, command: Command, input: File, output: File)
+
+interface CommandHandler  {
+    @Deprecated("This was renamed to handleCommandAsTag", ReplaceWith("handleCommandAsTag(command, element, input, output)"))
+    fun handleCommand(element: Element, command: Command, input: File, output: File) =
+        handleCommandAsTag(command, element, input, output)
+    fun handleCommandAsTag(command: Command, body: Element, input: File, output: File)
+    fun handleCommandAsComment(command: Command, body: List<Node>, input: File, output: File) { }
```
IgnatBeresnev, line 14: I think with these changes external plugins that override `CommandHandler` might be broken 

External plugins currently override `handleCommand`, but we only call `handleCommandAsTag` from Dokka. So nothing will happen for the plugins other than compilation error for not overriding `handleCommandAsTag`

Looks like we should call legacy `handleCommand` from `handleCommandAsTag` to give people time to migrate
```suggestion
    fun handleCommand(element: Element, command: Command, input: File, output: File) { }
    fun handleCommandAsTag(command: Command, body: Element, input: File, output: File) = handleCommand(body, command, in, out)
    fun handleCommandAsComment(command: Command, body: List<Node>, input: File, output: File) { }
```

vmishenev: So we will have no way to migrate to `handleCommandAsTag` if `handleCommand` is called.
In this case a user implementation of new interface  should have 
``` kt
    fun handleCommand(element: Element, command: Command, input: File, output: File) =
        handleCommandAsTag(command, element, input, output)
```

IgnatBeresnev (approved):

Issue #1599, opened 2020-10-29
Parsing java has failures in hashing 


asfalcone:
**Describe the bug**
Parsing java has failures in hashing 

**To Reproduce**
https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/fragment/source/ Using the source of the Android Jetpack Fragment library we often (but not always) see two kinds of errors in hashing from `findVirtualFileForTopLevelClass`

**Stacktrace 1**

```
Index 88 out of bounds for length 79
java.lang.ArrayIndexOutOfBoundsException: Index 88 out of bounds for length 79
	at gnu.trove.THashMap.rehash(THashMap.java:350)
	at gnu.trove.THash.postInsertHook(THash.java:286)
	at gnu.trove.THashMap.put(THashMap.java:174)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl.findVirtualFileForTopLevelClass(KotlinCliJavaFileManagerImpl.kt:306)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl.access$findVirtualFileForTopLevelClass(KotlinCliJavaFileManagerImpl.kt:48)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl$findPsiClass$1.invoke(KotlinCliJavaFileManagerImpl.kt:70)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl$findPsiClass$1.invoke(KotlinCliJavaFileManagerImpl.kt:48)
	at org.jetbrains.kotlin.util.PerformanceCounter.time(PerformanceCounter.kt:101)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl.findPsiClass(KotlinCliJavaFileManagerImpl.kt:69)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCliJavaFileManagerImpl.findClass(KotlinCliJavaFileManagerImpl.kt:135)
	at com.intellij.psi.impl.PsiElementFinderImpl.findClass(PsiElementFinderImpl.java:44)
	at com.intellij.psi.impl.JavaPsiFacadeImpl.doFindClass(JavaPsiFacadeImpl.java:110)
[...]

PR #2366, opened 2022-02-16
Remove various caches from `JvmDependenciesIndexImpl`

IgnatBeresnev:
Removed various caches and did some `dokkaHtmlMultiModule` benchmarking on [kotlinx-coroutines](https://github.com/Kotlin/kotlinx.coroutines).

I benchmarked 4 times: baseline (current master) + once for each removed cache (`#1`, `#2` and `#3`, will outline them by comments) to see the impact of each removed cache. 

[The results (html link)](https://ignatberesnev.github.io/static/dokka/removed-cache-benchmark/combined-results.html) show that there's no significant build time degradation, but the benchmarks are not as ideal as they could be:
1. Now that I understand how these caches work, perhaps `kotlinx-coroutines` wasn't the best project to benchmark on -- it doesn't have that many dependencies and roots, so the cache is relatively small in size. I wonder if the results will be any different on a large scale enterprise project
2. As far as I understand, the caches were introduced to reduce load on IO, so supposedly we'll see the biggest difference on slow aged HDDs, whereas the benchmarks were performed on a pristine new gen SSD with a lot of free memory, so the files were probably cached as well.

Alternatively, it's possible to add some basic thread safety to the current solution by synchronizing on individual `Cache` objects when working with them, but because there's a section where two caches need to be locked consecutively, I can come up with a hypothetical concurrent scenario in which a deadlock happens (whether it could actually happen is a good question, need more time to investigate). 

P.S., turns out, `JvmDependenciesIndexImpl` was [copy-pasted from the kotlin repo](https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/index/JvmDependenciesIndexImpl.kt) (it also has caches), which makes it a little scary to remove (hopefully it was introduced to kotlin based on some testing). I was unable to find the PR or any attached issue to introducing the caches, so not sure what the motivation was behind it, and its primary contributor does not work at JB anymore, so I can't ask directly.

### DIFF
--- a/kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/JvmDependenciesIndexImpl.kt
+++ b/kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/JvmDependenciesIndexImpl.kt
@@ -20,8 +20,6 @@ import com.intellij.ide.highlighter.JavaClassFileType
 import com.intellij.ide.highlighter.JavaFileType
 import com.intellij.openapi.vfs.VfsUtilCore
 import com.intellij.openapi.vfs.VirtualFile
-import com.intellij.util.containers.IntArrayList
-import gnu.trove.THashMap
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.FqName
 import java.util.*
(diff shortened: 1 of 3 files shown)


kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/JvmDependenciesIndexImpl.kt
```diff
@@ -36,38 +34,8 @@ class JvmDependenciesIndexImpl(_roots: List<JavaRoot>) : JvmDependenciesIndex {
...
-
-        // indices of roots that are known to contain this package
-        // if this list contains [1, 3, 5] then roots with indices 1, 3 and 5 are known to contain this package, 2 and 4 are known not to (no information about roots 6 or higher)
-        // if this list contains maxIndex that means that all roots containing this package are known
-        val rootIndices = IntArrayList(2)
-    }
-
-    // root "Cache" object corresponds to DefaultPackage which exists in every root. Roots with non-default fqname are also listed here but
-    // they will be ignored on requests with invalid fqname prefix.
-    private val rootCache: Cache by lazy {
```
IgnatBeresnev, line 53: Main cache **#1**, the one we've been having problems with. It's also nested.

Basically maps packages to an index inside `roots` (separate field at the top) to reduce lookup time (no need to iterate everything, just the needed packages).

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/JvmDependenciesIndexImpl.kt
```diff
@@ -36,38 +34,8 @@ class JvmDependenciesIndexImpl(_roots: List<JavaRoot>) : JvmDependenciesIndex {
...
-        Cache().apply {
-            roots.indices.forEach(rootIndices::add)
-            rootIndices.add(maxIndex)
-            rootIndices.trimToSize()
-        }
-    }
-
-    // holds the request and the result last time we searched for class
-    // helps improve several scenarios, LazyJavaResolverContext.findClassInJava being the most important
-    private var lastClassSearch: Pair<FindClassRequest, SearchResult>? = null
```
IgnatBeresnev, line 63: Cache **№2**. There are cache hits, actually more than I thought, so this could be useful. Could be made thread safe without much problem.

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/JvmDependenciesIndexImpl.kt
```diff
@@ -36,38 +34,8 @@ class JvmDependenciesIndexImpl(_roots: List<JavaRoot>) : JvmDependenciesIndex {
...
-
     override val indexedRoots by lazy { roots.asSequence() }
 
-    private val packageCache: Array<out MutableMap<String, VirtualFile?>> by lazy {
-        Array(roots.size) { THashMap<String, VirtualFile?>() }
-    }
```
IgnatBeresnev, line 69: Cache **№3**. Already synchronized on, so there shouldn't have been any problems with it. Seems to be the most important one as it saves most of the heavy lifting (stores the result of `findChildPackage`). I think this can be safely reverted

zarechenskiy (approved):

zarechenskiy:
Speaking about original motivation: this index was copied to fix https://github.com/Kotlin/dokka/issues/1599, there was a problem with the Trove library

vmishenev (approved):

IgnatBeresnev:
Stale. Will re-open if this is needed for tests again.

Issue #2227, opened 2021-11-16
Unable to hide/suppress k-doc in mixed(java+kotlin) project


macharya:
In my project, I've both kotlin and java files, I am trying to hide a package, I tried @hide @hidden @Suppress still cant hide a java class or public method. Added following 3 libs to my build.gradle and nothing seems to work. Also, what's the right way to use the package option in build.gradle not .kts. followed https://github.com/Kotlin/dokka/issues/492 but that did not solve this problem. Any help?

    classpath "org.jetbrains.dokka:dokka-gradle-plugin:1.4.30"
    classpath "org.jetbrains.dokka:kotlin-as-java-plugin:1.4.30"
    classpath "org.jetbrains.dokka:android-documentation-plugin:1.4.30"

vmishenev:
You can use `@suppress` to hide a method from the documentation.
Support of a package level will be in Dokka 1.6. (#2206)  

Now `matchingRegex` can be used to hide a package.  See [example](https://kotlin.github.io/dokka/1.5.30/user_guide/gradle/usage/)

bashbug:
I have the same issue with dokka 1.5.31

**Problem**
@hide works for all java functions/classes/etc. Except for functions which have a return value.

**How to reproduce**
Add to https://github.com/Kotlin/dokka/tree/master/examples/gradle/dokka-kotlinAsJava-example/src/main/kotlin/demo the following java class.
- setB() is hidden.
- getA() is shown, even with the @hide tag
```
public class JavaClass {
    private String a;
    private String b;
    JavaClass(String a){
        this.a = a;
    }

    /**
     * @hide
     */
    public void setB(String b) {
        this.b = b;
    }

    /**
     * @hide
     */
    public String getA() {
        return a;
    }
}
```

Is this also fixed @vmishenev with the upcoming 1.6 release?

macharya:
@bashbug how is @hide working for your java classes/methods? I tried 1.5.31 and still can't hide the java methods or class. Tried all three of these but no luck

    classpath "org.jetbrains.dokka:dokka-gradle-plugin:1.5.31"
    classpath "org.jetbrains.dokka:kotlin-as-java-plugin:1.5.31"
    classpath "org.jetbrains.dokka:android-documentation-plugin:1.5.31"

vmishenev:
@bashbug I suppose this fix will be in the next version after Dokka 1.6.
Dokka handles `setB` and `getA` as a setter and a getter so #2158 will fix the bug.
[...]

PR #2370, opened 2022-02-17
Clarify where to apply the android plugin

CharlesG-Branch:
This is in response to this comment https://github.com/Kotlin/dokka/issues/2227#issuecomment-974580635
We spent a fair amount of time trying to figure out why it wasn't working for us & putting this extra line
in the documentation would've saved us a lot of time. cc @IgnatBeresnev

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ apply(plugin= "org.jetbrains.dokka")
 ```
 
 ```kotlin
+dependencies {
+    dokkaHtmlPlugin("org.jetbrains.dokka:android-documentation-plugin:${dokka_version}")
+}
+
 dokkaHtml.configure {
     dokkaSourceSets {
         named("main") {
@@ -87,6 +91,9 @@ dokkaHtml.configure {
     }
 }
 ```
+
+**NOTE:** replace `Html` in the above code with whatever [output format](#output_formats) you need. 
+
 
 #### Multi-module projects
 For documenting Gradle multi-module projects, you can use `dokka${format}Multimodule` tasks.


README.md
```diff
@@ -79,6 +79,10 @@ apply(plugin= "org.jetbrains.dokka")
 ```
 
 ```kotlin
+dependencies {
+    dokkaHtmlPlugin("org.jetbrains.dokka:android-documentation-plugin:${dokka_version}")
+}
```
IgnatBeresnev, line 84: Since not everyone needs this plugin and there's no explanation of what it's used for, I'd rather have it described separately down below, in the end of the `Android` section.

Something like

> In order to improve documentation experience on Android platform (such as support for `@hide` tag), consider adding the [android plugin](https://kotlin.github.io/dokka/1.6.10/user_guide/android-plugin/android-plugin/):
> ```
> dependencies {
>    dokkaHtmlPlugin("org.jetbrains.dokka:android-documentation-plugin:${dokka_version}")
> }
> ```

And the same `dokkaHtmlPlugin("org.jetbrains.dokka:android-documentation-plugin:${dokka_version}")` could be added to that page as well (website doc files are in the `docs` module).

What do you think about that?

README.md
```diff
@@ -88,6 +92,9 @@ dokkaHtml.configure {
 }
 ```
 
+**NOTE:** replace `Html` in the above code with whatever [output format](#output_formats) you need.
```
IgnatBeresnev, line 95: I feel like this was explained in the `using the gradle plugin` and `applying plugins` sections:

> The plugin adds dokkaHtml, dokkaJavadoc, dokkaGfm and dokkaJekyll tasks to the project.

> Dokka plugin creates Gradle configuration for each output format in the form of dokka${format}Plugin:

I think it's a good idea to keep `README.md` short, do you really think it's worth to have it here?

This can definitely be added to [android plugin](https://kotlin.github.io/dokka/1.6.10/user_guide/android-plugin/android-plugin/) page, won't hurt

README.md
CharlesG-Branch: I'll respond to both of your comments here as they're pretty related.

I'd expect that the section on android has all of the configuration needed for an android project & that I'd just need to copy paste the example to get all of it.

My team found the plugin system pretty confusing. That might just be because we wanted to throw the config in gradle and not think about it that much, but I'd also argue that's how the vasty majority of people will be using the library. It could also be that whenever I see a `"org.jetbrains.dokka:android-documentation-plugin:${dokka_version}"` my lizard brain just immediately associates that with `implementation` commands etc, but IMHO that's another argument to make it more copy paste friendly.

After diving into the project more it did make sense.

PR #2353, opened 2022-02-10
Fix gathering inherited properties

BarkingBad:
Enables gathering super fields/properties by PSI's. This PR depends on model changes introduced in #2326 so I leave it as a draft (it is rebased onto that branch in order to build correctly. Nonetheless, it can be reviewed as a single commit.

PR #2377, opened 2022-02-25
Kotlin-as-Java fixes for external invocation

Kordyjan:
This PR brings few improvements to Kotlin-as-Java, especially in cases when it is not used as a transformer:

- java accessors retains information about the original property
- companion objects are transformed to java classes (this change is irrelevant to the normal execution of dokka, as the `companion` property of `DClass` was ignored when the KaJ plugin was enabled)
- generation of java accessors was unified for classes and objects
- lateinit properties are retaining their fields

**This is a follow-up PR to the #2353**

### DIFF
--- a/plugins/kotlin-as-java/src/main/kotlin/OriginalProperty.kt
+++ b/plugins/kotlin-as-java/src/main/kotlin/OriginalProperty.kt
@@ -1,0 +1,11 @@
+package org.jetbrains.dokka.kotlinAsJava
+
+import org.jetbrains.dokka.model.DFunction
+import org.jetbrains.dokka.model.DProperty
+import org.jetbrains.dokka.model.properties.ExtraProperty
+import org.jetbrains.dokka.model.properties.MergeStrategy
+
+class OriginalProperty(val original: DProperty) : ExtraProperty<DFunction> {
+    override val key = OriginalProperty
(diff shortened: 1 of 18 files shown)


plugins/kotlin-as-java/src/main/kotlin/OriginalProperty.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import org.jetbrains.dokka.model.DFunction
+import org.jetbrains.dokka.model.DProperty
+import org.jetbrains.dokka.model.properties.ExtraProperty
+import org.jetbrains.dokka.model.properties.MergeStrategy
+
+class OriginalProperty(val original: DProperty) : ExtraProperty<DFunction> {
+    override val key = OriginalProperty
+    companion object: ExtraProperty.Key<DFunction, OriginalProperty>
+}
```
IgnatBeresnev, line 11: Out of context, a bit unclear what `Original` is exactly. Original in relation to what?

After seeing the usage, it starts to make sense, but maybe it's worth adding a simple KDoc like `original function property before any merging or modificatoin took place` to save some brain fuel

Also, I didn't find any actual usages of this - is it something you'll need in the future or use in your plugins?

IgnatBeresnev:
Closing as stale, will revive if need arises

PR #2401, opened 2022-03-22
Minor refactoring of dokka versioning and publishing

IgnatBeresnev:
Unable to change dokka's version to RC without these changes, it'll fail build locally then.

* Added `RC` version type
* Removed `MC` version type as it seems to be unused (will revert if it's really needed for something)
* Added `gradle-plugin-portal` as a publication channel (only `RELEASE` and `RC` are allowed)
* Removed `dokka_version_base` in favour of `dokka_version`. Base version seems unused now, as TC passes value directly to `dokka_version` and automatically adding `-SNAPSHOT` is not that big of a deal, it'll be present by default anyway.
* Reverted condition for version check (now off by default, needs to by enabled by setting `ENABLE_VERSION_CHECK` to `true`)

I've also updated TeamCity publish configuration and added `env.ENABLE_VERSION_CHECK = true` and also added an `RC` option for `dokka_version`. Will test it when publishing RC to space

So version checking should still work now when publishing from TC, but is disabled locally.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -28,7 +28,7 @@ jobs:
           java-version: 11
       - uses: gradle/gradle-build-action@v2
       - name: Get current dokka version
-        run: echo "DOKKA_VERSION=`./gradlew :properties | grep '^version:.*' | cut -d ' ' -f 2 | cut -d '-' -f 1`" >> $GITHUB_ENV
+        run: echo "DOKKA_VERSION=`./gradlew :properties | grep '^version:.*' | cut -d ' ' -f 2`" >> $GITHUB_ENV
         if: github.event_name == 'release' || steps.filter.outputs.docs_changed == 'true'
         working-directory: ./dokka
       - name: Build docs
--- a/buildSrc/src/main/kotlin/org/jetbrains/DokkaPublicationChannel.kt
+++ b/buildSrc/src/main/kotlin/org/jetbrains/DokkaPublicationChannel.kt
@@ -22,6 +22,8 @@ enum class DokkaPublicationChannel {
 
     fun isMavenRepository() =  this == MAVEN_CENTRAL || this == MAVEN_CENTRAL_SNAPSHOT
 
+    fun isGradlePluginPortal() = this == GRADLE_PLUGIN_PORTAL
+
     companion object {
         fun fromPropertyString(value: String): DokkaPublicationChannel = when (value) {
             "space-dokka-dev" -> SPACE_DOKKA_DEV
--- a/buildSrc/src/main/kotlin/org/jetbrains/DokkaPublicationChannel.kt
+++ b/buildSrc/src/main/kotlin/org/jetbrains/DokkaPublicationChannel.kt
@@ -5,30 +5,29 @@ package org.jetbrains
 import org.gradle.api.Project
 
 enum class DokkaPublicationChannel {
-    SpaceDokkaDev,
-    MavenCentral,
-    MavenCentralSnapshot;
-
-    val isSpaceRepository get() = this == SpaceDokkaDev
-
-    val isMavenRepository
-        get() = when (this) {
-            MavenCentral, MavenCentralSnapshot -> true
-            else -> false
-        }
+    SPACE_DOKKA_DEV,
+    MAVEN_CENTRAL,
+    MAVEN_CENTRAL_SNAPSHOT,
+    GRADLE_PLUGIN_PORTAL;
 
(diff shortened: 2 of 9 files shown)


gradle.properties
```diff
@@ -1,6 +1,5 @@
 # Project Settings
-dokka_version_base=1.6.10
-dokka_publication_channels=maven-central-snapshot&space-dokka-dev
+dokka_version=1.6.20-SNAPSHOT
```
IgnatBeresnev, line 2: Had to add this because integration tests fail otherwise. Previously, it added `-SNAPSHOT` automatically and integration test projects used the SNAPSHOT version when running the tests.

If I revert it back to `1.6.10`, integration tests will use `1.6.10` from maven central, which doesn't have some features yet (like documentedVisibilities), and so the integration tests will fail.

I set it to `1.6.20-SNAPSHOT` specifically since this is the version now in master. I'll change it to `1.6.30-SNAPSHOT` after creating the `1.6.20` release branch.

buildSrc/src/main/kotlin/org/jetbrains/ValidatePublications.kt
```diff
@@ -11,31 +11,33 @@ import org.gradle.kotlin.dsl.findByType
...
                     subProject.assertPublicationVersion()
                 }
         }
     }
 
     private fun checkProjectDependenciesArePublished(project: Project) {
-        (project.configurations.findByName("implementation")?.allDependencies.orEmpty() +
-                project.configurations.findByName("api")?.allDependencies.orEmpty())
+        val implementationDeps = project.configurations.findByName("implementation")?.allDependencies.orEmpty()
+        val apiDependencies = project.configurations.findByName("api")?.allDependencies.orEmpty()
```
vmishenev, line 37: `apiDeps` would be similar to  `implementationDeps`

vmishenev (changes requested):
What is about building scripts of GH pages?

gradle.properties
vmishenev: Is it possible to change the integration test to use a special version like [here](https://github.com/Kotlin/dokka/blob/78f25296ad967cba3cfd5bc28a3fd4f5950f5ca0/integration-tests/gradle/projects/template.settings.gradle.kts#L17)?

buildSrc/src/main/kotlin/org/jetbrains/DokkaPublicationChannel.kt
```diff
@@ -5,30 +5,29 @@ package org.jetbrains
...
-
-    val isMavenRepository
-        get() = when (this) {
-            MavenCentral, MavenCentralSnapshot -> true
-            else -> false
-        }
+    SPACE_DOKKA_DEV,
+    MAVEN_CENTRAL,
+    MAVEN_CENTRAL_SNAPSHOT,
+    GRADLE_PLUGIN_PORTAL;
```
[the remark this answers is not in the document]
IgnatBeresnev: Added proper publication to gradle plugin portal via `dokkaPublish` and publication channels.

.github/workflows/gh-pages.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           java-version: 11
       - uses: gradle/gradle-build-action@v2
       - name: Get current dokka version
-        run: echo "DOKKA_VERSION=`./gradlew :properties | grep '^version:.*' | cut -d ' ' -f 2 | cut -d '-' -f 1`" >> $GITHUB_ENV
+        run: echo "DOKKA_VERSION=`./gradlew :properties | grep '^version:.*' | cut -d ' ' -f 2`" >> $GITHUB_ENV
```
IgnatBeresnev, line 31:
```
[ignat@jb dokka]$ ./gradlew :properties | grep '^version:.*' | cut -d ' ' -f 2
1.6.20-SNAPSHOT
```

vmishenev (approved):

buildSrc/src/main/kotlin/org/jetbrains/ValidatePublications.kt
```diff
@@ -53,22 +55,24 @@ open class ValidatePublications : DefaultTask() {
     }
 
     private fun Project.assertPublicationVersion() {
-        if (System.getenv("SKIP_VERSION_CHECK")?.contains("true", ignoreCase = true) == true)
+        val isVersionCheckEnabled = System.getenv("ENABLE_VERSION_CHECK").equals("true", ignoreCase = true)
```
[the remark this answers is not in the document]
vominhky: 58

PR #2414, opened 2022-03-31
Rework AnnotationTarget to be stricter.

owengray-google:
Rework AnnotationTarget to be stricter. Now the things that can be annotated are precisely the things which implement AnnotationTarget.

Also adds tests, though they currently don't do anything because of pre-existing bugs

This replaces github.com/Kotlin/dokka/pull/2210. I believe all the commends from that PR are addressed.

This is based on our desire to have PrimitiveJavaObject 

The two things that I think are bugs that limits the effectiveness of the tests:
1. Even after this change, `UnresolvedType` will not be annotable.

To allow that to happen would require a change in `org.jetbrains.kotlin.types.ErrorType` and `.UnresolvedType` to take and use an `annotations` parameter rather than explicitly override the class property it to empty, and then for that parameter to be backpropagated into `org.jetbrains.kotlin.types.ErrorUtils.createUnresolvedType` and `org.jetbrains.kotlin.resolve.resolveTypeElement.visitUserType`, using the there-local `annotations` property.

The justification for why you might want this is iffy.
a) Documentation generation can potentially be run on incomplete parts of the codebase, for speed, incrementality, (integration) tests, etc. In these circumstances, we expect to see unresolved types. Being more able to handle unresolved types as though they were real types would help us do such things.
b) Annotation processors might find it more convenient, to debug if nothing else?

2. It appears that dokka's `KotlinToJavaConverter` turns `Kotlin.Int` into `java.lang.Integer` not primitive `int` as dackka does. This means that you don't need the ability to have `PrimitiveJavaType` be `withExtraProperties`. However, `com.intellij.psi.PsiPrimitiveType` supports annotations, so it should be valid to support them in dokka.

### DIFF
--- a/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
@@ -13,8 +13,9 @@ import org.jetbrains.dokka.analysis.KotlinAnalysis
 import org.jetbrains.dokka.analysis.from
 import org.jetbrains.dokka.base.DokkaBase
 import org.jetbrains.dokka.base.parsers.MarkdownParser
-import org.jetbrains.dokka.base.translators.typeConstructorsBeingExceptions
 import org.jetbrains.dokka.base.translators.psi.parsers.JavadocParser
+import org.jetbrains.dokka.base.translators.psi.DefaultPsiToDocumentableTranslator
+import org.jetbrains.dokka.base.translators.typeConstructorsBeingExceptions
 import org.jetbrains.dokka.base.translators.unquotedValue
 import org.jetbrains.dokka.links.*
 import org.jetbrains.dokka.links.Callable
@@ -31,7 +32,6 @@ import org.jetbrains.dokka.utilities.DokkaLogger
 import org.jetbrains.dokka.utilities.parallelMap
 import org.jetbrains.dokka.utilities.parallelMapNotNull
 import org.jetbrains.kotlin.KtNodeTypes
-import org.jetbrains.dokka.model.BooleanConstant
 import org.jetbrains.kotlin.builtins.functions.FunctionClassDescriptor
 import org.jetbrains.kotlin.builtins.isBuiltinExtensionFunctionalType
 import org.jetbrains.kotlin.builtins.isExtensionFunctionType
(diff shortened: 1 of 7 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -68,16 +68,16 @@ import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 import java.nio.file.Paths
 import org.jetbrains.kotlin.resolve.constants.AnnotationValue as ConstantsAnnotationValue
 import org.jetbrains.kotlin.resolve.constants.ArrayValue as ConstantsArrayValue
-import org.jetbrains.kotlin.resolve.constants.EnumValue as ConstantsEnumValue
-import org.jetbrains.kotlin.resolve.constants.KClassValue as ConstantsKtClassValue
+import org.jetbrains.kotlin.resolve.constants.BooleanValue as ConstantsBooleanValue
```
owengray-google, line 71: I hit intelliJ autofix imports. It looks like this is a correct alphabetization? I hope this is fine.

IgnatBeresnev:
> This replaces github.com/https://github.com/Kotlin/dokka/pull/2210.

Can #2210 be closed then?

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -258,4 +260,56 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Disabled // The compiler throws away annotations on unresolved types upstream
```
vmishenev, line 264: Can you move it in `DefaultDescriptorToDocumentableTranslatorTest`?

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -258,4 +260,56 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
...
+    @Test
+    fun `Can annotate UnresolvedBound`() {
```
vmishenev, line 266: `UnresolvedBound` are created only in `DefaultPsiToDocumentableTranslator`. 

Here it can be replaced with `unresolved type`.

vmishenev (commented):
> To allow that to happen would require a change in org.jetbrains.kotlin.types.ErrorType and .UnresolvedType to take and use an annotations parameter rather than explicitly override the class property it to empty, and then for that parameter to be backpropagated into org.jetbrains.kotlin.types.ErrorUtils.createUnresolvedType and org.jetbrains.kotlin.resolve.resolveTypeElement.visitUserType, using the there-local annotations property.

I have checked. You are right. Thank you for investigation. 
It is part of Kotlin compiler. So feel free to contribute to [Kotlin repository](https://github.com/JetBrains/kotlin). 
The needed code is [here](https://github.com/JetBrains/kotlin/blob/04da0ba55d801ae6c0e0bc6d6918dd0e56fca435/compiler/frontend/src/org/jetbrains/kotlin/resolve/TypeResolver.kt#L259).

vmishenev (approved):

PR #2416, opened 2022-04-03
Added Mermaid plugin to the plugins-list.md

glureau:
Feel free to tell me if the plugin should handle some quality level before to be added in this list.
I'm currently working on supporting dark mode and be able to configure themes from gradle (not sure how I can do that latter).

EDIT: dark mode and theme configuration are supported now


IgnatBeresnev (approved):
Good job! :) I do remember what I promised about adding a proper extension point, hopefully we'll get to it sometime soon

vmishenev (approved):

Issue #1838, opened 2021-04-11
Module.md - Kotlin code sample with annotation fails


hfhbd:
**Describe the bug**
If a code sample inside the Module.md contains an annotation, like `@ExperimentalTime` this sample and all following content is missing.

I checked out the sample code from https://github.com/Kotlin/kotlin-examples/tree/master/gradle/dokka/dokka-gradle-example/src/main/kotlin.

**Expected behaviour**
Code sample containing the annotation

**Screenshots**
<img width="1894" alt="Bildschirmfoto 2021-04-12 um 00 42 08" src="https://user-images.githubusercontent.com/22521688/114323982-0cc27600-9b28-11eb-936e-5b6be8a89456.png">

**To Reproduce**
```markdown
# Module Dokka Gradle Example

This is an example of how you can write module documentation with Dokka.

## Testing Kotlin code

Sample 1/2

````kotlin
fun sample1() {

}
````
````markdown
Sample 2/2

````kotlin
@ExperimentalTime
fun sample2() {

}
````
````markdown
# Package demo

This package contains a few examples of Dokka usage.

````

**Dokka configuration**
Configuration of dokka used to reproduce the bug

```Kotlin
import org.jetbrains.dokka.gradle.DokkaTask
import java.net.URL

plugins {
    kotlin("jvm") version "1.4.32"
    id("org.jetbrains.dokka") version ("1.4.30")
}

repositories {
    mavenCentral()
    jcenter()
}

dependencies {
    testImplementation(kotlin("test-junit"))
}

tasks.withType<DokkaTask>().configureEach {
    dokkaSourceSets {
        configureEach {
            moduleName.set("Dokka Gradle Example")
            includes.from("Module.md")
            val sourceSetName = name
            println(sourceSetName)
            sourceLink {
                localDirectory.set(file("src/$sourceSetName/kotlin"))
                remoteUrl.set(URL("https://github.com/Kotlin/kotlin-examples/tree/master/" +
                        "gradle/dokka/dokka-gradle-example/src/main/kotlin"
                ))
                remoteLineSuffix.set("#L")
            }
        }
    }
}

```

**Installation**
- Operating system: macOS 11
- Build tool: Gradle v7.0
- Dokka version: 1.4.30

IgnatBeresnev:
Partially fixed in upcoming 1.6.20. The fix will solve the majority of problems, although some corner cases may still arise

PR #2418, opened 2022-04-04
Fix using `@` symbol inside code block

vmishenev:


plugins/base/src/main/kotlin/parsers/Parser.kt
```diff
@@ -49,10 +51,49 @@ abstract class Parser {
 
     private fun jkdocToListOfPairs(javadoc: String): List<Pair<String, String>> =
         "description $javadoc"
-            .split("\n@")
+            .splitIgnoredInsideBackticks("\n@")
```
IgnatBeresnev, line 54: A bit concerned about performance degradation after this change. Looking at the implementation of `split` and at yours, there's additional overhead added for each j/kdoc that's parsed, whereas by far not every doc comment contains code sections and may even have no tags (in this situation, split would do a lookup of the delimiter and just return the string if nothing is found, whereas `splitIgnoredInsideBackticks` would go char by char and copy it to the StringBuilder)

So it seems like this case is only relevant for comments where it contains both `@` symbol and backticks. I suggest we use the special function only if this condition holds true. Ideally, we'd want to check if there's at least one `@` is inside backticks (could be done via `indexOf` comparison), but I think it may be too much. Other than that, I think `contains` would call `nativeIndexOf` and it should be pretty fast.

```kotlin
private fun jkdocToListOfPairs(doc: String): List<Pair<String, String>> {
    val descriptionDoc = "description $doc"

    // annotations inside code blocks must not be regarded as kdoc/javadoc tags
    val mayContainAnnotationsInsideCodeSection = doc.contains("```") && doc.contains("\n@")
    val docTokens = if (mayContainAnnotationsInsideCodeSection) {
        descriptionDoc.splitIgnoredInsideBackticks()
    } else {
        descriptionDoc.split("\n@")
    }

    return docTokens.map { content ->
        val contentWithEscapedAts = content.replace("\\@", "@")
        val (tag, body) = contentWithEscapedAts.split(" ", limit = 2)
        tag to body
    }
}
```

plugins/base/src/test/kotlin/markdown/ParserTest.kt
```diff
@@ -951,6 +951,7 @@ class ParserTest : KDocTest() {
     fun `Multilined Code Block`() {
         val kdoc = """
         | ```kotlin
+        | @Suppress("UNUSED_VARIABLE")
```
IgnatBeresnev, line 954: I think it'd be better to have a separate small test that checks just `@` inside code blocks. In case this particular feature breaks it'd be easier to spot and understand what exactly broke.

plugins/base/src/main/kotlin/parsers/Parser.kt
vmishenev: This is a good point, but I have another solution.

hfhbd:
Fixes https://github.com/Kotlin/dokka/issues/1838 too

plugins/base/src/main/kotlin/parsers/Parser.kt
IgnatBeresnev: It's definitely better, but I still think it's overkill to migrate to this function exclusively for such a rare corner case. I believe Kotlin's `split` is better tested and outperforms `splitIgnoredInsideBackticks` for the majority of cases.

Any objections to what I proposed above about using `splitIgnoredInsideBackticks` only when backticks and `@` are present?

vmishenev: 1. I am not sure It is a rare corner case. See the attached issues.
2. I am not sure `split` outperforms `splitIgnoredInsideBackticks`. See `split` implementation in JVM. Anyway there are cases when we should use `splitIgnoredInsideBacktick`. It have to be tested as well.
3. The suggested code have 2 redundant  invocations of `contains` with linear complexity. 
4. This is a heuristic that make the code more complicated to understand a normal flow.

IgnatBeresnev: Discussed it in DMs. This method is only used for `module.md` files, so it's not as performance critical as I thought it was. I'll suggest a few renames, but other than that I think it'll be fine

plugins/base/src/main/kotlin/parsers/Parser.kt
```diff
@@ -49,10 +51,51 @@ abstract class Parser {
 
     private fun jkdocToListOfPairs(javadoc: String): List<Pair<String, String>> =
```
IgnatBeresnev, line 52: let's rename this to `parseTags` or something like this?

I've run dokkaHtml on kotlinx.coroutines and dumped the contents of the `javadoc` argument - it's all `module.md` contents, there's not a single kdoc/javadoc. The method and the argument names are really misleading

`parseTags` is a generic name that can be used for both k/jdoc and `module.md` files

plugins/base/src/main/kotlin/parsers/Parser.kt
```diff
@@ -49,10 +51,51 @@ abstract class Parser {
...
+
+        var isInCode = false
+        val result = mutableListOf<String>()
+        var rangeStart = 0
+        var rangeEnd = 0
+        var currentOffset = 0
+        while (currentOffset < length) {
+
+            if (get(currentOffset) == '`') {
+                countOfBackticks++
```
IgnatBeresnev, line 73: I would leave a comment here or somewhere close, something like
```
// as per commonmark spec, there can be any number of backticks, not only one or three
```
otherwise someone (like me) might think to optimize this to `indexOf("```")` or `indexOf("`")`

then your other comment would make more sense actually :+1:

plugins/base/src/main/kotlin/parsers/Parser.kt
```diff
@@ -49,10 +51,51 @@ abstract class Parser {
...
         "description $javadoc"
-            .split("\n@")
+            .splitIgnoredInsideBackticks("\n@")
             .map { content ->
                 val contentWithEscapedAts = content.replace("\\@", "@")
                 val (tag, body) = contentWithEscapedAts.split(" ", limit = 2)
                 tag to body
             }
+
+    private fun CharSequence.splitIgnoredInsideBackticks(delimiter: String): List<String> {
```
IgnatBeresnev, line 61: You were right when you said it's not a simple split anymore. Suggestions:

1. the name can be changed to something like `parseNewLineTagSections`
2. `delimiter` argument can be dropped, it's highly unlikely that there will be some other char, not `"\n@"`. Moreover, the whole idea of ignoring backticks loses it's meaning if another delimiter is used, and then simple `split` could be used instead.

plugins/base/src/main/kotlin/parsers/Parser.kt
vmishenev: > let's rename this to `parseTags` or something like this?

It does not parse tags entirely. Besides, we already have `parseTagWithBody` method.

> it's all module.md contents, there's not a single kdoc/javadoc. 

It may be not only the multi-module documentation, but module and package documentation as well. 
In first case the method parse the all `module.md` contents.

plugins/base/src/main/kotlin/parsers/Parser.kt
vmishenev: On the other hand, another delimiter can be used to parse, e.g. a md unordered list.

plugins/base/src/main/kotlin/parsers/Parser.kt
IgnatBeresnev: The naming is up to you, but `jkdoc` part should definitely be removed, it's misleading

IgnatBeresnev (approved):
👍👍

PR #2432, opened 2022-04-11
Decompose `base.ftl` template

vmishenev:


plugins/base/src/main/resources/dokka/templates/includes/page_heads.ftl
```diff
@@ -0,0 +1,6 @@
+<#macro display>
+    <title>${pageName}</title>
+    <@template_cmd name="pathToRoot">
+    <link href="${pathToRoot}images/logo-icon.svg" rel="icon" type="image/svg">
+    </@template_cmd>
+</#macro>
```
IgnatBeresnev, line 6: What do you think about renaming the file to something like `page_metadata.ftl` or `page_head_metadata.ftl`? I think having some words from the [html spec](https://www.w3schools.com/html/html_head.asp) would help in intuitive understanding of what this template does.

I'd say `page_heads.ftl` is 50/50 - you assume it's elements inside `<head>`, but not 100% sure and have to double check. Moreover, it conflicts with `header.ftl` a bit

plugins/base/src/main/resources/dokka/templates/includes/article.ftl
```diff
@@ -0,0 +1,3 @@
+<#macro display>
+        <@content/>
+</#macro>
```
IgnatBeresnev, line 3: Maybe rename the file it to `page_content`, `main_content` or just `content`? 

I think `article` is also 50/50, you think you know what it is, but still have to check it, whereas something with `content` would be 100% clear to me

docs/src/doc/docs/user_guide/base-specific/frontend.md
```diff
@@ -81,22 +81,24 @@ To customize HTML output, you can use the [default template](https://github.com/
     To change page assets, you can set properties `customAssets` and `customStyleSheets`.
     Assets are handled by Dokka itself, not FreeMaker.
 
-Currently, there is only one template file with predefined name `base.ftl`. It defines general design of all pages to render.  
-If `templatesDir` is defined, Dokka will find the `base.ftl` file there.
+There is a template file with predefined name `base.ftl`. It defines general design of all pages to render.
+`base.ftl` can import another templates that can be set by user as well.
```
IgnatBeresnev, line 85: I assume the user can overwrite individual templates, like only the footer? 

Can you please add which templates can be overridden and what it's used for? i.e
```
* footer.ftl - overrides the footer portion of the page, which includes the footer message and the copyright
```

vmishenev: > I assume the user can overwrite individual templates, like only the footer?
Yes.

> Can you please add which templates can be overridden and what it's used for? i.e

I suppose the description of templates is too verbose. 
A file path should be enough. WDYT?

IgnatBeresnev: Yeah, if it's easy to deduce meaning from the name, perhaps it's too much, you're right :+1: 

I'd say the only one which might raise questions is `article.ftl` (if it doesn't get renamed) - maybe it's worth to mention that it contains the main content of the page.

Anyhow, it's up to you -- I think listing all templates by name/path is the main thing

IgnatBeresnev (approved):

PR #2448, opened 2022-04-18
Update Jsoup to 1.14.3

IgnatBeresnev:
Maven central is giving a big warning about a [vulnarability](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-37714)

This ought to fix it


IgnatBeresnev:
JSoup has had some API changes:

* Better support for nullability. They've marked methods with `@Nullable` and `@NotNull`, so build started failing due to compilation error (needed to add some checks for tests). Also, Qodana asked me to remove some excessive null checks
* They've added `element.first()` and `element.last()` methods, which clashses with Kotlin's `collections.first()` and `collections.last()` methods.

plugins/base/base-test-utils/src/main/kotlin/renderers/SignatureUtils.kt
```diff
@@ -11,7 +11,8 @@ fun TestOutputWriter.renderedContent(path: String = "root/example.html"): Elemen
         .single()
 
 fun Element.signature(): Elements = select("div.symbol.monospace")
-fun Element.firstSignature(): Element = signature().first()
+fun Element.firstSignature(): Element = signature().first() ?: throw IllegalStateException("No signature found")
```
vmishenev, line 14: I suppose `NoSuchElementException` is better. See e.g.  `.first()` from stdlib

plugins/base/src/test/kotlin/resourceLinks/ResourceLinksTest.kt
```diff
@@ -184,7 +184,7 @@ class ResourceLinksTest : BaseAbstractTest() {
                     assertNull(writerPlugin.writer.contents["scripts/relativePath.js"])
                     assertNull(writerPlugin.writer.contents["styles/relativePath.js"])
                     Jsoup
-                        .parse(writerPlugin.writer.contents["root/example.html"])
+                        .parse(writerPlugin.writer.contents.getValue("root/example.html"))
```
vmishenev, line 187: it does not matter but above there is `[]`

vmishenev (approved):

plugins/base/base-test-utils/src/main/kotlin/renderers/SignatureUtils.kt
IgnatBeresnev: Indeed, will fix

plugins/base/src/test/kotlin/resourceLinks/ResourceLinksTest.kt
IgnatBeresnev: Yeah, but it doesn't lead to an error. On the contrary, it's quite logical. `map[key]` may return null, so above it's evaluated to `assertNull(null)`

Whereas `Jsoup.parse` accepts only `@NotNull`, thus `.getValue()`, which throws an error

PR #2449, opened 2022-04-18
Make a default value dependent on source set

vmishenev:
In StdLib there are many cases when arguments of actual functions have default values.
See https://github.com/JetBrains/kotlin/search?q=ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS

### DIFF
--- a/core/src/main/kotlin/model/defaultValues.kt
+++ b/core/src/main/kotlin/model/defaultValues.kt
@@ -3,9 +3,11 @@ package org.jetbrains.dokka.model
 import org.jetbrains.dokka.model.properties.ExtraProperty
 import org.jetbrains.dokka.model.properties.MergeStrategy
 
-class DefaultValue(val value: Expression): ExtraProperty<Documentable> {
+class DefaultValue(val value: SourceSetDependent<Expression>): ExtraProperty<Documentable> {
     companion object : ExtraProperty.Key<Documentable, DefaultValue> {
-        override fun mergeStrategyFor(left: DefaultValue, right: DefaultValue): MergeStrategy<Documentable> = MergeStrategy.Remove // TODO pass a logger somehow and log this
+        override fun mergeStrategyFor(left: DefaultValue, right: DefaultValue): MergeStrategy<Documentable> =
+            MergeStrategy.Replace(DefaultValue(left.value + right.value))
+
     }
 
     override val key: ExtraProperty.Key<Documentable, *>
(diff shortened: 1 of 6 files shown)


vmishenev:
Expect properties cannot be initialized. So a default value has to dependent on source set.

core/src/main/kotlin/model/defaultValues.kt
```diff
@@ -3,9 +3,15 @@ package org.jetbrains.dokka.model
 import org.jetbrains.dokka.model.properties.ExtraProperty
 import org.jetbrains.dokka.model.properties.MergeStrategy
 
-class DefaultValue(val value: Expression): ExtraProperty<Documentable> {
+class DefaultValue(val expression: SourceSetDependent<Expression>): ExtraProperty<Documentable> {
+
+    @Deprecated("Use `expression` property that depends on source set", ReplaceWith("expression.values.first()"))
```
IgnatBeresnev, line 8: Can you actually try out the replacement suggestion in IDEA? I think it should be with `this`, otherwise if you apply the fix the statement won't compile

There was such a similar problem with replacement for `documentable` vs `documentables`, so maybe it's the same here
```suggestion
    @Deprecated("Use `expression` property that depends on source set", ReplaceWith("this.expression.values.first()"))
```

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -117,6 +120,19 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
...
+        sourceSet: DokkaSourceSet
+    ) {
+        // a default value of parameter can be got from expect source set
+        d.extra[DefaultValue]?.expression?.let {
+            it[sourceSet] ?: if (d is DParameter) it[d.expectPresentInSet] else null
+        }
+            ?.let { expr ->
+                operator(" = ")
+                highlightValue(expr)
+            }
```
IgnatBeresnev, line 134: Gap looks strange, just a suggestion
```suggestion
        d.extra[DefaultValue]?.expression
            ?.let { it[sourceSet] ?: if (d is DParameter) it[d.expectPresentInSet] else null }
            ?.let { expr ->
                operator(" = ")
                highlightValue(expr)
            }
```

IgnatBeresnev (approved):

PR #2469, opened 2022-04-25
Fix reversed ordering of enum entries

### LINKED ISSUE #2466: Enum members are listed in backwards order.
**Describe the bug**
I have a Kotlin project that includes several enum classes. There is a single source file [`astronomy.kt`](https://github.com/cosinekitty/astronomy/blob/8696241c9098df43c9d15d0e9785de2344852d54/source/kotlin/src/main/kotlin/io/github/cosinekitty/astronomy/astronomy.kt) that contains all the code and documentation comments. I am using Dokka to generate "GitHub Flavored Markdown" (gfm). When Dokka encounters an enum declaration like this...

```Kotlin
/**
 * Selects whether to correct for atmospheric refraction, and if so, how.
 */
enum class Refraction {
    /**
     * No atmospheric refraction correction (airless).
     */
    None,

    /**
     * Recommended correction for standard atmospheric refraction.
     */
    Normal,

    /**
     * Used only for compatibility testing with JPL Horizons online tool.
     */
    JplHor,
}
```

... it generates markdown with the members listed in reverse order:

```
//[astronomy](../../../index.md)/[io.github.cosinekitty.astronomy](../index.md)/[Refraction](index.md)

# Refraction

[jvm]\
enum [Refraction](index.md) : [Enum](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/index.html)&lt;[Refraction](index.md)&gt; 

Selects whether to correct for atmospheric refraction, and if so, how.

## Entries

| | |
|---|---|
| [JplHor](-jpl-hor/index.md) | [jvm]<br>[JplHor](-jpl-hor/index.md)()<br>Used only for compatibility testing with JPL Horizons online tool. |
| [Normal](-normal/index.md) | [jvm]<br>[Normal](-normal/index.md)()<br>Recommended correction for standard atmospheric refraction. |
| [None](-none/index.md) | [jvm]<br>[None](-none/index.md)()<br>No atmospheric refraction correction (airless). |

## Properties

| Name | Summary |
|---|---|
| [name](../-node-event-kind/-ascending/index.md#-372974862%2FProperties%2F-1216412040) | [jvm]<br>val [name](../-node-event-kind/-ascending/index.md#-372974862%2FProperties%2F-1216412040): [String](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/index.html) |
| [ordinal](../-node-event-kind/-ascending/index.md#-739389684%2FProperties%2F-1216412040) | [jvm]<br>val [ordinal](../-node-event-kind/-ascending/index.md#-739389684%2FProperties%2F-1216412040): [Int](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-int/index.html) |
```

**Expected behaviour**
The enum members should be documented in the same order they appear in the source code. I chose the above example because it is relatively simple, but there are other cases where the order is more problematic. For example, I have an enumerated type for solar system bodies, and the planets are listed backwards, which is confusing for people using my project. Issue #47 might possibly be related?

**Screenshots**
N/A

**To Reproduce**
1. Clone the [Astronomy Engine](https://github.com/cosinekitty/astronomy) repository and change into its directory.
2. `git checkout kotlin`
3. `cd source/kotlin`
4. `./gradlew dokkaGfm`
5. Look at the file `source/kotlin/build/dokka/gfm/astronomy/io.github.cosinekitty.astronomy/-refraction/index.md`. It will list the enum values in the order JplHor, Normal, None. The correct order is None, Normal, JplHor.

**Dokka configuration**
Configuration of dokka used to reproduce the bug:

```Kotlin
plugins {
    java
    kotlin("jvm") version "1.6.10"
    `maven-publish`
    id("org.jetbrains.dokka") version "1.6.10"
}

group = "io.github.cosinekitty"
version = "0.0.1"

repositories {
    mavenCentral()
}

dependencies {
    dokkaHtmlPlugin("org.jetbrains.dokka:kotlin-as-java-plugin:1.6.10")
    val junit5Version = "5.8.2"
    testImplementation("org.junit.jupiter:junit-jupiter-api:$junit5Version")
    testImplementation("org.junit.jupiter:junit-jupiter-params:$junit5Version")
    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:$junit5Version")
    testImplementation(kotlin("test"))
}

tasks.test {
    useJUnitPlatform()
}

configure<JavaPluginExtension> {
    sourceCompatibility = JavaVersion.VERSION_1_8
}

val sourceJar by tasks.creating(Jar::class) {
    dependsOn(tasks["classes"])
    archiveClassifier.set("sources")
    from(sourceSets["main"].allSource)
}

task("fatJar", type = Jar::class) {
    from(configurations.runtimeClasspath.get().map(::zipTree))
    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
    with(tasks.jar.get())
}

publishing {
    publications {
        register("mavenJava", MavenPublication::class) {
            from(components["kotlin"])
            artifact(sourceJar)
        }
    }
}

tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
    kotlinOptions {
        allWarningsAsErrors = true
    }
}

tasks.dokkaGfm.configure {
    dokkaSourceSets {
        named("main") {
            includeNonPublic.set(false)
            reportUndocumented.set(true)
        }
    }
}

```

**Installation**
- Operating system: macOS/Windows/Linux: The bug happens in all 3: macOS, Windows 10, and multiple flavors of Debian Linux.
- Build tool: Gradle v7.4.1
- Dokka version: 1.6.10

**Additional context**
```
$ ./gradlew --version

------------------------------------------------------------
Gradle 7.4.1
------------------------------------------------------------

Build time:   2022-03-09 15:04:47 UTC
Revision:     36dc52588e09b4b72f2010bc07599e0ee0434e2e

Kotlin:       1.5.31
Groovy:       3.0.9
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.13 (JetBrains s.r.o. 11.0.13+7-b1751.25)
OS:           Linux 4.19.0-20-amd64 amd64
```

I am happy to provide whatever help the maintainers could use to reproduce/diagnose this bug. Thank you!

IgnatBeresnev:
Fixes #2466

Before
* ordered by name in navigation,
* reversed order on the enum page

Now - same order as in source, both for navigation and the enum page

For class
```kotlin
/**
 * Selects whether to correct for atmospheric refraction, and if so, how.
 */
enum class Refraction(s: String = "EMPTY") {
    EMPTY,
    /**
     * No atmospheric refraction correction (airless).
     */
    None("Hello"),

    /**
     * Recommended correction for standard atmospheric refraction.
     */
    Normal("World"),

    /**
     * Used only for compatibility testing with JPL Horizons online tool.
     */
    JplHor("GitHub"),
}
```

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
@@ -26,15 +26,21 @@ abstract class NavigationDataProvider {
             children = page.navigableChildren()
         )
 
-    private fun ContentPage.navigableChildren(): List<NavigationNode> =
-        when {
-            this !is ClasslikePageNode ->
-                children.filterIsInstance<ContentPage>().map { visit(it) }
-            documentables.any { it is DEnum } ->
-                children.filter { it is WithDocumentables && it.documentables.any { it is DEnumEntry } }
-                    .map { visit(it as ContentPage) }
-            else -> emptyList()
-        }.sortedBy { it.name.toLowerCase() }
+    private fun ContentPage.navigableChildren(): List<NavigationNode> {
+        return if (this !is ClasslikePageNode) {
+            children
+                .filterIsInstance<ContentPage>()
+                .map { visit(it) }
+                .sortedBy { it.name.toLowerCase() }
+        } else if (documentables.any { it is DEnum }) {
+            // no sorting for enum entries, should be the same as in source code
+            children
+                .filter { child -> child is WithDocumentables && child.documentables.any { it is DEnumEntry } }
+                .map { visit(it as ContentPage) }
+        } else {
+            emptyList()
+        }
+    }
 
     /**
      * Parenthesis is applied in 1 case:
--- a/plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
+++ b/plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
@@ -216,7 +216,7 @@ class DocumentableVisibilityFilterTransformer(val context: DokkaContext) : PreMe
         }
 
         private fun filterEnumEntries(entries: List<DEnumEntry>, filteredPlatforms: Set<DokkaSourceSet>): Pair<Boolean, List<DEnumEntry>> =
-            entries.foldRight(Pair(false, emptyList())) { entry, acc ->
+            entries.fold(Pair(false, emptyList())) { acc, entry ->
                 val intersection = filteredPlatforms.intersect(entry.sourceSets)
                 if (intersection.isEmpty()) Pair(true, acc.second)
                 else {
--- a/plugins/base/src/test/kotlin/enums/EnumsTest.kt
+++ b/plugins/base/src/test/kotlin/enums/EnumsTest.kt
@@ -5,9 +5,12 @@ import org.jetbrains.dokka.pages.*
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
 import org.jetbrains.dokka.model.*
 import org.jetbrains.kotlin.utils.addToStdlib.firstIsInstance
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
 import org.junit.jupiter.api.Assertions.*
 import org.junit.jupiter.api.Test
 import signatures.renderedContent
+import utils.TestOutputWriter
 import utils.TestOutputWriterPlugin
 
 class EnumsTest : BaseAbstractTest() {
@@ -159,7 +162,7 @@ class EnumsTest : BaseAbstractTest() {
             configuration,
             pluginOverrides = listOf(writerPlugin)
         ) {
-            renderingStage = { rootPage, context ->
+            renderingStage = { _, _ ->
                 val enumEntriesOnPage = writerPlugin.writer.renderedContent("root/testpackage/-test-enum/index.html")
                     .select("div.table[data-togglable=Entries]")
                     .select("div.table-row")
@@ -183,6 +186,58 @@ class EnumsTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Test
+    fun `should preserve enum source ordering for navigation menu`() {
+        val configuration = dokkaConfiguration {
+            sourceSets {
+                sourceSet {
+                    sourceRoots = listOf("src/")
+                }
+            }
+        }
+
+        val writerPlugin = TestOutputWriterPlugin()
+
+        testInline(
+            """
+            |/src/main/kotlin/basic/Test.kt
+            |package testpackage
+            |
+            |enum class TestEnum {
+            |   E1,
+            |   E2,
+            |   E3,
+            |   E4,
+            |   E5,
+            |   E6,
+            |   E7,
+            |   E8,
+            |   E9,
+            |   E10
+            |}
+        """.trimMargin(),
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
+                val sideMenu = writerPlugin.writer.navigationHtml().select("div.sideMenuPart")
+
+                assertEquals("E1", sideMenu.select("#root-nav-submenu-0-0-0").text())
+                assertEquals("E2", sideMenu.select("#root-nav-submenu-0-0-1").text())
+                assertEquals("E3", sideMenu.select("#root-nav-submenu-0-0-2").text())
+                assertEquals("E4", sideMenu.select("#root-nav-submenu-0-0-3").text())
+                assertEquals("E5", sideMenu.select("#root-nav-submenu-0-0-4").text())
+                assertEquals("E6", sideMenu.select("#root-nav-submenu-0-0-5").text())
+                assertEquals("E7", sideMenu.select("#root-nav-submenu-0-0-6").text())
+                assertEquals("E8", sideMenu.select("#root-nav-submenu-0-0-7").text())
+                assertEquals("E9", sideMenu.select("#root-nav-submenu-0-0-8").text())
+                assertEquals("E10", sideMenu.select("#root-nav-submenu-0-0-9").text())
+            }
+        }
+    }
+
+    fun TestOutputWriter.navigationHtml(): Element = contents.getValue("navigation.html").let { Jsoup.parse(it) }
 
     @Test
     fun `should handle companion object within enum`() {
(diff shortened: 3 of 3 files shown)


plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -26,15 +26,21 @@ abstract class NavigationDataProvider {
...
+        return if (this !is ClasslikePageNode) {
+            children
+                .filterIsInstance<ContentPage>()
+                .map { visit(it) }
+                .sortedBy { it.name.toLowerCase() }
+        } else if (documentables.any { it is DEnum }) {
+            // no sorting for enum entries, should be the same as in source code
+            children
+                .filter { child -> child is WithDocumentables && child.documentables.any { it is DEnumEntry } }
+                .map { visit(it as ContentPage) }
```
IgnatBeresnev, line 39: apparently we don't have any unit tests for navigation, only very basic, half-fake integration tests which are difficult to extend

I wonder if it's at all possible to write unit tests for navigation sidemenu, given how it's generated?

plugins/base/src/main/kotlin/transformers/documentables/DocumentableVisibilityFilterTransformer.kt
```diff
@@ -216,7 +216,7 @@ class DocumentableVisibilityFilterTransformer(val context: DokkaContext) : PreMe
         }
 
         private fun filterEnumEntries(entries: List<DEnumEntry>, filteredPlatforms: Set<DokkaSourceSet>): Pair<Boolean, List<DEnumEntry>> =
-            entries.foldRight(Pair(false, emptyList())) { entry, acc ->
+            entries.fold(Pair(false, emptyList())) { acc, entry ->
```
IgnatBeresnev, line 219: This caused elements to be reversed

plugins/base/src/test/kotlin/enums/EnumsTest.kt
```diff
@@ -98,11 +288,11 @@ class EnumsTest : BaseAbstractTest() {
 
         testInline(
             """
-            |/src/main/kotlin/basic/Test.kt
-            |package enums
+            |/src/main/kotlin/basic/TestEnum.kt
+            |package testpackage
```
IgnatBeresnev, line 292: `Test` and `enums` were rather generic and caused a bit of confusion. Decided to rename it

plugins/base/src/test/kotlin/enums/EnumsTest.kt
```diff
@@ -20,29 +25,55 @@ class EnumsTest : BaseAbstractTest() {
...
             """
             |/src/main/kotlin/basic/Test.kt
-            |package enums
+            |package testpackage
             |
-            |enum class Test {
+            |enum class TestEnum {
             |   E1,
-            |   E2
+            |   E2,
```
vmishenev, line 37: Can you make the order no alphabetical?

IgnatBeresnev: Good point, will do

plugins/base/src/test/kotlin/enums/EnumsTest.kt
```diff
@@ -20,29 +25,112 @@ class EnumsTest : BaseAbstractTest() {
             }
         }
 
+        val writerPlugin = TestOutputWriterPlugin()
+
         testInline(
             """
             |/src/main/kotlin/basic/Test.kt
-            |package enums
+            |package testpackage
```
vmishenev, line 33: Maybe it would be convenient to have a help function like
```kt
    @Test
    fun testEnum(code: String, block: T.() -> Unit) {
            sourceSets {
                sourceSet {
                    sourceRoots = listOf("src/")
                }
            }
        }

        val writerPlugin = TestOutputWriterPlugin()

        testInline(
            """
            |/src/main/kotlin/basic/Test.kt
            |package testpackage
            |$code
            """.trimMargin(),
            configuration,
            pluginOverrides = listOf(writerPlugin),
            block
        ) 
    }
```

vmishenev (approved):

plugins/base/src/test/kotlin/enums/EnumsTest.kt
IgnatBeresnev: Indeed. Enum tests generally need some refactoring and more test cases, I don't want to get too much into it in this PR.

I'm sure we'll get to improving enum support in the future, so let's refactor it then

plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt
```diff
@@ -26,15 +26,21 @@ abstract class NavigationDataProvider {
...
-            else -> emptyList()
-        }.sortedBy { it.name.toLowerCase() }
+    private fun ContentPage.navigableChildren(): List<NavigationNode> {
+        return if (this !is ClasslikePageNode) {
+            children
+                .filterIsInstance<ContentPage>()
+                .map { visit(it) }
+                .sortedBy { it.name.toLowerCase() }
+        } else if (documentables.any { it is DEnum }) {
+            // no sorting for enum entries, should be the same as in source code
```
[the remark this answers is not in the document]
IgnatBeresnev: Do you want to discuss it first or can I merge the PR already? :) I'll revert it if we decide it should be alphabetical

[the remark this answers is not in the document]
vmishenev: up to you)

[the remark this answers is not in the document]
IgnatBeresnev: I'll merge it then because another PR (related to spaces) depends on it, I'll need to rebase on these changes before merging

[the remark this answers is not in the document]
cosinekitty: > I am not sure we should have a different sorting for enum entries in a tree. We can discuss it on our meetings.

Hello. As the person who reported #2466, I wanted to make a case that enum members should always be listed in the order they appear in the source code, not alphabetical order. The order of enumerated type members often reflects important semantics about the problem domain. In my case, it is important to me that the planets are listed in order of distance from the Sun. Sorting them by their English names will confuse/distract people reading my documentation. In a sense, it makes their English names more important than their natural order, which is unhelpful for non-English speakers. I would also suggest that any developer who wants their enumerated types in alphabetical order can simply write their code with that order. If there is no consensus about this, I would hope I could at least have some kind of `@preserveOrder` tag I can put in the comments above my enum class. Thank you for your consideration in this.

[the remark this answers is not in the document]
IgnatBeresnev: Yeah, I actually totally agree with everything up above and wanted to convey the same message.

Another point is that if you have a bunch of classes up in the air - it's unclear how to sort them in a single list _other than_ alphabetically. This is not the case with enum entries though, which have strict linear ordering, some people even write code around `ordinal` property (which is bad, but still). So it only makes sense to preserve the ordering.

PR #2471, opened 2022-04-26
Fix `publishPlugin` task dependency on `dokkaPublish`

IgnatBeresnev:
It tries to find `publishPlugin` task for every subproject otherwise and fails

Should work now, at least it fails like I expect it to :) Will test next release :sweat_smile:

### DIFF
--- a/buildSrc/src/main/kotlin/org/jetbrains/publication.kt
+++ b/buildSrc/src/main/kotlin/org/jetbrains/publication.kt
@@ -84,10 +84,6 @@ fun Project.createDokkaPublishTaskIfNecessary() {
         if (publicationChannels.any { it.isMavenRepository() }) {
             dependsOn(tasks.named("publishToSonatype"))
         }
-
-        if (publicationChannels.any { it.isGradlePluginPortal() }) {
-            dependsOn(tasks.named("publishPlugins"))
-        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/org/jetbrains/publication.kt
```diff
@@ -84,10 +84,6 @@ fun Project.createDokkaPublishTaskIfNecessary() {
         if (publicationChannels.any { it.isMavenRepository() }) {
             dependsOn(tasks.named("publishToSonatype"))
         }
-
-        if (publicationChannels.any { it.isGradlePluginPortal() }) {
```
vmishenev, line 88: I would prefer having such checks  in one place. I suppose you can check the match of project or the existence of task here.

IgnatBeresnev: I agree with you, it would be nice to have a single place for publishing things, but it seems like it's a different case altogether, even though it's all related to publishing. 

`publishPlugins` is only relevant for the `gradle-plugin`, whereas `createDokkaPublishTaskIfNecessary` is called for other 20 plugins. I don't think there will ever be another module that will need this check, so having this here seems excessive and somewhat counter-intuitive

I don't think it's a huge problem since you can find it by the usage of `publicationChannels`, which is already [spread out a little](https://github.com/Kotlin/dokka/blob/52d066c10a50b0e34f1c390c965a89ae403e07a9/build.gradle.kts#L82)

This PR can certainly wait, so we can discuss it in person during a meeting if you still want to have it in this single place

Issue #2187, opened 2021-10-14
Inherited function type resolution is flaky


owengray-google:
(this debugging regards `kotlinSubclassFunction` as inherited from `KotlinSubclass` by `JavaLeafClass : KotlinSubclass`)
<img width="670" alt="Screen Shot 2021-10-14 at 9 34 55 AM" src="https://user-images.githubusercontent.com/56644404/137334086-8bf91854-8c84-4d16-998f-116a0582d5e0.png">

Sometimes, the type of `kotlinSubclassFunction` in `JavaLeafClass : KotlinSubclass` does not resolve and becomes `null`. This has been observed (so far exclusively, but I have suspicions some other cases are related) in java types inheriting Kotlin class functions that don't have explicit return types (i.e. use `fun foo() = "string"`).

**Expected behaviour**
The return type of `kotlinSubclassFunction` should be `String`, not `null`, as it is for kotlin types inheriting such things.
Our source is here:
https://cs.android.com/androidx/platform/tools/dokka-devsite-plugin/+/master:testData/inheritance/source/

Dokka version 1.6.0-dev-122

As this is flaky and seems likely to be a concurrency issue, maybe this is related to https://github.com/Kotlin/dokka/issues/2052?

vmishenev:
I have opened an issue for IDE team:
https://youtrack.jetbrains.com/issue/KT-50111
But it can be fixed not soon.

IgnatBeresnev:
[KT-50111](https://youtrack.jetbrains.com/issue/KT-50111) has been fixed and I believe it should've been included into 1.6.20.

Closing this for now, but do let us know if this problem occurs again.

PR #2474, opened 2022-04-27
Retain source language information in Documentable tree

owengray-google:
This will allow dackka to figure out if a DParameter is from Kotlin or from Java
Which is necessary to determine if a bare parameter is non-null K or platform J

Test: added assertions to translator tests and the multilanguage linker test

### DIFF
--- a/core/src/main/kotlin/model/documentableProperties.kt
+++ b/core/src/main/kotlin/model/documentableProperties.kt
@@ -46,3 +46,17 @@ data class CheckedExceptions(val exceptions: SourceSetDependent<List<DRI>>) : Ex
     }
     override val key: ExtraProperty.Key<Documentable, *> = CheckedExceptions
 }
+
+enum class Language {
+    JAVA, KOTLIN, UNKNOWN
+}
+
+data class SourceLanguage(val sourceLanguage: Language) : ExtraProperty<HasSourceLanguage> {
+    companion object : ExtraProperty.Key<HasSourceLanguage, SourceLanguage> {
+        override fun mergeStrategyFor(left: SourceLanguage, right: SourceLanguage) = MergeStrategy.Replace(
+            if (left.sourceLanguage == right.sourceLanguage) SourceLanguage(left.sourceLanguage)
+            else SourceLanguage(Language.UNKNOWN)
+        )
+    }
+    override val key: ExtraProperty.Key<HasSourceLanguage, *> = SourceLanguage
+}
--- a/plugins/all-modules-page/out/index.md
+++ b/plugins/all-modules-page/out/index.md
@@ -1,0 +1,10 @@
+/
+
+# Sample project
+
+Sample documentation with [external link](https://www.google.pl)
+
+## All modules:
+
+| Name |
+|---|
--- a/plugins/base/src/main/kotlin/translators/documentables/documentableLanguage.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/documentableLanguage.kt
@@ -2,14 +2,11 @@ package org.jetbrains.dokka.base.translators.documentables
 
 import org.jetbrains.dokka.DokkaConfiguration
 import org.jetbrains.dokka.analysis.PsiDocumentableSource
+import org.jetbrains.dokka.model.Language
 import org.jetbrains.dokka.model.WithSources
 
-internal enum class DocumentableLanguage {
-    JAVA, KOTLIN
-}
-
-internal fun WithSources.documentableLanguage(sourceSet: DokkaConfiguration.DokkaSourceSet): DocumentableLanguage =
+internal fun WithSources.documentableLanguage(sourceSet: DokkaConfiguration.DokkaSourceSet): Language =
     when (sources[sourceSet]) {
-        is PsiDocumentableSource -> DocumentableLanguage.JAVA
-        else -> DocumentableLanguage.KOTLIN
+        is PsiDocumentableSource -> Language.JAVA
+        else -> Language.KOTLIN
     }
--- a/plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
+++ b/plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
@@ -3,10 +3,8 @@ package translators
 import org.jetbrains.dokka.base.DokkaBase
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
 import org.jetbrains.dokka.links.DRI
-import org.jetbrains.dokka.model.Annotations
-import org.jetbrains.dokka.model.TypeConstructor
+import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.doc.Text
-import org.jetbrains.dokka.model.firstMemberOfType
 import org.jetbrains.dokka.plugability.DokkaPlugin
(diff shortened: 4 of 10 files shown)


owengray-google:
Still TODO is creating a kotlin-inheriting-from-Java test and verifying that `DescriptorToDocumentableTranslator` should be automatically deciding on Kotlin.

core/src/main/kotlin/model/documentableProperties.kt
```diff
@@ -46,3 +46,17 @@ data class CheckedExceptions(val exceptions: SourceSetDependent<List<DRI>>) : Ex
     }
     override val key: ExtraProperty.Key<Documentable, *> = CheckedExceptions
 }
+
+enum class Language {
+    JAVA, KOTLIN, UNKNOWN
+}
```
IgnatBeresnev, line 52: What do you think about renaming it to something like `SourceLanguage` (preferred) or `DocumentableLanguage` as before? 

My concern is that this enum will be used by someone as a feature/language toggle somewhere on a higher level, and I'd like to try to narrow down the scope of it's usage just to the source/documentable level

If it clashes with `data class SourceLanguage`, it can be renamed to `SourceLanguageExtra`

plugins/all-modules-page/out/index.md
```diff
@@ -0,0 +1,10 @@
+/
+
+# Sample project
+
+Sample documentation with [external link](https://www.google.pl)
+
+## All modules:
+
+| Name |
+|---|
```
IgnatBeresnev, line 10: This keeps appearing after building dokka, very annoying, we need to fix this :) Can you remove it from the PR?

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -240,12 +238,17 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
             |}
         """.trimMargin(),
             configurationWithNoJVM,
-            pluginOverrides = listOf(OnlyPsiPlugin()) // suppress a descriptor translator because of psi and descriptor translators work in parallel
+            //pluginOverrides = listOf(OnlyPsiPlugin()) // suppress a descriptor translator because of psi and descriptor translators work in parallel
```
IgnatBeresnev, line 241: Is this intentional or a leftover? If intentional, we might have a problem -- this is a recent test from https://github.com/Kotlin/dokka/pull/2400 that checks resolution of ultralight classes specifically, and from what I understand this plugin is needed to avoid conurrency issues

IgnatBeresnev (commented):
Overall LGTM, thanks! I've only got some minor comments

1. Can you run `./gradlew apiDump` and commit generated changes? Otherwise it fails the build
2. I see that you've added language-related asserts to existing tests. While that's good, having a separate test suite to check for this particular behaviour would be much better imo. Ideally, in a separate class `SourceLanguageResolutionTest` (or something like this), with simple checks (java, kotlin, groovy for unknown) and some corner cases like inheritance, maybe an extension function on a java type, etc. Just as a sanity check and so that the problem is blatant in case it does break

plugins/base/src/main/kotlin/translators/documentables/documentableLanguage.kt
```diff
@@ -2,14 +2,11 @@ package org.jetbrains.dokka.base.translators.documentables
...
 import org.jetbrains.dokka.analysis.PsiDocumentableSource
+import org.jetbrains.dokka.model.Language
 import org.jetbrains.dokka.model.WithSources
 
-internal enum class DocumentableLanguage {
-    JAVA, KOTLIN
-}
-
-internal fun WithSources.documentableLanguage(sourceSet: DokkaConfiguration.DokkaSourceSet): DocumentableLanguage =
+internal fun WithSources.documentableLanguage(sourceSet: DokkaConfiguration.DokkaSourceSet): Language =
```
vmishenev, line 8: Why do you not use this function to determine a source language?
It works with `Classlikes`, `DFunction`, `DProperty`.

owengray-google: Yes, but it does not work on DParameter, which is where we have been having issues. Specifically, with the receiver types of extension functions.

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
owengray-google: This is intentional. Without this, the DPackage does not contain the KotlinSubClass.

Concurrency issues sound problematic--wouldn't these two plugins both run under most conditions?

plugins/base/src/main/kotlin/translators/documentables/documentableLanguage.kt
owengray-google: Actually, `withSources` does something similar to what I am doing here: `descriptor.toSourceElement.containingFile.toString()`
Making `DParameter` `withSources` might be a cleaner solution. I'll take a look.

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
IgnatBeresnev: I might be wrong on the concurrency side of things, @vmishenev knows best

vmishenev: This test reproduces a fixed  bug #2187 that appeared under specific conditions. If it does not suppress a descriptor translator and is broken, it will be lead to the flaky test.

I suppose you can copy the test for your goals without the plugin `OnlyPsiPlugin`.

plugins/base/src/main/kotlin/translators/documentables/documentableLanguage.kt
IgnatBeresnev: Arguably, other plugin developers might need something similar, and having a `HasSourceLanguage` marker interface and some way to easily get the source language (like an extra) would be better than some internal function on `WithSources` that you don't know about

I would consider this to be an API improvement more than solving a specific problem for Dackka :)

core/src/main/kotlin/model/documentableProperties.kt
owengray-google: This could still be done, though there might also be a pre-existing alternative that is more comprehensive, perhaps in the compiler.

owengray-google:
Still has TODOs, especially adding dedicated tests.

owengray-google:
I have been looking around, and KSP [is doing something similar](https://github.com/google/ksp/blob/main/compiler-plugin/src/main/kotlin/com/google/devtools/ksp/symbol/impl/binary/KSPropertyAccessorDescriptorImpl.kt) to what I was trying here with their `origin` and `location` attributes.
They have a lot of code and difficulty doing this too.

I've landed a mitigation downstream for us, which determines source location and language in the cases of parameters defined in or inherited from source, but not anything from compiled code, which is the trickiest case.

What I have in this PR covers most compiled code cases, but there are a lot of edge cases and I kept running into more. If you were to choose to try to get this implemented fully and correctly in dokka (thereby allowing determination of nonobvious default-nullability cases), this might be of help.

I will be speaking with people on the Google side to see if we and KSP can ask for handling of this to get into the compiler itself.

PR #2475, opened 2022-04-27
Fix suddenly appearing all-modules-page/index.md file

IgnatBeresnev:
This annoying file used to appear whenever you ran tests

https://github.com/Kotlin/dokka/pull/2474/files/cf02a243237306930a7f6d149c43126699b33be6#diff-5b3183ca422abbdc8f2db7e6fd54b5b7d18a40d54c543db3e47b714698a9dc26

### DIFF
--- a/core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
+++ b/core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
@@ -28,6 +28,10 @@ abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : DokkaTestGe
         File("src/test/resources/$name").takeIf { it.exists() }?.toPath()
             ?: throw InvalidPathException(name, "Cannot be found")
 
+    /**
+     * @param useOutputLocationFromConfig if set to true, output location specified in [DokkaConfigurationImpl.outputDir]
+     *                                    will be used. If set to false, a temporary folder will be used instead.
+     */
     protected fun testFromData(
         configuration: DokkaConfigurationImpl,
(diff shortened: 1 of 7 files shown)


plugins/all-modules-page/src/test/kotlin/templates/MultiModuleDocumentationTest.kt
```diff
@@ -39,7 +39,7 @@ class MultiModuleDocumentationTest : MultiModuleAbstractTest() {
             includes = listOf(folder.root.resolve("README.md"))
         }
 
-        testFromData(configuration, preserveOutputLocation = true) {
+        testFromData(configuration) {
```
IgnatBeresnev, line 42: That's the culprit, because the flag was set to `true` - it used the output location from configuration, which is the `out` folder by default

core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
```diff
@@ -31,19 +31,21 @@ abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : DokkaTestGe
     protected fun testFromData(
         configuration: DokkaConfigurationImpl,
         cleanupOutput: Boolean = true,
-        preserveOutputLocation: Boolean = false,
+        useOutputLocationFromConfig: Boolean = false,
```
IgnatBeresnev, line 34: Having `cleanupOutput` and `preserveOutput` in one method is rather confusing as these are the two opposites.

Hopefully the new name represents the behaviour a little bit better

vmishenev (approved):

PR #2476, opened 2022-04-27
Move source links into signature

vmishenev:
image

BUT
image

### DIFF
--- a/core/src/main/kotlin/pages/ContentNodes.kt
+++ b/core/src/main/kotlin/pages/ContentNodes.kt
@@ -384,7 +384,7 @@ enum class TokenStyle : Style {
 enum class TextStyle : Style {
     Bold, Italic, Strong, Strikethrough, Paragraph,
     Block, Span, Monospace, Indented, Cover, UnderCoverText, BreakableAfter, Breakable, InlineComment, Quotation,
-    RightAligned
+    FloatingRight
 }
 
 enum class ContentStyle : Style {
--- a/core/src/main/kotlin/pages/ContentNodes.kt
+++ b/core/src/main/kotlin/pages/ContentNodes.kt
@@ -383,7 +383,8 @@ enum class TokenStyle : Style {
 
 enum class TextStyle : Style {
     Bold, Italic, Strong, Strikethrough, Paragraph,
-    Block, Span, Monospace, Indented, Cover, UnderCoverText, BreakableAfter, Breakable, InlineComment, Quotation
+    Block, Span, Monospace, Indented, Cover, UnderCoverText, BreakableAfter, Breakable, InlineComment, Quotation,
+    RightAligned
 }
 
 enum class ContentStyle : Style {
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/pages/ContentNodes.kt
```diff
@@ -383,7 +383,8 @@ enum class TokenStyle : Style {
 
 enum class TextStyle : Style {
     Bold, Italic, Strong, Strikethrough, Paragraph,
-    Block, Span, Monospace, Indented, Cover, UnderCoverText, BreakableAfter, Breakable, InlineComment, Quotation
+    Block, Span, Monospace, Indented, Cover, UnderCoverText, BreakableAfter, Breakable, InlineComment, Quotation,
+    RightAligned
```
IgnatBeresnev, line 387: I think it's better to put the common word first, i.e `AlignedRight`, so that when you type `Aligned`, you'll see more options in autocomplete, like `AlignedLeft`, `AlignedCenter`, etc

plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
```diff
@@ -35,15 +34,18 @@ class SourceLinksTransformer(val context: DokkaContext) : PageTransformer {
         if (sourceLinks.isEmpty()) {
             return input
         }
-        return input.transformContentPagesTree { node ->
+        context.logger.warn("sourceLinks.size " + sourceLinks.size)
```
IgnatBeresnev, line 37: probably a leftover

plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -1181,3 +1181,13 @@ div.runnablesample {
...
+.clearfix::after {
+    content: ' ';
+    clear: both;
+    display: block;
+    height: 0;
+}
+
+.right-aligned {
+    float: right;
+}
```
IgnatBeresnev, line 1193: It's a bit strange that `right-aligned` class sets `float`. I'd expect `TextStyles.RightAligned` to add `align-text: right`

Any particular reason why it shouldn't be called `FloatingRight`?

plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
```diff
@@ -152,6 +94,40 @@ class SourceLinksTransformer(val context: DokkaContext) : PageTransformer {
...
         return doc?.getLineNumber(textRange.startOffset)?.plus(1)
     }
+
+    private fun ContentNode.signatureGroupOrNull() =
+        (this as? ContentGroup)?.takeIf { it.dci.kind == ContentKind.Symbol }
+
+    private fun transformContent(
+        contentNode: ContentNode, sources: Map<DRI, List<Pair<DokkaSourceSet, String>>>
+    ): ContentNode =
+        contentNode.signatureGroupOrNull()?.let { cg ->
```
IgnatBeresnev, line 104: `signatureGroup` -> `sg` instead of `cg`?

IgnatBeresnev (commented):
Is it possible to add a simple rendering test with multiple signatures on the page (function overload) and assert that each signature has its own source button with the correct link, to avoid regressions later on?

PR #2485, opened 2022-05-03
dokkaGfm supports code blocks, inline code.

### LINKED ISSUE #2477: dokkaHtml preserves teletype formatting, but dokkaGfm unhelpfully strips it
**Describe the bug**
In my Kotlin comments, I often refer to a parameter name or other code symbol using backquotes, as is typical in Markdown to render teletype/monospace text. For example, my remarks about `eclipticSeparation` in:

```kotlin
/**
 * Determines visibility of a celestial body relative to the Sun, as seen from the Earth.
 *
 * This function returns an [ElongationInfo] object, which provides the following
 * information about the given celestial body at the given time:
 *
 * - `visibility` is an enumerated type that specifies whether the body is more easily seen
 *    in the morning before sunrise, or in the evening after sunset.
 *
 * - `elongation` is the angle in degrees between two vectors: one from the center of the Earth to the
 *    center of the Sun, the other from the center of the Earth to the center of the specified body.
 *    This angle indicates how far away the body is from the glare of the Sun.
 *    The elongation angle is always in the range [0, 180].
 *
 * - `eclipticSeparation` is the absolute value of the difference between the body's ecliptic longitude
 *   and the Sun's ecliptic longitude, both as seen from the center of the Earth. This angle measures
 *   around the plane of the Earth's orbit, and ignores how far above or below that plane the body is.
 *   The ecliptic separation is measured in degrees and is always in the range [0, 180].
 *
 * @param body
 * The celestial body whose visibility is to be calculated.
 *
 * @param time
 * The date and time of the observation.
 */
fun elongation(body: Body, time: Time): ElongationInfo {
    val relativeLongitude = pairLongitude(body, Body.Sun, time)
    val elongation = angleFromSun(body, time)
    return if (relativeLongitude > 180.0)
        ElongationInfo(time, Visibility.Morning, elongation, 360.0 - relativeLongitude)
    else
        ElongationInfo(time, Visibility.Evening, elongation, relativeLongitude)
}
```

dokkaHtml respects this notation and converts this into

```html
<code class="lang-kotlin">eclipticSeparation</code> is the absolute value of ...
```

This is the desired behavior. However, dokkaGfm strips out the backquotes and converts into plain text:

```markdown
- 
   eclipticSeparation is the absolute value of ...
```

**Expected behaviour**
Backquotes should be left intact in dokkaGfm's markdown output, so that it is clear which parts of the text are prose, and which are source code excerpts. For example:

```markdown
-  `eclipticSeparation` is the absolute value of ...
```

**Screenshots**
N/A

**To Reproduce**
1. Clone the [Astronomy Engine](https://github.com/cosinekitty/astronomy) repository and change into its directory.
2. `git checkout -b backquote 53f5540f09df94baae112c56b9fb5066175d8a0f`
3. `cd source/kotlin`
4. `./gradlew dokkaGfm dokkaHtml`
5. Edit the file `source/kotlin/build/dokka/gfm/astronomy/io.github.cosinekitty.astronomy/elongation.md`. Notice that the backquotes have been removed from the documented symbols "visibility", "elongation", and "eclipticSeparation".
6. However, dokkaHtml's output preserves the intended formatting. See the corresponding file `source/kotlin/build/dokka/html/astronomy/io.github.cosinekitty.astronomy/-astronomy/elongation.html`. This suggests the bug is specific to dokkaGfm.

**Dokka configuration**
(See repro steps above. The configuration is present in the directory `source/kotlin`.)

**Installation**
- Operating system: Happens in all 3: macOS/Windows/Linux.
- Build tool: Gradle v7.4.1
- Dokka version: 1.6.10

**Additional context**
```
$ ./gradlew --version

------------------------------------------------------------
Gradle 7.4.1
------------------------------------------------------------

Build time:   2022-03-09 15:04:47 UTC
Revision:     36dc52588e09b4b72f2010bc07599e0ee0434e2e

Kotlin:       1.5.31
Groovy:       3.0.9
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.13 (JetBrains s.r.o. 11.0.13+7-b1751.25)
OS:           Linux 4.19.0-20-amd64 amd64
```

I am happy to provide whatever help the maintainers could use to reproduce/diagnose this bug. Thank you!

cosinekitty:
Fixes #2477.

Inline code, text that is nested within a pair of backquotes,
is now converted into GitHub Flavored Markdown (gfm) without
stripping out the backquotes. For example:

    The parameter `sum` must be a non-negative real number.

Code blocks, which are any number of lines of literal text
between triple-backquotes, and an optional programming language
name, are now preserved. If absent, the programming language
is assumed to be "kotlin". This follows the behavior of the
html renderer. For example:

    Here is an example of calling the function:
    ```kotlin
    val sum = addThemUp(left, right)
    ```


vmishenev (approved):

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
```diff
@@ -306,6 +306,22 @@ open class CommonmarkRenderer(
...
     }
 
+    override fun StringBuilder.buildCodeBlock(code: ContentCodeBlock, pageContext: ContentPage) {
+        append("```")
+        append(code.language.ifEmpty { "kotlin" })
+        buildNewLine()
+        code.children.forEach { it.build(this, pageContext) }
+        buildNewLine()
+        append("```")
+        buildNewLine()
```
IgnatBeresnev, line 316: I wonder if this is necessary. Is there no newline or does it lead to some rendering bugs if we remove this?

IgnatBeresnev (approved):

plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/renderer/CommonmarkRenderer.kt
cosinekitty: It depends on whether we can count on a newline being added by some higher level code, which I am not familiar enough with Dokka to predict. It is definitely safer to have extraneous newlines than missing newlines. As an example here:
```
Here is a block of literal text between triple backquotes.
``` And here text on the same line as the closing backquotes.
Here is text on the following line.

When the triple backquotes are not followed immediately by a newline, the remainder of the text is rendered as teletype.

**Comment by IgnatBeresnev** [comment]

**Comment by IgnatBeresnev** [comment]
> Thank you for the support and help.

My pleasure :) 

> I am eager to try out the new release of Dokka when it includes this fix (and the other one for enumeration order). Is there a way for me to know when a new Dokka release is available?

Dokka usually releases soon after Kotlin, we try to follow the same major version release schedule. Beta/RC releases depend on whether we actually have changes that are worth releasing.

We'll soon (today/tomorrow) publish a dev build of current master (so including your changes and enum ordering) for internal use. If you want, I can tell you how you can try it out -- it requires adding our dev maven repository to the build scripts and changing the version.

**Comment by cosinekitty** [comment]
> If you want, I can tell you how you can try it out -- it requires adding our dev maven repository to the build scripts and changing the version.

Yes, that would be helpful. I would be able to verify the changes in the "real world" (I could not get the Maven local repository stuff to work in my Astronomy Engine project, no matter how much I tried).

Also, do your dev builds persist over time? I wonder if I could use a specific dev build of Dokka in my GitHub Actions CI process until the official release is available.

**Comment by IgnatBeresnev** [comment]
> Also, do your dev builds persist over time?

Yes, they do, but use it at your own risk: maven repo seems to be pretty reliable and we have good test coverage, so no major bugs should skip in, but still mo guarantees with dev releases :) 

To use it, add this to the repositories:
```kotlin
repositories {
    maven("https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev")
}
```

And use Dokka version `1.7.0-dev-164` - it's the latest master as of now, built with Kotlin 1.6.21.

PR #2496, opened 2022-05-10
Render quotes as blockquotes instead of code blocks.

### LINKED ISSUE #2491: Quotes are rendered as code blocks instead of blockquotes
**Describe the bug**
When I use quotes in KDocs, as such:
```
/** 
* > quote
*/
```

And render it with Dokka (HTML format), the quotes are rendered as code blocks (`<code>` elements with syntax highlighting).

**Expected behaviour**
The quotes should be rendered as such, using the `<blockquote>` element.

**Screenshots**
image
image

**To Reproduce**
1. Add KDocs to a function in your code, and include a quote (`> `) in them.
2. Render the docs using Dokka in the HTML format.
3. The docs should show a syntax-highlighted code block instead of a blockquote.

**Dokka configuration**

```Kotlin
plugins {
  id("org.jetbrains.dokka") version "latest.release"
}

tasks.dokkaHtml.configure {
    outputDirectory.set(projectDir.resolve("docs"))
}
```

**Installation**
- Operating system: Linux
- Build tool: Gradle v7.4.2
- Dokka version: latest (according to gradle's `latest.release` version)

TheOnlyTails:
This is not 100% done yet, I still want to make some changes to the styling and add some tests (as @IgnatBeresnev requested).
Blockquotes are being rendered though.
(this is my first PR to dokka, so please tell me if i'm doing something wrong without realizing it)

```
/**
* > Matches 0 or 1 of the preceding token, effectively making it optional.
*/ 
```
image

Closes #2491

### DIFF
--- a/plugins/base/base-test-utils/src/main/kotlin/renderers/JsoupUtils.kt
+++ b/plugins/base/base-test-utils/src/main/kotlin/renderers/JsoupUtils.kt
@@ -29,6 +29,8 @@ class A(vararg matchers: Any) : Tag("a", *matchers)
 class B(vararg matchers: Any) : Tag("b", *matchers)
 class I(vararg matchers: Any) : Tag("i", *matchers)
 class STRIKE(vararg matchers: Any) : Tag("strike", *matchers)
+
+class BLOCKQUOTE(vararg matchers: Any) : Tag("blockquote", *matchers)
 class Dl(vararg matchers: Any) : Tag("dl", *matchers)
 class Dt(vararg matchers: Any) : Tag("dt", *matchers)
 class Dd(vararg matchers: Any) : Tag("dd", *matchers)
--- a/plugins/base/base-test-utils/src/main/kotlin/renderers/JsoupUtils.kt
+++ b/plugins/base/base-test-utils/src/main/kotlin/renderers/JsoupUtils.kt
@@ -30,7 +30,7 @@ class B(vararg matchers: Any) : Tag("b", *matchers)
 class I(vararg matchers: Any) : Tag("i", *matchers)
 class STRIKE(vararg matchers: Any) : Tag("strike", *matchers)
 
-class BLOCKQUOTE(vararg matchers: Any) : Tag("blockquote", *matchers)
+class BlockQuote(vararg matchers: Any) : Tag("blockquote", *matchers)
 class Dl(vararg matchers: Any) : Tag("dl", *matchers)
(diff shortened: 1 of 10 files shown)


IgnatBeresnev:
Yeah, code-wise looks good so far, good job! Feel free to suggest alternative styling (css) for quotation if you wish, but this will also do

plugins/base/base-test-utils/src/main/kotlin/renderers/JsoupUtils.kt
```diff
@@ -29,6 +29,8 @@ class A(vararg matchers: Any) : Tag("a", *matchers)
 class B(vararg matchers: Any) : Tag("b", *matchers)
 class I(vararg matchers: Any) : Tag("i", *matchers)
 class STRIKE(vararg matchers: Any) : Tag("strike", *matchers)
+
+class BLOCKQUOTE(vararg matchers: Any) : Tag("blockquote", *matchers)
```
IgnatBeresnev, line 33: `STRIKE` seems to be the only one in all caps (not sure why), so I'd stick to the majority
```suggestion
class Blockquote(vararg matchers: Any) : Tag("blockquote", *matchers)
```

plugins/base/src/test/kotlin/renderers/html/TextStylesTest.kt
```diff
@@ -69,6 +69,14 @@ class TextStylesTest : HtmlRenderingOnlyTestBase() {
         files.contents.getValue("test-page.html").contains("&nbsp;<strike><b>styled text</b></strike>")
     }
 
+    @Test
+    fun `should include blockquote`() {
+        val page = testPage {
+            text("blockquote text", styles = setOf(TextStyle.Quotation))
+        }
```
IgnatBeresnev, line 76: I think this test will fail since we only wrap groups and not individual text nodes. I believe it should be like this:
```kotlin
testPage {
    group(styles = setOf(TextStyle.Quotation)) {
        text("blockquote text")
    }
}
```

IgnatBeresnev:
Oh, also, please run `./gradlew apiDump` once again after applying changes, it'll add the blockquote matcher to the exposed API  :)

IgnatBeresnev (approved):
Good job, thank you for the contribution! :)

TheOnlyTails:
thank you very much! You were super helpful!

vmishenev (approved):

PR #2498, opened 2022-05-15
Remove unused `collectorTaskSupported` parameter

3flex:


IgnatBeresnev (approved):
LGTM, thanks for noticing!

I thought maybe it was forgotten to be enabled for javadoc tasks, but it looks like `dokkaJavadocCollector` works just fine with javadocs, so it indeed looks like it's unused

PR #2501, opened 2022-05-15
Remove outdated suppressions in Gradle runner

3flex:


IgnatBeresnev (approved):
Thank you for noticing and taking the time to fix it :)

PR #2502, opened 2022-05-15
Remove unused & redundant dependencies for Gradle runner

3flex:


IgnatBeresnev:
Hi! 

Unfortunately, I'm completely out of context on these dependencies. How did you realize they were unused and redundant? Just want to make sure it doesn't break anything in some corner case

3flex:
There are no Jackson annotations used in the Gradle runner at all, so that's safe to remove.

`gradleApi()` brings in all of the dependencies that are provided by the Gradle API, including kotlin-stlib, kotlin-reflect etc, so there's no need to depend on `kotlin-reflect` separately.

Lastly, `gradleApi()` itself is added to the "api" configuration by the [Gradle Plugin Development Plugin](https://docs.gradle.org/current/userguide/java_gradle_plugin.html). You can see how the other configurations depend on the "api" configuration [here](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph), so that the compile, runtime, testCompile and testRuntime classpaths all derive from "api", and there's no need to add it separately to the `compileOnly` or `testImplementation` configurations.

Hope that makes sense!

IgnatBeresnev (approved):
LGTM, thanks for the contribution and the explanation :)

vmishenev (approved):

runners/gradle-plugin/build.gradle.kts
```diff
@@ -32,13 +28,6 @@ dependencies {
                 .classpath.asFiles.first()
         )
     )
-
-    constraints {
-        val kotlin_version: String by project
-        compileOnly("org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}") {
-            because("kotlin-gradle-plugin and :core both depend on this")
```
vmishenev, line 39: This is a [dependency constraint](https://docs.gradle.org/current/userguide/dependency_constraints.html). 
It allows to choose needed version of `kotlin-reflect` from transitive dependencies. Currently it makes no sense because the latest `kotlin-gradle-plugin` excludes `kotlin-reflect`.

vmishenev (approved):

PR #2504, opened 2022-05-16
Correction of a typo

PaladiRka:


IgnatBeresnev (approved):
Thank you for noticing and taking the time to correct it! :)

PR #2507, opened 2022-05-18
Fix some tests on Windows

3flex:
Unfortunately I still don't have a fully working build on Windows, but this does address some of the issues I've seen.

### DIFF
--- a/integration-tests/cli/src/integrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
+++ b/integration-tests/cli/src/integrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
@@ -267,9 +267,10 @@ class CliIntegrationTest : AbstractCliIntegrationTest() {
     fun `should accept json as input configuration`() {
         val dokkaOutputDir = File(projectDir, "output")
         assertTrue(dokkaOutputDir.mkdirs())
-        val jsonPath = javaClass.getResource("/my-file.json")?.path ?: throw IllegalStateException("No JSON found!")
+        val resourcePath = javaClass.getResource("/my-file.json")?.toURI() ?: throw IllegalStateException("No JSON found!")
+        val jsonPath = File(resourcePath).absolutePath
         PrintWriter(jsonPath).run {
-            write(jsonBuilder(dokkaOutputDir.path, basePluginJarFile.path, File(projectDir, "src").path, reportUndocumented = true))
+            write(jsonBuilder(dokkaOutputDir.invariantSeparatorsPath, basePluginJarFile.invariantSeparatorsPath, File(projectDir, "src").invariantSeparatorsPath, reportUndocumented = true))
             close()
         }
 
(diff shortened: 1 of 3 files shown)


IgnatBeresnev (approved):
Thanks for the fix, we indeed don't test it on windows much, so it's always appreciated :)

Just for the context: am I correct in understanding that previously it resolved to `\\` used as separator on Windows and because of that tests failed? Don't have a windows machine to test it on myself, unfortunately

3flex:
Because directory "C:\Users" is a thing on Windows, when some of the paths were substituted into the generated JSON file, it would contain the pattern `\U` which was treated as some kind of escape sequence and caused Jackson to choke.

I can get the actual error for you?

IgnatBeresnev:
If it's not too difficult for you, that'd be helpful in case it backfires somehow :)

3flex:
There were two errors actually. the first is because there's a `:` character in the file path on Windows:
```
Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/Users/3flex/IdeaProjects/dokka/integration-tests/cli/build/resources/integrationTest/my-file.json
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
	at java.base/java.nio.file.Path.of(Path.java:147)
	at java.base/java.nio.file.Paths.get(Paths.java:69)
	at org.jetbrains.dokka.MainKt.initializeConfiguration(main.kt:411)
	at org.jetbrains.dokka.MainKt.main(main.kt:426)
```
The second is the one I referred to in my previous comment:

```
Exception in thread "main" com.fasterxml.jackson.core.JsonParseException: Unrecognized character escape 'U' (code 85)
 at [Source: (String)"{
  "moduleName": "Dokka Example",
  "moduleVersion": null,
  "outputDir": "C:\Users\3flex\AppData\Local\Temp\junit12364076267926552043\project\output",
  "pluginsClasspath": ["C:\Users\3flex\AppData\Local\Temp\junit12364076267926552043\base-plugin.jar"],
  "cacheRoot": null,
  "offlineMode": false,
  "sourceLinks": [],
  "externalDocumentationLinks": [],
  "perPackageOptions": [],
  "sourceSets": [
    {
      "displayName": "jvm",
      "sourceSetID": {
        "scopeId": ":dokkaHtml",
       "[truncated 602 chars]; line: 4, column: 21]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:2337)
```
Both are fixed in the PR.

IgnatBeresnev:
Great, thanks!

vmishenev (approved):

Issue #2492, opened 2022-05-09
`kotlin-as-java` hides access modifiers


tripolkaandrey:
**Describe the bug**
I have a Java project where I am using Dokka. For the proper representation of the documentation for Java, I am using the `kotlin-as-java` plugin. The produced documentation does not have access modifiers for functions. I have found out that the problem is with the `kotlin-as-java` plugin.  Access modifiers are present when the plugin is not applied.

**Expected behaviour**
Access modifiers should be displayed.

**Screenshots**
Dokka output with `kotlin-as-java` applied.
<img width="1174" alt="with kotlin-as-java" src="https://user-images.githubusercontent.com/54888768/167422695-cc43559e-ee9a-4879-ae55-f8e86a42e74e.png">

Dokka output when `kotlin-as-java` is not applied.
<img width="1174" alt="when `kotlin-as-java` not applied" src="https://user-images.githubusercontent.com/54888768/167424835-ff295215-fb10-4125-91a3-4495505b9638.png">

**To Reproduce**
1) Configure Dokka with the provided below configuration.
2) Create a Java class and methods with different access modifiers.
3) Run `dokkaHtml`.

**Dokka configuration**
``` dokka.gradle.kts
dependencies {
  dokkaPlugin("org.jetbrains.dokka:kotlin-as-java-plugin:1.6.20")
}

tasks.withType<DokkaTask>().configureEach {
    dokkaSourceSets.configureEach {
        documentedVisibilities.set(
            setOf(
                Visibility.PUBLIC,
                Visibility.PROTECTED,
            )
        )
    }
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v7.4.2
- Dokka version: 1.6.20

PR #2510, opened 2022-05-24
Make the `kotlin-as-java` plugin include information about access modifiers for functions

tripolkaandrey:
### Preface
This PR fixes the problem described [here](https://github.com/Kotlin/dokka/issues/2492). Please take a look at the issue before moving on.

### What was the problem?
The `DFunction` successfully collected information about the access modifier, but this information was not considered when generating/rendering output.

### Additional changes
Additionally, a few of the `kotlin-as-java` plugin tests required modification. These tests failed as they expected no access modifiers for public functions in the output.


vmishenev (approved):

IgnatBeresnev (approved):
Well done! Thank you for the fix! :)

PR #2520, opened 2022-06-06
Support definitely non-nullable types

vmishenev:
[definitely non-nullable types](https://github.com/Kotlin/KEEP/blob/c72601cf35c1e95a541bb4b230edb474a6d1d1a8/proposals/definitely-non-nullable-types.md) are enabled since Kotlin 1.7

### DIFF
--- a/plugins/base/src/test/kotlin/signatures/SignatureTest.kt
+++ b/plugins/base/src/test/kotlin/signatures/SignatureTest.kt
@@ -186,6 +186,31 @@ class SignatureTest : BaseAbstractTest() {
     }
 
     @Test
+    fun `fun with definitely non-nullable types`() {
+        val source = source("fun <T> elvisLike(x: T, y: T & Any): T & Any = x ?: y")
+        val writerPlugin = TestOutputWriterPlugin()
+
+        testInline(
+            source,
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
+                writerPlugin.writer.renderedContent("root/example/elvis-like.html").firstSignature().match(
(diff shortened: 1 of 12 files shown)


plugins/base/src/test/kotlin/signatures/SignatureTest.kt
```diff
@@ -185,6 +185,31 @@ class SignatureTest : BaseAbstractTest() {
...
+    fun `fun with definitely non-nullable types`() {
+        val source = source("fun <T> elvisLike(x: T, y: T & Any): T & Any = x ?: y")
+        val writerPlugin = TestOutputWriterPlugin()
+
+        testInline(
+            source,
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
```
IgnatBeresnev, line 198: can you also add asserts for documentable model level? i.e make sure it resolves to `DefinitelyNonNullable` and assert DRI for `Any`. Here or maybe `DRITest` would be a better place

plugins/kotlin-as-java/src/main/kotlin/signatures/JavaSignatureProvider.kt
```diff
@@ -196,6 +196,7 @@ class JavaSignatureProvider internal constructor(ctcc: CommentsToContentConverte
         is Star -> operator("?")
 
         is Nullable -> signatureForProjection(p.inner)
+        is DefinitelyNonNullable -> signatureForProjection(p.inner)
```
IgnatBeresnev, line 199: Can you also add tests for kotlin-as-java? There are some signature-related ones in `KotlinAsJavaPluginTest` already

IgnatBeresnev (approved):
:+1:

PR #2532, opened 2022-06-13
Handle more corner cases for inherited accessors

IgnatBeresnev:


plugins/kotlin-as-java/src/main/kotlin/signatures/JavaSignatureProvider.kt
```diff
@@ -114,7 +114,7 @@ class JavaSignatureProvider internal constructor(ctcc: CommentsToContentConverte
             ) {
                 annotationsBlock(p)
                 p.visibility[it]?.takeIf { it !in ignoredVisibilities }?.name?.let { keyword("$it ") }
-                p.modifier[it]?.name?.let { keyword("$it ") }
+                p.modifier[it]?.takeIf { it !in ignoredModifiers }?.name?.let { keyword("$it ") }
```
IgnatBeresnev, line 117: Noticed it when testing, it lead to an excessive space symbol

IgnatBeresnev, line 117: Noticed it when testing, it lead to an excessive space symbol

plugins/base/src/main/kotlin/translators/descriptors/DescriptorAccessorConventionUtil.kt
```diff
@@ -10,29 +10,93 @@ import org.jetbrains.kotlin.load.java.propertyNamesBySetMethodName
...
+    val (javaMethods, kotlinFunctions) = functions.partition { it is JavaMethodDescriptor }
+    if (javaMethods.isEmpty()) {
+        return DescriptorFunctionsHolder(regularFunctions = kotlinFunctions, emptyMap())
+    }
+
     val propertiesByName = properties.associateBy { it.name.asString() }
-    val regularFunctions = mutableListOf<FunctionDescriptor>()
-    val accessors = mutableMapOf<PropertyDescriptor, MutableList<FunctionDescriptor>>()
-    functions.forEach { function ->
+    val regularFunctions = ArrayList<FunctionDescriptor>(kotlinFunctions)
```
vmishenev, line 54: Are there reasons to use `ArrayList` instead of `List`?

plugins/base/src/main/kotlin/translators/descriptors/DescriptorAccessorConventionUtil.kt
```diff
@@ -10,29 +10,93 @@ import org.jetbrains.kotlin.load.java.propertyNamesBySetMethodName
...
+                    accessorHolder?.copy(getter = function) ?: DescriptorAccessorHolder(getter = function)
+                else
+                    accessorHolder?.copy(setter = function) ?: DescriptorAccessorHolder(setter = function)
+            }
         } else {
             regularFunctions.add(function)
         }
     }
+
+    val accessorLookalikes = removeNonAccessorsReturning(accessors)
```
vmishenev, line 72: I would make a filter here. It is more obvious

vmishenev (approved):
I had a brief look at this.

plugins/base/src/main/kotlin/translators/descriptors/DescriptorAccessorConventionUtil.kt
IgnatBeresnev: Yeah, ArrayList's constructor takes another collection as an argument and adds all of its elements into the new list. It's more efficient as by default even `addAll` still adds elements one by one, causing resizes and so on

plugins/base/src/main/kotlin/translators/descriptors/DescriptorAccessorConventionUtil.kt
IgnatBeresnev: Unfortunately, one simple filter would not do :( We need to both remove elements from `accessors` and add removed elements to `regularFunctions` (so it involves mapping of elements also)

Filter would only return the elements which need to be added to `regularFunctions`, but it will not remove it from `accessors`. 

This could be done in multiple steps, something like this, but IMHO it involves many excessive operations which can be substituted with a single iteration
```kotlin
val nonAccessibleProperties: Map<PropertyDescriptor, DescriptorAccessorHolder> =
        accessors.filterValues { it.getter == null && it.setter != null }

val functionsOfNonAccessibleProperties = nonAccessibleProperties.values.map { it.setter!! }
regularFunctions.addAll(functionsOfNonAccessibleProperties)

nonAccessibleProperties.keys.forEach { accessors.remove(it) }
```

plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocIndexTest.kt
```diff
@@ -71,7 +71,7 @@ internal class JavadocIndexTest : AbstractJavadocTemplateMapTest() {
             AnnotationTarget.ANNOTATION_CLASS::class.java.methods.any { it.name == "describeConstable" }
 
         testIndexPages(commonTestQuery) { indexPages ->
-            assertEquals(if (hasAdditionalFunction()) 32 else 31, indexPages.sumBy { it.elements.size })
+            assertEquals(if (hasAdditionalFunction()) 31 else 30, indexPages.sumBy { it.elements.size })
```
IgnatBeresnev, line 74: `protected void finalize()` now property labeled as `protected` and is thus hidden by default. It used to resolve to `public`

PR #2540, opened 2022-06-16
Fix incorrectly labeling java properties as val/var

### LINKED ISSUE #2539: Public Java property is shown as `val`
For public properties without getter or setter Dokka shows keyword `val` instead of `var`

```java
public class JavaClass {
    public int publicProperty;
}
```

image

**Installation**
- Dokka version: 1.7.0-dev-170

IgnatBeresnev:
Fixes #2539

### DIFF
--- a/core/src/main/kotlin/model/documentableProperties.kt
+++ b/core/src/main/kotlin/model/documentableProperties.kt
@@ -39,6 +39,16 @@ object ObviousMember : ExtraProperty<Documentable>, ExtraProperty.Key<Documentab
     override val key: ExtraProperty.Key<Documentable, *> = this
 }
 
+/**
+ * Whether this [DProperty] is `var` or `val`, should be present both in Kotlin and in Java properties
+ *
+ * In case of properties that came from `Java`, [IsVar] is added if
+ * the java field has no accessors at all (plain field) or has a setter
+ */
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/model/documentableProperties.kt
```diff
@@ -39,6 +39,16 @@ object ObviousMember : ExtraProperty<Documentable>, ExtraProperty.Key<Documentab
...
 
+/**
+ * Whether this [DProperty] is `var` or `val`, should be present both in Kotlin and in Java properties
+ *
+ * In case of properties that came from `Java`, [IsVar] is added if
+ * the java field has no accessors at all (plain field) or has a setter
+ */
+object IsVar : ExtraProperty<DProperty>, ExtraProperty.Key<DProperty, IsVar> {
+    override val key: ExtraProperty.Key<DProperty, *> = this
+}
```
IgnatBeresnev, line 50: Alternatively, this could be made a field in `DProperty`, I think it could easily be an essential part of the documentation model

vmishenev:
Why do you not use  `if (this.getter == null && this.setter == null)` for this? 
Properties always have a getter. It works for Kotlin.

Although there is a case when we have a final field in Java. Can you add a such test?

vmishenev (approved):

PR #2547, opened 2022-06-21
Do not generate source links for synthetic elements

### LINKED ISSUE #2544: Regression: linking Java sources broken in 1.7.0
**Describe the bug**

Dokka used to be able to link Java sources, see, e.g., https://alchemistsimulator.github.io/reference/kdoc/alchemist/it.unibo.alchemist.core.implementations/-engine/add-output-monitor.html

Since 1.7.0, `dokkaHtmlCollector` fails with `textRange must not be null`.

**Expected behaviour**

Dokka does not fail, the generated documentation links Java sources as it used to do in 1.6.x

**To Reproduce**

```
git clone https://github.com/AlchemistSimulator/Alchemist.git
cd Alchemist
git checkout renovate/dokka
./gradlew dokkaHtmlCollector
```

**Dokka configuration**

```Kotlin
    tasks.withType<org.jetbrains.dokka.gradle.DokkaTask>().configureEach {
            jdkVersion.set(multiJvm.jvmVersionForCompilation)
            listOf("kotlin", "java") // remove "java" from this list to "fix" the bug!
                .map { "src/main/$it" }
                .map { it to File(projectDir, it) }
                .toMap()
                .filterValues { it.exists() }
                .forEach { (path, file) ->
                    sourceLink {
                        localDirectory.set(file)
                        val project = if (project == rootProject) "" else project.name
                        val url = "https://github.com/AlchemistSimulator/Alchemist/blob/master/$project/$path"
                        remoteUrl.set(uri(url).toURL())
                        remoteLineSuffix.set("#L")
                    }
                }
            configurations.implementation.get().dependencies.forEach { dep ->
                val javadocIOURLs = fetchJavadocIOForDependency(dep)
                if (javadocIOURLs != null) {
                    val (javadoc, packageList) = javadocIOURLs
                    externalDocumentationLink {
                        url.set(javadoc)
                        packageListUrl.set(packageList)
                    }
                }
            }
        }
        failOnWarning.set(true)
    }
```

**Note**: removing `"java"` from the list above "fixes" the issue (it just turns off the source linking for Java sources).

**Installation**
- see Gradle scan:  https://gradle.com/s/4easgt5kurdlw
- Dokka version: 1.7.0

**Additional context**

Mixed Java-Kotlin projects (originally in Java, new parts written in Kotlin, slowly transitating to Kt).

I have no idea what to fix, I need guidance at the very least.

IgnatBeresnev:
Fixes #2544

### DIFF
--- a/core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
+++ b/core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
@@ -68,7 +68,7 @@ abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : DokkaTestGe
         block: T.() -> Unit
     ) {
         val testMethods = testBuilder().apply(block).build()
-        val testDirPath = getTempDir(cleanupOutput).root.toPath()
+        val testDirPath = getTempDir(cleanupOutput).root.toPath().toAbsolutePath()
         val fileMap = query.toFileMap()
         fileMap.materializeFiles(testDirPath.toAbsolutePath())
         if (!cleanupOutput)
@@ -85,7 +85,7 @@ abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : DokkaTestGe
                     }.toSet(),
                     sourceLinks = sourceSet.sourceLinks.map { link ->
                         link.copy(
-                            localDirectory = testDirPath.toFile().resolve(link.localDirectory).canonicalPath
+                            localDirectory = testDirPath.toFile().resolve(link.localDirectory).absolutePath
                         )
                     }.toSet()
                 )
--- a/plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
+++ b/plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
@@ -24,36 +24,47 @@ import java.io.File
 
 class SourceLinksTransformer(val context: DokkaContext) : PageTransformer {
 
-    private val builder : PageContentBuilder =  PageContentBuilder(
+    private val builder : PageContentBuilder = PageContentBuilder(
         context.plugin<DokkaBase>().querySingle { commentsToContentConverter },
         context.plugin<DokkaBase>().querySingle { signatureProvider },
         context.logger
     )
 
-    override fun invoke(input: RootPageNode) =
-        input.transformContentPagesTree { node ->
+    override fun invoke(input: RootPageNode): RootPageNode {
+        val sourceLinks = getSourceLinksFromConfiguration()
+        if (sourceLinks.isEmpty()) {
+            return input
+        }
+        return input.transformContentPagesTree { node ->
             when (node) {
                 is WithDocumentables ->
-                    node.documentables.filterIsInstance<WithSources>().flatMap { resolveSources(it) }
-                    .takeIf { it.isNotEmpty() }
-                    ?.let { node.addSourcesContent(it) }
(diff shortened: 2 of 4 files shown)


plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
```diff
@@ -134,9 +135,13 @@ class SourceLinksTransformer(val context: DokkaContext) : PageTransformer {
...
     private fun PsiElement.lineNumber(): Int? {
-        val doc = PsiDocumentManager.getInstance(project).getDocument(containingFile)
-        // IJ uses 0-based line-numbers; external source browsers use 1-based
-        return doc?.getLineNumber(textRange.startOffset)?.plus(1)
+        return if (this is SyntheticElement) {
+            null // does not actually exist in code, thus no line number
+        } else {
+            val doc = PsiDocumentManager.getInstance(project).getDocument(containingFile)
+            // IJ uses 0-based line-numbers; external source browsers use 1-based
+            return doc?.getLineNumber(textRange.startOffset)?.plus(1)
```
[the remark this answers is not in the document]
IgnatBeresnev: It's not marked as nullable by contract in `PsiElement`, but abstract `LightElement` does return `null` by default. Other than that, usually `TextRange.EMPTY_RANGE` (0, 0) is returned from what I can see.

I thought about checking it for null instead of `SyntheticElement`, but I'm afraid it can lead to bugs where source links are not generated due to some change in the IntelliJ API, and we wouldn't even know, since source links for java projects are not covered by tests at all.

If null is all of a sudden returned here, I personally would prefer it failing loud during testing. The problem was that we don't test on java projects at all, nevermind with configured source links. Alchemist project from the issue is a good candidate for before-release testing, which would've caught it.

plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
```diff
@@ -24,36 +24,47 @@ import java.io.File
...
-    private val builder : PageContentBuilder =  PageContentBuilder(
+    private val builder : PageContentBuilder = PageContentBuilder(
         context.plugin<DokkaBase>().querySingle { commentsToContentConverter },
         context.plugin<DokkaBase>().querySingle { signatureProvider },
         context.logger
     )
 
-    override fun invoke(input: RootPageNode) =
-        input.transformContentPagesTree { node ->
+    override fun invoke(input: RootPageNode): RootPageNode {
```
IgnatBeresnev, line 33: A bit of refactoring to make it not run if sources are not configured and to not parse configuration on every documentable

core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
```diff
@@ -85,7 +85,7 @@ abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : DokkaTestGe
                     }.toSet(),
                     sourceLinks = sourceSet.sourceLinks.map { link ->
                         link.copy(
-                            localDirectory = testDirPath.toFile().resolve(link.localDirectory).canonicalPath
+                            localDirectory = testDirPath.toFile().resolve(link.localDirectory).absolutePath
```
IgnatBeresnev, line 88: The problem with paths was here. `canonicalPath` resolves symlinks, while `absolutePath` does not, so on windows in resulted in two different temp folder locations.

I'll refactor this bit later on in a separate PR once I get a hold of a windows machine

PR #2550, opened 2022-06-22
Give notice of alpha version for GFM, Javadoc, Jekyll formats

vmishenev:
Current log:
```
Initializing plugins
Dokka is performing: documentation for Dokka Gradle Example
Validity check
The Javadoc output format is in Alpha version
Creating documentation models
Merging documentation models
Creating pages
Transforming pages
Rendering
Running post-actions
```
The implementation works for all runners.
.

### DIFF
--- a/plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/GfmPlugin.kt
+++ b/plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/GfmPlugin.kt
@@ -9,9 +9,8 @@ import org.jetbrains.dokka.gfm.location.MarkdownLocationProvider
 import org.jetbrains.dokka.gfm.renderer.BriefCommentPreprocessor
 import org.jetbrains.dokka.gfm.renderer.CommonmarkRenderer
 import org.jetbrains.dokka.plugability.DokkaPlugin
+import org.jetbrains.dokka.renderers.PostAction
 import org.jetbrains.dokka.transformers.pages.PageTransformer
-import org.jetbrains.dokka.validity.PreGenerationChecker
-import org.jetbrains.dokka.validity.PreGenerationCheckerOutput
 
 class GfmPlugin : DokkaPlugin() {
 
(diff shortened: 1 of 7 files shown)


plugins/gfm/src/main/kotlin/org/jetbrains/dokka/gfm/GfmPlugin.kt
```diff
@@ -38,4 +40,13 @@ class GfmPlugin : DokkaPlugin() {
...
                 order { after(rootCreator) })
     }
+
+    val alphaVersionNotifier by extending {
+        CoreExtensions.preGenerationCheck providing { ctx ->
+            PreGenerationChecker {
+                ctx.logger.warn("The GFM output format is in Alpha version")
+                PreGenerationCheckerOutput(true, emptyList())
+            }
+        }
```
IgnatBeresnev, line 50: Want to propose a couple of things:

1. I think it's worth making it `internal` so that it doesn't leak into public API and we don't have to deal with binary compatibility issues when it's removed
2. I'd swap `preGenerationCheck` to `postActions` so that it's the last message that's printed. If the message is printed somewhere in the middle of a wall of text, it's less likely to be noticed. Having it as the very first message would also be good
3. The message is not scary enough :) "It's in alpha, so what?" Doesn't explain the consequences. We can take the description of what to expect from [Kotlin alpha stability level](https://kotlinlang.org/docs/components-stability.html#stability-levels-explained) (also, I think "the" and "version" can be omitted):

```suggestion
    internal val alphaVersionNotifier by extending {
        CoreExtensions.postActions providing { ctx ->
            PostAction {
                ctx.logger.warn("GFM output format is in Alpha, use at your own risk, expect bugs and migration issues")
            }
        }
    }
```

vmishenev: I left  `version` to understand what `alpha` means.

IgnatBeresnev (approved):
:+1:

PR #2553, opened 2022-06-23
Add enum synthetic functions to documentable model

IgnatBeresnev:
Previously `values` and `valueOf` enum functions were not present in the documentable model

Even though we skip obvious members for our formats, other plugins might want to use it

### DIFF
--- a/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
@@ -59,6 +59,7 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.annotationClass
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameOrNull
 import org.jetbrains.kotlin.resolve.descriptorUtil.parents
 import org.jetbrains.kotlin.resolve.scopes.MemberScope
+import org.jetbrains.kotlin.resolve.scopes.StaticScopeForKotlinEnum
 import org.jetbrains.kotlin.resolve.source.KotlinSourceElement
 import org.jetbrains.kotlin.resolve.source.PsiSourceElement
 import org.jetbrains.kotlin.resolve.source.PsiSourceFile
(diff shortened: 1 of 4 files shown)


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -301,6 +301,16 @@ private class DokkaDescriptorVisitor(
         }
     }
 
+    private fun ClassDescriptor.getEnumDescriptorsWithKind(): DescriptorsWithKind {
+        val descriptorsWithKind = this.unsubstitutedMemberScope.getDescriptorsWithKind()
+        val staticScopeForKotlinEnum = (this.staticScope as? StaticScopeForKotlinEnum) ?: return descriptorsWithKind
+
+        // synthetic values() and valueOf() functions are not present among average class functions
+        val enumSyntheticFunctions = staticScopeForKotlinEnum.getContributedDescriptors { true }
```
vmishenev, line 309: Are you sure it will contain only functions?! 
Anyway it is not a blocker to merge the PR.

`staticScope` is `MemberScope` too.
Yet another implementation:
```kt
val descriptorsFromMemberScope = unsubstitutedMemberScope.getDescriptorsWithKind()
val descriptorsFromStaticScope = staticScope.getDescriptorsWithKind()
returns descriptorsFromMemberScope  + descriptorsFromStaticScope // a plus operator - TODO
```

vmishenev (approved):

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
IgnatBeresnev: The return type for `getContributedDescriptors` is `SimpleFunctionDescriptor`, so it should be just functions, yeah

I thought about what you suggested, but it seems like calling `getDescriptorsWithKind` on `staticScope` would add some overhead and require writing even more code to support it, so I think it's best to keep it simple

PR #2557, opened 2022-06-27
Implement a workaround for PathTreeWalk @SinceKotlin #KT-52977

qurbonzoda:


qurbonzoda:
It was decided to implement this workaround in `kotlin-stdlib-docs` build scripts.

PR #2570, opened 2022-07-14
Another take at removing the kotlin-stdlib dependency from the Gradle Plugin

martinbonnin:
Follow up from https://github.com/Kotlin/dokka/pull/2543
See also [this kotlinlang slack discussion](https://kotlinlang.slack.com/archives/C19FD9681/p1656593570021959)

Looks like the "com.gradle.plugin-publish" creates its own pom that doesn't find "configuration-level" excludes but requires "dependency-level" exclude. This PR mimics the behaviour from the Kotlin Gradle Plugin. It's hard to test end to end, especially since "com.gradle.plugin-publish" is closed source but I could verify that the file at `dokka/runners/gradle-plugin/build/publish-generated-resources/pom.xml` now contains excludes 🤞

### DIFF
--- a/runners/gradle-plugin/build.gradle.kts
+++ b/runners/gradle-plugin/build.gradle.kts
@@ -32,13 +32,27 @@ dependencies {
 
 // Gradle will put its own version of the stdlib in the classpath, do not pull our own we will end up with
 // warnings like 'Runtime JAR files in the classpath should have the same version'
-configurations.api {
-    exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib")
-    exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-jdk7")
-    exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-jdk8")
-    exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-common")
+configurations.api.configure {
+    excludeGradleCommonDependencies()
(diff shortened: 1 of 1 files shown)


IgnatBeresnev:
Thank you yet again for helping with this! it's of enormous help :) 

It is indeed difficult to test it. For maven repos we can publish it to the snapshot repository and test it through it, but there's nothing like `gradlePluginSnapshotRepository()`, right?

Anyhow I think it's worth giving it a shot. Let's wait for integration tests to pass, and I'll test it manually for maven repos, make sure it's working at least for maven central

martinbonnin:
> For maven repos we can publish it to the snapshot repository and test it through it, but there's nothing like gradlePluginSnapshotRepository(), right?

Not that I'm aware of unfortunately :(. The portal allows you to delete a version shortly it has been uploaded so I guess one way to test, if you have access to the the portal credentials, is to upload manually and to end-to-end testing. Not ideal but I don't really see another way.

> Let's wait for integration tests to pass, and I'll test it manually for maven repos, make sure it's working at least for maven central, and we'll merge it

Sounds good 👍 . It's the same code as the Kotlin plugin + `dokka/runners/gradle-plugin/build/publish-generated-resources/pom.xml` did change so 🤞 it's all good this time.

IgnatBeresnev (approved):
Looks good, thanks once again for the contribution!

Issue #2576, opened 2022-07-20
UI improvements


IgnatBeresnev:
An umbrella issue for many small UI improvements that are planned for the near future. Since most of them are really small and easy to do, I think creating separate issues for each one is excessive.

Emoji statuses to keep track of progress:
* :hourglass: Not started
*  :eyes: In progress / being investigated
* :heavy_check_mark:  Done

### Tasks:

* :heavy_check_mark: For package level pages stop writing it’s package as it’s too long, takes too much space and, actually, there is a package on the top. Write something like “package level declarations” instead (#2586)
* ~~Don’t underline type parameters and functions names that refer to the same page, see [findAnnotationOnBean()](https://docs.spring.io/spring-framework/docs/current/kdoc-api/spring-beans/org.springframework.beans.factory/find-annotation-on-bean.html)~~ Won't fix, too much work required for this change
* :heavy_check_mark: “See also” section: make it the same as Authors & Since section, fix size & alingment. Fix references, see [getBean()](https://docs.spring.io/spring-framework/docs/current/kdoc-api/spring-beans/org.springframework.beans.factory/get-bean.html) (#2587)
* :heavy_check_mark: “Throws” section: make it the same as Authors & Since section, fix references, see [decodeFromJsonElement()](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/decode-from-json-element.html) (#2587)
* :heavy_check_mark: Make Authors/Since/See also/Throws sections smaller, remove extra indents (#2587)
* :heavy_check_mark: Drop copy signature functionality at all (#2577)
* :heavy_check_mark: Save position in navigation tree (#2575)
* :heavy_check_mark: Try to add icons into navigation tree (#2578)
* :heavy_check_mark: Sources: link to sources should be near its signature, we should try to remove source tab from the bottom of a page completely as it’s not intuitive and have problems with MPP, see [awaitAll()](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-all.html) (#2467)
* :heavy_check_mark: Make current breadcrumb element not clickable and of default font color (#2588)
[...]

PR #2577, opened 2022-07-20
Disable copy button for signatures

IgnatBeresnev:
Copy button itself and the script for it are still present for two scenarios:

* Copying anchor links
* Copying contents of code blocks. These are usually usage examples, so I think it could be useful

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -95,7 +95,6 @@ open class HtmlRenderer(
             }
             node.dci.kind in setOf(ContentKind.Symbol) -> div("symbol $additionalClasses") {
                 childrenCallback()
-                if (node.hasStyle(TextStyle.Monospace)) copyButton()
             }
             node.hasStyle(TextStyle.BreakableAfter) -> {
                 span { childrenCallback() }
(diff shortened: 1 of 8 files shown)


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -95,7 +95,6 @@ open class HtmlRenderer(
             }
             node.dci.kind in setOf(ContentKind.Symbol) -> div("symbol $additionalClasses") {
                 childrenCallback()
-                if (node.hasStyle(TextStyle.Monospace)) copyButton()
```
IgnatBeresnev, line 98: This is the only real change in this PR. The rest is either formatting or to make tests pass

vmishenev (commented):
I suppose [some styles](https://github.com/Kotlin/dokka/blob/b6a8e58832929cd08d7fcd9f92e7fd010990f5e0/plugins/base/src/main/resources/dokka/styles/style.css#L346), that are related to `symbol`,  are redundant.

IgnatBeresnev:
Indeed, hopefully I got all of it

IgnatBeresnev:
> And [here](https://github.com/Kotlin/dokka/blob/b6a8e58832929cd08d7fcd9f92e7fd010990f5e0/plugins/base/src/main/resources/dokka/styles/style.css#L346)

Are you sure you're looking at the right commit? Or maybe the link is wrong? See https://github.com/Kotlin/dokka/pull/2577/files#diff-316e3e9b3d50ff73dcc848844e5eaac72b2d38b7f1e84dca8a383a8b2c5a162c

vmishenev:
My fault, the link is wrong. 
https://github.com/Kotlin/dokka/blob/b6a8e58832929cd08d7fcd9f92e7fd010990f5e0/plugins/base/src/main/resources/dokka/styles/style.css#L404

IgnatBeresnev:
Jesus :) Yeah, removed

vmishenev (approved):

PR #2578, opened 2022-07-20
Add member icons to navigation menu

IgnatBeresnev:
Navigation icons to help distinguish between members. Styling is TBD

Live example: https://ignatberesnev.github.io/static/dokka/icons/intellij/neutral-java-style/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/index.html


IgnatBeresnev:
Documentables with Java sources will have Java-styled icons, otherwise Kotlin-styled icons will be displayed.

Also added handling of long text links and package names in the navigation tree, it's displayed properly now, with icon always on the left (see [this class](https://ignatberesnev.github.io/static/dokka/icons/alchemist-java/alchemist-incarnation-biochemistry/it.unibo.alchemist.model.interfaces/-environment-supporting-deformable-cells/index.html) for instance)

Current pre-review version for a mixed language project: https://ignatberesnev.github.io/static/dokka/icons/alchemist-java/alchemist-incarnation-biochemistry/it.unibo.alchemist.model.interfaces/-environment-supporting-deformable-cells/index.html

plugins/base/src/test/kotlin/enums/KotlinEnumsTest.kt
```diff
@@ -191,56 +191,6 @@ class KotlinEnumsTest : BaseAbstractTest() {
         }
     }
 
-    @Test
-    fun `should preserve enum source ordering for navigation menu`() {
```
IgnatBeresnev, line 195: Enum entries have been dropped from the navigation menu

plugins/base/src/test/kotlin/enums/JavaEnumsTest.kt
```diff
@@ -57,4 +68,50 @@ class JavaEnumsTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Test
+    fun `should mark synthetic functions generated for Kotlin as obvious`() {
```
IgnatBeresnev, line 73: Looks like some classes got duplicated for some reason, probably due to automatic conflict resolution. Deleted `JavaEnumTest` and moved this test to `JavaEnumsTest`, that was the only difference

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+     *  - page only contains functions (therefore documentable from this page is [DFunction])
+     */
+    private fun ContentPage.displayableName(): String =
+        if (this is WithDocumentables && documentables.all { it is DFunction }) {
+            "$name()"
+        } else {
+            name
+        }
+
+    private fun chooseNavigationIcon(contentPage: ContentPage): NavigationNodeIcon? {
```
IgnatBeresnev, line 47: This method is new

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+
+import org.jetbrains.dokka.analysis.PsiDocumentableSource
+import org.jetbrains.dokka.base.renderers.sourceSets
+import org.jetbrains.dokka.base.transformers.documentables.isException
+import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.pages.*
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.transformers.pages.PageTransformer
+
+open class NavigationPageInstaller(val context: DokkaContext) : NavigationDataProvider(), PageTransformer {
```
IgnatBeresnev, line 11: The class was getting a bit too long for `htmlPreprocessors.kt`, extracted it. Will point out which parts are new

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+    private fun ContentPage.navigableChildren(): List<NavigationNode> {
+        return if (this !is ClasslikePageNode) {
+            children
+                .filterIsInstance<ContentPage>()
+                .map { visit(it) }
+                .sortedBy { it.name.toLowerCase() }
+        } else {
+            emptyList()
+        }
+    }
```
IgnatBeresnev, line 91: Removed enum entries from here, so it's not in the navigation tree anymore

vmishenev (commented):
So I am waiting a source dependent icons.
And it would be great to combine the icons into a SVG sprite.

vmishenev (approved):
Have you checked it in mobile!?

IgnatBeresnev:
> Have you checked it in a mobile browser!?

I did, it works and doesn't look worse than before. Is there something wrong with it?

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+        if (this is WithDocumentables && documentables.all { it is DFunction }) {
+            "$name()"
+        } else {
+            name
+        }
+
+    private fun chooseNavigationIcon(contentPage: ContentPage): NavigationNodeIcon? {
+        return if (contentPage is WithDocumentables) {
+            val documentable = contentPage.documentables.firstOrNull()
+            val isJava = (documentable as? WithSources)?.sources?.any { it.value is PsiDocumentableSource } ?: false
```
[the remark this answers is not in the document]
IgnatBeresnev: It's a bit ugly and has some excessive operations if you want to use `documentableLanguage()` since it has a `sourceSet` parameter, and we don't have any particular source set here

But I guess avoiding code duplication is more important. Extracted it into a separate function, makes it somewhat better

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
IgnatBeresnev: I don't think it necessarily _should_ be in the same file, sooner or later it'll grow so big that all of them will have to be extracted into their own ones. Imho, grouping classes by features rather than domain is easier for navigating around

Doesn't make a huge difference now, so returned it

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
IgnatBeresnev: Unfortunately, had to return enum entries to navigation. 

If someone has a direct link to an enum entry, it's unclear where to scroll - from entry's page I couldn't figure out a clean solution for selecting parent's navigation menu link. 

The only dirty way is to iterate all of the sideMenu and compare links, but that looks wrong on so many levels.

Maybe enum entries should not have pages at all. I'll bring this up during sync so we can brainstorm together

vmishenev:
> I did, it works and doesn't look worse than before. Is there something wrong with it?

No, I have not tested it there.

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
vmishenev: Mu concern: If I open `htmlPreprocessors.kt` I expect to see ALL preprocessors there.

> Imho, grouping classes by features rather than domain is easier for navigating around

Anyway now there are two different grouping. It is misleading.

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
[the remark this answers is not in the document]
vmishenev: > But I guess avoiding code duplication is more important. Extracted it into a separate function, makes it somewhat better

Yeah, I do not like low level things like `PsiDocumentableSource` (non core API) here. It is not obviously what it means.

PR #2585, opened 2022-07-25
Add a distinct divider between function overloads

IgnatBeresnev:
By far not the best solution, but I found no other _reliable_ way to add a divider for function overloads especially considering different sourcesets. This adds no new public API, so I think it's not a big deal - will probably have to redesign this down the road anyway. 

#  For function overloads with some description

___

# For function overloads with the same description

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -14,16 +14,13 @@ import org.jetbrains.dokka.base.resolvers.anchors.SymbolAnchorHint
 import org.jetbrains.dokka.base.resolvers.local.DokkaBaseLocationProvider
 import org.jetbrains.dokka.base.templating.*
 import org.jetbrains.dokka.links.DRI
-import org.jetbrains.dokka.model.DisplaySourceSet
+import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.properties.PropertyContainer
-import org.jetbrains.dokka.model.sourceSetIDs
-import org.jetbrains.dokka.model.withDescendants
 import org.jetbrains.dokka.pages.*
 import org.jetbrains.dokka.pages.HtmlContent
 import org.jetbrains.dokka.plugability.*
 import org.jetbrains.dokka.utilities.htmlEscape
 import org.jetbrains.kotlin.utils.addIfNotNull
-import java.net.URI
 
 internal const val TEMPLATE_REPLACEMENT: String = "###"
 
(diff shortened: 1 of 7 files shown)


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -768,9 +794,6 @@ open class HtmlRenderer(
             content(this, page)
         }
 
-    private val String.isAbsolute: Boolean
-        get() = URI(this).isAbsolute
-
```
IgnatBeresnev, line 773: Unused

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -321,6 +313,40 @@ open class HtmlRenderer(
         }
     }
 
+    private fun groupDivergentInstancesWithSourceSet(
```
IgnatBeresnev, line 316: This made `buildDivergent` considerably longer, even though this particular function needed no context from it and shared nothing in common, so I don't see a reason why it has to be a nested function.

Extracted for the sake of readability of `buildDivergent`

IgnatBeresnev:
Proposing to go with this solution as it adds no new public API, so it will be easy to remove when the page is redesigned, which is just a matter of time.

I'll post the final result to #kotlin-libs and ask for feedback. If this version is not liked for some reason, it can be changed before release

vmishenev (changes requested):
1. https://dokka-snapshots.s3.eu-central-1.amazonaws.com/fuction-overloads-dinstinct/serialization/8779f50/kotlinx-serialization-core/kotlinx.serialization.builtins/serializer.html 
There are not overloads.
2. Regression: There is extra space in a declaration list. It requires a check `node.dci.kind == ContentKind.Main` that it is removed in this PR.

image

IgnatBeresnev:
> There are not overloads.

Not sure what you mean. These are different functions with different description, but with the same function name, so for me it makes an overload :thinking:

vmishenev:
> Not sure what you mean. These are different functions with different description, but with the same function name, so for me it makes an overload thinking

Sorry(  I wrote a wrong sentence.

For https://dokka-snapshots.s3.eu-central-1.amazonaws.com/fuction-overloads-dinstinct/stdlib/4ce210b/kotlin-stdlib/kotlin.collections/indices.html, that has several extension properties﻿, it does not have a divider.

IgnatBeresnev:
> has several extension properties

Silly me, for some reason I thought class named `Same**Method**NamePageMergerStrategy` only merged methods :smiling_face_with_tear:

Previously there was a check that it was only functions, changed it to accept for all members, so both functions and properties.

IgnatBeresnev:
After some discussion, we decided to drop "Overload" header and just stick with a simple horizontal line

Updated the PR, @vmishenev please have a look

vmishenev (approved):

PR #2586, opened 2022-07-25
Replace package name on cover with "Package-level declarations"

IgnatBeresnev:
See #2576

### DIFF
--- a/plugins/base/src/main/resources/dokka/styles/style.css
+++ b/plugins/base/src/main/resources/dokka/styles/style.css
@@ -192,8 +192,6 @@ html ::-webkit-scrollbar-thumb {
 }
 
 .cover > .platform-hinted {
-    padding-top: 12px;
-    margin-top: 12px;
     padding-bottom: 12px;
 }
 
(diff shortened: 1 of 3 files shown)


plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -192,8 +192,6 @@ html ::-webkit-scrollbar-thumb {
 }
 
 .cover > .platform-hinted {
-    padding-top: 12px;
-    margin-top: 12px;
```
IgnatBeresnev, line 196: This is not even noticable since there's bottom padding from `h1.cover` already. Makes the page a little bit more compact and grabs less attention

plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -550,8 +548,8 @@ h1 {
 
 
 h1.cover {
-    font-size: 60px;
-    line-height: 64px;
+    font-size: 52px;
+    line-height: 56px;
```
IgnatBeresnev, line 552: Also barely noticable, makes longer names fit a bit better and doesn't grab as much attention compared to the rest of the page.

vmishenev (approved):

PR #2587, opened 2022-07-26
Apply the same style to all KDoc tags, including throws/see/parameters

IgnatBeresnev:

* Now all KDoc tag sections should look the same, including throws/see/parameters
* Decreased space between tag header and values

* Fixed a problem where longer link text had many word breaks:
    * See/Throws link text is now just the class name (if DRI is resolved), full path if unresolved; previously always full path
    * Increased first column width for See/Throws/Parameters a bit.
    * (I have a complete JS script to dynamically adjust the width of the first column based on link width, will discuss it during sync and maybe file it as another PR)

### DIFF
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -469,8 +469,10 @@ open class DefaultPageCreator(
                     customTags.forEach { (_, sourceSetTag) ->
                         sourceSetTag[platform]?.let { tag ->
                             customTagContentProviders.filter { it.isApplicable(tag) }.forEach { provider ->
-                                with(provider) {
-                                    contentForDescription(platform, tag)
+                                group(sourceSets = setOf(platform), styles = setOf(ContentStyle.KDocTag)) {
+                                    with(provider) {
+                                        contentForDescription(platform, tag)
+                                    }
                                 }
                             }
                         }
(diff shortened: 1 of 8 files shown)


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -537,7 +537,7 @@ open class DefaultPageCreator(
                 val params = tags.withTypeNamed<Param>()
                 val availablePlatforms = params.values.flatMap { it.keys }.toSet()
 
-                header(2, "Parameters", kind = ContentKind.Parameters, sourceSets = availablePlatforms)
+                header(4, "Parameters", kind = ContentKind.Parameters, sourceSets = availablePlatforms)
```
vmishenev, line 540: I think it is better to make a constant to avoid magic number.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -555,7 +555,9 @@ open class DefaultPageCreator(
                                                 kind = ContentKind.Parameters,
                                                 styles = mainStyles + ContentStyle.RowTitle
                                             )
-                                            comment(it.root)
+                                            if (it.root.children.isNotEmpty()) {
```
vmishenev, line 558: I would create a helper like `TagWrapper.isEmpty() = children.isEmpty()` (Law of Demeter)

vmishenev (changes requested):
Before:
image
After:
image

IgnatBeresnev:
> After (there is space after Return)

I've fixed `Since Kotlin` block, totally forgot about that :+1: but I don't understand what's wrong with `Return`

vmishenev (approved):

PR #2589, opened 2022-07-27
Style section and platform tabs like on kotlinlang.org

IgnatBeresnev:
See #2576

It should now visually look the same as on kotlinlang.org, for instance like on [this page](https://kotlinlang.org/docs/getting-started.html#create-your-powerful-application-with-kotlin)

Changing only platform tabs made it inconsistent with section tabs, so I changed section tabs as well

Live demo: https://ignatberesnev.github.io/static/dokka/kotlinlang-tabs/index.html


vmishenev (approved):
Does [that](https://github.com/Kotlin/dokka/blob/51764340e58dbc6e3a1ecc1a50884e1ed68e1c83/plugins/base/src/main/resources/dokka/styles/style.css#L707) need?

PR #2593, opened 2022-07-29
Fix parsing of static imports in annotation params

### LINKED ISSUE #2580: Missing annotation values for static imports
Some annotation values which are statically imported constants are not showing up in docs. For example, [this function](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core/src/main/java/androidx/core/location/LocationManagerCompat.java;l=480) has the annotation `@RequiresPermission(ACCESS_FINE_LOCATION)` where the constant comes from `import static android.Manifest.permission.ACCESS_FINE_LOCATION;`, but the annotation value [shows up as the empty string](https://developer.android.com/reference/kotlin/androidx/core/location/LocationManagerCompat#registerGnssStatusCallback(android.location.LocationManager,java.util.concurrent.Executor,androidx.core.location.GnssStatusCompat.Callback)) in the docs. Some annotations in the same file use a [set of constant values](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core/src/main/java/androidx/core/location/LocationManagerCompat.java;l=231), which are [not showing up at all](https://developer.android.com/reference/kotlin/androidx/core/location/LocationManagerCompat#requestLocationUpdates(android.location.LocationManager,java.lang.String,androidx.core.location.LocationRequestCompat,androidx.core.location.LocationListenerCompat,android.os.Looper)).

It seems this was addressed in #1845 with the `JvmAnnotationAttribute.toValue` function in `DefaultPsiToDocumentableTranslator`, but in this case the `attributeValue` of the `JvmAnnotationAttribute` is not a `JvmAnnotationEnumFieldValue` (it is a `PsiAnnotationConstantValue`), so that solution doesn't work here:

Screen Shot 2022-07-21 at 11 11 59 AM

**Installation**
Operating system: macOS
Build tool: Gradle v7.3
Dokka version: 1.6.20-dev-154

IgnatBeresnev:
Fixes #2580


plugins/base/src/main/kotlin/translators/psi/PsiPackageHelper.kt
```diff
@@ -0,0 +1,15 @@
+/**
+ * Pretend to be from `com.intellij.psi` in order to access package-private declarations
+ */
+@file:Suppress("PackageDirectoryMismatch")
+package com.intellij.psi
+
+import com.intellij.lang.jvm.annotation.JvmAnnotationAttributeValue
+
+/*
+ * For some reason `PsiAnnotationConstantValue` is package private,
```
vmishenev, line 10: Can you use a public interface `JvmAnnotationConstantValue` instead of `PsiAnnotationConstantValue`?

vmishenev (changes requested):

plugins/base/src/main/kotlin/translators/psi/PsiPackageHelper.kt
IgnatBeresnev: Wow, yeah, totally. 

I was looking for something like this, but when you open up `PsiAnnotationConstantValue` and go to declaration of `JvmAnnotationConstantValue` from the signature - it goes to the kotlin-compiler artifact, where the interface is empty, so I thought there was no other way :facepalm: 

Importing it seems to be fine though

vmishenev (approved):

PR #2600, opened 2022-08-02
Add test failing multiplatform test for LongSparseArray

asfalcone:
At `documentablesCreationStage` the package exists but there are no classlikes or functions.

### DIFF
--- a/plugins/base/src/test/kotlin/multiplatform/BasicMultiplatformTest.kt
+++ b/plugins/base/src/test/kotlin/multiplatform/BasicMultiplatformTest.kt
@@ -26,6 +26,44 @@ class BasicMultiplatformTest : BaseAbstractTest() {
     }
 
     @Test
+    fun dataTestLongSparseArray() {
+        val testDataDir = getTestDataDir("multiplatform/longSparseArray").toAbsolutePath()
+
+        val configuration = dokkaConfiguration {
+            sourceSets {
+                sourceSet {
+                    name = "jvm"
+                    displayName = "JVM"
+                    sourceRoots = listOf("$testDataDir/jvmMain/")
+                }
+            }
+            sourceSets {
+                sourceSet {
+                    name = "common"
+                    displayName = "common"
+                    analysisPlatform = "common"
+                    sourceRoots = listOf("$testDataDir/commonMain/")
+                }
+            }
+            sourceSets {
+                sourceSet {
+                    name = "native"
+                    displayName = "Native"
+                    analysisPlatform = "native"
+                    sourceRoots = listOf("$testDataDir/nativeMain/")
+                }
+            }
+        }
+
+        testFromData(configuration) {
+            documentablesCreationStage = {
+                assertEquals(1, it[0].packages.size)
+                assertEquals(1, it[0].packages[0].classlikes.size)
+            }
+        }
+    }
+
+    @Test
     fun inlineTestExample() {
         val configuration = dokkaConfiguration {
             sourceSets {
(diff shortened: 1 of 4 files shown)


plugins/base/src/test/kotlin/multiplatform/BasicMultiplatformTest.kt
```diff
@@ -25,6 +25,44 @@ class BasicMultiplatformTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    fun dataTestLongSparseArray() {
+        val testDataDir = getTestDataDir("multiplatform/longSparseArray").toAbsolutePath()
+
+        val configuration = dokkaConfiguration {
+            sourceSets {
```
vmishenev, line 33: If you use `expect`/`actual` declarations, you need to put dependency on a common source set like `dependentSourceSets = setOf(common.value.sourceSetID)`.

The full working code:
```
                val common =  sourceSet {
                    name = "common"
                    displayName = "common"
                    analysisPlatform = "common"
                    sourceRoots = listOf("$testDataDir/commonMain/")
                }
                sourceSet {
                    name = "jvm"
                    displayName = "JVM"
                    sourceRoots = listOf("$testDataDir/jvmMain/")
                    dependentSourceSets = setOf(common.value.sourceSetID)
                }
                sourceSet {
                    name = "native"
                    displayName = "Native"
                    analysisPlatform = "native"
                    sourceRoots = listOf("$testDataDir/nativeMain/")
                    dependentSourceSets = setOf(common.value.sourceSetID)
                }
```

plugins/base/src/test/kotlin/multiplatform/BasicMultiplatformTest.kt
```diff
@@ -25,6 +25,44 @@ class BasicMultiplatformTest : BaseAbstractTest() {
...
+                    displayName = "Native"
+                    analysisPlatform = "native"
+                    sourceRoots = listOf("$testDataDir/nativeMain/")
+                }
+            }
+        }
+
+        testFromData(configuration) {
+            documentablesCreationStage = {
+                assertEquals(1, it[0].packages.size)
```
vmishenev, line 60: It is not obvious. In creation stage modules come from PSI/Descriptors translators for each source set. We have two translators: `DefaultDescriptorToDocumentableTranslator` and `DefaultPSIToDocumentableTranslator`. In you case there are 3 source set. Then we will have 6 unmerged modules here:
`modules[0]` - java code of `jvm` source set.
`modules[1]` - kotlin code of `jvm` source set.
`modules[2]` - java code of `common` source set.
etc.

It is easier to use `documentablesMergingStage` instead of `documentablesCreationStage`.

plugins/base/src/test/kotlin/multiplatform/BasicMultiplatformTest.kt
```diff
@@ -25,6 +25,52 @@ class BasicMultiplatformTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    fun dataTestLongSparseArray() {
+        val testDataDir = getTestDataDir("multiplatform/longSparseArray").toAbsolutePath()
+
+        val configuration = dokkaConfiguration {
+            
+            val common =  sourceSet {
```
vmishenev, line 34: This should be in `sourceSets` block.

plugins/base/src/test/kotlin/multiplatform/BasicMultiplatformTest.kt
asfalcone: `documentablesMergingStage` and `documentablesTransformationStage` both only have jvm and native sourceSets

vmishenev:
[commonMain/androidx/collection/LongSparseArray.kt](https://raw.githubusercontent.com/Kotlin/dokka/102e968371ee0a87e5efb1c0b5ae6eb258ff5a0b/plugins/base/src/test/resources/multiplatform/longSparseArray/commonMain/androidx/collection/LongSparseArray.kt%20%20) has 2 redundant spaces in its name.

IgnatBeresnev:
Hi! What's the status of this PR? Wondering if there's anything left to do or if this can be closed (it seems the initial problem is no longer relevant)

JFYI we have some [multiplatform tests in integration-tests module](https://github.com/Kotlin/dokka/tree/master/integration-tests/gradle/projects/it-multiplatform-0)

asfalcone:
It can be closed. Thanks

Issue #2590, opened 2022-07-28
How do I select a Java file？


2017398956:
**Question**
when I  select a kotlin file ,like this:
     sourceRoots.setFrom(file("./xxx/xxx.kt"))
it works well.
But when i select a java file :
    sourceRoots.setFrom(file("./xxx/xxxx.java"))
it not woking .

Task:dokkaHtml

**Installation**
- Operating system: macOS 12.4
- Build tool: Gradle v7.3.3
- Dokka version: 1.7.10
- IDE:Android Studio Chipmunk | 2021.2.1 Patch 1

IgnatBeresnev:
Hi! Not sure why the behaviour is different for kotlin and java files.

Out of curiosity, is there any specific reason you need to select only one file? 

It works if you specify the folder path, like 
```kotlin
sourceRoots.setFrom(file("./xxx/xxx"))
```

2017398956:
> Hi! Not sure why the behaviour is different for kotlin and java files.
> 
> Out of curiosity, is there any specific reason you need to select only one file?
> 
> Pretty sure it works if you specify the folder path, like
> 
> ```kotlin
> sourceRoots.setFrom(file("./xxx/xxx"))
> ```

Yes, a java folder path is ok but  I only need partly java files of some folders , if using the whole folder is too many classes to my users.

daymxn:
> 

FWIW, you can also utilize `perPackageOption.matchingRegex` to filter files, which could be useful if you're needing a portion of files in a directory. Although, if you only need one or two files- it would definitely be overkill.

2017398956:
> > > Hi! Not sure why the behaviour is different for kotlin and java files.
> > > Out of curiosity, is there any specific reason you need to select only one file?
> > > Pretty sure it works if you specify the folder path, like
> > > ```kotlin
> > > sourceRoots.setFrom(file("./xxx/xxx"))
> > > ```
> > 
> > 
> > Yes, a java folder path is ok but I only need partly java files of some folders , if using the whole folder is too many classes to my users.
> 
> FWIW, you can also utilize `perPackageOption.matchingRegex` to filter files, which could be useful if you're needing a portion of files in a directory. Although, if you only need one or two files- it would definitely be overkill.

I've tried with this:

`tasks.withType(DokkaTask.class).configureEach { /*...*/
    dokkaSourceSets {
        moduleName.set("test")
        named("main"){
            sourceRoots.setFrom(file("./xxx/utils/"))
        }
[...]

PR #2602, opened 2022-08-03
Resolve issue where Java files cannot be selected separately

2017398956:
See https://github.com/Kotlin/dokka/issues/2590

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -248,7 +248,7 @@ open class HtmlRenderer(
                 }
             }.stripDiv()
             sourceSet to createHTML(prettyPrint = false).prepareForTemplates()
-                .div(classes = "content sourceset-depenent-content") {
+                .div(classes = "content sourceset-dependent-content") {
                     if (counter++ == 0) attributes["data-active"] = ""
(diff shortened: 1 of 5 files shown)


vmishenev (changes requested):
Thank you for contributing. 
To fix this properly, i think, it is better to:
1. Fix a type [here](https://github.com/2017398956/dokka/blob/47c87aa69ddc10fa08bbb4f3aebdb4e37cdfb14e/kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/AnalysisEnvironment.kt#L542). "java" should be instead of ".java".
2. Then needed files will be in `sourceRoots`. And a [function of `sourceRoots` processing ](https://github.com/Kotlin/dokka/blob/47c87aa69ddc10fa08bbb4f3aebdb4e37cdfb14e/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt#L85)  should be changed a little bit.
3. Can you add a test like [this](https://github.com/Kotlin/dokka/blob/47c87aa69ddc10fa08bbb4f3aebdb4e37cdfb14e/plugins/base/src/test/kotlin/linking/EnumValuesLinkingTest.kt#L21)?

2017398956:
> See #2590

Yes, after insteading **.java** with **java** it works well and I had commit a new pr : https://github.com/Kotlin/dokka/pull/2604

PR #2604, opened 2022-08-04
Fixed Java file extension spelling error and add a test

2017398956:
See https://github.com/Kotlin/dokka/issues/2590


IgnatBeresnev (approved):
Neat, thanks for investigating this!

PR #2611, opened 2022-08-07
Use OS color scheme to init light/dark mode (#2591)

### LINKED ISSUE #2591: Automatic dark mode
Follow up from https://github.com/Kotlin/dokka/issues/1731

The dark mode is nice 👏 . Would be nice to have an "automatic mode" that detects the current OS setting. This seems to be possible. See [this stackoverflow question](https://stackoverflow.com/questions/56393880/how-do-i-detect-dark-mode-using-javascript) for a possible implementation.

The dark mode would be a 3 way setting:
- light
- dark
- auto

pt2121:
Closes #2591

I tried to implement what mentioned in the issue:

> I think if you open up Dokka for the first time, it could choose light/dark mode based on OS theme, and then if the person decides to switch it to something else - remember that choice.

### DIFF
--- a/plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
+++ b/plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
@@ -32,15 +32,12 @@ const darkModeSwitch = () => {
     const localStorageKey = "dokka-dark-mode"
     const storage = localStorage.getItem(localStorageKey)
     const osDarkSchemePreferred = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
-    const savedDarkMode = storage ? JSON.parse(storage) : osDarkSchemePreferred
+    const darkModeEnabled = storage ? JSON.parse(storage) : osDarkSchemePreferred
     const element = document.getElementById("theme-toggle-button")
-    const enabledClasses = document.getElementsByTagName("html")[0].classList
-    initPlayground(savedDarkMode ? samplesDarkThemeName : samplesLightThemeName)
-    if (savedDarkMode != enabledClasses.contains('theme-dark')) {
-        enabledClasses.toggle("theme-dark");
-    }
+    initPlayground(darkModeEnabled ? samplesDarkThemeName : samplesLightThemeName)
 
     element.addEventListener('click', () => {
+        const enabledClasses = document.getElementsByTagName("html")[0].classList
         enabledClasses.toggle("theme-dark")
 
         //if previously we had saved dark theme then we set it to light as this is what we save in local storage
--- a/plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
+++ b/plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
@@ -31,12 +31,16 @@ window.addEventListener('load', () => {
 const darkModeSwitch = () => {
     const localStorageKey = "dokka-dark-mode"
     const storage = localStorage.getItem(localStorageKey)
-    const savedDarkMode = storage ? JSON.parse(storage) : false
+    const osDarkSchemePreferred = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+    const savedDarkMode = storage ? JSON.parse(storage) : osDarkSchemePreferred
     const element = document.getElementById("theme-toggle-button")
+    const enabledClasses = document.getElementsByTagName("html")[0].classList
     initPlayground(savedDarkMode ? samplesDarkThemeName : samplesLightThemeName)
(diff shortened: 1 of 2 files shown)


plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
```diff
@@ -31,12 +31,16 @@ window.addEventListener('load', () => {
...
     const storage = localStorage.getItem(localStorageKey)
-    const savedDarkMode = storage ? JSON.parse(storage) : false
+    const osDarkSchemePreferred = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+    const savedDarkMode = storage ? JSON.parse(storage) : osDarkSchemePreferred
     const element = document.getElementById("theme-toggle-button")
+    const enabledClasses = document.getElementsByTagName("html")[0].classList
     initPlayground(savedDarkMode ? samplesDarkThemeName : samplesLightThemeName)
+    if (savedDarkMode != enabledClasses.contains('theme-dark')) {
+        enabledClasses.toggle("theme-dark");
+    }
```
IgnatBeresnev, line 41: If you have a look at each `.html` page, you'll see this script in the header (inspect [this page](https://kotlinlang.org/api/kotlinx.coroutines/)):
```js
<script>const storage = localStorage.getItem("dokka-dark-mode")
const savedDarkMode = storage ? JSON.parse(storage) : false
if(savedDarkMode === true){
    document.getElementsByTagName("html")[0].classList.add("theme-dark")
}</script>
```
This script is added in [this template](https://github.com/Kotlin/dokka/blob/master/plugins/base/src/main/resources/dokka/templates/base.ftl#L12-L18) with the following comment:
> This script doesn't need to be there but it is nice to have
> since app in dark mode doesn't 'blink' (class is added before it is rendered) 

So I'm thinking maybe it's better to have toggling of themes in one place only, that is in the header script of each page (so add it to the template). This way it's consistent + possibly avoids theme blinking. Something like

```js
const storage = localStorage.getItem("dokka-dark-mode")
if (storage == null) {
    // use OS preferred theme
} else {
    // use theme from storage
}
```

_______

Code inside this file (`platform-content-handler.js`) is mostly adding a click event listener + initizling playground (a.k.a. runnable samples) with the correct theme, so your code is good here. I would just move the class toggling part and leave everything else

plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
```diff
@@ -31,12 +31,16 @@ window.addEventListener('load', () => {
 const darkModeSwitch = () => {
     const localStorageKey = "dokka-dark-mode"
     const storage = localStorage.getItem(localStorageKey)
-    const savedDarkMode = storage ? JSON.parse(storage) : false
+    const osDarkSchemePreferred = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches
+    const savedDarkMode = storage ? JSON.parse(storage) : osDarkSchemePreferred
```
IgnatBeresnev, line 35: I think `savedDarkMode` can be renamed to something like `darkModeEnabled` since fallback value is `osDarkSchemePreferred`, and OS theme is not really "saved"

IgnatBeresnev (commented):
Hi! I think you're on the right track! Thanks for deciding to implement this

plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
pt2121: Thanks for the detailed explanation! I will move the logic for toggling the theme into the template.

plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
pt2121: Agree, I'll rename it.

pt2121:
I updated my PR.
If a user has never select a theme, do you want to change the theme when the user changes the OS theme as well?
If so, I can add the listener and toggle the class.

IgnatBeresnev:
> If a user has never selected the Dokka theme, do you want to change the theme when the user changes the OS theme as well?

If you want, I don't think we'll object :) But I think it's fine the way it is - navigating to any page would trigger the script in the header, so the new theme should be picked up. We'll probably also have to re-do it anyway if/when Dokka migrates to react routing

IgnatBeresnev (approved):
LGTM, although I'm unable to test it properly (running a custom linux theme, so it would mess everything up). @vmishenev if you are able and have the theme toggle, could you please check?

Thanks once again for taking the initiative!

vmishenev (approved):
I have tested it via simulation in a debugger. It works if a theme switcher has not been used.

PR #2613, opened 2022-08-08
Do not render platform tabs for common-only content

IgnatBeresnev:
See #2576 

The main idea is that declarations without platform tabs can be considered to be "common" by default

The problem with rendering single "common" platform tab everywhere is that it attracts too much attention with the current changes, but doesn't bring that much value. This ought to make package overview page more minimalistic.

### DIFF
--- a/core/src/main/kotlin/model/DisplaySourceSet.kt
+++ b/core/src/main/kotlin/model/DisplaySourceSet.kt
@@ -8,12 +8,14 @@ import org.jetbrains.dokka.utilities.SelfRepresentingSingletonSet
 data class DisplaySourceSet(
     val sourceSetIDs: CompositeSourceSetID,
     val name: String,
-    val platform: Platform
+    val platform: Platform,
+    val dependentSourceSetIDs: Set<DokkaSourceSetID> = emptySet()
 ) : SelfRepresentingSingletonSet<DisplaySourceSet> {
     constructor(sourceSet: DokkaSourceSet) : this(
         sourceSetIDs = CompositeSourceSetID(sourceSet.sourceSetID),
         name = sourceSet.displayName,
-        platform = sourceSet.analysisPlatform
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/model/DisplaySourceSet.kt
```diff
@@ -8,12 +8,14 @@ import org.jetbrains.dokka.utilities.SelfRepresentingSingletonSet
 data class DisplaySourceSet(
     val sourceSetIDs: CompositeSourceSetID,
     val name: String,
-    val platform: Platform
+    val platform: Platform,
+    val dependentSourceSetIDs: Set<DokkaSourceSetID> = emptySet()
```
IgnatBeresnev, line 12: not ideal, but I guess this could be useful if someone decides to build a graph of source set dependencies or something

plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -953,8 +953,8 @@ td.content {
     padding-bottom: 8px;
 }
 
-.table-row .with-platform-tabs .sourceset-dependent-content .brief,
-.table-row .with-platform-tabs .sourceset-dependent-content .inline-comment {
+.table-row .platform-hinted .sourceset-dependent-content .brief,
+.table-row .platform-hinted .sourceset-dependent-content .inline-comment {
```
IgnatBeresnev, line 957: Because there's no tabs, it renders the pages the same way it would for single platform projects, and there's currently the same problem - no padding. See [this page](https://nomisrev.github.io/kotlinx-serialization-jsonpath/-kotlin-x%20-serialization%20-json-path/io.github.nomisrev/index.html#993988515%2FFunctions%2F788909594) for instance

This should solve it for both single and multi platform pages

vmishenev (changes requested):

vmishenev (approved):

Issue #2596, opened 2022-07-29
Documentation generation is flaky


owengray-google:
**Describe the bug**
When running dokka, some files are not generated. This can apparently happen to any file, though it seems orders of magnitude more common for `index.html` and other files near the root of the output hierarchy. I suspect that this is due to file lock contention by the output writer.

We first discovered this by adding code that enforces that all expected files in our dackka integration tests ("goldens") were actually generated (previously we had only enforced that all generated files were expected).

This seems like it would be a huge problem for us and for you, because it could result in missing files in `developer.android.com`/`kotlinlang.org` (depending on how your update process works and how it removes files that are *intended* to cease being generated from one version to the next.)

From a test and a rough estimate, out of 130,000 files generated by running unmodified upstream dokka on `androidx` `framework/support`, 400 files were generated in the second run that were not generated in the first. Around a quarter of these (100 files) were `index.html`, representing a flake rate for `index.html` of around `20%`, and for other files of `0.2%`.

- Dokka version: `1.7.10`

Replication process:
-Copy `frameworks/support` to an `upstream-dokka` checkout.
--Delete `compose/ui/ui-test/src/androidAndroidTest` because dokka fails if you don't (throws in `org.jetbrains.kotlin.types.expressions`--looks like the compiler).
-run `./gradlew shadowJar` in the upstream checkout.
-run `java -jar ./runners/cli/build/libs/dokka-cli-1.7.10-SNAPSHOT.jar -pluginsClasspath "./integration-tests/cli/build/libs/fat-base-plugin-1.7.10-SNAPSHOT.jar" -sourceSet "-src ./frameworks/support/"` in the upstream checkout.
-add `dokka/root` to `git` or otherwise copy it for later comparison.
-run `java -jar ./runners/cli/build/libs/dokka-cli-1.7.10-SNAPSHOT.jar -pluginsClasspath "./integration-tests/cli/build/libs/fat-base-plugin-1.7.10-SNAPSHOT.jar" -sourceSet "-src ./frameworks/support/"` again.
-notice that there is a diff (e.g. `git status`) between the docs generated by each run. I found roughly `400` out of `130,000` `.html` files that were generated by the second run but not by the first.

owengray-google:
When we run this through dackka, we notice the heightened flake rate for all files in the root of the package tree like `index.html`, such as `package-list`, `classes.html`, and `packages.html`, which upstream dokka does not seem to generate.
I think this points further towards file lock contention.

[...]

PR #2619, opened 2022-08-10
Fix generic types caching

vmishenev:
It caused flaky generation of type links for generic type params  (#2596)

### DIFF
--- a/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
@@ -534,7 +534,8 @@ class DefaultPsiToDocumentableTranslator(
                             )
 
                             else -> {
-                                // cache types that has no annotation and no type parameter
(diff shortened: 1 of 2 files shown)


plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -521,33 +532,40 @@ class DefaultPsiToDocumentableTranslator(
...
                                 extra = type.annotations()
                             )
-                            else -> GenericTypeConstructor(
-                                DRI.from(resolved),
-                                type.parameters.map { getProjection(it) },
-                                extra = type.annotations()
-                            )
+
+                            else -> {
+                                // cache types that has no annotation and no type parameter
```
IgnatBeresnev, line 537: Side note: this comments explains the code, but doesn't explain the intention behind it, so it's not really clear why we can't cache type parameters. Describing a use case where this can hurt would be nice

IgnatBeresnev (approved):

PR #2627, opened 2022-08-17
Fix incorrect handling of static members used within `@see` tag

### LINKED ISSUE #2626: Multiple `@see` tags for imported static functions are missing from documentation
If you specify multiple `@see` tags that all have imported static functions from the same class, only the last one will be visible in generated documentation, and with the wrong link (to the class intead of the function)

### Reproducer

File 1:
```kotlin
package me.beresnev.dokka.debug

import me.beresnev.dokka.debug.util.CollectionExtensions.emptyMap
import me.beresnev.dokka.debug.util.CollectionExtensions.emptyList
import me.beresnev.dokka.debug.util.CollectionExtensions.emptySet

/**
 * @see [List] check the listing!
 * @see [Map] paging Lewis and Clark
 * @see [emptyMap] empty map
 * @see [emptyList] empty list
 * @see [emptySet] empty set
 */
fun foo() {}
```

File 2:
```kotlin
package me.beresnev.dokka.debug.util

object CollectionExtensions {
    fun emptyList() {}
    fun emptyMap() {}
    fun emptySet() {}
}
```

### Actual

Initially reported [in Slack](https://kotlinlang.slack.com/archives/C0F4UNJET/p1660594888084379) by @evanchooly

IgnatBeresnev:
Fixes #2626

### DIFF
--- a/plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
+++ b/plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
@@ -12,6 +12,7 @@ import org.intellij.markdown.flavours.gfm.GFMFlavourDescriptor
 import org.intellij.markdown.flavours.gfm.GFMTokenTypes
 import org.jetbrains.dokka.base.parsers.factories.DocTagsFromIElementFactory
 import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.links.PointingToDeclaration
 import org.jetbrains.dokka.model.doc.*
 import org.jetbrains.dokka.model.doc.Suppress
 import org.jetbrains.kotlin.kdoc.parser.KDocKnownTag
@@ -51,7 +52,7 @@ open class MarkdownParser(
                 val dri = externalDri(referencedName)
                 See(
                     parseStringToDocNode(content.substringAfter(' ')),
-                    dri?.fqName() ?: referencedName,
+                    dri?.fqDeclarationName() ?: referencedName,
                     dri
                 )
             }
@@ -59,7 +60,7 @@ open class MarkdownParser(
                 val dri = externalDri(content.substringBefore(' '))
                 Throws(
                     parseStringToDocNode(content.substringAfter(' ')),
-                    dri?.fqName() ?: content.substringBefore(' '),
+                    dri?.fqDeclarationName() ?: content.substringBefore(' '),
                     dri
                 )
             }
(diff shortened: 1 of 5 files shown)


plugins/base/src/main/kotlin/parsers/MarkdownParser.kt
```diff
@@ -567,8 +568,20 @@ open class MarkdownParser(
         }
 
         //Horrible hack but since link resolution is passed as a function i am not able to resolve them otherwise
+        @kotlin.Suppress("DeprecatedCallableAddReplaceWith")
+        @Deprecated("This function makes wrong assumptions and is missing a lot of corner cases related to generics, " +
+                "parameters and static members. This is not supposed to be public API and will not be supported in the future")
         fun DRI.fqName(): String? = "$packageName.$classNames".takeIf { packageName != null && classNames != null }
```
IgnatBeresnev, line 574: The overall problem is basically here.

For `com.example.Object.function()` it would return `com.example.Object`, so if you have multiple `@see` tags that point to multiple different functions from the same `object` - there would be name collisions, and some members would be discarded by `groupBy` further down below.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -590,26 +591,31 @@ open class DefaultPageCreator(
...
                                             sourceSets = setOf(platform),
                                             kind = ContentKind.Comment,
                                             styles = this@group.mainStyles,
                                         ) {
-                                            it.address?.let { dri ->
+                                            seeTag.address?.let { dri ->
+                                                // only declaration links are expected to be fully qualified
+                                                val linkText = dri.takeIf { it.target is PointingToDeclaration }
+                                                    ?.let { seeTag.name.removePrefix("${it.packageName}.") }
+                                                    ?: seeTag.name
```
IgnatBeresnev, line 604: Turns out, `@see` can also be used with local params (we have tests for it) and generics. And with functions, of course, so using `classNames` was wrong here, my bad

plugins/base/src/test/kotlin/content/seealso/ContentForSeeAlsoTest.kt
```diff
@@ -517,76 +517,73 @@ class ContentForSeeAlsoTest : BaseAbstractTest() {
     }
 
     @Test
-    fun `undocumented see also from java`(){
```
IgnatBeresnev, line 520: Moved two tests to `JavadocParserTest` as they are not about content, but verify documentables model.

vmishenev (approved):

PR #2633, opened 2022-08-21
Add hotkey for opening search dialog

### LINKED ISSUE #2265: Hotkey to open HTML search
It's normal to have a way to open the search bar using a hotkey e.g. `/` or Cmd-K.

atyrin:
Add a `Cmd + K`/`Ctrl + K` hotkey for opening the search window.

* Meta key `Cmd`/`Ctrl` depends on OS. For macOS -- `Cmd`, for every other -- `Ctrl`.
* The search button have a tooltip with the name of key combination. The tooltip appears after 500 ms on mouse over.
* The combination prevents default hotkey actions like the kotlinlang.org or github.com behave. E.g. the `Cmd+K` combination is default option to focus on the omnibox in Firefox.

* Tested on Chrome/Filefox/Safary on macOS, on Edge/Firefox on Windows

Fix #2265

### DIFF
--- a/plugins/base/frontend/src/main/components/utils/hotkey.ts
+++ b/plugins/base/frontend/src/main/components/utils/hotkey.ts
@@ -1,0 +1,53 @@
+import {detectOsKind, OsKind} from "./os";
+
+type ModifierKey = {
+    name: string
+    keyArg: string
+}
+
+class ModifierKeys {
+    static metaKey: ModifierKey = {name: "Command", keyArg: "Meta"}
+    static ctrlKey: ModifierKey = {name: "Ctrl", keyArg: "Control"}
+    static altKey: ModifierKey = {name: "Alt", keyArg: "Alt"}
+    static shiftKey: ModifierKey = {name: "Shift", keyArg: "Shift"}
+}
+
+const setOfKeys = [ModifierKeys.altKey, ModifierKeys.shiftKey, ModifierKeys.ctrlKey, ModifierKeys.metaKey]
+
+export class Hotkey {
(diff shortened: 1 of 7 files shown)


atyrin:
There is an issue, that the hotkey will be triggered even if the key combination is a subset of another key combination (e.g. `Cmd` + `Shift` + `K`)

plugins/base/frontend/src/main/components/utils/hotkey.ts
```diff
@@ -0,0 +1,53 @@
...
+    public registerHotkeyWithAccel = (event: () => void, letter: string) => {
+        const osMetaKey = this.getOsAccelKey()
+        document.onkeydown = (keyDownEvent) => {
+            const isMetaKeyPressed = keyDownEvent.getModifierState(osMetaKey.keyArg)
+            const isOtherModifierKeyPressed = setOfKeys
+                .filter(key => key !== osMetaKey)
+                .map((otherKeys: ModifierKey) => keyDownEvent.getModifierState(otherKeys.keyArg))
+                .some(value => value)
+
+            if (isMetaKeyPressed && !isOtherModifierKeyPressed && keyDownEvent.key === letter) {
```
IgnatBeresnev, line 37: I'd leave a comment here like you did in the PR comments - it might not be obvious at first sight. Something like

> We don't want to trigger action for hotkeys like Ctrl+K if is a subset of another key combination (e.g. Ctrl + Shift + K)

IgnatBeresnev (approved):
Nicely done! Especially the tooltip :)

vmishenev (approved):

PR #2641, opened 2022-08-24
Enable Gradle Enterprise

Goooler:
https://docs.gradle.com/enterprise/gradle-plugin/


Goooler:
We can use this to speed up builds, if necessary we can publish the build scan on CI only like this:
https://github.com/apache/groovy/blob/3405c7be6d64f462a76b6a864aecdbdcbbe80e89/gradle/build-scans.gradle#L20-L55

IgnatBeresnev:
> if necessary we can publish the build scan on CI only like this

Oooh, yes, that would make me happy :) Certainly would not hurt

Goooler:
https://github.com/Kotlin/dokka/actions/runs/2920255208

image

vmishenev (approved):
As for me, it is extremely useful. Thank you for contributing.

IgnatBeresnev (approved):
Thanks!

PR #2642, opened 2022-08-24
Do not render constructor pages and blocks and for annotation classes

IgnatBeresnev:
See #2576

TBD what to do with primary constructors


IgnatBeresnev:
Primary constructor in the main class signature is here to stay. It's probably what most people look at and it shows default values, which you don't see in "Properties" block

vmishenev (approved):

PR #2643, opened 2022-08-24
Underline `@param` tag key for more consistency

IgnatBeresnev:
Styling TBD

### DIFF
--- a/plugins/base/src/main/resources/dokka/styles/style.css
+++ b/plugins/base/src/main/resources/dokka/styles/style.css
@@ -686,6 +686,24 @@ a {
     text-decoration: none;
 }
 
+u {
+    /*text-decoration: none;*/
+    /*padding-bottom: 2px;*/
+    /*border-bottom: 1px solid var(--border-color);*/
+
+    text-decoration-line: underline;
+    text-underline-offset: 5px;
+    text-decoration-color: #585858;
+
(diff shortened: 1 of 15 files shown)


plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -1038,7 +1056,7 @@ td.content {
 
 @media print, screen and (min-width: 960px) {
     .keyValue {
-        grid-template-columns: 25% 75%;
+        grid-template-columns: 20% 80%;
```
IgnatBeresnev, line 1059: Revert to 1.7.10 behaviour. It looks good if `keyValue` table is part of the main content, but first column takes too much space if parameters are within tabbed content (there's not much space as it is)

IgnatBeresnev:
Researched making parameter names clickable (focus signature on click), and it seems a bit too complicated to implement, not sure it's worth the effort now.

The main problem is when you have multiple signatures on the same page and multiple param blocks, especially if you have 4 signatures and only 2 param blocks. Ideally we'd need to add anchor ids/tags to signatures for them to be focusable, and write some js code around it - preferably so that it can be reused by other components, so something universal.

If it becomes a usability problem where people expect param keys to be clickable links, it can be implemented, but I think it can be skipped for now to not overcomplicate things

vmishenev (approved):

PR #2645, opened 2022-08-29
Add `DynamicTypeDeserializer`

vmishenev:


integration-tests/gradle/projects/it-js-ir-0/src/main/kotlin/RootPackageClass.kt
```diff
@@ -18,4 +20,6 @@ fun test(list: MutableList<Int>) = "list"
 
 @JsModule("is-sorted")
 @JsNonModule
-external fun <T> sorted(a: Array<T>): Boolean
\ No newline at end of file
+external fun <T> sorted(a: Array<T>): Boolean
+
+external interface TextLinkProps: AnchorHTMLAttributes<HTMLAnchorElement>
```
IgnatBeresnev, line 25: If something were to break again, this would catch it, right? Not a JS expert, so I fail to understand why - is it considered to be a dynamic type or something? 

Maybe leaving a comment there would not hurt, so that other people do not remove it. i.e
```
// do not remove, this declaration is used to check presence of dynamic type deserializer
```

IgnatBeresnev (approved):
:+1:

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/resolve/DokkaJsResolverForModuleFactory.kt
```diff
@@ -82,7 +90,7 @@ internal class DokkaJsResolverForModuleFactory(
                 moduleInfo.kotlinLibrary
                     .createKlibPackageFragmentProvider(
                         storageManager = moduleContext.storageManager,
-                        metadataModuleDescriptorFactory = metadataFactories.DefaultDeserializedDescriptorFactory,
```
IgnatBeresnev, line 85: Also a bit unclear where `DefaultDeserializedDescriptorFactory` went - is it now included in `DefaultDescriptorFactory`?

PR #2647, opened 2022-08-29
Handle `PsiImmediateClassType` as `PsiClassType` to resolve bounds

### LINKED ISSUE #2646: Java Enums used in declaration are rendered as Any
`public JavaEnum returningEnumType()` is rendered as `fun returningEnumType(): Any`

### Reproducer:

```java
// same package

// File 1: JavaEnum.java
public enum JavaEnum {
    FOO, BAR
}

// File 2: ContainingEnumType.java
public class ContainingEnumType {

    public JavaEnum returningEnumType() {
        return null;
    }

    public JavaEnum[] returningEnumTypeArray() {
        return null;
    }

    public void acceptingEnumType(JavaEnum javaEnum) {}
}
```

IgnatBeresnev:
Fixes #2646

### DIFF
--- a/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
@@ -513,7 +513,7 @@ class DefaultPsiToDocumentableTranslator(
             }
 
             return when (type) {
-                is PsiClassReferenceType ->
+                is PsiClassType ->
                     type.resolve()?.let { resolved ->
                         when {
                             resolved.qualifiedName == "java.lang.Object" -> type.cacheBoundIfHasNoAnnotation { annotations -> JavaObject(annotations.annotations()) }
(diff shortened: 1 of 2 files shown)


plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -564,8 +564,6 @@ class DefaultPsiToDocumentableTranslator(
 
                 is PsiPrimitiveType -> if (type.name == "void") Void
                 else type.cacheBoundIfHasNoAnnotation { annotations -> PrimitiveJavaType(type.name, annotations.annotations()) }
-                is PsiImmediateClassType ->
```
IgnatBeresnev, line 567: Not sure why it was done this way. Probably an exception was thrown and mapping `PsiImmediateClassType -> JavaObject` was added without researching.

`PsiImmediateClassType` and `PsiClassReferenceType` are both extending `PsiClassType` (`resolve()` is also part of the latter), so hopefully it's all good

vmishenev (approved):

PR #2652, opened 2022-09-03
Migrate buildSrc to composite build

Goooler:
Ref https://github.com/detekt/detekt/pull/4090

https://docs.gradle.org/7.5.1/userguide/composite_builds.html


build.gradle.kts
```diff
@@ -5,12 +5,14 @@ plugins {
     kotlin("jvm") apply false
     id("java")
     id("org.jetbrains.dokka") version "1.7.20"
-    id("io.github.gradle-nexus.publish-plugin")
+    id("io.github.gradle-nexus.publish-plugin") version "1.1.0"
+    id("build-logic") apply false
```
IgnatBeresnev, line 9: Is there any specific reason that `build-logic` is not applied here, but applied within `allprojects {}`? Makes me think it'll break something if I remove `apply false` here and `apply(plugin = "build-logic")` down below, but nothing bad seems to happen

build-logic/build.gradle.kts
```diff
@@ -6,8 +6,17 @@ repositories {
...
 }
 
+gradlePlugin {
+    plugins {
+        create("build-logic") {
+            id = "build-logic"
+            implementationClass = "org.jetbrains.BuildLogic"
+        }
+    }
+}
```
IgnatBeresnev, line 16: I've had a look at the referece pull requests with the same change, and none of them create a separate plugin. 

However, I have noticed that all of them have separate build script files with the name pattern like `kotlin-jvm-convention.gradle.kts` that they then apply as a plugin in `apply(plugin = "kotlin-jvm-convention")`. 

I came to the conclusion that it's the idiomatic way of sharing build logic, and what Dokka is doing with directly referencing the code from `buildSrc` it not very idiomatic, but it needs to work as before - that's why this plugin, which imports all classes, was created. 

Is that the case?

IgnatBeresnev (commented):
Sorry it took so long to get to, I'm learning Gradle from your PRs and had to read up on it :)

Overall looks good and useful, I only have minor comments/questions

build-logic/build.gradle.kts
```diff
@@ -6,8 +6,17 @@ repositories {
     gradlePluginPortal()
```
IgnatBeresnev, line 6: Comment for the `build-logic` directory: this build-logic pattern will be obvious for people who are proficient with Gradle, but not everyone is, unfortunately. 

Can you please create a `README.md` in the root of the `build-logic` directory and add some description for why this module exists and what it does? It would help with making Dokka beginner and contributor friendly, and will hopefully help the future generations of maintainers.

The text is up to you, but here are the things worth mentioning in my opinion:

* State that it allows to encapsulate complex build logic that also can be shared between projects, similar to [buildSrc](https://docs.gradle.org/7.6/userguide/organizing_gradle_projects.html#sec:build_sources).
* State that `build-logic` is a common pattern, and it does not cause the whole project to become out of date when changes are made, as opposed to `buildSrc`. Attaching a link to an article would also be useful, for instance https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3

build.gradle.kts
Goooler: If so, we should apply it in `plugins` and `subprojects`, to avoid redundant applying.

Actually, we should just apply `build-logic` in where needed instead of all projects

build-logic/build.gradle.kts
Goooler: Yep, they depend on plugins in `.gradle.kts` but we use functions, it not the same.

IgnatBeresnev:
Awesome, thanks! 

I noticed that #2704 is very related and it should hopefully make the `BuildLogic` class unnecessary. However, because it has more changes, it'll be more difficult to rebase than this PR, which mostly contains moved classes.

So if you don't mind, let's wait for #2704 to be merged first, and then hopefully it'll be easy for you to rebase onto it and just move all new/old classes without having to deal with merge conflicts

Goooler:
No problem.

Goooler:
Maybe we no need this anymore once https://github.com/Kotlin/dokka/pull/2652 merged, see https://docs.gradle.org/8.0/release-notes.html#improvements-for-buildsrc-builds.

3flex:
> Maybe we no need this anymore

It's still required. None of the improvements in Gradle 8 change the fact that buildSrc changes cause the whole project to become out of date - there's still a warning about this in the latest Gradle docs.

Goooler:
Yeah, this PR will go on after #2704 merged.

IgnatBeresnev:
#2704 has been finally merged :) Could you please rebase onto master or merge it in? I'll have a look right away

Since we utilize convention plugins, `class BuildLogic : Plugin<Project>` won't be needed anymore, right?

Goooler:
Addressed to #2912.

PR #2659, opened 2022-09-08
Wrap long signatures dynamically based on taken client width

IgnatBeresnev:
See #2576

### DIFF
--- a/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
+++ b/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
@@ -22,6 +22,7 @@ const wrapSymbolParameters = (symbol) => {
         setTimeout(function() {
             wrapSymbolParameters(symbol);
         }, 100)
+        return
     }
 
     let innerTextWidth = Array.from(symbol.children)
--- a/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
+++ b/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
@@ -1,0 +1,63 @@
+// helps with some corner cases where <wbr> starts working already,
+// but the signature is not yet long enough to be wrapped
+const leftPaddingPx = 60
+
+const wrapAllSymbolParameters = () => {
+    document.querySelectorAll("div.symbol").forEach(symbol => wrapSymbolParameters(symbol))
+}
+
+const wrapSymbolParameters = (symbol) => {
+    let symbolBlockWidth = symbol.clientWidth
+    let innerTextWidth = Array.from(symbol.children)
+        .map(it => it.getBoundingClientRect().width).reduce((a, b) => a + b, 0)
+
+    // if signature text takes up more than a single line, wrap params for readability
+    let shouldWrapParams = innerTextWidth > (symbolBlockWidth - leftPaddingPx)
+    if (shouldWrapParams) {
+        let parameters = symbol.querySelector("span.parameters")
+        if (parameters != null) {
+            parameters.classList.add("wrapped")
+            parameters.querySelectorAll("span.parameter").forEach(param => {
+                // has to be a physical indent so that it can be copied. styles like
+                // paddings and `::before { content: "    " }` do not work for that
+                param.prepend(createNbspIndent())
+            })
+        }
+    }
+}
+
+const createNbspIndent = () => {
+    let indent = document.createElement("span")
+    indent.append(document.createTextNode("\u00A0\u00A0\u00A0\u00A0"))
+    indent.classList.add("nbsp-indent")
+    return indent
+}
+
+const resetAllSymbolParametersWrapping = () => {
+    document.querySelectorAll("div.symbol").forEach(symbol => resetSymbolParametersWrapping(symbol))
+}
+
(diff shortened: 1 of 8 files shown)


plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
```diff
@@ -0,0 +1,63 @@
...
+
+    // if signature text takes up more than a single line, wrap params for readability
+    let shouldWrapParams = innerTextWidth > (symbolBlockWidth - leftPaddingPx)
+    if (shouldWrapParams) {
+        let parameters = symbol.querySelector("span.parameters")
+        if (parameters != null) {
+            parameters.classList.add("wrapped")
+            parameters.querySelectorAll("span.parameter").forEach(param => {
+                // has to be a physical indent so that it can be copied. styles like
+                // paddings and `::before { content: "    " }` do not work for that
```
vmishenev, line 22: https://stackoverflow.com/questions/16552397/css-how-to-add-white-space-before-elements-content

It can improve performance.

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
```diff
@@ -0,0 +1,63 @@
...
+// but the signature is not yet long enough to be wrapped
+const leftPaddingPx = 60
+
+const wrapAllSymbolParameters = () => {
+    document.querySelectorAll("div.symbol").forEach(symbol => wrapSymbolParameters(symbol))
+}
+
+const wrapSymbolParameters = (symbol) => {
+    let symbolBlockWidth = symbol.clientWidth
+    let innerTextWidth = Array.from(symbol.children)
```
vmishenev, line 11: Should it ignore annotations? 
image

IgnatBeresnev: Yeah, good point! it should probably ignore all blocks

also on your screnshot it looks like there's 8 spaces instead of 4 :thinking:

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
IgnatBeresnev: I tried JS fiddle from the answer, but when you select the next - it's not selecting spaces and thus not copying it to clipboard (it will copy the text without spaces), which is the biggest problem :( 

I'll give it a shot, but it doesn't look like it's going to work

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
vmishenev: Sorry, it happened due to my debugger.

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
IgnatBeresnev: Yeah, no luck:


And here's if I copy (Ctrl+C) it: 
```kotlin
fun <T> CoroutineScope.async(
context: CoroutineContext = EmptyCoroutineContext, 
start: CoroutineStart = CoroutineStart.DEFAULT, 
block: suspend CoroutineScope.() -> T
): Deferred<T>
```

vmishenev (approved):
Ok, but the content is shifted after I resize a window.

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
```diff
@@ -13,6 +13,17 @@ const wrapSymbolParameters = (symbol) => {
     }
 
     let symbolBlockWidth = symbol.clientWidth
+
+    // Even though the script is marked as `defer` and we wait for `DOMContentLoaded` event,
+    // it can happen that `symbolBlockWidth` is 0, indicating that something hasn't been loaded.
+    // Re-try after some time, should work. Should not go into infinite recursion because
```
vmishenev, line 19: Have you tried [to force restyle](https://stackoverflow.com/questions/18564942/clean-way-to-programmatically-use-css-transitions-from-js/31862081#31862081)?

IgnatBeresnev: I did try using `getComputedStyle(elem)` when I started implementing this change and came across similar problems, but it didn't work (returned all 0s), so I had to make the script `defer`, which helped. Haven't tried it in this scenario as it's only reproducible if hosted on GitHub pages for some reason (and not always).

Anyhow, relying on undocumented behaviour from 2012 seems shady, `setTimeout` for 100 millis should be more than enough and should work in all browsers.

But if you propose a specific code suggestion, we can try. Will have to test it in all browsers and pray to god nothing changes though.

IgnatBeresnev:
@vmishenev updated the PR. Got rid of timeout logic.

JFYI, may be useful in the future for these kinds of things: turns out, in Chrome's dev menu, in Network tab, you can enable network throttling, so resources take much more time to load. With this enabled, the issue is easily reproducible even locally.

vmishenev (approved):

PR #2661, opened 2022-09-08
Trim four-spaces inside indented code block

vmishenev:
By [CommonMark spec](https://spec.commonmark.org/0.30/#indented-code-blocks), we should omit 4 spaces of indentation.
It should be fixed in IDEA as well:
image


IgnatBeresnev (approved):
:+1:

PR #2664, opened 2022-09-12
Extract classpath from `KotlinSharedNativeCompilation` as well

vmishenev:
e.g. StdLib was missed.

### DIFF
--- a/integration-tests/gradle/projects/it-multiplatform-0/gradle.properties
+++ b/integration-tests/gradle/projects/it-multiplatform-0/gradle.properties
@@ -1,3 +1,5 @@
 dokka_it_kotlin_version=1.7.10
+#these flags are enabled by default since 1.6.20.
+#remove when this test is executed with Kotlin >= 1.6.20
 kotlin.mpp.enableGranularSourceSetsMetadata=true
 kotlin.native.enableDependencyPropagation=false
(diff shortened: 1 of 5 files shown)


integration-tests/gradle/projects/it-multiplatform-0/gradle.properties
```diff
@@ -1 +1,3 @@
 dokka_it_kotlin_version=1.7.10
+kotlin.mpp.enableGranularSourceSetsMetadata=true
+kotlin.native.enableDependencyPropagation=false
```
IgnatBeresnev, line 3: What happens if you don't specify these two flags, will it not work?

I don't think our average user is going to have that enabled

vmishenev: These flags are enabled by default since Kotlin 1.6.20. See [HMPP](https://kotlinlang.org/docs/multiplatform-hierarchy.html)
It will not work for earlier Kotlin versions.

IgnatBeresnev: Oh, sweet. 

A comment would not hurt, imo. Otherwise future maintainers might be scared to remove it. Something along the lines of
```
// these flags are enabled by default since 1.6.20.
// remove when this test is executed with Kotlin >= 1.6.20
```

IgnatBeresnev (approved):

PR #2670, opened 2022-09-14
Update `prism.js` to match new webhelp highlight colors

IgnatBeresnev:


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -792,7 +800,6 @@ open class HtmlRenderer(
             TextStyle.Strong -> strong { body() }
             TextStyle.Var -> htmlVar { body() }
             TextStyle.Underlined -> underline { body() }
-            is TokenStyle -> span("token " + styleToApply.toString().toLowerCase()) { body() }
```
IgnatBeresnev, line 795: Previously, if a text element had >2 `TokenStyle` styles, it was rendered as
```html
span("token first") {
    span("token second") {
        text
    }
}
```

plugins/base/src/main/resources/dokka/styles/prism.css
```diff
@@ -1,40 +1,92 @@
...
-	--property-color: #9876aa;
-	--function-color: #ffc66d;
-}
-
+/*
+ * Custom Dokka styles
+ */
 code .token {
-    white-space: pre;
+	white-space: pre;
```
IgnatBeresnev, line 5: (note: will remove excessive whitespaces before making ready for review)

IgnatBeresnev:
JFYI: For some reason the new version of `prism.js` (1.29.0) parses inline comments incorrectly and includes everything post-comment under the comment token.

Unable to reproduce it on their [playground](https://prismjs.com/test.html#language=kotlin), so maybe it's a bug of Dokka (incorrect line breaks or something) - needs to be investigated furtner, can be postponed, so I downgraded it to the version that we had before that.

vmishenev:
> JFYI: For some reason the new version of prism.js (1.29.0) parses inline comments incorrectly and includes everything post-comment under the comment token.

I suppose you have not copied [this hook.](https://github.com/Kotlin/dokka/blob/e88a21e933ca7cb58898c924afe53e08130abfe9/plugins/base/src/main/resources/dokka/scripts/prism.js#L13)

plugins/base/src/test/kotlin/content/HighlightingTest.kt
```diff
@@ -71,9 +71,10 @@ class HighlightingTest : BaseAbstractTest() {
                     assertTrue(children?.get(it.first)?.style?.contains(it.second) == true)
                 val annotation = children?.first()?.children?.first()
 
-                assertTrue(annotation?.children?.get(0)?.style?.contains(TokenStyle.Annotation) == true)
-                assertTrue(annotation?.children?.get(1)?.children?.first()?.style?.contains(TokenStyle.Annotation) == true)
+                val annotationStyles = setOf(TokenStyle.Annotation, TokenStyle.Builtin)
```
vmishenev, line 74: What does `TokenStyle.Builtin` mean?
I am not sure the output-specific problem should be fixed here, in the content model rather than in the render stage.

plugins/base/src/main/resources/dokka/styles/prism.css
```diff
@@ -64,25 +62,28 @@ https://prismjs.com/download.html#themes=prism&languages=clike+java+javadoclike+
...
 	color: #9a6e3a;
 	/* This background color was intended by the author of this theme. */
-	/*background: hsla(0, 0%, 100%, .5);*/
+	background: hsla(0, 0%, 100%, 0.5);
 }
 
 .token.atrule,
 .token.attr-value,
 .token.keyword {
-	color: var(--keyword-color);
```
vmishenev, line 73: Why do you not use CSS variables? It is more convenient when there are several themes.

plugins/base/src/test/kotlin/content/HighlightingTest.kt
IgnatBeresnev: > What does TokenStyle.Builtin mean?

It's a good question, you were the one that added the type :) It was already in Public API, so I just used it.

> I am not sure the output-specific problem should be fixed here, in the content model rather than in the render stage.

Sure, I can fix it for html-only, but let's then delete `TokenStyle.Builtin` entry so there's no confusion?

plugins/base/src/main/resources/dokka/styles/prism.css
IgnatBeresnev: I didn't write it out by hand, it's a direct copy-paste from webhelp's PR, so you best address that question to them :)

I could go out of my way and substitute colors with variables, but I'd prefer to keep changes to a minimum (and document them as I did in the header) so that it's easier to update it in the future. 

That was actually the whole reason why I needed to copy all of the styles instead of making selective changes.

plugins/base/src/test/kotlin/content/HighlightingTest.kt
vmishenev: > It's a good question, you were the one that added the type :) It was already in Public API, so I just used it.

I hoped somebody tells me what it means)
You can remove it since it is not used anywhere.

plugins/base/src/main/resources/dokka/styles/prism.css
vmishenev: It is the easiest way, but I am not sure WH styles will be consistent with Dokka further.

IgnatBeresnev: Yeah, good point :+1:  Although if we do change any of the copy-pasted code from webhelp, during updates it'll be very difficult to tell which styles were copy-pasted and should be replaced, and which were custom.

So there's a need for Dokka-specific style changes, I think it should be done as overrides at the bottom of the file. This way it can be both easily kept in sync with upstream and allow for some custom Dokka colors

IgnatBeresnev:
* Upgraded prism.js, added the hook and left a comment for future generations
* Removed `TokenStyle.Builtin` entry, now adding `builtin` class only for html pages
* Reverted some of the unnecessary changes (the ones before review)

vmishenev (approved):

PR #2681, opened 2022-09-26
Suppress useless companion for Kotlin-As-Java

### LINKED ISSUE #200: Feature request: a way to suppress useless companion objects from javadoc
Companion objects that consist of no fields and all `@JvmStatic` methods are effectively redundant, especially for Java users who will want to use static methods as normal. Currently these things clutter up the generated javadocs, making them ugly.

Whilst arguably the Kotlin compiler should optimise them out entirely in this case, that's a separate discussion (which I raised on youtrack). For now it'd be nice to recognise this pattern and suppress emission of references to the companions.

### LINKED ISSUE #2086: readonly companion field with @JvmField doesn't in outer class in javadoc even with kotlin-as-java
**Describe the bug**
A clear and concise description of what the bug is 

With this class, the field should exist in `Class` with `public static final` but exists in `Class.Companion` with `getClass()` function.

```
class Class {
  companion object {
    /** Doc */
    @JvmField
    val field = "value"
  }
}
```

**Expected behaviour**
A clear and concise description of what you expected to happen

`field` should exist in `Class` with `public static final`.

**Screenshots**
If applicable, add screenshots to help explain your problem

image
image

**To Reproduce**
Steps to reproduce the behavior, ideally with an example project

**Dokka configuration**

```Kotlin

tasks.withType(AbstractDokkaLeafTask::class).all {
    moduleName.set("class-ast tree")
    dependencies {
        plugins("org.jetbrains.dokka:kotlin-as-java-plugin:1.5.0")
    }
    dokkaSourceSets["main"].includes.from("kdoc.md")
}

```

**Installation**
- Operating system: macOS
- Build tool: Gradle v7.1
- Dokka version: 1.5.0

**Additional context**
Add any other context about the problem here

Providing a PR can drastically speed up the process of fixing this bug

atyrin:
* hide a companion object if it contains only `const`/`@JvmField`/`lateinit` properties or `@JvmStatic` functions
* lift the declarations to outer class from the companion
* mark lifted declarations as static
* mark `const`/`@JvmField`/`lateinit` properties as static for named objects
* remove getter/setter for `const`/`lateinit` properties in named objects/companions
* add the companion property to the outer class

Fix #200, fix #2086

### DIFF
--- a/plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
+++ b/plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
@@ -1,0 +1,48 @@
+package org.jetbrains.dokka.kotlinAsJava.converters
+
+import org.jetbrains.dokka.model.DFunction
+import org.jetbrains.dokka.model.DObject
+import org.jetbrains.dokka.model.DProperty
+
+
+internal class KotlinCompanion(private val companion: DObject?) {
+
+    fun staticFunctions(): List<DFunction> {
+        if(companion == null) return emptyList()
+        return companion.functions.filter { it.isJvmStatic }
+    }
+
+    /**
+     * @return properties that will be visible as static for java.
+     * See [Static fields](https://kotlinlang.org/docs/java-to-kotlin-interop.html#static-fields)
+     */
+    fun staticProperties(): List<DProperty> {
+        if(companion == null) return emptyList()
+        return companion.properties.filter { it.isJvmField || it.isConst || it.isLateInit}
+    }
+
+    fun asJava():DObject? {
+        if(companion == null) return null
+        if(companion.hasNothingToRender()) return null
+
+        return companion.asJava(
+            excludedProps = staticProperties(),
+            excludedFunctions = staticFunctions()
+        )
+    }
+
+    /**
+     * Hide companion object if there isn't members of parents.
+     * Properties and functions that are moved to outer class are not counted as members.
+     */
+    private fun DObject.hasNothingToRender(): Boolean{
+        val nonStaticPropsCount = properties.minus(staticProperties().toSet()).size
+        val nonStaticFunctionsCount = functions.minus(staticFunctions().toSet()).size
+        val classLikesCount = classlikes.size
+        val superTypesCount = supertypes.values.firstOrNull()?.size ?: 0
+
+        return nonStaticFunctionsCount + nonStaticPropsCount +
+                classLikesCount + superTypesCount == 0
+    }
+}
+
--- a/plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
+++ b/plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
@@ -6,73 +6,69 @@ import org.jetbrains.dokka.model.properties.PropertyContainer
 
 private const val DEFAULT_COMPANION_NAME = "Companion"
 
-internal class KotlinCompanion(private val companion: DObject?) {
-
-    fun staticFunctions(): List<DFunction> {
-        if(companion == null) return emptyList()
-        return companion.functions.filter { it.isJvmStatic }
-    }
-
-    /**
-     * @return properties that will be visible as static for java.
-     * See [Static fields](https://kotlinlang.org/docs/java-to-kotlin-interop.html#static-fields)
-     */
-    fun staticProperties(): List<DProperty> {
-        if(companion == null) return emptyList()
-        return companion.properties.filter { it.isJvmField || it.isConst || it.isLateInit}
-    }
-
-    fun companionInstancePropertyInJava(): List<DProperty>{
-        if(companion == null) return emptyList()
-        if(companion.hasNothingToRender()) return emptyList() // do not show if companion not rendered
-
-        with(companion){
-            return listOf(DProperty(
-                name = name ?: DEFAULT_COMPANION_NAME,
-                modifier = sourceSets.associateWith { JavaModifier.Final },
-                dri = dri.copy(callable = Callable(name ?: DEFAULT_COMPANION_NAME, null, emptyList())),
-                documentation = emptyMap(),
-                sources = emptyMap(),
-                visibility = companion.sourceSets.associateWith {
-                    JavaVisibility.Public
-                },
-                type = GenericTypeConstructor(dri, emptyList()),
-                setter = null,
-                getter = null,
-                sourceSets = sourceSets,
-                receiver = null,
-                generics = emptyList(),
-                expectPresentInSet = expectPresentInSet,
-                isExpectActual = false,
-                extra = PropertyContainer.withAll(sourceSets.map {
-                    mapOf(it to setOf(ExtraModifiers.JavaOnlyModifiers.Static)).toAdditionalModifiers()
-                })
-            ))
-        }
-    }
-
-    fun asJava():DObject? {
-        if(companion == null) return null
-        if(companion.hasNothingToRender()) return null
-
-        return companion.asJava(
-            excludedProps = staticProperties(),
-            excludedFunctions = staticFunctions()
-        )
-    }
-
-    /**
-     * Hide companion object if there isn't members of parents.
-     * Properties and functions that are moved to outer class are not counted as members.
-     */
-    private fun DObject.hasNothingToRender(): Boolean{
-        val nonStaticPropsCount = properties.minus(staticProperties().toSet()).size
-        val nonStaticFunctionsCount = functions.minus(staticFunctions().toSet()).size
-        val classLikesCount = classlikes.size
-        val superTypesCount = supertypes.values.firstOrNull()?.size ?: 0
-
-        return nonStaticFunctionsCount + nonStaticPropsCount +
-                classLikesCount + superTypesCount == 0
-    }
+internal fun DObject?.staticFunctionsForJava(): List<DFunction> {
+    if (this == null) return emptyList()
+    return functions.filter { it.isJvmStatic }
 }
 
+/**
+ * @return properties that will be visible as static for java.
+ * See [Static fields](https://kotlinlang.org/docs/java-to-kotlin-interop.html#static-fields)
+ */
+internal fun DObject?.staticPropertiesForJava(): List<DProperty> {
+    if (this == null) return emptyList()
+    return properties.filter { it.isJvmField || it.isConst || it.isLateInit }
+}
+
+internal fun DObject?.companionInstancePropertyForJava(): List<DProperty> {
+    if (this == null) return emptyList()
+    if (hasNothingToRender()) return emptyList() // do not show if companion not rendered
+
+    return listOf(
+        DProperty(
+            name = name ?: DEFAULT_COMPANION_NAME,
+            modifier = sourceSets.associateWith { JavaModifier.Final },
+            dri = dri.copy(callable = Callable(name ?: DEFAULT_COMPANION_NAME, null, emptyList())),
+            documentation = emptyMap(),
+            sources = emptyMap(),
+            visibility = sourceSets.associateWith {
+                JavaVisibility.Public
+            },
+            type = GenericTypeConstructor(dri, emptyList()),
+            setter = null,
+            getter = null,
+            sourceSets = sourceSets,
+            receiver = null,
+            generics = emptyList(),
+            expectPresentInSet = expectPresentInSet,
+            isExpectActual = false,
+            extra = PropertyContainer.withAll(sourceSets.map {
+                mapOf(it to setOf(ExtraModifiers.JavaOnlyModifiers.Static)).toAdditionalModifiers()
+            })
+        )
+    )
+}
+
+internal fun DObject?.companionAsJava(): DObject? {
+    if (this == null) return null
+    if (hasNothingToRender()) return null
+
+    return asJava(
+        excludedProps = staticPropertiesForJava(),
+        excludedFunctions = staticFunctionsForJava()
+    )
+}
+
+/**
+ * Hide companion object if there isn't members of parents.
+ * Properties and functions that are moved to outer class are not counted as members.
+ */
+private fun DObject.hasNothingToRender(): Boolean {
+    val nonStaticPropsCount = properties.size - staticPropertiesForJava().size
+    val nonStaticFunctionsCount = functions.size - staticFunctionsForJava().size
+    val classLikesCount = classlikes.size
+    val superTypesCount = supertypes.values.firstOrNull()?.size ?: 0
+
+    return nonStaticFunctionsCount + nonStaticPropsCount +
+            classLikesCount + superTypesCount == 0
+}
--- a/plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
+++ b/plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
@@ -20,8 +20,7 @@ internal fun DObject?.staticPropertiesForJava(): List<DProperty> {
     return properties.filter { it.isJvmField || it.isConst || it.isLateInit }
 }
 
-internal fun DObject?.companionInstancePropertyForJava(): DProperty? {
-    if (this == null) return null
+internal fun DObject.companionInstancePropertyForJava(): DProperty? {
     if (hasNothingToRender()) return null // do not show if companion not rendered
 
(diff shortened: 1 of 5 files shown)


plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.jetbrains.dokka.links.Callable
+import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.model.properties.PropertyContainer
+
+private const val DEFAULT_COMPANION_NAME = "Companion"
+
+internal class KotlinCompanion(private val companion: DObject?) {
+
+    fun staticFunctions(): List<DFunction> {
+        if(companion == null) return emptyList()
```
zarechenskiy, line 12: Missing space after `if`

zarechenskiy: Note that the same problem is with other `if` statements

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,78 @@
...
+        return companion.functions.filter { it.isJvmStatic }
+    }
+
+    /**
+     * @return properties that will be visible as static for java.
+     * See [Static fields](https://kotlinlang.org/docs/java-to-kotlin-interop.html#static-fields)
+     */
+    fun staticProperties(): List<DProperty> {
+        if(companion == null) return emptyList()
+        return companion.properties.filter { it.isJvmField || it.isConst || it.isLateInit}
```
zarechenskiy, line 22: Missing space after `it.isLateInit`

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun staticFunctions(): List<DFunction> {
+        if(companion == null) return emptyList()
+        return companion.functions.filter { it.isJvmStatic }
+    }
+
+    /**
+     * @return properties that will be visible as static for java.
+     * See [Static fields](https://kotlinlang.org/docs/java-to-kotlin-interop.html#static-fields)
+     */
+    fun staticProperties(): List<DProperty> {
```
zarechenskiy, line 20: I'd rename it to something like `staticPropertiesForJvm` as there is no entity as static in Kotlin and can be slightly confusing

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,78 @@
+package org.jetbrains.dokka.kotlinAsJava.converters
+
+import org.jetbrains.dokka.links.Callable
+import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.model.properties.PropertyContainer
+
+private const val DEFAULT_COMPANION_NAME = "Companion"
+
+internal class KotlinCompanion(private val companion: DObject?) {
```
zarechenskiy, line 9: Looks like there is no need in this class as it doesn't have any state as is. Instead, we could move members to top-level and make them extension to `DObject`

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun staticProperties(): List<DProperty> {
+        if(companion == null) return emptyList()
+        return companion.properties.filter { it.isJvmField || it.isConst || it.isLateInit}
+    }
+
+    fun companionInstancePropertyInJava(): List<DProperty>{
+        if(companion == null) return emptyList()
+        if(companion.hasNothingToRender()) return emptyList() // do not show if companion not rendered
+
+        with(companion){
```
zarechenskiy, line 29: If `companionInstancePropertyInJava` will become an extension one to `DObject`, then there will be no need in this `with`

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,78 @@
...
+        )
+    }
+
+    /**
+     * Hide companion object if there isn't members of parents.
+     * Properties and functions that are moved to outer class are not counted as members.
+     */
+    private fun DObject.hasNothingToRender(): Boolean{
+        val nonStaticPropsCount = properties.minus(staticProperties().toSet()).size
+        val nonStaticFunctionsCount = functions.minus(staticFunctions().toSet()).size
```
zarechenskiy, line 70: Can't we just compare size of these collections without _possibly_ complex calculation of `equals`/`hashCode`? I.e. `properties.size - staticProperties().toSet().size`

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -259,24 +266,56 @@ internal fun DClass.asJava(): DClass = copy(
...
+/**
+ * Companion objects requires some custom logic for rendering as Java.
+ * They are excluded from usual classlikes rendering and added after.
+ */
+internal fun DClass.classlikesInJava(): List<DClasslike> {
+    val companionAsJava = companion.asKotlinCompanion().asJava()
+    val filterOutCompanion = { it: DClasslike -> it.name != companion?.name }
+
+    val classlikes = classlikes
+        .filter { filterOutCompanion(it) }
```
zarechenskiy, line 287: I believe it'd be better to inline `filterOutCompanion`, it doesn't big and its implementation is pretty clear

zarechenskiy (changes requested):
I've finished with the first review round, nice job 👍

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
atyrin: I thought about it but didn't want to add confusion with additional methods to usual named objects

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -76,7 +77,11 @@ internal fun DPackage.asJava(): DPackage {
     )
 }
 
-internal fun DProperty.asJava(isTopLevel: Boolean = false, relocateToClass: String? = null) =
+internal fun DProperty.asJava(
+    isTopLevel: Boolean = false,
+    relocateToClass: String? = null,
+    isBelongToObjectOrCompanion: Boolean = false
```
IgnatBeresnev, line 83: nitpick: `isBelong..` is not grammatically correct, what do you think about `isFrom..`?

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -259,24 +262,51 @@ internal fun DClass.asJava(): DClass = copy(
...
+ * Companion objects requires some custom logic for rendering as Java.
+ * They are excluded from usual classlikes rendering and added after.
+ */
+internal fun DClass.classlikesInJava(): List<DClasslike> {
+    val companionAsJava = companion.companionAsJava()
+
+    val classlikes = classlikes
+        .filter { it.name != companion?.name }
+        .map { it.asJava() }
+    return if (companionAsJava != null) classlikes.plus(companionAsJava) else classlikes
```
IgnatBeresnev, line 284: also nitpick: I'd move `companionAsJava` closer to where it's used, otherwise it makes you think that it's used for something else (i.e for computing `classlikes` val)
```suggestion
    val classlikes = classlikes
        .filter { it.name != companion?.name }
        .map { it.asJava() }
        
    val companionAsJava = companion.companionAsJava()
    return if (companionAsJava != null) classlikes.plus(companionAsJava) else classlikes
```

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -332,23 +362,33 @@ internal fun DEnum.asJava(): DEnum = copy(
...
 )
 
-internal fun DObject.asJava(): DObject = copy(
+/**
+ * Parameters [excludedProps] and [excludedFunctions] used for rendering companion objects
+ * where some members (that lifted to outer class) are not rendered
+ */
+internal fun DObject.asJava(
+    excludedProps: List<DProperty> = emptyList(),
+    excludedFunctions: List<DFunction> = emptyList()
```
IgnatBeresnev, line 371: Wondering if having a `Set` here would be better, or if these lists are usually so small that it wouldn't make any difference..

plugins/kotlin-as-java/src/test/kotlin/CompanionAsJavaTest.kt
```diff
@@ -0,0 +1,471 @@
...
+                    .firstOrNull { it.name == "MyClass" } as DClass
+
+                assertNotNull(parentClass.properties.none {it.name == COMPANION_NAME})
+            }
+        }
+    }
+
+
+
+    private fun `companion object not rendered for declaration`(declaration: String) {
```
IgnatBeresnev, line 376: [same comment](https://github.com/Kotlin/dokka/pull/2691#discussion_r993795261) here about callable method naming

I'd rename it to something like `assertCompanionObjectNotRendered`.

upd: oh, I see that you already have `assertCompanionNotRendered` method.. [Same comment](https://github.com/Kotlin/dokka/pull/2691#discussion_r993802472) as in the other PR as well :) 

This could be extracted into a separate helper function and re-used
```
val parentClass = module.packages.flatMap { it.classlikes }
                    .firstOrNull { it.name == "MyClass" } as DClass
```

other than that I don't see any real and harmful copy-paste

I personally would rather have tests with plain structure and _some_ copy-paste, but without having to hop around too much to understand what it checks. In current form it would also be difficult to modify these tests if 1 of 5 cases changes or requires some additional checks, and would probably lead to more input params/flags to this function

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @return properties that will be visible as static for java.
+ * See [Static fields](https://kotlinlang.org/docs/java-to-kotlin-interop.html#static-fields)
+ */
+internal fun DObject?.staticPropertiesForJava(): List<DProperty> {
+    if (this == null) return emptyList()
+    return properties.filter { it.isJvmField || it.isConst || it.isLateInit }
+}
+
+internal fun DObject?.companionInstancePropertyForJava(): List<DProperty> {
```
IgnatBeresnev, line 23: Is there a reason this has to return a list? The name suggest that it's a single property as well

If you need to sum lists somewhere and don't wanna deal with nulls, you could wrap it in `listOfNotNull`

IgnatBeresnev (commented):
Well done! Left some minor comments to discuss

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
atyrin: I've left the `List`. I don't expect a huge amount of entries in the number of properties or functions that can affect performance. Also, the list that passes here was used in the other places (but probably it was possible to painless switch to `Set` even there).

plugins/kotlin-as-java/src/test/kotlin/CompanionAsJavaTest.kt
atyrin: Thanks! I've returned the plain tests. I will stick to the idea that code duplication in the tests sometimes is a better decision.

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
atyrin: True, I wouldn't say I liked it but was unable to find a better decision. I changed the return type to `DProperty ` and wrapped it in the suggested function on the use side.

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
```diff
@@ -259,24 +262,51 @@ internal fun DClass.asJava(): DClass = copy(
...
+    val propertiesFromCompanion = companion
+        .staticPropertiesForJava()
+        .filterNot { it.hasJvmSynthetic() }
+        .map { it.asJava(isFromObjectOrCompanion = true) }
+    val companionInstanceProperty = listOfNotNull(companion.companionInstancePropertyForJava())
+    val ownProperties = properties
+        .filterNot { it.hasJvmSynthetic() }
+        .map { it.asJava() }
+
+    return propertiesFromCompanion + ownProperties + companionInstanceProperty
```
IgnatBeresnev, line 307:
```suggestion
    val companionInstanceProperty = companion.companionInstancePropertyForJava()
    val ownProperties = properties
        .filterNot { it.hasJvmSynthetic() }
        .map { it.asJava() }

    return propertiesFromCompanion + ownProperties + (companionInstanceProperty ?: emptyList())
```

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
```diff
@@ -0,0 +1,72 @@
...
+
+internal fun DObject?.companionAsJava(): DObject? {
+    if (this == null) return null
+    if (hasNothingToRender()) return null
+
+    return asJava(
+        excludedProps = staticPropertiesForJava(),
+        excludedFunctions = staticFunctionsForJava()
+    )
+}
```
IgnatBeresnev, line 58: I would make functions `companionAsJava`, `companionInstancePropertyForJava`, `staticPropertiesForJava` and `staticFunctionsForJava` be extensions of non nullable type, i.e `DObject.companionAsJava()`, and extract the null check one level higher.

This would help get rid of multiple occurrences of `if (this == null) return emptyList()` . You'd only have to correct usages to be a safe call, i.e `companion?.companionAsJava`, and not much else from what I see

IgnatBeresnev (approved):
Left some minor suggestion, the rest LGTM for the most part :+1:

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
atyrin: Not sure that the suggested change is valid. Since `companionInstanceProperty ` has the type `DProperty`  and the `emptyList` is a `List` of something then the expression in braces has the type `Any`.

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinCompanion.kt
atyrin: Agree with changes for `companionAsJava`, `companionInstancePropertyForJava`.
But for both `static*` extensions, there is more than one place where I expect a list so it will require more null checks on a higher level than just one here.

plugins/kotlin-as-java/src/main/kotlin/converters/KotlinToJavaConverter.kt
IgnatBeresnev: Sorry, I should've given more context for why my proposed change

The main reason is the name of the val. When I see `companionInstanceProperty`, judging solely by the name, I expect it to be a single object, not a list, especially not an empty one. This doesn't hurt much in this context, but if someone were to add more code that would involve using `companionInstanceProperty`, it could lead to some ugly solutions

There's also a very minor optimization that comes as a nice bonus (avoid creating an extra list and an array for vararg), but it's not substantial and hopefully escape analysis can take care of it

But it's a minor thing at this point, I don't mind it being the way it is now, so it's up to you

IgnatBeresnev: Oh, I see what you meant, `(companionInstanceProperty ?: emptyList())` does not compile, sorry. The suggestion can be dissmissed then. The main point was about the name, so it can probably be inlined to `propertiesFromCompanion + ownProperties + listOfNotNull(companionInstanceProperty)`

PR #2686, opened 2022-09-28
Fix inheriting docs in case of diamond inheritance

vmishenev:


plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -1007,7 +1009,13 @@ private class DokkaDescriptorVisitor(
             org.jetbrains.kotlin.types.Variance.OUT_VARIANCE -> Covariance(this)
         }
 
-    private fun DeclarationDescriptor.getDocumentation() = (findKDoc()?.let {
+    fun descriptorToAnyDeclaration(descriptor: DeclarationDescriptor): PsiElement? {
```
IgnatBeresnev, line 1012: Can visibility be reduced to private?

IgnatBeresnev (approved):

PR #2692, opened 2022-10-04
Update Kotlin to 1.7.20

IgnatBeresnev:
Current Kotlin plugins version (`213-1.7.20-199-IJ6777.52`) is already the latest. It was built a week before the release, and the only difference between this version and the release version is the changelog, which should not affect us in any way.


vmishenev (approved):
[There](https://github.com/Kotlin/dokka/actions/runs/3180773512/jobs/5184743032) is OOME: Metaspace.

Issue #2700, opened 2022-10-11
Improve compatibility with the Gradle API, and follow best practices


aSemy:
**Is your feature request related to a problem? Please describe**

Hi, this is related to the Dokka Gradle plugin, which I would like to help improve.

Improving usage of Gradle would help with the following issues:

* https://github.com/Kotlin/dokka/issues/2231
* https://github.com/Kotlin/dokka/issues/2515
* https://github.com/Kotlin/dokka/issues/1833
* https://github.com/Kotlin/dokka/issues/1864
* https://github.com/Kotlin/dokka/issues/1752
* https://github.com/Kotlin/dokka/issues/2612
* https://github.com/Kotlin/dokka/issues/1217
* https://github.com/Kotlin/dokka/issues/1382
* https://github.com/Kotlin/dokka/issues/2286
* https://github.com/Kotlin/dokka/pull/830
* https://github.com/Kotlin/dokka/issues/2822
* https://github.com/Kotlin/dokka/pull/2756
* https://github.com/Kotlin/dokka/issues/2903

These PRs are related:

* https://github.com/Kotlin/dokka/pull/2499
* https://github.com/Kotlin/dokka/pull/2799
* https://github.com/Kotlin/dokka/pull/2807
* https://github.com/Kotlin/dokka/pull/2816

**Describe the solution you'd like**

* use Kotlin DSL helpers to make the code more Kotlin-esque, for example 
   https://github.com/Kotlin/dokka/blob/83174361becb2af227159834cdf6e14db9300c53/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/gradleConfigurations.kt#L10
* Tasks that require downloading files should cache and re-use files https://github.com/Kotlin/dokka/issues/2695
* Use configurations for sharing outputs between projects https://docs.gradle.org/7.3/samples/sample_jvm_multi_project_with_code_coverage.html
* Remove non-idiomatic helper functions https://github.com/Kotlin/dokka/blob/master/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaProperty.kt
* Rename files to match the contents, e.g. https://github.com/Kotlin/dokka/blob/master/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/main.kt
[...]

PR #2702, opened 2022-10-11
apply kotlin-dsl plugin

aSemy:
Part of #2700

apply kotlin-dsl plugin, and update code to use idiomatic Gradle API 

This helps because the code in the plugin will more closely match the code used in `*.gradle.kts` files


aSemy:
My pleasure! 

For reference the results of applying `kotlin-dsl` are specified here: https://docs.gradle.org/7.5.1/userguide/kotlin_dsl.html#sec:kotlin-dsl_plugin

One of the changes that isn't explicitly stated is that the Kotlin language level will be changed to 1.4, to match the compatibility defined here: https://docs.gradle.org/current/userguide/compatibility.html#kotlin

IgnatBeresnev:
Looks like integration tests fail for Gradle 5.6, but I don't think we have to support it as KGP supports 6.7.1 onwards.

I've made an attempt to fix it in #2739, need to wait until it gets merged and rebased onto

IgnatBeresnev:
An update on this: I like the change and the DSL, but we cannot merge it until after the 1.8.10 release, but it should hopefully be soon.

The reason is that `kotlin-dsl` doesn't seem to work with Gradle 5.6, but we use 5.6 for the stdlib integration test, and we cannot upgrade it there. We agreed to remove/update the test after 1.8.10 though, so it should unblock this PR.

runners/gradle-plugin/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ import org.jetbrains.*
 
 plugins {
     `java-gradle-plugin`
```
hfhbd, line 5: You could remove this too, the `kotlin-dsl` applies the `JavaGradlePluginPlugin`.

IgnatBeresnev: I'll merge this PR now to get the whole chain moving, but @aSemy if you want to apply the suggestion, I think we can do it in #2833, since it'll have to be rebased anyway

IgnatBeresnev:
The stdlib integration tests has been removed, so this PR has been unblocked (finally) :tada: 

I've tested this branch locally with the latest master, all the tests pass, so I'm merging it

runners/gradle-plugin/build.gradle.kts
aSemy, line 5:
```suggestion
```

IgnatBeresnev: That works too, thanks :)

Issue #2703, opened 2022-10-11
Update build config to be more modular, using buildSrc convention plugins


aSemy:
**Is your feature request related to a problem? Please describe**

Presently the Dokka Gradle config uses a lot of `allprojects {}` and `subprojects {}` blocks to share configuration, but this is discouraged by Gradle

* https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration
* https://melix.github.io/blog/2021/12/composition-in-gradle.html

Using inheritance makes configuring the project difficult, as it prevents subprojects from being ordered nicely (for example, it's not easily possible to have 'empty' subprojects, as the `subprojects{}` block will add the plugins to them). It also means configuration is applied, even when not desired. For example, Dokka always applies the `kotlin("jvm")` plugin

https://github.com/Kotlin/dokka/blob/83174361becb2af227159834cdf6e14db9300c53/build.gradle.kts#L44

But that's not required for the Gradle plugin project, as Gradle will supply the correct Kotlin plugin when the `kotlin-dsl` plugin is applied.

**Describe the solution you'd like**

Dokka uses buildSrc convention plugins to configure and apply configuration in a modular way.

* https://docs.gradle.org/current/samples/sample_convention_plugins.html
* https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins

**Describe alternatives you've considered**

**Additional context**

This would help with implementation of #2700, so the Gradle Plugin subproject would not have unecessary configuration applied to it (like the `kotlin("jvm")` plugin.

**Are you willing to provide a PR?**

I have made a start, but there's a lot of work to do!

aSemy:
Other improvements I think can me made

* use the Java Test Fixtures plugin to share test-support code https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures
* Directly depending on another project's task can cause issue - this can be removed easily enough https://github.com/Kotlin/dokka/blob/c065412fdc7ee94c0be64b48eedb9819965df307/integration-tests/maven/build.gradle.kts#L14
* The dependencies are a bit scattered. For example, JUnit Jupiter is manually defined as 5.6.0, but some versions are defined in `gradle.properties`. Dependencies could be set in one place using Gradle Version Catalog, and aligned using a Gradle Version Platform https://docs.gradle.org/current/userguide/platforms.html#sub:using-platform-to-control-transitive-deps

[...]

Issue #2546, opened 2022-06-21
[1.7.0] Runtime JAR files in the classpath should have the same version


IgnatBeresnev:
Build fails if `Dokka` is used as a dependency in `buildSrc`, initially reported in the community slack: [link to message](https://kotlinlang.slack.com/archives/C0F4UNJET/p1655743911527999)

The same problem is relevant for [kotlinx.serialization](https://youtrack.jetbrains.com/issue/KT-52811).

### Workarounds

* Update to Gradle `7.5-rc-2`, it will solve the problem
* Exclude stdlib:
```kotlin
implementation("org.jetbrains.dokka:dokka-gradle-plugin:1.7.0") {
    exclude("org.jetbrains.kotlin","kotlin-stdlib-jdk8")
}
```

### Fix

Should hopefully be solved by https://github.com/Kotlin/dokka/pull/2543 in 1.7.10, will post updates here.

### Stacktrace

```
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    /opt/rjaros/.gradle/wrapper/dists/gradle-7.4.2-all/9uukhhbclvbegdvsww0j0cr3p/gradle-7.4.2/lib/kotlin-stdlib-1.5.31.jar (version 1.5)
    /opt/rjaros/.gradle/wrapper/dists/gradle-7.4.2-all/9uukhhbclvbegdvsww0j0cr3p/gradle-7.4.2/lib/kotlin-stdlib-common-1.5.31.jar (version 1.5)
    /opt/rjaros/.gradle/wrapper/dists/gradle-7.4.2-all/9uukhhbclvbegdvsww0j0cr3p/gradle-7.4.2/lib/kotlin-stdlib-jdk7-1.5.31.jar (version 1.5)
    /opt/rjaros/.gradle/wrapper/dists/gradle-7.4.2-all/9uukhhbclvbegdvsww0j0cr3p/gradle-7.4.2/lib/kotlin-stdlib-jdk8-1.5.31.jar (version 1.5)
    /opt/rjaros/.gradle/wrapper/dists/gradle-7.4.2-all/9uukhhbclvbegdvsww0j0cr3p/gradle-7.4.2/lib/kotlin-reflect-1.5.31.jar (version 1.5)
```

IgnatBeresnev:
[...]

PR #2704, opened 2022-10-11
build config: Use buildSrc convention plugins to configure the Dokka subprojects

aSemy:
This is a large part of the implementation for #2703

Status: ready for review

* set up buildSrc convention plugins
* move shared configuration into the convention plugins
* tidy up the subproject build files to remove duplicated configuration
* set up [centralized repo definitions](https://docs.gradle.org/current/userguide/declaring_repositories.html#sub:centralized-repository-declaration) 

Important to note:

* This should have no impact on the actual source code or published artifacts, this is only updating the build config
* All Gradle plugins must be defined in `./buildSrc/build.gradle.kts`, so they are aligned across all subprojects

Related

* This PR duplicates the change proposed in https://github.com/Kotlin/dokka/pull/1321


aSemy:
I made some good progress but then I got stuck when I reached the integration test config, because it's a bit of a beast.

https://github.com/Kotlin/dokka/blob/83174361becb2af227159834cdf6e14db9300c53/integration-tests/build.gradle.kts

aSemy:
@IgnatBeresnev I see you've added the 'Gradle Plugin' label, and while this PR would help, it's more about the improving the general Gradle build config of the entire project :)

IgnatBeresnev:
Yeah, true. I don't want to create too many labels, but I definitely want to mark all Gradle-related PRs in case we ask for anyone's help with review and whatnot

If it makes anyone uneasy, we can certainly create a separate one or rename the existing

aSemy:
With regards to labels, what about a 'build config' or 'ci/cd' label that would cover the build scripts, and also GitHub Actions, Qodana, TeamCity etc?

aSemy:
I've made some good progress! All shared build-config now comes from buildSrc convention plugins, not from `subprojects{}` or `allprojects{}` blocks. This makes the build much more modular, because it's not restricted by inheritance, and easier to configure because a build script is more self contained - it doesn't have config applied from some arbitrary parent.

plugins/android-documentation/build.gradle.kts
```diff
@@ -1,9 +1,22 @@
...
+    implementation(kotlin("stdlib"))
+    implementation(kotlin("reflect"))
+
     implementation(project(":plugins:base"))
     testImplementation(project(":plugins:base"))
     testImplementation(project(":plugins:base:base-test-utils"))
+    testImplementation(project(":test-utils"))
+    testImplementation(project(":core:test-api"))
+    testImplementation("org.junit.jupiter:junit-jupiter:5.6.0")
 }
```
aSemy, line 20: All of the 'plugins' subprojects had these dependencies defined, because they were defined in a `subprojects {}` block

```kotlin
dependencies {
    compileOnly(project(":core"))
    implementation(kotlin("stdlib-jdk8"))
    implementation(kotlin("reflect"))

    testImplementation(project(":test-utils"))
    testImplementation(project(":core:test-api"))
    testImplementation("org.junit.jupiter:junit-jupiter:5.6.0")
}
```

In order to keep the code working I've copy and pasted them, to recreate the same result. However, it would be nicer if these dependencies weren't so repetitive. 

Perhaps all of these dependencies aren't needed in all 'plugins' projects and some can be removed?

Alternatively, perhaps these dependencies could be defined as API dependencies in another project (like `:plugins:base`?)

```kotlin
dependencies {
    compileOnlyApi(project(":core"))
    api(kotlin("stdlib-jdk8"))
    api(kotlin("stdlib"))
    api(kotlin("reflect"))
    testApi(project(":test-utils"))
    testApi(project(":core:test-api"))
    testApi("org.junit.jupiter:junit-jupiter:5.6.0")
}
```

then each plugin project would automatically get these dependencies.

IgnatBeresnev:
After researching  #2652, I think I now finally understand why the proposed changes are good and what they do! 

However, if we first merge #2652 - it would introduce conflicts. @aSemy would you be OK with that? Looks like you would be able to just move the files over into `build-logic`.

aSemy:
Either merging this PR or #2652 first would work, but since there are fewer changes in #2652, and converting buildSrc to an included build is a more 'logical' upgrade, I think it would be best if this PR were merged first, and either #2652 is updated, or we create a new similar PR

settings.gradle.kts
```diff
@@ -60,3 +53,42 @@ gradleEnterprise {
...
+        maven("https://cache-redirector.jetbrains.com/intellij-dependencies")
+        maven("https://www.myget.org/F/rd-snapshots/maven/")
+    }
+
+    pluginManagement {
+        repositories {
+            gradlePluginPortal()
+            mavenCentral()
+        }
+    }
```
Goooler, line 93: `pluginManagement` should be placed out of `dependencyResolutionManagement`.

settings.gradle.kts
```diff
@@ -60,3 +53,42 @@ gradleEnterprise {
         publishAlwaysIf(isCiBuild)
     }
 }
+
+@Suppress("UnstableApiUsage") // Central declaration of repositories is an incubating feature
```
Goooler, line 57: I believe no need to suppress these `UnstableApiUsage`s.

build.gradle.kts
```diff
@@ -82,8 +27,27 @@ nexusPublishing {
...
+            "frontend",
+
+            "integration-tests",
+            "gradle",
+            "cli",
+            "maven",
+
+            "test-utils",
+        )
+    )
```
Goooler, line 52:
```suggestion
    ignoredProjects += setOf(
        "search-component",
        "compiler-dependency",
        "kotlin-analysis",
        "intellij-dependency",
        "frontend",

        "integration-tests",
        "gradle",
        "cli",
        "maven",

        "test-utils",
    )
```

settings.gradle.kts
aSemy: haha wow, I've been putting `pluginManagement` in the wrong place for about a year, thanks

settings.gradle.kts
aSemy: it's still incubating https://docs.gradle.org/7.6/userguide/declaring_repositories.html#sub:centralized-repository-declaration

> Central declaration of repositories is an incubating feature



without it, there are several warnings

image

Goooler: There might be more places that need to be suppressed, personally think it's unnecessary.

settings.gradle.kts
aSemy: Ah, if `pluginManagement` is placed outside of `dependencyResolutionManagement` then the block needs to be moved, otherwise there's an error. 

```
dokka/settings.gradle.kts:45:1: Unexpected `plugins` block found. `plugins` can not appear before `pluginManagement`.
> Unexpected `plugins` block found. `plugins` can not appear before `pluginManagement`.
```

Which is okay, I will move it, but personally I think it's neater if all the repositories are together inside one block.

Goooler: It should be at the top of `settings.gradle`.

plugins/all-modules-page/build.gradle.kts
```diff
@@ -29,4 +39,8 @@ dependencies {
 
     val jsoup_version: String by project
     implementation("org.jsoup:jsoup:$jsoup_version")
-}
\ No newline at end of file
+
+    testImplementation(project(":test-utils"))
+    testImplementation(project(":core:test-api"))
+    testImplementation("org.junit.jupiter:junit-jupiter:5.6.0")
```
Goooler, line 45: Can we put these into `org.jetbrains.conventions.kotlin-jvm`? Might include deps  above:

```kotlin
    compileOnly(project(":core"))
    implementation(kotlin("stdlib-jdk8"))
    implementation(kotlin("reflect"))
```

runners/cli/build.gradle.kts
```diff
@@ -1,7 +1,10 @@
 import org.jetbrains.DokkaPublicationBuilder.Component.Shadow
 import org.jetbrains.registerDokkaArtifactPublication
 
+
```
Goooler, line 4: Extra line.

runners/gradle-plugin/build.gradle.kts
```diff
@@ -83,6 +80,13 @@ pluginBundle {
     }
 }
 
+val javadocJar by tasks.creating(Jar::class) {
+    group = JavaBasePlugin.DOCUMENTATION_GROUP
+    description = "Assembles java doc to jar"
+    archiveClassifier.set("javadoc")
+    from(tasks.javadoc)
+}
```
Goooler, line 88: Can we defer this creating?

buildSrc/src/main/kotlin/org/jetbrains/conventions/base.gradle.kts
```diff
@@ -0,0 +1,12 @@
...
+plugins {
+    base
+}
+
+// common Gradle configuration that should be applied to all projects
+
+if (project != rootProject) {
+    project.group = rootProject.group
+    project.version = rootProject.version
+}
```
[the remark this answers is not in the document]
aSemy: More configuration will be added here in the future, for example

* [reproducible archives](https://docs.gradle.org/7.6/userguide/working_with_files.html#sec:reproducible_archives)
* setting a default for test logging


`allprojects {}` is [discouraged by Gradle](https://docs.gradle.org/7.6/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration)

https://docs.gradle.org/7.6/userguide/multi_project_configuration_and_execution.html#sec:decoupled_projects

> using key Gradle features like the allprojects and subprojects keywords automatically cause your projects to be coupled.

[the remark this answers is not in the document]
Goooler: Gotcha.

plugins/all-modules-page/build.gradle.kts
aSemy: There's no hard and fast rule for this, but I think it's best not to define dependencies in convention plugins (even test dependencies). 

It's much easier to understand what a subproject requires and the links between subprojects when the dependencies are defined locally, because it's easy to look at a subproject's `build.gradle.kts` to check. If the dependencies are applied somewhere else, then they could be really tricky to trace down and understand.

Either way - first let's to convert the existing project to use convention plugins, and to match the existing setup as closely as possible and to make this PR as straightforward as possible

runners/gradle-plugin/build.gradle.kts
aSemy: Good point, yes. It's deferred at the moment.

https://github.com/Kotlin/dokka/blob/fe360f7fadc7173be3ae170778a93ebd01962149/build.gradle.kts#L65-L69

I think I used `creating` because the Gradle Publishing and Signing plugins are janky, and don't fully support deferred config. But I added the Javadoc JAR to the integration test publication, so that will make sure it's tested.

Goooler:
@aSemy Rebasing?

Goooler (commented):
LGTM

IgnatBeresnev:
@Goooler thanks for the review! I'm also taking a look now

runners/gradle-plugin/build.gradle.kts
```diff
@@ -118,3 +126,7 @@ afterEvaluate { // Workaround for an interesting design choice https://github.co
     configureSonatypePublicationIfNecessary("pluginMaven", "dokkaGradlePluginPluginMarkerMaven")
     createDokkaPublishTaskIfNecessary()
 }
+
+tasks.processResources {
+    duplicatesStrategy = DuplicatesStrategy.WARN
+}
```
IgnatBeresnev, line 132: I see that it migrated from another place. Any idea what this was/is used for? Are there any specific problems if it gets deleted?

runners/gradle-plugin/gradle.properties
```diff
@@ -1 +0,0 @@
-kotlin.stdlib.default.dependency=false
```
IgnatBeresnev, line 1: Is this not relevant anymore? It was a recent fix (#2543)

https://mbonnin.medium.com/the-different-kotlin-stdlibs-explained-83d7c6bf293

runners/gradle-plugin/build.gradle.kts
```diff
@@ -83,6 +80,13 @@ pluginBundle {
     }
 }
 
+val javadocJar by tasks.registering(Jar::class) {
+    group = JavaBasePlugin.DOCUMENTATION_GROUP
+    description = "Assembles Javadoc into a JAR"
+    archiveClassifier.set("javadoc")
+    from(tasks.javadoc)
+}
```
IgnatBeresnev, line 88: I believe this will now publish Java's Javadoc format, whereas before it published Dokka's HTML format. I've checked the `javadoc.jar` for the [latest release](https://search.maven.org/artifact/org.jetbrains.dokka/dokka-gradle-plugin/1.7.20/jar)

The previous `javadocJar` task with Dokka's HTML format is declared in `kotlin-jvm.gradle.kts`, but if you extract it to `maven-publish` as I mentioned, I think it should be accessible here

settings.gradle.kts
```diff
@@ -60,3 +60,36 @@ gradleEnterprise {
...
+dependencyResolutionManagement {
+
+    repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)
```
IgnatBeresnev, line 67: I actually wonder if the default `PREFER_PROJECT` would be better

It looks like we have a few projects (namely, `intellij-dependency` and `frontend`) that need very specific repositories (like the for the gradle-node plugin) that are unlikely to be used in any other project. Similar to dependencies, having very specific repositories be declared in that project locally would make it somewhat easier to reason about

Having truly common repositories (i.e used in >1 module) here totally makes sense though, so no questions here

wdyt?

IgnatBeresnev (commented):
Superb work! Thank you!

I only have minor questions and comments

I ran the integration tests locally with success, and will trigger a CI build and test publishing before merging

settings.gradle.kts
Goooler: How about `FAIL_ON_PROJECT_REPOS`?

plugins/android-documentation/build.gradle.kts
aSemy: Yes I agree - postpone trying to optimise the dependencies.

build.gradle.kts
```diff
@@ -82,8 +27,25 @@ nexusPublishing {
...
+        "kotlin-analysis",
+        "intellij-dependency",
+        "frontend",
+
+        "integration-tests",
+        "gradle",
+        "cli",
+        "maven",
+
+        "test-utils",
```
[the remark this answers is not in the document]
aSemy: > It's a shame it doesn't accept the project path. When I looked at `gradle`, I thought it was somehow referencing our gradle runner, not realizing there's a `gradle` project under integration tests

https://github.com/Kotlin/binary-compatibility-validator/issues/16

> also, we're not tracking the public API for the gradle plugin, but it's a good idea to start doing it. Can you run `apiDump` to generate it? it [fails the check now](https://github.com/Kotlin/dokka/actions/runs/4241592462/jobs/7372011661) because it's expecting the file, but can't find it

Yeah I was surprised that the Gradle Plugin was missing an API dump. And it's also odd that it hasn't been required until now. But anyway, I've added it now.

> Adding some comments would help

Done!

kotlin-analysis/intellij-dependency/build.gradle.kts
```diff
@@ -2,18 +2,9 @@ import org.jetbrains.DokkaPublicationBuilder.Component.Shadow
...
-}
-
-repositories {
-    mavenCentral()
-    maven(url = "https://www.jetbrains.com/intellij-repository/snapshots")
-    maven(url = "https://www.jetbrains.com/intellij-repository/releases")
-    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide")
-    maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies")
-    maven("https://cache-redirector.jetbrains.com/intellij-dependencies")
-    maven("https://www.myget.org/F/rd-snapshots/maven/")
```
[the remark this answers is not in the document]
aSemy: Thanks for the reminder! I've made the change.

buildSrc/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+
+tasks.dokkaHtml {
+    onlyIf { !isLocalPublication }
+    outputDirectory.set(layout.buildDirectory.dir("dokka").map { it.asFile })
+}
+
+val javadocJar by tasks.registering(Jar::class) {
+    archiveClassifier.set("javadoc")
+    from(tasks.dokkaHtml)
+}
```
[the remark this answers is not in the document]
aSemy: Good question! I think explaining this is a demo of Gradle concepts.

Moving this task to the `maven-publish` convention is a little bit tricky because that convention plugin doesn't have the Dokka plugin applied, so `tasks.dokkaHtml` might not exist. Even if a subproject did apply both Dokka and `maven-publish`, if it applies them in the wrong order, the task still might not exist. (This is why the Gradle Kotlin DSL is great - it will protect you against using missing tasks by only generating the accessor when the task is present.)

You _could_ move this task to the `maven-publish` convention but wrap it in a block that will only register the task when the Dokka plugin is present

```kotlin
plugins.withId("org.jetbrains.dokka") {
    // define javadocJar task...
}
```

(or an equivalent 'plugin reaction config' in the `kotlin-jvm` that will only define the javadocJar task if `maven-publish` is present)

But... both of these aren't really necessary. It's not really a problem if a subproject has a task that it doesn't need, because the task will only be run when it's required. And the javadocJar task is only required when a publication is triggered.

https://github.com/Kotlin/dokka/blob/8c0344e682fc4ba0e44fc740269c7ba308c47c35/buildSrc/src/main/kotlin/org/jetbrains/publication.kt#L36

buildSrc/src/main/kotlin/org/jetbrains/conventions/dokka-integration-test.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+
+val integrationTestImplementation by configurations.getting {
+    extendsFrom(configurations.implementation.get())
+}
+
+val integrationTestRuntimeOnly by configurations.getting {
+    extendsFrom(configurations.runtimeOnly.get())
+}
+
+val integrationTest by tasks.registering(Test::class) {
```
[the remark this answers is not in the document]
aSemy: Oops, good spot. I thought I copied it over but I guess it got lost during a merge. I've added it back.

runners/gradle-plugin/build.gradle.kts
aSemy: If you run `./gradlew assemble` you can see the warnings

I don't know about this one:

```
> Task :runners:maven-plugin:jar
Encountered duplicate path "META-INF/maven-plugin.kotlin_module" during copy operation configured with DuplicatesStrategy.WARN
```



This warning should probably be resolved by deleting `META-INF/gradle-plugins/org.jetbrains.dokka.properties` - the plugin identifier is provided by the Gradle config, 

```
> Task :runners:gradle-plugin:processResources
Encountered duplicate path "META-INF/gradle-plugins/org.jetbrains.dokka.properties" during copy operation configured with DuplicatesStrategy.WARN
```

runners/gradle-plugin/build.gradle.kts
aSemy: Ahh okay, now I understand your suggestion to move the javadocJar task to the `maven-publish` convention. It needs to be present in both `kotlin-dsl` projects, and `kotlin-jvm` projects. But the Gradle plugin must only have the `kotlin-dsl` plugin applied, not both.

runners/gradle-plugin/gradle.properties
IgnatBeresnev: I see. It'd be nice to have one less flag. The problem was somewhat difficult to track and test (#2546), so if you don't mind - let's restore it just to be on the safe side and so that it doesn't block this PR?

I've written it down to be investigated later, but if you want and have the time - it can be extracted into a separate PR, we'll need it anyway

runners/gradle-plugin/build.gradle.kts
IgnatBeresnev: We could extract Dokka into its own convention, would that help? And then use the convention from `maven-publish`?

aSemy: Good ideas, done. Take a look at let me know if it's as you'd like!

runners/gradle-plugin/gradle.properties
aSemy: Also, the chance of accidentally including the wrong stdlib could be improved with included-builds, which is on the list #2652. The regular Dokka code could be built completely separately from the Gradle plugin. (It's not a critical requirement, just extra security)

settings.gradle.kts
aSemy: `PREFER_PROJECT` is the default behaviour, so I'll set it to be explicit, and I've added a comment to explain.

I've added in some config to tell Gradle that the Yarn/Node repos can only provide Yarn/Node - so Gradle can avoid using those repositories for other dependencies.

buildSrc/build.gradle.kts
```diff
@@ -6,8 +8,24 @@ repositories {
...
+val kotlinVersion = properties["kotlin_version"]
+
 dependencies {
+    // Import Gradle Plugins that will be used in the buildSrc pre-compiled script plugins, and any `build.gradle.kts`
+    // files in the project.
+    // Use their Maven coordinates (plus versions), not Gradle plugin IDs!
+    // This should be the only place that Gradle plugin versions are defined, so they are aligned across all build scripts
+
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
     implementation("com.github.jengelman.gradle.plugins:shadow:2.0.4")
```
Goooler, line 25:
```suggestion
```

Goooler: Already been replaced by `implementation("gradle.plugin.com.github.johnrengelman:shadow:7.1.2")`.

buildSrc/src/main/kotlin/org/jetbrains/conventions/dokka-integration-test.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+ *    However, if the test is cacheable, it produces no outputs, so no documentation
+ *    to showcase. It needs to be broken into two separate tasks: one cacheable for running
+ *    the tests and producing HTML output, and another non-cacheable for copying the output.
+ *
+ * @see [org.jetbrains.dokka.it.TestOutputCopier] for more details on showcasing documentation
+ */
+@DisableCachingByDefault(because = "Contains incorrect inputs/outputs configuration, see the KDoc for details")
+abstract class NonCacheableIntegrationTest : Test()
+
+val integrationTest by tasks.register<NonCacheableIntegrationTest>("integrationTest") {
```
Goooler, line 44:
```suggestion
val integrationTest by tasks.registering(NonCacheableIntegrationTest::class) {
```

buildSrc/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import java.util.*
```
Goooler, line 1:
```suggestion
import java.util.Properties
```

buildSrc/build.gradle.kts
```diff
@@ -6,8 +8,23 @@ repositories {
...
-    implementation("com.github.jengelman.gradle.plugins:shadow:2.0.4")
-    implementation("org.jetbrains.kotlinx:binary-compatibility-validator:0.10.1")
+    // Import Gradle Plugins that will be used in the buildSrc pre-compiled script plugins, and any `build.gradle.kts`
+    // files in the project.
+    // Use their Maven coordinates (plus versions), not Gradle plugin IDs!
+    // This should be the only place that Gradle plugin versions are defined, so they are aligned across all build scripts
+
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
+    implementation("gradle.plugin.com.github.johnrengelman:shadow:7.1.2")
+    implementation("org.jetbrains.kotlinx:binary-compatibility-validator:0.12.1")
```
Goooler, line 26:
```suggestion
    implementation("org.jetbrains.kotlinx:binary-compatibility-validator:0.13.0")
```

aSemy: I'm not opposed to bumping this, but I want to try and limit this PR as much as possible. I think it would be better to update in a separate PR.

IgnatBeresnev (approved):
:+1: :+1:  :+1:

I've started integration tests, and will publish a dev build for comparison, just to make sure publication works and hasn't been affected

Will merge if everything's OK, or get back with details

IgnatBeresnev:
Found a minor problem when comparing two versions:

* 1.8.10 RC - [1.8.10-dev-203](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev/org/jetbrains/dokka/dokka-gradle-plugin/1.8.10-dev-203/)
* This branch - [1.8.20-dev-204](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev/org/jetbrains/dokka/dokka-gradle-plugin/1.8.20-dev-204/)

The configuration for `subprojects {}` was extracted into the kotlin-jvm convention. Previously, it was applied to the gradle plugin as well, whereas now it's not. This leads to:

* [KotlinCompile configuration](https://github.com/Kotlin/dokka/blob/6603d557211f6e35c9615c901956ae468125ed1b/buildSrc/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts#L15) is not applied for the Gradle plugin, so if you compare the `.module` files for the plugin you'll see that `"org.gradle.jvm.version": 8` was changed to `11`
* The sources for the gradle plugin are not published. I'll leave a comment on the line that I think causes it

___

The rest seems to work fine :) 

[Optional] Dokka used for publishing used to be 1.7.20, now it's 1.7.10 - got a bit confused when comparing published javadocs :) I'll update the versions after the release anyway, so it's ok to skip it

buildSrc/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+        allWarningsAsErrors = true
+        languageVersion = language_version
+        apiVersion = language_version
+        jvmTarget = "1.8"
+    }
+}
+
+// Gradle metadata
+java {
+    withSourcesJar()
```
IgnatBeresnev, line 33: If I understand correctly, `withSourcesJar` adds a task and it's also added as a dependency for `assemble`, so `sources.jar` will be built for projects that have this applied

`kotlin-jvm` is not applied for the gradle plugin and its sources are not published, so it looks like this `withSourcesJar` needs to be a part of the `maven-publish` convention

aSemy: Thanks for checking, and good catches.

I've split out the shared Java config into a new convention plugin, `base-java.gradle.kts`. It sets up the Java Toolchain for all subprojects to be Java 8, and, if the Maven Publish plugin is present, it enables the sources jar.

IgnatBeresnev:
Hi! Not sure if you noticed that the integration tests for coroutines are unable to be run because of the toolchain being set to 1.8 globally. This refactoring has uncovered quite a number of problems already, which is good, I guess :)

```
A problem occurred evaluating root project 'kotlinx.coroutines'.
> Failed to apply plugin 'jdk-convention'.
> Project required JDK 11+, but found 1.8
```

Can you please fix it when you have the time, so that the checks pass? I believe that's the last thing holding up the merge

aSemy:
thanks for checking @IgnatBeresnev, I didn't realise that was the problem.

I set the toolchain for Dokka to use Java 8 because that matches the existing `jvmTarget`

https://github.com/Kotlin/dokka/blob/ded804e5772399f1495016d598573cb20b673b58/build.gradle.kts#L33

If I set the toolchain to 11, then the `:runners:maven-plugin` build fails:

```
```

It looks like kotlinx.coroutines has a custom check for the JDK version...

https://github.com/Kotlin/kotlinx.coroutines/blob/4116d4a178f32b1241db8247a38cd2823fb8b03e/buildSrc/src/main/kotlin/jdk-convention.gradle.kts#L7-L14

I'll look into it a little more. Perhaps the kotlinx.coroutines int. test requires a little more config.

IgnatBeresnev:
> I set the toolchain for Dokka to use Java 8 because that matches the existing jvmTarget

Yeah, that's the most important part - we cannot break compatibility with Java 8 for now.

We can set `jvmTarget` to 8 in the kotlin convention + the gradle plugin, but run the whole project with java 11 - I don't mind that

buildSrc/build.gradle.kts
aSemy: btw I found out why this star-import appeared: the Kotlin official style defines `java.util` as package that should be star-imported.

image

IgnatBeresnev:
> @IgnatBeresnev Do you know why the jvm-version is set to 12 in the example workflows?

It was copied from `s3-snapshots.yml`, but why it's set there to 12 - zero clue :( I think it's fine setting it to 11 like you did

> @IgnatBeresnev Tests have passed, it's ready for the TeamCity build!

Thank you once again! Triggered the builds, will merge once they pass :)

aSemy:
@IgnatBeresnev By the way, good job on the GitHub labels! They look really smart, and they're helpful.

IgnatBeresnev:
@aSemy yeah, going through all of the issues :) Part of the reason is actually for the dokkatoo DSL. In particular, the `configuration` and the `Gradle plugin` tags could be of interest to go along with the re-write since we're breaking compatibility anyway

buildSrc/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+
+plugins {
+    `java`
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
+    }
+}
```
Goooler, line 18: If we apply `toolchain` for all modules, they'll run tests on Java 8 instead of the JDK matrix on CI, we need to apply more toolchains in `Test` which will slow  down the build, I believe we should not use `toolchain` for now. Check

- https://github.com/detekt/detekt/pull/5735
- https://github.com/Kotlin/dokka/blob/f5581c51d0de23d2b038e00b1a152faebaef0920/.github/workflows/gradle-test.pr.yml#L12-L13

aSemy: Good catch!

Toolchains can still help with this though, because it's possible to [set them per-task](https://docs.gradle.org/current/userguide/toolchains.html#toolchains_for_tasks). I did something similar for MockK - I'll put together a PR.

aSemy: Ah right, yes that would be slower. I agree, using GitHub parallel workflows is best, not multiple Gradle tasks.

ilya-g:
Hi @aSemy. 
I've noticed that after this change the artifact `dokka-analysis` started publishing an additional `shadowRuntimeElements` variant with the following attributes:
```
    {
      "name": "shadowRuntimeElements",
      "attributes": {
        "org.gradle.category": "library",
        "org.gradle.dependency.bundling": "shadowed",
        "org.gradle.libraryelements": "jar",
        "org.gradle.usage": "java-runtime"
      },
      "files": [
        {
          "name": "dokka-analysis-1.8.20-local-1-all.jar",
          "url": "dokka-analysis-1.8.20-local-1-all.jar",
          "size": 100899229,
          ...
        }
      ]
    }
```

This is unfortunate, because it causes variant resolution ambiguity with the `runtimeElements` variant in `...Plugin` configurations of manually created `DokkaTaskPartial` tasks in the stdlib docs build.

Is this an expected consequence of this change?
- if no, could you investigate why it has happened and how to stop publishing this variant?
- if yes, could you provide a guidance how to overcome this variant ambiguity in a project?

PR #2701, opened 2022-10-11
rename DokkaPlugin file

aSemy:
minor fix: rename file to match the class within

part of #2700

PR #2705, opened 2022-10-11
Move Gradle tasks into task package

aSemy:
part of #2700 

depends on 

* #2701 
* #2702 

This helps organise the package to make it a little more understandable. All the Gradle tasks are stored in one package.

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaLeafTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaLeafTask.kt
@@ -1,3 +1,5 @@
+@file:Suppress("PackageDirectoryMismatch")
+
 package org.jetbrains.dokka.gradle
 
 import org.gradle.api.NamedDomainObjectContainer
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaParentTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaParentTask.kt
@@ -1,3 +1,5 @@
+@file:Suppress("PackageDirectoryMismatch")
+
 package org.jetbrains.dokka.gradle
 
 import org.gradle.api.Project
@@ -99,4 +101,3 @@ abstract class AbstractDokkaParentTask : AbstractDokkaTask() {
         )
     }
 }
-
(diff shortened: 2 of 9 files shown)


IgnatBeresnev:
While I agree that it'd be nice to change structure here, I think we're going to decline this change since package names are part of public API already, it's used to configure Dokka:

```kotlin
import org.jetbrains.dokka.gradle.DokkaTask

val dokkaHtml by getting(DokkaTask::class) {
    outputDirectory.set(buildDir.resolve("dokka"))
    ...
}
```

Changing the package as proposed might break builds upon update. 

We could take this step, but we plan to integrate Dokka into Kotlin Gradle Plugin at some point (so it's built in, and not a separate dependency), so these tasks might or might not be moved there or some place else, we're not sure. I'd like to avoid introducing too many compatibility problems before then.

We can get back to it as soon as we start on integrating Dokka into KGP if it's still relevant

aSemy:
Understood, this would be a breaking change. 

As a compromise, what about moving the files so they are visually organised, but retaining the package name so the code stays the same? Although I'm not convinced of this myself, I think it would lead to more confusion and perhaps problems.

aSemy:
Alternatively, use a more convention method of leaving the classes in the current package and marking them with `@Deprecated` and `@ReplaceWith`. Because the tasks are abstract classes, they could even extend the moved classes.

I'll update the PR as a demo (because it's a small amount of work) - but I would understand if you still want to decline this PR.

IgnatBeresnev:
:+1:  I also thought about using typealiases to mitigate the breaking changes.

I would still postpone this PR until we have some other breaking changes that require the users to change their configuration, like #2752 or maybe some other Gradle-related refactorings. So that there's a single breaking release, instead of multiple ones in a row

We'll discuss our plans for the near future and try to find a place for it.

aSemy:
> 👍 I also thought about using typealiases to mitigate the breaking changes.
> 
> I would still postpone this PR until we have some other breaking changes that require the users to change their configuration, like #2752 or maybe some other Gradle-related refactorings. So that there's a single breaking release, instead of multiple ones in a row
> 
> We'll discuss our plans for the near future and try to find a place for it.

Whenever you think is best 👍  But the breaking change would be the PR that removes typealiases. I think it's best to share the deprecations as soon as possible, because then it gives more opportunity for the deprecation warning to be seen and acted on.

IgnatBeresnev:
> But the breaking change would be the PR that removes typealiases.

Sorry, "breaking change" was not the best term to use. I meant to describe the situation when a user has to take action. If we merge this and they update Dokka, they'll start seeing deprecation warnings (or maybe even build failures due to `failOnWarning`?), so they'll have to go and investigate. Given that some people update dependencies via automatic bots, it can take more time/effort to update Dokka than expected.

For me it makes sense to group such changes in one release, and there's a very big we'll have something similar in the next one. We'll see after we have an internal planning after the release, hopefully I won't forget to give an update in this PR.

IgnatBeresnev:
The plans for #2752 or any other "breaking" changes are still unclear, but the changes from this PR would improve maintainability of the Gradle plugin significantly, so I really want to merge it one way or another

Proposing to just move the task classes into the physical directory `org.jetbrains.dokka.gradle.tasks`, but leave the package statement as `package org.jetbrains.dokka.gradle`, so that it doesn't break source compatibility. 

The warnings will have to be suppressed with a comment though, which can be a link to this message/PR

IgnatBeresnev (approved):
Thanks!

PR #1194, opened 2020-07-21
Simplify Gradle Plugin

sellmair:
Closes #1108 
Closes #1107 
Closes #995 
Closes #1113 
Closes #1114 
Closes #1285 
Closes #1286
Closes #1287 
Closes #1288 
Closes #1289 
Closes #1290 
Closes #1291 
Closes #1292
Closes #1274
Closes #1106

PR #2706, opened 2022-10-11
Remove 'by' property setter helper function

aSemy:
Part of #2700 

I found these extension functions very confusing. I'm not sure what their purpose is. Do they only rename an existing function? 

It's confusing because 'by' is already a Kotlin keyword for delegation, but these extension functions don't do anything with delegation. I think using the normal `.set()` function makes the code more readable and understandable because `.set()` is a common Gradle function.

This PR includes changes from 

* #2702 
* #2701  
* #2705


IgnatBeresnev:
Not sure why these methods were introduced (the PR itself is huge, #1194), but I agree that it can be confused with Kotlin's `by` keyword, and overall would make the code a little easier to understand.

I'm ready to approve it once it has been rebased and contains no changes from other PRs, so let's wait for the other ones to be merged/closed first

Goooler:
Use `by` to set values in Kotlin dsl is reasonable, no need to replace it.

IgnatBeresnev:
> Use by to set values in Kotlin dsl is reasonable, no need to replace it.

That's not part of the DSL though, and I honestly fail to see what value it brings. If it's considered to be a good practice, could you link to examples/docs/etc?

`by` makes me think it's a delegate of some sort, or that there's some complicated logic with how the value is assigned or computed, whereas it's a simple `set`

Goooler:
Alright, I haven't found strong evidence for using `by` in setting properties, it might due to my false memory.

aSemy:
> Alright, I haven't found strong evidence for using `by` in setting properties, it might due to my false memory.

Are you thinking about how you can do `val someTask by tasks.registering { }`? https://docs.gradle.org/current/userguide/kotlin_dsl.html#using_kotlin_delegated_properties

IgnatBeresnev:
I'd be happy to merge it now the changes don't include the postponed #2702 and #2705

aSemy:
done https://github.com/Kotlin/dokka/pull/2834/files

PR #2709, opened 2022-10-13
Dokka Gradle Plugin: Implement Gradle BuildService for Dokka execution

aSemy:
Refactor the way dokka-gradle executes a Dokka generation task to use a Gradle BuildService, and `javaexec {}`. This replaces the existing 'reflection magic'. I think this is a lot more simple than the previous method of using reflection to create an instance.

Benefits:

* idiomatic Gradle
* no reflection magic
* easier to configure/debug/test
* BuildServices can be restricted so they don't run in parallel (reduces risk of Dokka running twice and corrupting output)
* no API changes for users - this should 'just work' (though this needs testing!)

I haven't tested extensively so there might be some breaking changes. Of particular note is the logging output.

Possible improvements:

* Instead of a new main class `DokkaGeneratorMainRunner.kt`, the Gradle Plugin could add dokka-cli to the runtime classpath and re-use the `main()` method within? This would help with re-using code, and helps align dokka-cli and dokka-gradle.

This PR includes #2702


aSemy:
I'm closing this in favour of #2740, which I find more understandable and idiomatic. However, this is dependent on the compatibility of #2740 - this approach is equally valid, and both this PR and the other are a solid improvement over the current situation.

PR #2710, opened 2022-10-13
infrastructure: add concurrency cancellation for test/build GitHub Actions

aSemy:
I see that in the GitHub Actions the build queue can get 'stacked up' if there are a few smaller commits in quick succession. Since the actions take some time, some small commits can cause the GHA build queue to get congested.

This PR adds a 'concurrency' check for GitHub actions. It means that if a newer trigger comes in (for example, I make two commits in a row on a PR), then the first job will be cancelled, allowing for the new job to run.

This helps save energy, and for faster feedback in PRs.

Docs: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run


.github/workflows/check.yml
```diff
@@ -2,6 +2,10 @@ name: Check
 
 on: pull_request
 
+concurrency:
+  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
```
IgnatBeresnev, line 6: Having `@` and spaces around it is a bit confusing, makes me think it's also some special expression syntax that does something. 

In the linked documentation example they don't use spaces and use `-` instead to connect context variables, which looks more familiar to me

Is there any specific reason to using `@`?

aSemy: No, there's no specific reason. I copied the format from this example: https://docs.github.com/en/actions/examples/using-concurrency-expressions-and-a-test-matrix

It just makes the string a little prettier 

For example: 

```
Canceling since a higher priority waiting request for 'Tests @ aSemy:feat/embedded_static_libs' exists
```

image


But in practice I never ever look at the computed value. The action gets cancelled, and a new one is triggered. So if you want a different format, then go ahead.

IgnatBeresnev: It does indeed make the log messages look more readable :+1: For some reason I thought the group string was an internal thing only

IgnatBeresnev (approved):
Thank you once again!

PR #2711, opened 2022-10-13
enable Gradle build cache by default

aSemy:
Enables [Gradle Build Cache](https://docs.gradle.org/current/userguide/build_cache.html). This should massively increase build speed on CI/CD as Gradle will be able to avoid doing repetitive work.

Since all of the GitHub Actions use the `gradle/gradle-build-action` action, this will automatically share the Gradle Build Cache. (I also added `gradle/gradle-build-action` in the Maven workflow, because I expect the `$USER_HOME/.m2` will be shared and re-used - if not this can be revisited later).

I expect that the GitHub Action build times will improve dramatically, especially for PRs like this where there are no code changes!

(Note that the caching won't be enabled on this PR, because by default only the master/main branch will cache. This is to avoid multiple jobs updating the cache concurrently.)

### DIFF
--- a/.github/workflows/dokka-examples.yml
+++ b/.github/workflows/dokka-examples.yml
@@ -123,6 +123,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: 11
+      - uses: gradle/gradle-build-action@v2 # provides a cache of .m2 dependencies
       - run: mvn compile dokka:dokka
         working-directory: examples/maven
         if: steps.filter.outputs.examples_changed == 'true'
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,3 +17,4 @@ kotlin.code.style=official
 # Gradle settings
 org.gradle.jvmargs=-Xmx6g -XX:MaxMetaspaceSize=2g
 org.gradle.parallel=true
+org.gradle.caching=true


aSemy:
> Is there any source point that `gradle/gradle-build-action` can cache `.m2` folder?

In [the documentation](https://github.com/marketplace/actions/gradle-cache#configuration) it describes how to disable caching `~/.m2/repository`

> ```yaml
>  # ...
>     # Disable caching of ~/.m2/repository/
>     save-maven-dependencies-cache: false
> ```

Goooler:
> In [the documentation](https://github.com/marketplace/actions/gradle-cache#configuration) it describes how to disable caching ~/.m2/repository

But this is `burrunan/gradle-cache-action` instead of `gradle/gradle-build-action`, I can't find anything that contains `.m2` or `save-maven-dependencies-cache` in the Gradle's.

Goooler:
I believe you just enable `org.gradle.caching` is ok.

aSemy:
You are right! Thanks for checking. I've removed it.

I see there's Maven caching for setup-java, which I must have gotten mixed up with. I'll make another PR for enabling that. https://github.com/actions/setup-java#caching-maven-dependencies

PR #2714, opened 2022-10-15
Fix markdown image width.

NightlyNexus:


IgnatBeresnev:
Hi! Current approach seems to work (i.e it produces expected result) - is there something wrong with it?

NightlyNexus:
The current does not actually change the width with GitHub markdown, unfortunately, and leaves the image unscaled, at full width.



**Before:**

Screenshot_20221213_010509





**After:**

Screenshot_20221213_010524

IgnatBeresnev:
Oh, I see, thanks

These markdown files are used to build MkDocs-based documentation ([link here](https://kotlin.github.io/dokka/1.7.20/community/plugins-list/#mathjax)), so it's not really intended to be viewed from the GitHub, although it's a nice bonus

If I change it to the proposed solution, images won't be displayed in the generated web documentation:



There seems to be something wrong with the paths. It works in web documentation if I change `../images` to `../../images`, but then it doesn't work in GFM.

And, unfortunately, I don't see any other way of using images and setting their attributes in [mkdocs documentation](https://squidfunk.github.io/mkdocs-material/reference/images/#usage), so it might be a limitation.

You're welcome to tinker with it yourself, maybe you can find a way. To build docs locally, run 

```
./gradlew clean :docs:mkdocsBuild
```

then the website will be generated under `docs/build/mkdocs/index.html`

IgnatBeresnev:
I'll close the PR as it breaks our main documentation and there doesn't seem to be a way to have it work in both cases.

Thank you for taking the time though, much appreciated. Let us know if I missed something, we'll happily re-open it.

PR #2718, opened 2022-10-18
Add missing DokkaConfiguration import statement

SimonMarquis:


IgnatBeresnev (approved):
Big thanks for noticing and fixing it!

PR #2722, opened 2022-10-19
[ci/cd] GitHub Actions - enabled setup-java Maven caching

aSemy:
Enable caching of Maven dependencies using https://github.com/actions/setup-java#caching-maven-dependencies

I expect that this will improve the build times for GitHub Actions, because Dokka uses Maven in the Maven plugin, and in the integration tests. I've enabled the cache for all usages for consistency, and because the GHA cache is fast there's no real disadvantage to caching, even if it isn't needed.

Related to #2711

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,6 +17,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: 11
+          cache: 'maven'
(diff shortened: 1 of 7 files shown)


IgnatBeresnev (approved):
Have no idea how I missed it, it's quite straightforward, thanks

PR #2724, opened 2022-10-19
Fix source link to the line with name

vmishenev:


plugins/base/src/main/kotlin/transformers/pages/sourcelinks/SourceLinksTransformer.kt
```diff
@@ -92,8 +93,9 @@ class SourceLinksTransformer(val context: DokkaContext) : PageTransformer {
     }
 
     private fun PsiElement.lineNumber(): Int? {
+        val skippedKDoc = children.firstOrNull { it !is KDoc } ?: this
```
IgnatBeresnev, line 96: Will this work for Java sources the same way that it will for Kotlin? Shouldn't there also be something like `it !is JavaDoc` (or a common parent)?

IgnatBeresnev (approved):
:+1: 

The only thing is if there's an annotation on top of a class/method, source link will lead to the line with the annotation, not the name. Either PR's title or behaviour needs to be corrected to avoid confusion, as at the moment it's not 100% accurate

Issue #2729, opened 2022-10-26
Dokka spends 35%+ of its time in GC on kotlinx.coroutines


qwwdfsad:
Hi, I did a short profiling session of Dokka in `kotlinx.coroutines`.
The methodology was pretty straightforward: warmup Gradle daemon, run `./gradlew cleanDokkaHtmlMultiModule cleanDokkaHtmlPartial  dokkaHtmlMultiModule`, profile it with `async-profiler` in CPU/alloc modes, the corresponding flame graphs attached.

The root cause is quite straightforward -- for text-based HTML blocks Dokka invokes `parseWithNormalisedSpaces`, which is harmless in itself, but under the hood, it invokes Jsoup-parser unconditionally, which is not only a slow operation by itself, but also allocates a 64K temporary buffer **for each text element**.

PR #2730 invokes Jsoup conditionally by optimistically looking for `&` in the text first and applying Jsoup only when necessary (also, it's probably worth doing it manually anyway, but it's beyond the scope of my change).

It's quite hard to measure the impact of the change on coroutines because there are a lot of other tasks happening (ktor is probably a better candidate to test against, GC there takes 90% of the CPU time).

My numbers are the following:
* No outstanding `char[]` allocations in the profile (new hotspots are identified)
* 10-15% less GC in the profile (35% -> ~20-25% of all execution time)
* In `no-daemon` mode, it saves around ~15% of CPU time and peak CPU consumption by ~200% (on 16 core OS X)

![dokka](https://user-images.githubusercontent.com/2780116/198056287-39c175b0-6e80-4f29-8cd3-425cd7cc0fbb.svg)
![dokka2](https://user-images.githubusercontent.com/2780116/198056307-680a9bca-42b4-46f2-a49b-b50cb3482155.svg)

qwwdfsad:
The load average is pretty much dropped from 6 to 5.
It probably should be tested on a less performant machine or with cores restriction (which OS X doesn't have, but Linux has via `taskset`)

qwwdfsad:
Closing this one as it is superseded with #2903

PR #2730, opened 2022-10-26
Improve performance of 'parseWithNormalisedSpaces' that is unconditio…

qwwdfsad:
…nally invoked for each HTML text element

Addresses #2729


qwwdfsad:
I'm totally good. 

It cuts a significant portion of allocations **and** CPU, but alas the actual execution time is dominated by classloader per task stuff. So the only downside of merging it now is that we won't be able to brag with "we improved performance by 30%" (as if we did it after Worker isolation)

PR #2733, opened 2022-10-31
Introduce jackson-databind constraint for 2.12.7.1

ryanlewis:
Fixes [CVE-2022-42003](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-42003) without going to 2.13.X branch.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -15,6 +15,12 @@ dependencies {
     val jackson_version: String by project
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jackson_version")
     implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-xml:$jackson_version")
+    val jackson_databind_version: String by project
+    constraints {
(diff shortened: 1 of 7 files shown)


IgnatBeresnev:
Thanks for noticing and submitting the PR!

It looks like this version exists for `jackson-databind` only, and doesn't exist for `jackson-module-kotlin`, which causes build problems since the property is re-used, see CI logs

I think we can introduce a separate property `jackson_databind_version` for a limited time and add an explicit dependency on `jackson-databind`, at least until the next update in which it'll be synced

Do you have the time to implement that?

ryanlewis:
@IgnatBeresnev thanks, will take a look, shouldn't take long

IgnatBeresnev (approved):
Well done, thank you so much!

I'll start lengthy integration tests which should take a few hours to complete, and we'll merge if they pass

IgnatBeresnev:
JFYI integration tests failed due to a different problem with androidx documentation, it should be fixed soon and I'll re-run it

Issue #215, opened 2017-09-11
Intellij IDE doesn't support dokka javadoc format


aleksander-stanczak:
I have following configuration in build.gradle file:

```
buildscript {
    ext.dokka_version='0.9.15'
    repositories {
        jcenter()
    }

    dependencies {
        classpath "org.jetbrains.dokka:dokka-gradle-plugin:$dokka_version"
    }
}

apply plugin: 'org.jetbrains.dokka'

task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
    outputFormat = 'javadoc'
    outputDirectory = "$buildDir/dokkaJavadoc"
}

task packageJavadoc(type: Jar, dependsOn: 'dokkaJavadoc') {
    from "$buildDir/dokkaJavadoc"
    classifier = 'javadoc'
}

artifacts {
    archives packageJavadoc
}
```

This generates properly looking javadoc jar ("${moduleName}-javadoc.jar"). However when imported with Intellij Idea, while it is detected as javdoc jar, IDE support doesn't work. Is it format incompatibility with javadoc produced by dokka or just IDE bug?

yole:
What exactly doesn't work? There is no support for showing documentation from external javadoc for Kotlin declarations, it's only supported for Java.

gszecsi:
I think the next question would be that does IDEA/AndroidStudio support any kind of external documentation for Kotlin declarations? None of the provided Dokka output formats seems to be working.

yole:
IntelliJ IDEA does not support any external documentation for Kotlin sources. Given that the vast majority of Kotlin code available today is open-source, it's unclear what benefit exactly the support for external docs would bring.

gszecsi:
The benefit would be that even non-open-source projects could allow subcontractors or employees to use their libraries written in Kotlin, and still be able to use the Quick Documentation feature of IDEA/Android Studio.
Kotlin is a great language and environment to overcome Java's issues, so I think it is a good point of view that more and more people will consider using it against Java. You are saying that "vast majority of Kotlin code available", but here both the terms "vast majority" and "available" are quite subjective.

I personally (and our subcontractors also) really would like if IDEA/Android Studio supported showing Kotlin external documentation.

Alefas:
[...]

PR #2734, opened 2022-11-01
Fixes "The following documentation URLs were checked" message in Android Studio (IntelliJ)

bomberaya:
### Fixed wrong Javadoc links in Android Studio (IntelliJ)

Before:
Screen Shot 2022-11-01 at 23 27 52

After:
Screen Shot 2022-11-02 at 00 03 20

---
  
### Fixed missing package name in rendered html pages

Before:
Screen Shot 2022-11-01 at 23 56 33 2

After:
Screen Shot 2022-11-01 at 23 56 33

### DIFF
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -5,7 +5,7 @@ import org.jetbrains.dokka.model.*
 internal fun JavadocFunctionNode.getAnchor(): String =
     "$name(${parameters.joinToString(",") {
         when (val bound = if (it.typeBound is Nullable) it.typeBound.inner else it.typeBound) {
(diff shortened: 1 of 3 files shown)


IgnatBeresnev:
Sweet, thanks for fixing it! People have been frustrated with the IntelliJ integration for a long time, so it's much appreciated

Can you please have a look at the failing test and see if there are any other?

```
JavadocLocationTest > resolved signature to one argument function() FAILED
```

bomberaya:
You're welcome :) Is there already an open issue for this?

I will fix the tests as well in a couple of days.

IgnatBeresnev:
> Is there already an open issue for this?

I don't think this particular problem has been reported, but there's an umbrella issue #215

You'll probably find more bugs as you continue using it :sweat_smile: I think the integration has not been tested properly and it was developed mostly based on the visuals. Reporting such bugs and especially fixing them would be much appreciated!

IgnatBeresnev (approved):
LGTM, thanks for the fix!

PR #2736, opened 2022-11-03
Revise README documentation for examples and plugins

IgnatBeresnev:
Preview:
* [Examples](https://github.com/Kotlin/dokka/tree/revise-readme-docs/examples)
* [Plugins](https://github.com/Kotlin/dokka/tree/revise-readme-docs/plugins)


plugins/versioning/README.md
```diff
@@ -0,0 +1,332 @@
...
+import org.jetbrains.dokka.versioning.VersioningConfiguration
+
+buildscript {
+    dependencies {
+        classpath("org.jetbrains.dokka:versioning-plugin:1.7.20")
+    }
+}
+
+dependencies {
+    dokkaPlugin("org.jetbrains.dokka:versioning-plugin:$dokkaVersion")
```
atyrin, line 298: Probably, `$dokkaVersion` should be replaced with the exact version as everywhere

IgnatBeresnev: Yep, you're right! Fixed

PR #2728, opened 2022-10-26
Revise documentation

IgnatBeresnev:
[Link for preview](https://ignatberesnev.github.io/static/dokka/docs/kotlinlang/overview.html).

Alternatively, download and install [Stardust plugin](https://helpserver.labs.jb.gg/help/stardust/installing-stardust.html). Open any `.md` file with documentation, for instance `gradle.md`. On the right sidebar you'll see `Stardust Article Preview`, open it and you'll see rendered documentation.

Changes:
* Renamed `docs` module to `mkdocs` to avoid confusion. You can ignore changes in `mkdocs` module, it's all just moved files.
* Added a new `docs` Stardust module for new documentation.

Fixes #2202
Fixes #1929
Fixes #653
Fixes #614
Fixes #486
Fixes #2269
Fixes #2598

PR #2737, opened 2022-11-03
Revise root README.md documentation

IgnatBeresnev:
It is pretty much ready, but this version has many outgoing links to the main documentation, so we have to wait until #2728 is merged and published, after which proper links can be added.


IgnatBeresnev:
This is ready for review, but not yet ready to be merged. I'll need to update the links once our revamped documentation goes live

PR #2738, opened 2022-11-04
Make not-found-version page dynamic

vmishenev:
image


plugins/versioning/src/main/resources/dokka/scripts/version-navigation.js
```diff
@@ -0,0 +1,40 @@
...
+        appended.onload = function() {
+            const messageBox = appended.contentWindow.document.querySelector(".sub-title")
+            messageBox.textContent="The declaration is unavaibaile in " +
+            elem.textContent + " version, but this exists in "
+            let link = appended.contentWindow.document.createElement("a")
+            link.href = lastVersion.href
+            link.textContent =  lastVersion.textContent
+            link.target = "_parent"
+            messageBox.append(link)
+        };
```
IgnatBeresnev, line 35: If you have stdlib built locally with these changes, could you please publish it somewhere? 

I wanted to check the following cases:

1. Does it always lead to the latest version?  Case: I have a declaration in 1.1, which was removed in 1.2, and the latest version is 1.3. What happens if I go to version 1.1, choose the removed declaration and select 1.2? Will it tell me it exists in 1.3?
2. If you click on the proposed latest version, will it load the whole page or only the main content in the middle? In other words, you will see navigation from which version, initial or proposed?

vmishenev: https://vmishenev.github.io/kotlin-dokka-stdlib/prefinal/

IgnatBeresnev: Not sure how the linked docs were built - it it a bug of the versioning plugin or of configuration? 

`decapitalize` was not deprecated in 1.0

IgnatBeresnev: Also not sure how this happened, but `isISOcontrol` exists in 1.0 and 1.5-1.7. 

If you open it for 1.8, then choose 1.4, it'll say that it exists in 1.0, which is true, but I'd say it's less useful then if it pointed to the latest version

IgnatBeresnev: Navigation can also be confusing. You go to 1.8 and click on `isISOControl`, then choose version 1.1, it changes the dropdown on top to 1.1, but if you click anywhere in the navigation it will lead to 1.8

I reckon it should either not change the version on top, or re-draw navigation menu to fit the newly selected version

plugins/versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersionsNavigationCreator.kt
```diff
@@ -68,9 +67,13 @@ class HtmlVersionsNavigationCreator(private val context: DokkaContext) : Version
...
                                         versioningStorage.currentVersion.dir.resolve("not-found-version.html")
 
-                                a(href = absolutePath?.toRelativeString(position) +
-                                        if (!isExistsFile) "?v=" + version.urlEncoded() else "") {
-                                        text(version)
+                                val href =
+                                    absolutePath?.toRelativeString(position) + if (!isExistsFile) "?v=" + version.urlEncoded() else ""
+                                a(
+                                    href = href,
+                                    classes = if (!isExistsFile) "unavailable-version" else null
```
IgnatBeresnev, line 74: I initially thought that the whole _version_ itself was unavailable, when in reality the _declaration_ for this version is unavailable.

So proposing to change it to `declaration-unavailable`. I also wonder if it's good UX to highlight such versions red, preventing the user from clicking it

Something like this, but prettier

vmishenev: >  I also wonder if it's good UX to highlight such versions red, preventing the user from clicking it

The style of top menu should not depend on a page.

plugins/versioning/src/main/resources/dokka/scripts/version-navigation.js
vmishenev: > Not sure how the linked docs were built - it it a bug of the versioning plugin or of configuration?

None of them. It can be done in the filter plugin, but now it works like old Dokka: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/decapitalize.html
Anyway, we should display the version of Kotlin for `DeprecatedSinceKotlin`.

vmishenev: > Also not sure how this happened, but isISOcontrol exists in 1.0 and 1.5 - 1.7 only.

It is a bug in the filter plugin and not relevant to this PR. I will fix it after merging.

PR #2739, opened 2022-11-04
Update tested versions for integration tests

IgnatBeresnev:


integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/kotlin/StdlibGradleIntegrationTest.kt
```diff
@@ -17,7 +17,7 @@ class StdlibGradleIntegrationTest(override val versions: BuildVersions) : Abstra
         @get:JvmStatic
         @get:Parameterized.Parameters(name = "{0}")
         val versions = BuildVersions.permutations(
-            gradleVersions = listOf("5.6"),
+            gradleVersions = listOf("6.7.1"),
```
IgnatBeresnev, line 20: This is the main change and reason for this PR. Looks like Gradle 5 does not support Kotlin DSL plugin, see #2702 

I think our minimum supported Gradle version should be the same as in KGP, which is 6.7.1, so there's no point testing against 5.6

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
```diff
@@ -2,28 +2,32 @@ package org.jetbrains.dokka.it.gradle
...
         ) + BuildVersions.permutations(
-            gradleVersions = listOf("6.9", *ifExhaustive("6.1.1", "5.6.4")),
+            androidGradlePluginVersions = listOf("4.2.0"),
+            gradleVersions = listOf("6.7.1"),
             kotlinVersions = listOf("1.7.20", "1.6.21", "1.5.31", "1.4.32"),
-            androidGradlePluginVersions = listOf("4.0.0", *ifExhaustive("3.6.3"))
+        ) + BuildVersions.permutations(
+            androidGradlePluginVersions = listOf(*ifExhaustive("3.6.4")),
+            gradleVersions = listOf(*ifExhaustive("6.7.1")),
+            kotlinVersions = listOf(*ifExhaustive("1.7.20", "1.6.21", "1.5.31", "1.4.32")),
```
IgnatBeresnev, line 30: Made some changes which should hopefully be compatible. This may help repair `isExhaustive` flagged runs.

1. Latest AGP with latest Gradle and supported Kotlin versions. Version 7 of AGP works with Gradle 7 only
2. Previous major AGP version with its supported Gradle version (AGP 4.2+ works with Gradle 6.7)
3. Really old version of AGP, not sure if it'll work

vmishenev (approved):

aSemy:
The tests in https://github.com/Kotlin/dokka/pull/2702 are now passing - so perhaps this PR isn't needed? Although dropping support for ancient Gradle versions is not a bad thing!

IgnatBeresnev:
@aSemy the tests are passing because of #2826 - they're just not run at all :)

IgnatBeresnev:
There will be merge conflicts after updating to 1.8.0, and it should be revisited after merging #2841 anyway, so I'm closing it for now, but will either re-open it or make a new one after 1.8.10 release.

Issue #2716, opened 2022-10-17
Mark plugin API as Experimental/Unstable


IgnatBeresnev:
Currently there's no indication that Dokka's plugin API is not stable

An experimental/unstable opt-in annotation should be created and put on an entry point, probably `DokkaPlugin`, which all plugins should override.

PR #2743, opened 2022-11-09
Mark Plugin Api as experimental

vmishenev:
#2716

### DIFF
--- a/core/src/main/kotlin/plugability/DokkaPlugin.kt
+++ b/core/src/main/kotlin/plugability/DokkaPlugin.kt
@@ -9,6 +9,11 @@ import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KProperty
 import kotlin.reflect.KProperty1
 
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING, message = "The Plugin API is experimental. It can be incompatibly changed in the future.")
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.FIELD)
+@Retention(AnnotationRetention.BINARY)
+annotation class ExperimentalDokkaPluginApi
+
 abstract class DokkaPlugin {
     private val extensionDelegates = mutableListOf<KProperty<*>>()
     private val unsafePlugins = mutableListOf<Lazy<Extension<*, *, *>>>()
@@ -25,8 +30,10 @@ abstract class DokkaPlugin {
         )
     }
 
+    @OptIn(ExperimentalDokkaPluginApi::class)
     protected fun <T : Any> extending(definition: ExtendingDSL.() -> Extension<T, *, *>) = ExtensionProvider(definition)
 
+    @ExperimentalDokkaPluginApi
     protected class ExtensionProvider<T : Any> internal constructor(
         private val definition: ExtendingDSL.() -> Extension<T, *, *>
     ) {
@@ -56,15 +63,18 @@ interface WithUnsafeExtensionSuppression {
 
 interface ConfigurableBlock
 
+@ExperimentalDokkaPluginApi
 inline fun <reified P : DokkaPlugin, reified E : Any> P.query(extension: P.() -> ExtensionPoint<E>): List<E> =
     context?.let { it[extension()] } ?: throwIllegalQuery()
 
+@ExperimentalDokkaPluginApi
 inline fun <reified P : DokkaPlugin, reified E : Any> P.querySingle(extension: P.() -> ExtensionPoint<E>): E =
     context?.single(extension()) ?: throwIllegalQuery()
 
 fun throwIllegalQuery(): Nothing =
     throw IllegalStateException("Querying about plugins is only possible with dokka context initialised")
 
+@ExperimentalDokkaPluginApi
 inline fun <reified T : DokkaPlugin, reified R : ConfigurableBlock> configuration(context: DokkaContext): R? =
     context.configuration.pluginsConfiguration.firstOrNull { it.fqPluginName == T::class.qualifiedName }
         ?.let { configuration ->
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,6 +22,7 @@ allprojects {
         kotlinOptions {
             freeCompilerArgs = freeCompilerArgs + listOf(
                 "-opt-in=kotlin.RequiresOptIn",
+                "-opt-in=org.jetbrains.dokka.plugability.ExperimentalDokkaPluginApi",
                 "-Xjsr305=strict",
                 "-Xskip-metadata-version-check",
                 // need 1.4 support, otherwise there might be problems with Gradle 6.x (it's bundling Kotlin 1.4)
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -4378,6 +4378,9 @@ public final class org/jetbrains/dokka/plugability/DokkaPluginKt {
 	public static final fun throwIllegalQuery ()Ljava/lang/Void;
 }
 
+public abstract interface annotation class org/jetbrains/dokka/plugability/ExperimentalDokkaPluginApi : java/lang/annotation/Annotation {
+}
+
 public final class org/jetbrains/dokka/plugability/ExtendingDSL {
 	public fun <init> (Ljava/lang/String;Ljava/lang/String;)V
 	public final fun applyIf (Lorg/jetbrains/dokka/plugability/Extension;Lkotlin/jvm/functions/Function1;)Lorg/jetbrains/dokka/plugability/Extension;


plugins/build.gradle.kts
```diff
@@ -27,4 +28,11 @@ subprojects {
             showStackTraces = true
         }
     }
+    tasks.withType(KotlinCompile::class).all {
+        kotlinOptions {
+            freeCompilerArgs = freeCompilerArgs + listOf(
+                "-opt-in=org.jetbrains.dokka.plugability.ExperimentalDokkaPluginApi"
+            )
+        }
+    }
```
IgnatBeresnev, line 37: While convenient, it might hide user annoyances/problems since we wouldn't be dogfooding our own plugin API

If current changes require plugin authors to add too many annoying `OptIn` annotations and they use similar approaches to just suppress all warnings (present and future), it makes the annotation kind of useless

Let's remove it and just opt in place where it's needed? Then we can see better how it'll be used and it would add transparency to our own plugins

core/src/main/kotlin/plugability/DokkaPlugin.kt
```diff
@@ -9,6 +9,11 @@ import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KProperty
 import kotlin.reflect.KProperty1
 
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING, message = "The Plugin API is experimental. It can be incompatibly changed in the future.")
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.FIELD)
+@Retention(AnnotationRetention.BINARY)
+annotation class ExperimentalDokkaPluginApi
```
IgnatBeresnev, line 15: I wonder if it should be named `PluginApiPreview` instead (or something similar with `Preview`). Two reasons:

`ExperimentalDokkaPluginApi` sounds too generic, like something that should be used selectively and for features/api with smaller scope. If we introduce it now with this name and make users suppress it on plugin/file/project level, they will probably never remove it as long as there's some experimental API and the annotation is "used". However, they won't know that some API became stable and some other API (including new) became experimental.

With `Preview`, we're clearly grouping a set of API (all of it in our case) under one annotation. After we revise plugin API, we can remove the `Preview` annotation and the user should hopefully notice it (useless `OptIn` annotation inspection).

However, I am almost certain that after we revise the API, not all of it will become stable - this is where `ExperimentalDokkaPluginApi` would come in handy. We'd make, for example, 90% of API stable, and then mark remaining 10% with `ExperimentalDokkaPluginApi` (or something similar, maybe also under a group if possible).

___

`Experimental` sounds scary, as if we can remove it altogether without providing a migration path or even an alternative. I would be scared to build plugins around that. `Preview` indicates to me that some thought was put into designing the API, but it's not final yet and it might change, but not to the point where your plugin cannot exist anymore. The latter seems to be our case - we'll most likely just refactor it: name/package changes, data class to class migration, enum to sealed object, move extension points around, etc. All of these changes, even though breaking, are manageable with proper migration guides. So `Preview` to me better describes awaiting changes

core/src/main/kotlin/plugability/DokkaPlugin.kt
```diff
@@ -9,6 +9,12 @@ import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KProperty
 import kotlin.reflect.KProperty1
 
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING, message = "The Plugin API is preview. It is not final yet and it might change, but not to the point where a Dokka plugin cannot exist anymore.")
```
IgnatBeresnev, line 12: Feels like it should sound more official

Took inspiration from `FlowPreview`

```suggestion
@RequiresOptIn(
    level = RequiresOptIn.Level.WARNING,
    message = "All of Dokka's plugin API is in preview and it can be changed " +
            "in a backwards-incompatible manner with a best-effort migration. " +
            "By opting in, you acknowledge the risks of relying on preview API."
)
```

IgnatBeresnev:
In case you missed it:

```text
> Task :plugins:all-modules-page:compileTestKotlin FAILED
e: warnings found and -Werror specified
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (21, 29): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (21, 44): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (26, 5): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (26, 23): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (32, 28): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (32, 69): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\MultiModuleDokkaTestGenerator.kt: (32, 87): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\templates\MultiModuleDocumentationTest.kt: (42, 9): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\templates\ResolveLinkCommandResolutionTest.kt: (62, 9): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\templates\ResolveLinkCommandResolutionTest.kt: (91, 9): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\templates\ResolveLinkGfmCommandResolutionTest.kt: (57, 9): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.
w: D:\a\dokka\dokka\plugins\all-modules-page\src\test\kotlin\templates\ResolveLinkGfmCommandResolutionTest.kt: (57, 55): All of Dokka's plugin API is in preview and it can be changed in a backwards-incompatible manner with a best-effort migration. By opting in, you acknowledge the risks of relying on preview API.

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':plugins:all-modules-page:compileTestKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
   > Compilation error. See log for more details
```

plugins/build.gradle.kts
vmishenev: For a test task, it is the easiest way to add it in the script, otherwise, we would have `OptIn` in every file of tests.

IgnatBeresnev: It is the easiest for us, but it hides this and future problems. 

If plugin writers have to use `@OptIn` in every single test file, it's annoying, and they will either have to do it for every single class, or will have to suppress it like above, which is not obvious and needs to be googled.

Is there no way to suppress it (opt-in) for the base `testInline` or `BaseAbstractTest` declarations, so that it doesn't require the users to opt in?

vmishenev: `DokkaPlugin` is a part of `testInline` signature. 
Changing  `testInline` is not worth it

core/src/main/kotlin/plugability/DokkaPlugin.kt
```diff
@@ -9,13 +9,25 @@ import kotlin.properties.ReadOnlyProperty
...
+
 abstract class DokkaPlugin {
     private val extensionDelegates = mutableListOf<KProperty<*>>()
     private val unsafePlugins = mutableListOf<Lazy<Extension<*, *, *>>>()
 
     @PublishedApi
     internal var context: DokkaContext? = null
 
+    @PreviewDokkaPluginApi
+    protected abstract fun previewDokkaPluginApiEmptyMethod()
```
IgnatBeresnev, line 30: Some concerns with this method:

1. Whenever you override the method in IDEA, it propagates the annotation by default. You can even see it [below](https://github.com/Kotlin/dokka/pull/2743/files#diff-09b22991ead3a3b6b27039712b0391a89941978b85ae77f60016bf9ae39226ab) in one of our own plugins. So the user doesn't have to opt-in, and they don't even see the opt-in message.
2. The name doesn't reflect the intention, it's not very clear what this "empty method" is supposed to mean.
3. The return type can be confusing. Having to override a method that has an empty body feels wrong for some reason.
4. No documentation for the method. The opt-in annotation has the main message, but I think it's worth adding a simple explanation for why the method exists.

I like what Gradle did with Gradle Enterprise, where you have to set "yes" for `termsOfServiceAgree`:
https://github.com/Kotlin/dokka/blob/7544a215fb580ae0c47d1f397334f150d1a1ec65/settings.gradle.kts#L56-L62

It'd be nice to have something similar if we stick to the abstract function idea.

Maybe there's an obvious or just a better way to address the concerns I have (cc @qwwdfsad), but in the short time that I have I could only come up with the following:

```kotlin
enum class PluginApiPreviewAcknowledgement {
    @DokkaPluginApiPreview
    YES
}

abstract class DokkaPlugin {
    ...
    abstract fun pluginApiPreviewAcknowledgement(): PluginApiPreviewAcknowledgement
    ...
}
```

It could then be overridden in a plugin with

```kotlin
class GfmPlugin : DokkaPlugin() {
    ...
    @OptIn(DokkaPluginApiPreview::class)
    override fun pluginApiPreviewAcknowledgement(): PluginApiPreviewAcknowledgement {
        return PluginApiPreviewAcknowledgement.YES
    }
    ...
}
```

There's only one option to choose from, and when you use it - you have to opt in, which that makes you read the message, and it doesn't require to use the opt-in annotation in more than one place.

IgnatBeresnev (commented):
We should check for backward compatibility as well in case you haven't. That is, what happens if a plugin that was written for Dokka 1.7.20 (an existing artifact) is used in conjunction with with Dokka 1.8.10

core/src/main/kotlin/plugability/DokkaPlugin.kt
vmishenev: As for me, I do not against it.

Can you clarify this point? I have not understood this case. What do you mean?
> So the user doesn't have to opt-in, and they don't even see the opt-in message.

IgnatBeresnev: > Can you clarify this point? I have not understood this case. What do you mean?

You don't have to write `@OptIn(DokkaPluginApiPreview::class)` with the current approach, see [AndroidDocumentationPlugin](https://github.com/Kotlin/dokka/pull/2743/files#diff-09b22991ead3a3b6b27039712b0391a89941978b85ae77f60016bf9ae39226ab) or any other

qwwdfsad: Nice catch with IDE propagation!

>Can you clarify this point? I have not understood this case. What do you mean?

Right now, if one implements `DokkaPlugin` and uses IDEA to autogenerate method bodies (alt+enter -> Implement members), IDEA automatically opt-ins:

```
class Foo : DokkaPlugin {
   
}
```

alt+enter -> enter -> enter ->

```
class Foo : DokkaPlugin {
    @DokkaPluginApiPreview // <- THIS LINE
   override fun previewDokkaPluginApiEmptyMethod() {}
}
```

and the user effectively does not receive opt-in message at all.

I am not really fond of additional steps with `pluginApiPreviewAcknowledgement = YES` (hello there, Gradle build scans), but it seems to be the only option.

TL;DR

* We have a preview annotation
* Abstract method itself is not marked with this annotation
* Method return type is singleton (`object`) with a meaningful name, so users don't any options except return this object immediately and explicitly opt-in.

The name of the method and object is subject to bikeshed, so pick any you like. The key here is to have a meaningful documentation (that mentions the purpose of the method, the fact it is not used and the example of implementation) and meaningful opt-in message

qwwdfsad: (on a side note, we already have the concept of "enum with a single entry to emulate singleton", it's `object` :))

IgnatBeresnev: Note: since it marks the whole function as `@RequiresOptIn`, we'll have to suppress it on our end so that we don't see the warning in our logs.

<details>
<summary>screenshot</summary>



</details>

IgnatBeresnev: (just in case: I'm fine with both object and enum entry solutions, whichever one you prefer)

vmishenev: In case of entry, IDE generetes :

core/src/main/kotlin/plugability/DokkaPlugin.kt
```diff
@@ -9,13 +9,32 @@ import kotlin.properties.ReadOnlyProperty
...
+
 abstract class DokkaPlugin {
     private val extensionDelegates = mutableListOf<KProperty<*>>()
     private val unsafePlugins = mutableListOf<Lazy<Extension<*, *, *>>>()
 
     @PublishedApi
     internal var context: DokkaContext? = null
 
+    @DokkaPluginApiPreview
+    protected abstract fun warnDokkaPluginApiIsInPreviewEmptyMethod(): PluginApiPreviewAcknowledgement
```
IgnatBeresnev, line 37: It's unclear what the `EmptyMethod` in the name is supposed to indicate. The intended behaviour, I think, is that the user returns `PluginApiPreviewAcknowledgement`, so it will not be empty. I don't think it can considered to be empty even if it has the default `TODO()`. Proposing to rename it so that it's the same as the enum - `pluginApiPreviewAcknowledgement`.

Can you also add a `@see [PluginApiPreviewAcknowledgement]` to the KDoc (or vise versa)? Users will most likely first look at the documentation of the method itself, not the return type

IgnatBeresnev (approved):

PR #2744, opened 2022-11-13
Fix a typo in the root documentation page

ansman:


IgnatBeresnev (approved):
Thank you for the fix!

PR #2755, opened 2022-11-25
Dispose `AnalysisEnvironment`

vmishenev:


kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/EnvironmentAndFacade.kt
```diff
@@ -51,7 +51,10 @@ class DokkaMessageCollector(private val logger: DokkaLogger) : MessageCollector
 }
 
 // It is not data class due to ill-defined equals
-class EnvironmentAndFacade(val environment: KotlinCoreEnvironment, val facade: DokkaResolutionFacade) {
+class EnvironmentAndFacade(val environment: KotlinCoreEnvironment, val facade: DokkaResolutionFacade, private val analysisEnvironment: AnalysisEnvironment) {
     operator fun component1() = environment
     operator fun component2() = facade
+    protected fun finalize() {
+        analysisEnvironment.dispose()
+    }
```
IgnatBeresnev, line 59: oof, there's a great deal of problems and uncertainties with `finalize`, but the biggest one is that it's deprecated for removal, so it will fail sooner or later: https://openjdk.org/jeps/421

Is it possible to close it explicitly? Make the class implement `Closeable` and wrap usages in `.use {}` or close it some other way when the generation has been completed? 

This could even be an extension point or a feature where plugins register closeable components, and they are all closed at the end of documentation generation or as soon as documentable model has been built (probably somewhere in `SingleModuleGeneration#generate`)

IgnatBeresnev (changes requested):

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/KotlinAnalysis.kt
```diff
@@ -34,20 +59,30 @@ class DokkaAnalysisConfiguration(
...
 }
 
+/**
+ * It does not close [defaultKotlinAnalysis].
+ */
 internal class KotlinAnalysisImpl(
-    private val environments: SourceSetDependent<EnvironmentAndFacade>
+    private val environments: SourceSetDependent<EnvironmentAndFacade>,
+    private val defaultKotlinAnalysis: KotlinAnalysis? = null
 ) : KotlinAnalysis, SourceSetDependent<EnvironmentAndFacade> by environments {
```
IgnatBeresnev, line 73: If I understand correctly, there can be two cases:
1. `KotlinAnalysisImpl` used as the main analysis for the project. Here `defaultKotlinAnalysis` would be `null`.
2. `KotlinAnalysisImpl` used for analysis of some samples, which requires more/different configuration of the classpath/source roots. Here `defaultKotlinAnalysis` will NOT be null and will have the value from the first case.

If my understanding is correct, I would propose to create separate implementations for them. 

To demonstrate:

```kotlin
internal class ProjectKotlinAnalysis(
    private val environments: SourceSetDependent<EnvironmentAndFacade>
) : KotlinAnalysis, SourceSetDependent<EnvironmentAndFacade> by environments {

    override fun get(key: DokkaSourceSet): EnvironmentAndFacade {
        return environments[key] ?: throw IllegalStateException("Missing EnvironmentAndFacade for sourceSet $key")
    }

    override fun get(sourceSetID: DokkaSourceSetID): EnvironmentAndFacade {
        return environments.entries.first { (sourceSet, _) -> sourceSet.sourceSetID == sourceSetID }.value
    }

    override fun close() {
        environments.values.forEach(EnvironmentAndFacade::close)
    }
}

/**
 * Samples analysis is used for analyzing samples bla bla bla
 * ...
 * Sample analysis is short-lived, so it is closed before [projectKotlinAnalysis].
 * ...
 * [projectKotlinAnalysis] needs to be closed separately.
 */
internal class SamplesKotlinAnalysis(
    private val sampleEnvironments: SourceSetDependent<EnvironmentAndFacade>,
    private val projectKotlinAnalysis: ProjectKotlinAnalysis
) : KotlinAnalysis, SourceSetDependent<EnvironmentAndFacade> by sampleEnvironments {
    
    override fun get(key: DokkaSourceSet): EnvironmentAndFacade {
        return sampleEnvironments[key] ?: projectKotlinAnalysis[key]
    }

    override fun get(sourceSetID: DokkaSourceSetID): EnvironmentAndFacade {
        return sampleEnvironments.entries.firstOrNull { (sourceSet, _) -> sourceSet.sourceSetID == sourceSetID }?.value
            ?: projectKotlinAnalysis[sourceSetID]
    }

    override fun close() {
        sampleEnvironments.values.forEach(EnvironmentAndFacade::close)
    }
}
```

plugins/base/src/main/kotlin/DokkaBase.kt
```diff
@@ -280,6 +281,10 @@ class DokkaBase : DokkaPlugin() {
         externalClasslikesTranslator providing ::DefaultDescriptorToDocumentableTranslator
     }
 
+    val disposeKotlinAnalysisPostAction by extending {
+        CoreExtensions.postActions with PostAction { this@DokkaBase.querySingle { kotlinAnalysis }.close() }
+    }
```
IgnatBeresnev, line 286: Can it be `internal`?

plugins/base/src/main/kotlin/transformers/pages/samples/SamplesTransformer.kt
```diff
@@ -38,42 +34,26 @@ abstract class SamplesTransformer(val context: DokkaContext) : PageTransformer {
...
         runBlocking(Dispatchers.Default) {
-            val analysis = setUpAnalysis(context)
-
-            input.transformContentPagesTree { page ->
-                val samples = (page as? WithDocumentables)?.documentables?.flatMap {
-                    it.documentation.entries.flatMap { entry ->
-                        entry.value.children.filterIsInstance<Sample>().map { entry.key to it }
+            val analysis = KotlinAnalysisForSamples(sourceSets = context.configuration.sourceSets,
+                logger = context.logger,
+                defaultKotlinAnalysis = context.plugin<DokkaBase>().querySingle { kotlinAnalysis })
```
IgnatBeresnev, line 39:
```suggestion
            val analysis = KotlinAnalysisForSamples(
                sourceSets = context.configuration.sourceSets,
                logger = context.logger,
                defaultKotlinAnalysis = context.plugin<DokkaBase>().querySingle { kotlinAnalysis }
            )
```

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/EnvironmentAndFacade.kt
```diff
@@ -8,30 +8,48 @@ import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSourceLocation
...
 
-        val environment = createCoreEnvironment()
+    val environment = createCoreEnvironment()
 
-        val (facade, _) = createResolutionFacade(environment, analysisConfiguration.ignoreCommonBuiltIns)
-        EnvironmentAndFacade(environment, facade)
-    }
+    val (facade, _) = createResolutionFacade(environment, analysisConfiguration.ignoreCommonBuiltIns)
+    EnvironmentAndFacade(environment, facade, this)
+}
```
IgnatBeresnev, line 52: The very first silly thought I had was that `this` was somehow leaking through a lambda passed to the constructor

It seems like within the same `run` block there's configuration of `AnalysisEnvironment` and construction of objects for `EnvironmentAndFacade` with omitted return - I guess that confused me. Let's separate the two?

Something like that
```kotlin
internal fun createEnvironmentAndFacade(
    logger: DokkaLogger,
    classpath: List<File>,
    sourceRoots: Set<File>,
    sourceSet: DokkaConfiguration.DokkaSourceSet,
    analysisConfiguration: DokkaAnalysisConfiguration
): EnvironmentAndFacade {

    val analysisEnvironment = AnalysisEnvironment(
        messageCollector = DokkaMessageCollector(logger),
        analysisPlatform = sourceSet.analysisPlatform
    ).apply {
        if (analysisPlatform == Platform.jvm) {
            configureJdkClasspathRoots()
        }
        addClasspath(classpath)
        addSources(sourceRoots)
        loadLanguageVersionSettings(sourceSet.languageVersion, sourceSet.apiVersion)
    }

    val environment = analysisEnvironment.createCoreEnvironment()
    val (facade, _) = analysisEnvironment.createResolutionFacade(environment, analysisConfiguration.ignoreCommonBuiltIns)

    return EnvironmentAndFacade(environment, facade, analysisEnvironment)
}
```

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/EnvironmentAndFacade.kt
```diff
@@ -51,7 +69,21 @@ class DokkaMessageCollector(private val logger: DokkaLogger) : MessageCollector
 }
 
 // It is not data class due to ill-defined equals
-class EnvironmentAndFacade(val environment: KotlinCoreEnvironment, val facade: DokkaResolutionFacade) {
+class EnvironmentAndFacade(
+    environment: KotlinCoreEnvironment,
+    facade: DokkaResolutionFacade,
+    private val analysisEnvironment: AnalysisEnvironment
+) : Closeable {
```
IgnatBeresnev, line 76: Looks like this should now be named `EnvironmentAndFacadeAndAnalysis` :) 

What do you think about renaming it to something more universal and something that you would expect to be closeable (i.e to contain some logic) as opposed to just being a holder/pair for two  objects?

Maybe `AnalysisContext` or something?

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/KotlinAnalysis.kt
```diff
@@ -34,20 +62,39 @@ class DokkaAnalysisConfiguration(
...
+) : KotlinAnalysisImpl(sampleEnvironments, projectKotlinAnalysis)
 
-    override fun get(key: DokkaSourceSet): EnvironmentAndFacade {
-        return environments[key] ?: throw IllegalStateException("Missing EnvironmentAndFacade for sourceSet $key")
+/**
+ * It does not close [defaultKotlinAnalysis].
+ */
+internal open class KotlinAnalysisImpl(
+    private val environments: SourceSetDependent<AnalysisContext>,
+    private val defaultKotlinAnalysis: KotlinAnalysis? = null
```
IgnatBeresnev, line 84: Sorry, I didn't go into too much detail with my suggestion in the previous comment -- unfortunately, it's is still relevant.

The reason I suggested to separate the implementation into `ProjectKotlinAnalysis` and `SamplesKotlinAnalysis` is to avoid having an optional nullable field with a confusing name (`defaultKotlinAnalysis`).

It's unclear what `defaultKotlinAnalysis` is - it's default in relation to what? How is it limited to being "default"? As far as I see, it's just some kotlin analysis, can be anything.

```kotlin
internal class ProjectKotlinAnalysis(
    private val environments: SourceSetDependent<AnalysisContext>
)

internal class SamplesKotlinAnalysis(
    private val sampleEnvironments: SourceSetDependent<AnalysisContext>,
    private val projectKotlinAnalysis: KotlinAnalysis
)
```

^ Here you don't have that problem. `ProjectAnalysis` only has it's own environments. `SamplesAnalysis` has project's analysis and environments for samples, which is logical.

Additionally, if more optional/nullable fields get added, it'll be even more difficult to reason about (i.e in which case is field 1 null and field 2 not null or vise versa), which can be avoided early.

So my suggestion with separating the implementation  completely (without the base class) still stands, even if there's a little bit of "copy-paste".

Is there anything specific you don't like about it?

vmishenev: Probably `defaultKotlinAnalysis`  is not a good name here. I will rename it to `fallbackKotlinAnalysis `.



>Additionally, if more optional/nullable fields get added, it'll be even more difficult to reason about (i.e in which case is field 1 null and field 2 not null or vise versa), which can be avoided early.

What do you mean? Why should we avoid nullable fields in Kotlin? As for me, it is ok to have no  `fallbackKotlinAnalysis here.`  

> Is there anything specific you don't like about it?

The current implementation is more general and does not have  "copy-paste".

IgnatBeresnev: Discussed it separately and agreed on another solution, marking as resolved

kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/KotlinAnalysis.kt
```diff
@@ -34,20 +62,53 @@ class DokkaAnalysisConfiguration(
...
+    }
+
+    operator fun get(sourceSetID: DokkaSourceSetID): AnalysisContext {
+        return find(sourceSetID)
+            ?: parent?.get(sourceSetID)
+            ?: throw IllegalStateException("Missing EnvironmentAndFacade for sourceSet ${sourceSetID.sourceSetName}")
+    }
+
+    protected abstract fun find(sourceSetID: DokkaSourceSetID): AnalysisContext?
+    protected abstract fun find(sourceSet: DokkaSourceSet): AnalysisContext?
```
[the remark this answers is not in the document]
vmishenev: `DokkaSourceSetID` can not be unique.  Anyway, we do not check it. And it is a more efficient search.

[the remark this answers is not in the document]
IgnatBeresnev: Why do you mean by `more efficient search`? It's an abstract method :) If you're talking about the implementation - maybe, and maybe not. `DokkaSourceSetImpl` has 20+ fields, half of which are collections. I'm not sure comparing it by `equals()` is much faster than iterating over a list once and comparing strings.

Also, can you please give an example of when it's not unique? How do you then identify a unique `DokkaSourceSet`? I'm asking because we rely on `DokkaSourceSetID` for source set dependencies, which might lead to bugs in this case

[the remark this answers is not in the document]
vmishenev: > It's an abstract method :)

Yes, you could have a more efficient implementation than `find(sourceSet.sourceSetID)`.

> I'm not sure comparing it by equals() is much faster than iterating over a list once and comparing strings.

`equals()` can be not needed since we have a hash table. Otherwise, they will be compared by a reference first (our cases).

> can you please give an example of when it's not unique?

https://github.com/Kotlin/dokka/pull/2755/commits/833a9dc41a78a97be78c96cb88f9df82f2e36eb3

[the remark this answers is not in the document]
IgnatBeresnev: It can surely be non unique in our test cases where we set the values manually. Can it be non unique in a real production library/project? If it can, I don't think this method should exist at all

> Yes, you could have a more efficient implementation than find(sourceSet.sourceSetID)

I'm sorry, I'm not buying it. What stops the implementation from having a `Map<DokkaSourceSetID, value>`? How is it "more efficient" on the interface level? Even if so, it doesn't make sense to have two methods that do the same thing from the outside, but one is more efficient than the other - how does the user know which one to call? It's a leaky abstraction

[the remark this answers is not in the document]
vmishenev: > How is it "more efficient" on the interface level? 

I left it due to correctness rather than performance. 

I think it is better to remove 
``protected abstract fun find(sourceSetID: DokkaSourceSetID): AnalysisContext?``
Objections?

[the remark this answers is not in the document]
IgnatBeresnev: Sure 👍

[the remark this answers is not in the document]
IgnatBeresnev: Sorry, you've deprecated it with the reason `sourceSetID can be non unique`, which is a serious statement, but I still don't understand in which situations in can be non unique. Can you please give an example?

I had a closer look at the [commit that you linked above](https://github.com/Kotlin/dokka/commit/833a9dc41a78a97be78c96cb88f9df82f2e36eb3), and it's definitely a configuration error that has not been fixed. In the [test source set builder](https://github.com/Kotlin/dokka/blob/8c218ff4dd5f970233c43845c19299fc74256389/core/test-api/src/main/kotlin/testApi/testRunner/TestDokkaConfigurationBuilder.kt#L112), the ID is constructed as `DokkaSourceSetID(moduleName, name)`, which by default is `(root, main)`. In the commit, `displayName` was added, which helps in `DokkaSourceSet#equals`, but there are still two source sets with the same name `main`, which is invalid configuration.

IgnatBeresnev (approved):

PR #2756, opened 2022-11-25
Bump Gradle to 7.6

Goooler:
https://docs.gradle.org/7.6/release-notes.html


Goooler:
@IgnatBeresnev Maybe it's time to merge this first.

IgnatBeresnev:
@Goooler indeed, thanks for the reminder! 

I've started teamcity-based integration tests just in case (our GitHub ones are broken due to #2826). It'll take some hours to complete, and I'll get back to it check on it on Monday

IgnatBeresnev:
For some reason the stdlib integration test is failing when run via GitHub actions (it's supposed to be broken, just not in this way..), but the same test run on TeamCity passes, so I think it's safe to merge. Thanks!

PR #2740, opened 2022-11-04
Dokka Gradle Plugin: Implement Gradle Worker Daemon for Dokka generation

aSemy:
Refactor the way dokka-gradle executes a Dokka generation task to use a [Gradle Worker Daemon](https://docs.gradle.org/current/userguide/worker_api.html#creating_a_worker_daemon). The Worker is configured to use the same classpath as the existing reflection-based process.

Using a Worker Daemon completely replaces the existing 'reflection magic'. I think this is a lot more simple and easier to understand than the previous method of using reflection to create an instance.

This PR is a further improvement upon #2709, which required a main method. Using the Gradle Worker API means the dokka-core code can be called idiomatically. However, I am not sure about the compatibility with Gradle versions. #2709 might be more compatible...

Benefits:

* idiomatic Kotlin
* idiomatic Gradle
* no reflection magic
* easier to configure/debug/test
* no API changes for users - this should 'just work'

I haven't tested extensively so there might be some breaking changes. Of particular note:

* is the logging output still present and as expected
* is the Worker API compatible with the versions of Gradle that Dokka must support?

This PR includes 

* #2702 
* #2759

PR #2759, opened 2022-11-28
Gradle Plugin: Expose `finalizeCoroutines` as Gradle task option

aSemy:
* Allows the `finalizeCoroutines` option to be configurable from Gradle tasks
* fixes flakey tests in the Kotlinx Serialization and Kotlin Coroutines integration tests

This PR was triggered by an issue in https://github.com/Kotlin/dokka/pull/2740#issuecomment-1329499420. Kotlinx Serialization and Kotlinx Coroutines requires `finalizeCoroutines` to be disabled, else Dokka generation fails

Example error: https://github.com/Kotlin/dokka/actions/runs/3567615789/jobs/5995523064#step:4:4032

```
 Caused by: java.util.concurrent.RejectedExecutionException: DefaultDispatcher was terminated
```

Currently the Kotlinx Serialization and Kotlin Coroutines integration tests are flakey because of this error.

But with this PR, the option can be configured:

```kotlin
tasks.withType<AbstractDokkaTask>().configureEach {
    finalizeCoroutines.set(false)
}
```

I've fixed the flakey tests by setting `finalizeCoroutines` to `false` with a Gradle init script.

Since I would like to see more idiomatic Gradle config #2700, the new code in this PR follows the Gradle API more closely than pre-existing code (for example, using `abstract val` for properties, and setting the default value using `convention()`, so `getSafe()` is not necessary).

This PR includes 

* https://github.com/Kotlin/dokka/pull/2702


IgnatBeresnev:
An update on this: if we do start using Gradle workers from #2740, it looks like this configuration option won't be needed at all for the Gradle plugin users, can just set it to `false` by default. 

Disabling it for coroutines and serialization does make the tests pass, but if you try to run Dokka with gradle workers on any other project - I think it'll also fail until you turn set `finalizeCoroutines` to `false` as well.

Let's wait until we get to #2740 though and decide and test it there

adam-enko:
This PR is no longer necessary. `finalizeCoroutines` is exposed as an option in DGPv2.

Issue #2620, opened 2022-08-11
Do not include sources in Documentable equals/hashCode


owengray-google:
```
    @Test
    fun `hashcode Behavior`() {
        val moduleK = """
            |object Foo {
            |  fun bar() = Unit
            |  const val baz = "baz"
            |}
        """.render()
        val a = moduleK.children.first().children.first() as DObject
        val a2 = a.copy()
        val b = a.copy(sources = emptyMap())
        assertThat(a.equals(a2)).isTrue()
        assertThat(a.hashCode() == a2.hashCode()).isTrue()
        // These lines fail
        assertThat(a.equals(b)).isTrue()
        assertThat(a.hashCode() == b.hashCode()).isTrue()
    }
```

**Describe the bug**
Documentables are data classes, which use autogenerated `equals()` and `hashCode()` that use all of their properties.
The `sources` field of Documentables is a) not useful for conceptual equality, and b) extremely huge when running in practice, and therefore very expensive to evaluate.

There may be other fields that have both of these qualities as well.

**Expected behaviour**
`equals()` and `hashCode()` are efficient and more useful than referential equality.

**Possible solutions**
```
    data class DSomething(val args: String, val sources: String)
    // becomes
    data class DBetterHash private constructor(val args: String) {
        lateinit var sources: String
        constructor(args: String, sources: String) : this(args) { this.sources = sources }
    }
```
The above switch would not require any rewriting outside of `Documentable.kt`. Alternatively, if the constructor swap were not used, it would require rewriting every place a `WithSources` was constructed to later set the `lateinit var sources`.
(this would be much nicer if it could use varargs somehow)

Alternatively, simply override hashCode and equals.

owengray-google:
An overarching bug here is to make `hashCode` and `equals` faster.

PR #2780, opened 2022-12-16
Deduplicate extensions

vmishenev:
image
It happens when we have more than one class with the same name. (i.e. `mergeImplicitExpectActualDeclarations`).

I decided to write no test since #2764 breaks it.

### DIFF
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -354,7 +354,7 @@ open class DefaultPageCreator(
             val extensions = (classlikes as List<WithExtraProperties<DClasslike>>).flatMap {
                 it.extra[CallableExtensions]?.extensions
                     ?.filterIsInstance<Documentable>().orEmpty()
-            }.distinct()
+            }.distinctBy{ it.sourceSets to it.dri} // [Documentable] has expensive equals/hashCode at the moment, see #2620
 
             // Extensions are added to sourceSets since they can be placed outside the sourceSets from classlike
             // Example would be an Interface in common and extension function in jvm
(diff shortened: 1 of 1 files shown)


atyrin:
Shouldn't it be wrapped into platform tabs?

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -354,7 +354,7 @@ open class DefaultPageCreator(
             val extensions = (classlikes as List<WithExtraProperties<DClasslike>>).flatMap {
                 it.extra[CallableExtensions]?.extensions
                     ?.filterIsInstance<Documentable>().orEmpty()
-            }
+            }.distinct()
```
IgnatBeresnev, line 357: Given that we'll be comparing documentables, this might cause performance issues described in #2620 

Can we compare just the DRIs?

```suggestion
            }.distinctBy { it.dri }
```

vmishenev:
> Shouldn't it be wrapped into platform tabs?

Yes, it's already wrapped.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -354,7 +354,7 @@ open class DefaultPageCreator(
             val extensions = (classlikes as List<WithExtraProperties<DClasslike>>).flatMap {
                 it.extra[CallableExtensions]?.extensions
                     ?.filterIsInstance<Documentable>().orEmpty()
-            }
+            }.distinctBy{ it.sourceSets to it.dri} // comparing of [Documentable] is expensive
```
IgnatBeresnev, line 357: Very good idea to leave a comment! I'd stress that it's expensive at the time of writing this code, so that we or other maintainers are not afraid to change it in the future. 

Hopefully, one day `Documentable` will be migrated from a data class to an average class, or at the very least have an inexpensive equals/hashcode implemented

```suggestion
            }.distinctBy { it.sourceSets to it.dri} // [Documentable] has expensive equals/hashCode at the moment, see #2620
```

IgnatBeresnev (approved):

PR #2795, opened 2022-12-27
Default Java constructor

### LINKED ISSUE #2765: Autogenerate default Java constructors
**Describe the bug**
You do not autogenerate the default Java constructor. We autogenerate it downstream, but that causes us issues with the explicitly-hidden-constructor case.

(the first set of assertions is the `Documentable` tree, the second set is the component tree we/dackka generate from it)
```
@Test
fun `Default constructors are autogenerated when no explicit constructor is present`() {
    val moduleJ = """
        public class Foo {}
    """.trimIndent().render(java = true)
    val classlikeJ = moduleJ.page("Foo").data.content
    val moduleK = """
        public class Foo {}
    """.trimIndent().render(java = false)
    val classlikeK = moduleK.page("Foo").data.content

    // test the Documentables tree. The correct behavior would be both isNotEmpty
    for (module in listOf(moduleJ, moduleK)) {
        val classlike = module.explicitClasslike("Foo")
        if (module == moduleK)
            assertThat((classlike as DClass).constructors).isEmpty()
        else
            assertThat((classlike as DClass).constructors).isNotEmpty()
    }

    // test the Documentables tree. The correct behavior is both isNotEmpty
    for (classlike in listOf(classlikeJ, classlikeK)) {
        assertThat(classlike.data.publicConstructorsDetails).isNotEmpty()
    }
}

@Test
fun `Default constructors are not autogenerated when private-constuctor pattern is used`() {
    val moduleJ = """
        public class Foo { private Foo() {} }
    """.trimIndent().render(java = true)
    val classlikeJ = moduleJ.page("Foo").data.content
    val moduleK = """
        public class Foo private constructor() {}
    """.trimIndent().render(java = false)
    val classlikeK = moduleK.page("Foo").data.content

    // test the Documentables tree. The correct behavior is both isEmpty
    for (module in listOf(moduleJ, moduleK)) {
        val classlike = module.explicitClasslike("Foo")
        assertThat((classlike as DClass).constructors).isEmpty()
    }

    // test our Components tree. The correct behavior is both isEmpty.
    // Our fix for the next test breaks us in this case. I think this is unavoidably upstream.
    for (classlike in listOf(classlikeJ, classlikeK)) {
        if (classlike == classlikeK)
            assertThat(classlike.data.publicConstructorsDetails).isNotEmpty()
        else
            assertThat(classlike.data.publicConstructorsDetails).isEmpty()        }
}
```

atyrin:
Fix #2765

If a class doesn't have an explicit constructor (with any visibility), a default synthetic one will be provided.
Affects only `DClass`.

### DIFF
--- a/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
+++ b/plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
@@ -16,14 +16,15 @@ import org.jetbrains.dokka.analysis.KotlinAnalysis
 import org.jetbrains.dokka.analysis.PsiDocumentableSource
 import org.jetbrains.dokka.analysis.from
 import org.jetbrains.dokka.base.DokkaBase
-import org.jetbrains.dokka.base.translators.psi.parsers.JavaDocumentationParser
 import org.jetbrains.dokka.base.translators.psi.parsers.JavadocParser
 import org.jetbrains.dokka.base.translators.typeConstructorsBeingExceptions
 import org.jetbrains.dokka.base.translators.unquotedValue
-import org.jetbrains.dokka.links.*
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.links.nextTarget
+import org.jetbrains.dokka.links.withClass
+import org.jetbrains.dokka.links.withEnumEntryExtra
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.AnnotationTarget
-import org.jetbrains.dokka.model.Nullable
 import org.jetbrains.dokka.model.doc.DocumentationNode
 import org.jetbrains.dokka.model.doc.Param
 import org.jetbrains.dokka.model.properties.PropertyContainer
@@ -45,7 +46,7 @@ import org.jetbrains.kotlin.utils.addToStdlib.safeAs
 import java.io.File
 
 class DefaultPsiToDocumentableTranslator(
-    context: DokkaContext
+    context: DokkaContext,
 ) : AsyncSourceToDocumentableTranslator {
 
     private val kotlinAnalysis: KotlinAnalysis = context.plugin<DokkaBase>().querySingle { kotlinAnalysis }
@@ -94,8 +95,8 @@ class DefaultPsiToDocumentableTranslator(
 
     class DokkaPsiParser(
         private val sourceSetData: DokkaSourceSet,
-        facade: DokkaResolutionFacade,
-        private val logger: DokkaLogger
+        private val facade: DokkaResolutionFacade,
+        private val logger: DokkaLogger,
     ) {
         private val javadocParser = JavadocParser(logger, facade)
         private val syntheticDocProvider = SyntheticElementDocumentationProvider(javadocParser, facade)
@@ -276,10 +277,19 @@ class DefaultPsiToDocumentableTranslator(
                             JavaClassKindTypes.CLASS
                         )
                     }
-                }) + ancestry.interfaces.map { TypeConstructorWithKind(it.typeConstructor, JavaClassKindTypes.INTERFACE) }).toSourceSetDependent()
+                }) + ancestry.interfaces.map {
+                    TypeConstructorWithKind(
+                        it.typeConstructor,
+                        JavaClassKindTypes.INTERFACE
+                    )
+                }).toSourceSetDependent()
                 val modifiers = getModifier().toSourceSetDependent()
                 val implementedInterfacesExtra =
                     ImplementedInterfaces(ancestry.allImplementedInterfaces().toSourceSetDependent())
+
+                fun Array<PsiMethod>.defaultCtorIfEmpty() =
+                    if (isEmpty()) arrayOf(psi.createDefaultConstructor()) else this
+
                 when {
                     isAnnotationType ->
                         DAnnotation(
@@ -303,6 +313,7 @@ class DefaultPsiToDocumentableTranslator(
                                     .toAnnotations()
                             )
                         )
+
                     isEnum -> DEnum(
                         dri = dri,
                         name = name.orEmpty(),
@@ -341,6 +352,7 @@ class DefaultPsiToDocumentableTranslator(
                                 .toAnnotations()
                         )
                     )
+
                     isInterface -> DInterface(
                         dri = dri,
                         name = name.orEmpty(),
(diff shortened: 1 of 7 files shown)


plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -427,7 +455,8 @@ class DefaultPsiToDocumentableTranslator(
                 },
                 documentation = docs.toSourceSetDependent(),
                 expectPresentInSet = null,
-                sources = PsiDocumentableSource(psi).toSourceSetDependent(),
+                // isPhysical detects the virtual methods without real sources. Here it used for default constructor
+                sources = if (psi.isPhysical) PsiDocumentableSource(psi).toSourceSetDependent() else emptyMap(),
```
IgnatBeresnev, line 459: Nice! If you have the time, let's fix it for other documentables as well? There's 3 usages total

Could do the same type of extension as proposed above, something along the lines of:

```kotlin
fun PsiNamedElement.parseSources(): SourceSetDependent<DocumentableSource> {
    return when {
        psi.isPhysical -> PsiDocumentableSource(this).toSourceSetDependent()
        else -> emptyMap()
    }
}
```

plugins/base/src/test/kotlin/signatures/InheritedAccessorsSignatureTest.kt
```diff
@@ -367,9 +383,13 @@ class InheritedAccessorsSignatureTest : BaseAbstractTest() {
 
                 writerPlugin.writer.renderedContent("root/test/-java-class/index.html").let { javaClassContent ->
                     val signatures = javaClassContent.signature().toList()
-                    assertEquals(2, signatures.size, "Expected 2 signatures: class signature and property")
+                    assertEquals(
+                        3,
+                        signatures.size,
+                        "Expected 2 signatures: class signature, default constructor and property"
```
IgnatBeresnev, line 389:
```suggestion
                        "Expected 3 signatures: class signature, default constructor and property"
```

plugins/base/src/test/kotlin/signatures/SignatureTest.kt
```diff
@@ -976,9 +976,13 @@ class SignatureTest : BaseAbstractTest() {
 
                 writerPlugin.writer.renderedContent("root/test/-java-class/index.html").let { kotlinClassContent ->
                     val signatures = kotlinClassContent.signature().toList()
-                    assertEquals(2, signatures.size, "Expected 2 signatures: class signature and property")
+                    assertEquals(
+                        3,
+                        signatures.size,
+                        "Expected 2 signatures: class signature, default constructor and property"
```
IgnatBeresnev, line 982:
```suggestion
                        "Expected 3 signatures: class signature, default constructor and property"
```

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -1,5 +1,6 @@
 package translators
 
+import com.jetbrains.rd.util.first
```
IgnatBeresnev, line 3: That's an interesting package, I actually have no idea what `rd` stands for. But it seems to be coming from one of the intellij dependencies.

To avoid possible compatibility problems in the future (can be deleted/moved/changed any day for all we know), I would change it to an stdlib alternative. 

For example, instead of doing
```kotlin
map.first().value
``` 

I think you can do
```kotlin
map.values.first()
```

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -741,4 +741,93 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
...
+                val testedClass = module.findClasslike(packageName = "test", "A") as DClass
+
+                assertEquals(1, testedClass.constructors.size, "Expect 1 declared constructor")
+                assertEquals(JavaVisibility.Private, testedClass.constructors.first().constructorVisibility())
+            }
+        }
+    }
 }
+
+private fun DFunction.constructorVisibility() = visibility.first().value
```
IgnatBeresnev, line 833: Same thing about scope - it's an extension for any function, there's nothing limiting it to constructors only as far as I can see

So I would change it to just `DFunction.visibility()` or `DClass.constructorVisibility()`

IgnatBeresnev (commented):
Nicely done!

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -276,10 +277,19 @@ class DefaultPsiToDocumentableTranslator(
...
+                        it.typeConstructor,
+                        JavaClassKindTypes.INTERFACE
+                    )
+                }).toSourceSetDependent()
                 val modifiers = getModifier().toSourceSetDependent()
                 val implementedInterfacesExtra =
                     ImplementedInterfaces(ancestry.allImplementedInterfaces().toSourceSetDependent())
+
+                fun Array<PsiMethod>.defaultCtorIfEmpty() =
+                    if (isEmpty()) arrayOf(psi.createDefaultConstructor()) else this
```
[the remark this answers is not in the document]
atyrin: Not sure this function will be reused somewhere else. Annotations and interfaces don't have constructors. And Enums don't have a default one with the same logic as the classes.

I've extracted all constructor logic for PsiClass in the functions, but not sure is it ok to duplicate `when` inside the function.
https://github.com/Kotlin/dokka/pull/2795/files#diff-668d7a8ff4c6e4cbebe200868a974fefc008dd4d1a8158a6d9b6db6603b72ec7R404

plugins/base/src/main/kotlin/translators/psi/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -450,6 +485,18 @@ class DefaultPsiToDocumentableTranslator(
             )
         }
 
+        /**
+         * `isPhysical` detects the virtual declarations without real sources.
+         * Otherwise, [PsiDocumentableSource] initialization will fail: non-physical declarations doesn't have `virtualFile`.
+         * The check is required since in [parseConstructors] we create a virtual method for default constructor.
```
IgnatBeresnev, line 491: nitpick:

The last line is quite specific and it can rot with time if the implementation of `parseConstructors` changes and we forget to update this comment (we probably will forget it)

I'd change it to something more generic, like `This check protects from accidentally requesting sources for synthetic / virtual declarations.`

And it can be helpful to move the comment closer to `isPhysical` and make it an implementation comment (`// text`), since more generic logic for parsing sources can be added to this function in the future.

IgnatBeresnev (approved):
:+1: :+1:

PR #2797, opened 2023-01-02
Update Kotlin to 1.8.0

IgnatBeresnev:


examples/gradle/dokka-multiplatform-example/build.gradle.kts
```diff
@@ -19,7 +19,7 @@ kotlin {
     jvm() // Create a JVM target with the default name "jvm"
     linuxX64("linux")
     macosX64("macos")
-    js()
+    js(BOTH)
```
IgnatBeresnev, line 22: `js()` is considered to be legacy and throws an error. 

```
* What went wrong:

==========
This project currently uses the Kotlin/JS Legacy compiler backend, which has been deprecated and will be removed in a future release.
Please migrate the project to the new IR-based compiler (https://kotl.in/jsir) by adding:
- kotlin.js.compiler=ir to your gradle.properties file.
- js(IR) { ... } to your build file.
You can continue to use the deprecated legacy compiler in the current version of the toolchain by specifying it explicitly.
==========
```

However, it seems like you can still use the legacy backend, so I decided to go with `BOTH` to check for backwards compatibility in this and future releases. I don't think everyone will migrate at once, so it's worth giving some time. We can consider changing it to `IR` if/when it starts causing problems.

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
```diff
@@ -39,7 +39,7 @@ private fun Project.compileClasspathOf(compilation: KotlinCompilation): FileColl
     if (compilation.target.isAndroidTarget()) {
         // This is a workaround for https://youtrack.jetbrains.com/issue/KT-33893
         @Suppress("DEPRECATION") // for compatibility
-        return compilation.compileKotlinTask.cast<KotlinCompile>().classpath
+        return compilation.compileKotlinTask.cast<KotlinCompile>().libraries
```
IgnatBeresnev, line 42: `classpath` is deprecated with error, so no other choice.

`libraries` exists only in KGP >= 1.7.0, so Dokka 1.8 is not backwards compatible with Kotlin (KGP) 1.6.21 and earlier.

This needs to be discussed additionally and stressed in documentation or somewhere. I've asked the build tools team for more info until then

plugins/base/src/main/kotlin/translators/descriptors/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -310,6 +310,7 @@ private class DokkaDescriptorVisitor(
 
         // synthetic values() and valueOf() functions are not present among average class functions
         val enumSyntheticFunctions = staticScopeForKotlinEnum.getContributedDescriptors { true }
+            .filterIsInstance<FunctionDescriptor>()
```
IgnatBeresnev, line 313: This now returns all descriptors and not only functions

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/kotlin/StdlibGradleIntegrationTest.kt
```diff
@@ -46,6 +46,7 @@ class StdlibGradleIntegrationTest(override val versions: BuildVersions) : Abstra
     }
 
     @Test
+    @Ignore
```
IgnatBeresnev, line 49: Since we lost backwards compatibility with KGP <= 1.6.21, this test is broken as it's still using Kotlin 1.4.

If we don't do anything to preserve backwards compatibility, I would actually like to nuke this test and the submodule. Am I right in understanding that [kotlin-dokka-stdlib](https://github.com/JetBrains/kotlin-dokka-stdlib) repo is deprecated and will/should not be updated?

integration-tests/gradle/projects/coroutines/coroutines.diff
```diff
@@ -1,62 +1,60 @@
 diff --git a/build.gradle b/build.gradle
```
IgnatBeresnev, line 1: Had to update `kotlinx.coroutines` submodule to latest `develop` (as of now), as `master` is still using Kotlin 1.6.21 which is not supported in this branch.

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
```diff
@@ -2,30 +2,36 @@ package org.jetbrains.dokka.it.gradle
 
 internal object TestedVersions {
 
-    val LATEST = BuildVersions("7.4.2", "1.7.20")
-    val BASE =
+    val LATEST = BuildVersions("7.4.2", "1.8.0")
+
+    /**
+     * All supported Gradle/Kotlin versions, including [LATEST]
```
vmishenev, line 8: Can you put an URL to the compatibility matrix?

vmishenev (commented):
Ok, but we should keep compatibility with Kotlin (KGP) 1.6.21, i.e. since Kotlin 1.5.0.

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
IgnatBeresnev: Good call, I've added two links:

* Link to the [Kotlin/Gradle compatibility matrix](https://docs.gradle.org/current/userguide/compatibility.html#kotlin) 
* Link to [AGP/Gradle compatibility matrix](https://developer.android.com/studio/releases/gradle-plugin#updating-gradle)

IgnatBeresnev:
There will be merge conflicts after updating to 1.8.0, and it should be revisited after #2841 anyway, so I'm closing it for now, but will either re-open it or make a new one after 1.8.10 release.

IgnatBeresnev:
(happened by mistake)

IgnatBeresnev:
* Bumped Kotlin and IDE plugins to 1.8.10
* Preserved compatibility with KGP <= 1.7

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/KotlinGradlePluginVersion.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Accepts a full version string that contains the major, minor
+ * and patch versions divided by dots, such as "1.7.10".
+ *
+ * Does NOT parse and store custom suffixes, so `1.8.20-RC2`
+ * or `1.8.20-dev-42` will be viewed as `1.8.20`.
+ */
+internal fun parseKotlinVersion(fullVersionString: String): KotlinVersion? {
+    val versionParts = fullVersionString
+        .split(".", "-", limit = 4)
```
vmishenev, line 19: why is `4` rather than `3`?

vmishenev (approved):

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/KotlinGradlePluginVersion.kt
IgnatBeresnev: Limit sets the max number of substrings, so for 1.7.20-RC it'll be

[0] 1
[1] 7
[2] 20
[3] RC

If you use 3, then it'll be

[0] 1
[1] 7
[2] 20-RC

PR #2799, opened 2023-01-07
Fix configure version eagerly

JavierSegoviaCordoba:
A typical approach in some version plugins is using an object which uses internally a `Property<String>`, and the `toString()` method unwraps it.

With `provider { … }` unwrapping it too soon is avoided.


IgnatBeresnev:
For some additional context: the problem was reported by another user in the Kotlin Community Slack ([link to the message](https://kotlinlang.slack.com/archives/C0F4UNJET/p1672925338799409)), whose project version was all of a sudden changed to `0.4.1-dev.57.uncommitted+c39fefc` when running Dokka with GitHub Actions. However, the version was OK if run locally.

IgnatBeresnev (approved):
Thank you so much! 

I've started lengthy integration tests just in case, let's wait for them to complete and then it'll be ready to get merged.

(We should really remember to double check all of the other optoins, this isn't our first time running into this :sweat_smile:)

IgnatBeresnev:
Looks good, thanks once again!

JavierSegoviaCordoba:
Your welcome 🤗

PR #2806, opened 2023-01-10
Add missing group and description for Multimodule tasks

TWiStErRob:
image


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaPlugin.kt
```diff
@@ -74,6 +75,8 @@ open class DokkaPlugin : Plugin<Project> {
                 }
 
                 project.tasks.register<DefaultTask>("${name}Multimodule") {
+                    group = JavaBasePlugin.DOCUMENTATION_GROUP
```
IgnatBeresnev, line 78: I'm not sure if it's a good idea to put deprecated tasks into the documentation group, as it would somewhat encourage people to use it without reading the description. I'd leave them in `other` with a bunch of other forgotten and internal things. Wdyt?

It is definitely worth adding the description though :+1:

IgnatBeresnev (commented):
I noticed that the `ReplaceWith` for `DokkaMultimoduleTask` contains an error - it suggests to replace it with itself. Could you please correct it while we're at it? 

And changing the deprecation level to `ERROR` would be nice as well, it's been under `WARNING` for 2.5 years, I think it's enough time to migrate :)

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaPlugin.kt
TWiStErRob: `all dokka tasks are part of the documentation group` test was pretty explicit, it just wasn't validating **all** the Dokka tasks because of the multi-module condition :)

The deprecated task has deprecation (warn) logging to "encourage" non-use? And the huge all-uppercase DEPRECATED is right next to the name in `gradlew tasks` (although in IDEA's Gradle view I think it doesn't show, only maybe as tooltip). It should be pretty eye-catching even in the `documentation` group. Adding to the same group to the deprecated task, it and the encouraged version would be right next to each other.

"Other" is for tasks that don't have metadata set, usually unintentionally. If the above benefits are not enough, how about meeting halfway and setting `group = "deprecated"`? I'll of course adjust the tests too.

IgnatBeresnev: I totally agree with you, and it should've been done 2 years ago, but adding them now to only remove them in one major release could cause confusion. 

I wouldn't mind if you didn't have to hover over the task and wait a second to see the description - I don't know if people are patient enough 😄 Seeing two tasks side by side that only differ in one uppercase/lowercase letter to me would be confusing

If there's a deprecated group or It can be created - that's even better! I just don't want them the old and the new ones to be next to each other at this point

TWiStErRob: > I actually didn't know they existed at all, not sure how you found it

What are you referring to as "they"/"it" here?

>  Seeing two tasks side by side that only differ in one uppercase/lowercase letter to me would be confusing, and I'd waste some time

true that.

> wait a second to see the description

confirmed, tooltip shows (but I've never used it for real :D)

> If there's a deprecated group or It can be created

Yeah, you can do all kinds of crazy things with groups, for example use spaces 🤯 and uppercase letters:
https://github.com/GoogleCloudPlatform/app-gradle-plugin/blob/1b1f053f3194ecc8cc8225e3724b07f51f960731/src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java#L41

image

I'll adjust the PR to create the `deprecated` group which fits into the built-in Gradle naming.

IgnatBeresnev: > What are you referring to as "they"/"it" here?

I meant the deprecated tasks - somehow, I never ran into them

IgnatBeresnev (approved):
Sweet, thanks for taking the time and addressing review comments!

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaPlugin.kt
TWiStErRob: Just curiosity, I didn't understand why some Dokka tasks are under documentation and some under other.

PR #2807, opened 2023-01-10
Avoid snapshotting sourceLink.localDirectory input

ilya-g:
A typical use case in a project with complex source root setup is to set `sourceLink.localDirectory` to the root of the project checkout dir and remoteUrl to the root of the project VCS url respectively.
However, since this property was marked as `InputDirectory`, in that case Gradle snapshotted hashes of all files in this entire project directory before running the Dokka task (including for example .git and .gradle directories). This could be very slow in big projects and also result in various Gradle warnings.

Now `localDirectory` is not an input property, but its relativized path is stored in another read-only input string property. Therefore, it invalidates task outputs when its value changes, but not when any content inside it.

### DIFF
--- a/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
+++ b/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
@@ -24,6 +24,25 @@ class BasicCachingIntegrationTest(override val versions: BuildVersions) : Abstra
         runAndAssertOutcome(TaskOutcome.FROM_CACHE)
     }
 
+    @Test
+    fun localDirectoryPointingToRoot() {
+        fun String.findAndReplace(oldValue: String, newValue: String): String {
+            assertTrue(oldValue in this, "Expected to replace '$oldValue'")
+            return replace(oldValue, newValue)
+        }
+        val projectKts = projectDir.resolve("build.gradle.kts")
+        val projectKtsText = projectKts.readText()
+            .findAndReplace("localDirectory.set(file(\"src/main\"))", "localDirectory.set(projectDir)")
+            .findAndReplace("integration-tests/gradle/projects/it-basic/src/main", "integration-tests/gradle/projects/it-basic")
+        projectKts.writeText(projectKtsText)
+
+        runAndAssertOutcome(TaskOutcome.SUCCESS)
+        projectDir.resolve("unrelated.txt").writeText("modified")
+        // despite projectDir is used as an input in localDirectory, changing its contents shouldn't invalidate the cache
+        runAndAssertOutcome(TaskOutcome.FROM_CACHE)
+    }
+
+
     private fun runAndAssertOutcome(expectedOutcome: TaskOutcome) {
         val result = createGradleRunner(
             "clean",
--- a/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
+++ b/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
@@ -20,8 +20,8 @@ class BasicCachingIntegrationTest(override val versions: BuildVersions) : Abstra
 
     @Test
     fun execute() {
-        runAndAssertOutcome(TaskOutcome.SUCCESS)
-        runAndAssertOutcome(TaskOutcome.FROM_CACHE)
+        runAndAssertOutcomeAndContents(TaskOutcome.SUCCESS)
+        runAndAssertOutcomeAndContents(TaskOutcome.FROM_CACHE)
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
```diff
@@ -24,6 +24,25 @@ class BasicCachingIntegrationTest(override val versions: BuildVersions) : Abstra
...
+        val projectKts = projectDir.resolve("build.gradle.kts")
+        val projectKtsText = projectKts.readText()
+            .findAndReplace("localDirectory.set(file(\"src/main\"))", "localDirectory.set(projectDir)")
+            .findAndReplace("integration-tests/gradle/projects/it-basic/src/main", "integration-tests/gradle/projects/it-basic")
+        projectKts.writeText(projectKtsText)
+
+        runAndAssertOutcome(TaskOutcome.SUCCESS)
+        projectDir.resolve("unrelated.txt").writeText("modified")
+        // despite projectDir is used as an input in localDirectory, changing its contents shouldn't invalidate the cache
+        runAndAssertOutcome(TaskOutcome.FROM_CACHE)
```
IgnatBeresnev, line 42: Nice test!

What do you think about adding a similar test or an additional step to this test to verify that changing the value of `localDirectory` invalidates the task? 

I doubt it'll happen, but Gradle could optimize out unused `@Input` properties or change the default behaviour - I'd like it to be caught with a test

Something like that:

```kotlin
...
runAndAssertOutcome(TaskOutcome.FROM_CACHE)

projectKts.readText()
            .findAndReplace("localDirectory.set(projectDir)", "localDirectory.set(file(\"src\"))")

runAndAssertOutcome(TaskOutcome.SUCCESS)
```

IgnatBeresnev (commented):
Thanks for taking the initiative!

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
ilya-g: I thought about it, but changing `localDirectory` this way will also change source links in the resulting html, so the content assertions won't pass. And changing `remoteUrl` to compensate that will make this change useless because it would be unclear which of these two properties has invalidated the task

IgnatBeresnev: > the content assertions won't pass

Oh, I didn't know it checked for the actual values of the links, I thought it only asserted the task result... 

It doesn't look like we need to assert the output in this case though, this particular test can simply check that the cache was invalidated

WDYT about extracting the `run` part from `runAndAssertOutcome` into a separate function (without content checking), and using it for checking cache invalidation? Let it generate the wrong links as long as it **re**-generates them

IgnatBeresnev:
```kotlin
    private fun runAndAssertOutcome(expectedOutcome: TaskOutcome) {
        runTask(expectedOutcome)
        File(projectDir, "build/dokka/html").assertHtmlOutputDir()
    }

    private fun runTask(expectedOutcome: TaskOutcome) {
            "clean",
            "dokkaHtml",
            "-i",
            "-s",
            "-Dorg.gradle.caching.debug=true",
            "--build-cache"
        ).buildRelaxed()

        assertEquals(expectedOutcome, assertNotNull(result.task(":dokkaHtml")).outcome)
    }
```

and then

```kotlin
        val projectKtsWithUpdatedLocalDirectory = projectKts.readText()
            .findAndReplace("localDirectory.set(projectDir)", "localDirectory.set(file(\"src/main/kotlin\"))")
        projectKts.writeText(projectKtsWithUpdatedLocalDirectory)
        runTask(TaskOutcome.SUCCESS)
```

ilya-g: ok, I've followed this approach and split `runAndAssertOutcome` function into two.

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/GradleSourceLinkBuilder.kt
```diff
@@ -30,16 +31,24 @@ class GradleSourceLinkBuilder(
...
      */
-    @InputDirectory
-    @PathSensitive(PathSensitivity.RELATIVE)
+    @Internal
     val localDirectory: Property<File?> = project.objects.safeProperty()
 
+    @Suppress("unused")
+    @get:Input
+    internal val localDirectoryPath: Provider<String?> =
+        localDirectory.map { it.relativeToOrSelf(project.projectDir).invariantSeparatorsPath }
```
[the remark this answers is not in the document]
ilya-g:
```suggestion
    @Internal // changing contents of the directory should not invalidate the task
    val localDirectory: Property<File?> = project.objects.safeProperty()
    
    /**
     * The relative path to [localDirectory] from the project directory. Declared as an input to invalidate the task if that path changes.
     * Should not be used anywhere directly.
     */
    @Suppress("unused")
    @get:Input
    internal val localDirectoryPath: Provider<String?> =
        localDirectory.map { it.relativeToOrSelf(project.projectDir).invariantSeparatorsPath }
```

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/BasicCachingIntegrationTest.kt
IgnatBeresnev: Thanks!

IgnatBeresnev (approved):
Thanks!

ilya-g:
Note: I assigned this PR to me because I would have liked to hit merge myself in order to correct the message of the squashed commit.

IgnatBeresnev:
> Note: I assigned this PR to me because I would have liked to hit merge myself in order to correct the message of the squashed commit.

Oh, sorry, will bear this in mind for next time :+1: Never seen assigns used in that way and we don't use the field ourselves, so it didn't trigger anything in my brain, I just assumed you assigned it to yourself automatically.

I thought about leaving the more detailed comment in the commit description, but I felt like in-code comments captured it well, if that's what you wanted to add.

PR #2810, opened 2023-01-12
Push classlike tabs reorganization research results

IgnatBeresnev:
Bad code ahead, not to be merged, it's just a demonstration of research results in addition to written review

Diff is best viewed from IntelliJ, it will highlight what truly changed. Because of added blocks / extracted methods it might seem like a lot of code was added/deleted, but actually not

Must be closed with #2764

### DIFF
--- a/plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
+++ b/plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
@@ -174,10 +174,7 @@ function initTabs() {
 function showCorrespondingTabBody(element) {
     const buttonWithKey = element.querySelector("button[data-active]")
     if (buttonWithKey) {
-        const key = buttonWithKey.getAttribute("data-togglable")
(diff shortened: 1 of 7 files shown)


plugins/base/src/main/resources/dokka/scripts/platform-content-handler.js
```diff
@@ -174,10 +174,7 @@ function initTabs() {
 function showCorrespondingTabBody(element) {
```
IgnatBeresnev, line 174: This whole file is the same as in the original PR

plugins/base/src/main/resources/dokka/styles/style.css
```diff
@@ -942,6 +942,7 @@ td.content {
 }
 
 .platform-hinted > .content:not([data-active]),
+.tabs-section-body *[data-togglable]:not([data-active]),
```
IgnatBeresnev, line 945: From the original PR

PR #2813, opened 2023-01-13
Do not leak implementation details in generated Javadoc links

### LINKED ISSUE #2803: Strange internal reference IDs in HTML
**Describe the bug**
`<a id="..."` and `<a name="..."` and `<a href="..."` are containing weird values.
Here's an example (yes, this whole thing is ONE `id`, except it's invalid because it contains quotes and they're not escaped):
```
<a id="variantConfig(TypeAliased(typeAlias=GenericTypeConstructor(dri=net.twisterrob.gradle.common/BaseVariant///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={org.jetbrains.dokka.model.Annotations$Companion@3b321821=Annotations(myContent={DokkaSourceSetImpl(displayName=jvm, sourceSetID=:common:dokkaJavadoc/main, classpath=[P:\caches\gradle\caches\7.6\generated-gradle-jars\gradle-api-7.6.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-ant-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-astbuilder-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-console-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-datetime-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-dateutil-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-groovydoc-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-json-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-nio-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-sql-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-templates-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-test-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-xml-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\javaparser-core-3.17.0.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\gradle-installation-beacon-7.6.jar, P:\projects\workspace\net.twisterrob.gradle\compat\agp-base\build\libs\twister-compat-agp-base-0.15-SNAPSHOT.jar, P:\projects\workspace\net.twisterrob.gradle\compat\gradle\build\libs\twister-compat-gradle-0.15-SNAPSHOT.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk8\1.4.32\3302f9ec8a5c1ed220781dbd37770072549bd333\kotlin-stdlib-jdk8-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-reflect\1.4.32\ce852b166d97f0f1991b5130c2bb02e2ef6c554e\kotlin-reflect-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk7\1.4.32\3546900a3ebff0c43f31190baf87a9220e37b7ea\kotlin-stdlib-jdk7-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib\1.4.32\461367948840adbb0839c51d91ed74ef4a9ccb52\kotlin-stdlib-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.gradle\gradle-kotlin-dsl\6.1.1\4f811c43a0688fb76681c357033c734fa740d898\gradle-kotlin-dsl-6.1.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\gradle\7.3.1\e05fdef1532d40d90ef6d2f80f76dfc78d098bb9\gradle-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools\common\30.3.1\6d00d4eaf6e80712b579a71e50f574483e6aef31\common-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\guava\30.1-jre\d0c3ce2311c9e36e73228da25a6e99b2ab826f\guava-30.1-jre.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.code.findbugs\jsr305\3.0.2\25ea2e8b0c338a877313bd4672d3fe056ea78f0d\jsr305-3.0.2.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains\annotations\23.1.0\d095a317cafe983999c7c8ad14da1a503288b55e\annotations-23.1.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-common\1.4.32\ef50bfa2c0491a11dcc35d9822edbfd6170e1ea2\kotlin-stdlib-common-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\builder\7.3.1\3e23401ece1637cecbae6be5c02cc18ab7e452e5\builder-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\builder-model\7.3.1\fddc6058293b89467794f8fcd6bfb35e37739c48\builder-model-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\gradle-api\7.3.1\a2723750277a10ca522477752d98f996653ebf14\gradle-api-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools\annotations\30.3.1\bf7c0de4844ebdce4eec08b202d24d8f1f7ac520\annotations-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\manifest-merger\30.3.1\71f4364c86bbe1ff328c6e5eac880d07c8c92943\manifest-merger-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android\zipflinger\7.3.1\16999dcec6ec960f9dc53c76081936bd545b4dc9\zipflinger-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\apksig\7.3.1\46b1d756b1dd2f41714f24c84b611306d39ec065\apksig-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\apkzlib\7.3.1\5b1abe37db589a8c57e1023c0c709d41424d87dd\apkzlib-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.squareup\javawriter\2.5.0\81241ff7078ef14f42ea2a8995fa09c096256e6b\javawriter-2.5.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.ow2.asm\asm\9.1\a99500cf6eea30535eeac6be73899d048f8d12a8\asm-9.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\failureaccess\1.0.1\1dcf1de382a0bf95a3d8b0849546c88bac1292c9\failureaccess-1.0.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\listenablefuture\9999.0-empty-to-avoid-conflict-with-guava\b421526c5f297295adef1c886e5246c39d4ac629\listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.checkerframework\checker-qual\3.5.0\2f50520c8abea66fbd8d26e481d3aef5c673b510\checker-qual-3.5.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.errorprone\error_prone_annotations\2.3.4\dac170e4594de319655ffb62f41cbd6dbb5e601e\error_prone_annotations-2.3.4.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.j2objc\j2objc-annotations\1.3\ba035118bc8bac37d7eff77700720999acd9986d\j2objc-annotations-1.3.jar], sourceRoots=[P:\projects\workspace\net.twisterrob.gradle\common\src\main\java, P:\projects\workspace\net.twisterrob.gradle\common\src\main\kotlin], dependentSourceSets=[], samples=[], includes=[], includeNonPublic=false, reportUndocumented=false, skipEmptyPackages=true, skipDeprecated=false, jdkVersion=8, sourceLinks=[], perPackageOptions=[], externalDocumentationLinks=[ExternalDocumentationLinkImpl(url=https://docs.oracle.com/javase/8/docs/api/, packageListUrl=https://docs.oracle.com/javase/8/docs/api/package-list), ExternalDocumentationLinkImpl(url=https://kotlinlang.org/api/latest/jvm/stdlib/, packageListUrl=https://kotlinlang.org/api/latest/jvm/stdlib/package-list)], languageVersion=null, apiVersion=null, noStdlibLink=false, noJdkLink=false, suppressedFiles=[], analysisPlatform=jvm, documentedVisibilities=[PUBLIC])=[Annotation(dri=kotlin/Suppress///PointingToDeclaration/, params={names=ArrayValue(value=[TYPEALIAS_EXPANSION_DEPRECATION])}, mustBeDocumented=false, scope=DIRECT)]})})), inner=GenericTypeConstructor(dri=com.android.build.gradle.api/BaseVariant///PointingToDeclaration/, projections=[], presentableName=null, extra=PropertyContainer(map={org.jetbrains.dokka.model.Annotations$Companion@3b321821=Annotations(myContent={DokkaSourceSetImpl(displayName=jvm, sourceSetID=:common:dokkaJavadoc/main, classpath=[P:\caches\gradle\caches\7.6\generated-gradle-jars\gradle-api-7.6.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-ant-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-astbuilder-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-console-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-datetime-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-dateutil-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-groovydoc-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-json-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-nio-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-sql-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-templates-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-test-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-xml-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\javaparser-core-3.17.0.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\gradle-installation-beacon-7.6.jar, P:\projects\workspace\net.twisterrob.gradle\compat\agp-base\build\libs\twister-compat-agp-base-0.15-SNAPSHOT.jar, P:\projects\workspace\net.twisterrob.gradle\compat\gradle\build\libs\twister-compat-gradle-0.15-SNAPSHOT.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk8\1.4.32\3302f9ec8a5c1ed220781dbd37770072549bd333\kotlin-stdlib-jdk8-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-reflect\1.4.32\ce852b166d97f0f1991b5130c2bb02e2ef6c554e\kotlin-reflect-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk7\1.4.32\3546900a3ebff0c43f31190baf87a9220e37b7ea\kotlin-stdlib-jdk7-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib\1.4.32\461367948840adbb0839c51d91ed74ef4a9ccb52\kotlin-stdlib-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.gradle\gradle-kotlin-dsl\6.1.1\4f811c43a0688fb76681c357033c734fa740d898\gradle-kotlin-dsl-6.1.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\gradle\7.3.1\e05fdef1532d40d90ef6d2f80f76dfc78d098bb9\gradle-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools\common\30.3.1\6d00d4eaf6e80712b579a71e50f574483e6aef31\common-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\guava\30.1-jre\d0c3ce2311c9e36e73228da25a6e99b2ab826f\guava-30.1-jre.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.code.findbugs\jsr305\3.0.2\25ea2e8b0c338a877313bd4672d3fe056ea78f0d\jsr305-3.0.2.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains\annotations\23.1.0\d095a317cafe983999c7c8ad14da1a503288b55e\annotations-23.1.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-common\1.4.32\ef50bfa2c0491a11dcc35d9822edbfd6170e1ea2\kotlin-stdlib-common-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\builder\7.3.1\3e23401ece1637cecbae6be5c02cc18ab7e452e5\builder-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\builder-model\7.3.1\fddc6058293b89467794f8fcd6bfb35e37739c48\builder-model-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\gradle-api\7.3.1\a2723750277a10ca522477752d98f996653ebf14\gradle-api-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools\annotations\30.3.1\bf7c0de4844ebdce4eec08b202d24d8f1f7ac520\annotations-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\manifest-merger\30.3.1\71f4364c86bbe1ff328c6e5eac880d07c8c92943\manifest-merger-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android\zipflinger\7.3.1\16999dcec6ec960f9dc53c76081936bd545b4dc9\zipflinger-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\apksig\7.3.1\46b1d756b1dd2f41714f24c84b611306d39ec065\apksig-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\apkzlib\7.3.1\5b1abe37db589a8c57e1023c0c709d41424d87dd\apkzlib-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.squareup\javawriter\2.5.0\81241ff7078ef14f42ea2a8995fa09c096256e6b\javawriter-2.5.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.ow2.asm\asm\9.1\a99500cf6eea30535eeac6be73899d048f8d12a8\asm-9.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\failureaccess\1.0.1\1dcf1de382a0bf95a3d8b0849546c88bac1292c9\failureaccess-1.0.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\listenablefuture\9999.0-empty-to-avoid-conflict-with-guava\b421526c5f297295adef1c886e5246c39d4ac629\listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.checkerframework\checker-qual\3.5.0\2f50520c8abea66fbd8d26e481d3aef5c673b510\checker-qual-3.5.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.errorprone\error_prone_annotations\2.3.4\dac170e4594de319655ffb62f41cbd6dbb5e601e\error_prone_annotations-2.3.4.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.j2objc\j2objc-annotations\1.3\ba035118bc8bac37d7eff77700720999acd9986d\j2objc-annotations-1.3.jar], sourceRoots=[P:\projects\workspace\net.twisterrob.gradle\common\src\main\java, P:\projects\workspace\net.twisterrob.gradle\common\src\main\kotlin], dependentSourceSets=[], samples=[], includes=[], includeNonPublic=false, reportUndocumented=false, skipEmptyPackages=true, skipDeprecated=false, jdkVersion=8, sourceLinks=[], perPackageOptions=[], externalDocumentationLinks=[ExternalDocumentationLinkImpl(url=https://docs.oracle.com/javase/8/docs/api/, packageListUrl=https://docs.oracle.com/javase/8/docs/api/package-list), ExternalDocumentationLinkImpl(url=https://kotlinlang.org/api/latest/jvm/stdlib/, packageListUrl=https://kotlinlang.org/api/latest/jvm/stdlib/package-list)], languageVersion=null, apiVersion=null, noStdlibLink=false, noJdkLink=false, suppressedFiles=[], analysisPlatform=jvm, documentedVisibilities=[PUBLIC])=[Annotation(dri=kotlin/Suppress///PointingToDeclaration/, params={names=ArrayValue(value=[TYPEALIAS_EXPANSION_DEPRECATION])}, mustBeDocumented=false, scope=DIRECT)]})})), extra=PropertyContainer(map={org.jetbrains.dokka.model.Annotations$Companion@3b321821=Annotations(myContent={DokkaSourceSetImpl(displayName=jvm, sourceSetID=:common:dokkaJavadoc/main, classpath=[P:\caches\gradle\caches\7.6\generated-gradle-jars\gradle-api-7.6.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-ant-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-astbuilder-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-console-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-datetime-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-dateutil-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-groovydoc-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-json-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-nio-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-sql-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-templates-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-test-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\groovy-xml-3.0.13.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\javaparser-core-3.17.0.jar, P:\caches\gradle\wrapper\dists\gradle-7.6-all\9f832ih6bniajn45pbmqhk2cw\gradle-7.6\lib\gradle-installation-beacon-7.6.jar, P:\projects\workspace\net.twisterrob.gradle\compat\agp-base\build\libs\twister-compat-agp-base-0.15-SNAPSHOT.jar, P:\projects\workspace\net.twisterrob.gradle\compat\gradle\build\libs\twister-compat-gradle-0.15-SNAPSHOT.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk8\1.4.32\3302f9ec8a5c1ed220781dbd37770072549bd333\kotlin-stdlib-jdk8-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-reflect\1.4.32\ce852b166d97f0f1991b5130c2bb02e2ef6c554e\kotlin-reflect-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk7\1.4.32\3546900a3ebff0c43f31190baf87a9220e37b7ea\kotlin-stdlib-jdk7-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib\1.4.32\461367948840adbb0839c51d91ed74ef4a9ccb52\kotlin-stdlib-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.gradle\gradle-kotlin-dsl\6.1.1\4f811c43a0688fb76681c357033c734fa740d898\gradle-kotlin-dsl-6.1.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\gradle\7.3.1\e05fdef1532d40d90ef6d2f80f76dfc78d098bb9\gradle-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools\common\30.3.1\6d00d4eaf6e80712b579a71e50f574483e6aef31\common-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\guava\30.1-jre\d0c3ce2311c9e36e73228da25a6e99b2ab826f\guava-30.1-jre.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.code.findbugs\jsr305\3.0.2\25ea2e8b0c338a877313bd4672d3fe056ea78f0d\jsr305-3.0.2.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains\annotations\23.1.0\d095a317cafe983999c7c8ad14da1a503288b55e\annotations-23.1.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-common\1.4.32\ef50bfa2c0491a11dcc35d9822edbfd6170e1ea2\kotlin-stdlib-common-1.4.32.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\builder\7.3.1\3e23401ece1637cecbae6be5c02cc18ab7e452e5\builder-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\builder-model\7.3.1\fddc6058293b89467794f8fcd6bfb35e37739c48\builder-model-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\gradle-api\7.3.1\a2723750277a10ca522477752d98f996653ebf14\gradle-api-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools\annotations\30.3.1\bf7c0de4844ebdce4eec08b202d24d8f1f7ac520\annotations-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\manifest-merger\30.3.1\71f4364c86bbe1ff328c6e5eac880d07c8c92943\manifest-merger-30.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android\zipflinger\7.3.1\16999dcec6ec960f9dc53c76081936bd545b4dc9\zipflinger-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\apksig\7.3.1\46b1d756b1dd2f41714f24c84b611306d39ec065\apksig-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.android.tools.build\apkzlib\7.3.1\5b1abe37db589a8c57e1023c0c709d41424d87dd\apkzlib-7.3.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.squareup\javawriter\2.5.0\81241ff7078ef14f42ea2a8995fa09c096256e6b\javawriter-2.5.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.ow2.asm\asm\9.1\a99500cf6eea30535eeac6be73899d048f8d12a8\asm-9.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\failureaccess\1.0.1\1dcf1de382a0bf95a3d8b0849546c88bac1292c9\failureaccess-1.0.1.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.guava\listenablefuture\9999.0-empty-to-avoid-conflict-with-guava\b421526c5f297295adef1c886e5246c39d4ac629\listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar, P:\caches\gradle\caches\modules-2\files-2.1\org.checkerframework\checker-qual\3.5.0\2f50520c8abea66fbd8d26e481d3aef5c673b510\checker-qual-3.5.0.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.errorprone\error_prone_annotations\2.3.4\dac170e4594de319655ffb62f41cbd6dbb5e601e\error_prone_annotations-2.3.4.jar, P:\caches\gradle\caches\modules-2\files-2.1\com.google.j2objc\j2objc-annotations\1.3\ba035118bc8bac37d7eff77700720999acd9986d\j2objc-annotations-1.3.jar], sourceRoots=[P:\projects\workspace\net.twisterrob.gradle\common\src\main\java, P:\projects\workspace\net.twisterrob.gradle\common\src\main\kotlin], dependentSourceSets=[], samples=[], includes=[], includeNonPublic=false, reportUndocumented=false, skipEmptyPackages=true, skipDeprecated=false, jdkVersion=8, sourceLinks=[], perPackageOptions=[], externalDocumentationLinks=[ExternalDocumentationLinkImpl(url=https://docs.oracle.com/javase/8/docs/api/, packageListUrl=https://docs.oracle.com/javase/8/docs/api/package-list), ExternalDocumentationLinkImpl(url=https://kotlinlang.org/api/latest/jvm/stdlib/, packageListUrl=https://kotlinlang.org/api/latest/jvm/stdlib/package-list)], languageVersion=null, apiVersion=null, noStdlibLink=false, noJdkLink=false, suppressedFiles=[], analysisPlatform=jvm, documentedVisibilities=[PUBLIC])=[Annotation(dri=kotlin/Suppress///PointingToDeclaration/, params={names=ArrayValue(value=[TYPEALIAS_EXPANSION_DEPRECATION])}, mustBeDocumented=false, scope=DIRECT)]})})))">
```

**Expected behaviour**
Short and valid references within the docs. Not exposing local file paths of the build machine / CI that creates the docs (it might be a security risk based on the threat model).

**Screenshots**

| Raw | Inspect |
| - | - |
| image | image |

**To Reproduce**
1. Clone https://github.com/TWiStErRob/net.twisterrob.gradle/tree/16913c5a1d91c3f3ceed82decc25b845e70e093c
2. `gradlew dokkaJavadoc`
3. View raw text of `common\build\dokka\javadoc\net\twisterrob\gradle\common\VariantTaskCreator.html` or inspect `variantConfig` method's link or anchor.

**Dokka configuration**
[Configuration of dokka used to reproduce the bug](https://github.com/TWiStErRob/net.twisterrob.gradle/blob/16913c5a1d91c3f3ceed82decc25b845e70e093c/gradle/plugins/src/main/kotlin/net/twisterrob/gradle/build/PublishingPlugin.kt#L37-L43)

**Installation**
- Operating system: Windows 10
- Build tool: Gradle 7.6
- Dokka version: 1.7.20

**Additional context**
Looks like a deep `toString` of some DSL object, so it might be a trivial fix. Also note that the value is used raw without any HTML escaping, so it's generates invalid/unsafe HTML with easy injection point.

IgnatBeresnev:
Fixes #2803

### DIFF
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -1,10 +1,11 @@
 package org.jetbrains.dokka.javadoc.pages
 
 import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.utilities.htmlEscape
 
 internal fun JavadocFunctionNode.getAnchor(): String {
     val parameters = parameters.joinToString(",") { it.typeBound.asString() }
-    return "$name($parameters)"
+    return "$name($parameters)".htmlEscape()
 }
 
 private fun Bound.asString(): String = when (this) {
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -1,7 +1,6 @@
 package org.jetbrains.dokka.javadoc.pages
 
 import org.jetbrains.dokka.model.*
-import org.jetbrains.dokka.utilities.htmlEscape
 
 /**
  * Returns an unencoded, unescaped function anchor.
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -5,7 +5,7 @@ import org.jetbrains.dokka.utilities.htmlEscape
 
 internal fun JavadocFunctionNode.getAnchor(): String {
     val parameters = parameters.joinToString(",") { it.typeBound.asString() }
-    return "$name($parameters)".htmlEscape()
+    return "$name($parameters)".htmlEscape() // can be used in HTML as is, thus some escaping to avoid invalid HTML
 }
 
 private fun Bound.asString(): String = when (this) {
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -2,14 +2,20 @@ package org.jetbrains.dokka.javadoc.pages
 
 import org.jetbrains.dokka.model.*
 
-internal fun JavadocFunctionNode.getAnchor(): String =
-    "$name(${parameters.joinToString(",") {
-        when (val bound = if (it.typeBound is Nullable) it.typeBound.inner else it.typeBound) {
-            is TypeConstructor -> listOf(bound.dri.packageName, bound.dri.classNames).joinToString(".")
-            is TypeParameter -> bound.name
-            is PrimitiveJavaType -> bound.name
-            is UnresolvedBound -> bound.name
-            is JavaObject -> "Object"
-            else -> bound.toString()
-        }
-    }})"
+internal fun JavadocFunctionNode.getAnchor(): String {
+    val parameters = parameters.joinToString(",") { it.typeBound.asString() }
+    return "$name($parameters)"
+}
+
+private fun Bound.asString(): String = when (this) {
+    is Nullable -> this.inner.asString()
+    is DefinitelyNonNullable -> this.inner.asString()
+    is TypeConstructor -> listOf(this.dri.packageName, this.dri.classNames).joinToString(".")
+    is TypeParameter -> this.name
+    is PrimitiveJavaType -> this.name
+    is UnresolvedBound -> this.name
+    is TypeAliased -> this.typeAlias.asString()
+    is JavaObject -> "Object"
+    Dynamic -> "dynamic"
+    Void -> "void"
+}
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -3,9 +3,16 @@ package org.jetbrains.dokka.javadoc.pages
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.utilities.htmlEscape
 
+/**
+ * Returns an unencoded, unescaped function anchor.
+ *
+ * Should be URL encoded / HTML escaped at call site,
+ * depending on usage.
+ */
+// see the discussion in #2813 related to encoding/escaping this value for ids/hrefs
 internal fun JavadocFunctionNode.getAnchor(): String {
     val parameters = parameters.joinToString(",") { it.typeBound.asString() }
-    return "$name($parameters)".htmlEscape() // can be used in HTML as is, thus some escaping to avoid invalid HTML
+    return "$name($parameters)"
 }
 
 private fun Bound.asString(): String = when (this) {
--- a/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
+++ b/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
@@ -24,6 +24,16 @@ private fun Bound.asString(): String = when (this) {
     is UnresolvedBound -> this.name
     is TypeAliased -> this.typeAlias.asString()
     is JavaObject -> "Object"
-    Dynamic -> "dynamic"
+
+    // Void bound is currently used for return type only,
+    // which is not used in the anchor generation, but
+    // the handling for it is added regardless, just in case.
+    // Note: if you accept `Void` as a param, it'll be a TypeConstructor
     Void -> "void"
+
+    // Javadoc format currently does not support multiplatform projects,
+    // so in an ideal world we should not see Dynamic here, but someone
+    // might disable the checker or the support for it might be added
+    // by Dokka or another plugin, so the handling is added just in case.
+    Dynamic -> "dynamic"
 }
--- a/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/location/JavadocLocationTest.kt
+++ b/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/location/JavadocLocationTest.kt
@@ -47,6 +47,15 @@ class JavadocLocationTest : BaseAbstractTest() {
             |  * Referencing element from another package: [javadoc.test.Test]
             | */
             |class Referenced {}
+            |
+            |/jvmSrc/javadoc/test/FunctionParameters.kt
+            |package javadoc.test.functionparams
+            |
+            |typealias FunctionParamTypealias = String
+            |
+            |class FunctionParameters {
+            |    fun foo(param: FunctionParamTypealias) {}
+            |}
         """.trimIndent(),
             config,
             cleanupOutput = false,
@@ -148,6 +157,25 @@ class JavadocLocationTest : BaseAbstractTest() {
                     testClassNode
                 )
             )
+        }
+    }
+
+    @Test
+    fun `should resolve typealias function parameter`() {
+        locationTestInline { rootPageNode, dokkaContext ->
+            val transformer = htmlTranslator(rootPageNode, dokkaContext)
+            val containingClass = rootPageNode
+                .firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test.functionparams" }
+                .firstChildOfType<JavadocClasslikePageNode> { it.name == "FunctionParameters" }
+
+            val methodWithTypealiasParam = containingClass.methods.single { it.name == "foo" }
+            val methodSignatureHtml = transformer.htmlForContentNode(methodWithTypealiasParam.signature, null)
+
+            val expectedSignatureHtml = "final <a href=https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/index.html>Unit</a> " +
+                    "<a href=javadoc/test/functionparams/FunctionParameters.html#foo(javadoc.test.functionparams.FunctionParamTypealias)>foo</a>" +
+                    "(<a href=https://docs.oracle.com/javase/8/docs/api/java/lang/String.html>String</a> param)"
+
+            assertEquals(expectedSignatureHtml, methodSignatureHtml)
         }
     }
 
--- a/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/location/JavadocLocationTest.kt
+++ b/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/location/JavadocLocationTest.kt
@@ -16,57 +16,16 @@ import org.junit.jupiter.api.Assertions.assertEquals
 
 class JavadocLocationTest : BaseAbstractTest() {
 
-    private fun locationTestInline(testHandler: (RootPageNode, DokkaContext) -> Unit) {
-        val config = dokkaConfiguration {
-            format = "javadoc"
-            sourceSets {
-                sourceSet {
-                    sourceRoots = listOf("jvmSrc/")
-                    externalDocumentationLinks = listOf(
-                        DokkaConfiguration.ExternalDocumentationLink.jdk(8),
-                        DokkaConfiguration.ExternalDocumentationLink.kotlinStdlib()
-                    )
-                    analysisPlatform = "jvm"
-                }
-            }
-        }
-        testInline(
-            """
+    @Test
+    fun `resolved signature with external links`() {
+        val query = """
             |/jvmSrc/javadoc/test/Test.kt
             |package javadoc.test
             |import java.io.Serializable
-            |class Test<A>() : Serializable, Cloneable {
-            |   fun test() {}
-            |   fun test2(s: String) {}
-            |   fun <T> test3(a: A, t: T) {}
-            |}
-            |
-            |/jvmSrc/another/javadoc/example/Referenced.kt
-            |package javadoc.example.another
-            |/**
-            |  * Referencing element from another package: [javadoc.test.Test]
-            | */
-            |class Referenced {}
-            |
-            |/jvmSrc/javadoc/test/FunctionParameters.kt
-            |package javadoc.test.functionparams
-            |
-            |typealias FunctionParamTypealias = String
-            |
-            |class FunctionParameters {
-            |    fun foo(param: FunctionParamTypealias) {}
-            |}
-        """.trimIndent(),
-            config,
-            cleanupOutput = false,
-            pluginOverrides = listOf(JavadocPlugin())
-        ) { renderingStage = testHandler }
-    }
+            |class Test : Serializable, Cloneable {}
+        """.trimIndent()
 
-    @Test
-    fun `resolved signature with external links`() {
-
-        locationTestInline { rootPageNode, dokkaContext ->
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
             val testClass = rootPageNode.firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test" }
                 .firstChildOfType<JavadocClasslikePageNode>()
@@ -79,8 +38,15 @@ class JavadocLocationTest : BaseAbstractTest() {
 
     @Test
     fun `resolved signature to no argument function`() {
+        val query = """
+            |/jvmSrc/javadoc/test/Test.kt
+            |package javadoc.test
+            |class Test {
+            |   fun test() {}
+            |}
+        """.trimIndent()
 
-        locationTestInline { rootPageNode, dokkaContext ->
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
             val testClassNode = rootPageNode.firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test" }
                 .firstChildOfType<JavadocClasslikePageNode> { it.name == "Test" }
@@ -97,8 +63,15 @@ class JavadocLocationTest : BaseAbstractTest() {
 
     @Test
     fun `resolved signature to one argument function`() {
+        val query = """
+            |/jvmSrc/javadoc/test/Test.kt
+            |package javadoc.test
+            |class Test {
+            |   fun test2(s: String) {}
+            |}
+        """.trimIndent()
 
-        locationTestInline { rootPageNode, dokkaContext ->
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
             val testClassNode = rootPageNode.firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test" }
                 .firstChildOfType<JavadocClasslikePageNode> { it.name == "Test" }
@@ -115,8 +88,15 @@ class JavadocLocationTest : BaseAbstractTest() {
 
     @Test
     fun `resolved signature to generic function`() {
+        val query = """
+            |/jvmSrc/javadoc/test/Test.kt
+            |package javadoc.test
+            |class Test<A>() {
+            |   fun <T> test3(a: A, t: T) {}
+            |}
+        """.trimIndent()
 
-        locationTestInline { rootPageNode, dokkaContext ->
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
             val testClassNode = rootPageNode.firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test" }
                 .firstChildOfType<JavadocClasslikePageNode> { it.name == "Test" }
@@ -133,8 +113,13 @@ class JavadocLocationTest : BaseAbstractTest() {
 
     @Test
     fun `resolved package path`() {
+        val query = """
+            |/jvmSrc/javadoc/test/Test.kt
+            |package javadoc.test
+            |class Test {}
+        """.trimIndent()
 
-        locationTestInline { rootPageNode, dokkaContext ->
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val locationProvider = dokkaContext.plugin<JavadocPlugin>().querySingle { locationProviderFactory }
                 .getLocationProvider(rootPageNode)
             val packageNode = rootPageNode.firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test" }
@@ -146,7 +131,21 @@ class JavadocLocationTest : BaseAbstractTest() {
 
     @Test
     fun `resolve link from another package`(){
-        locationTestInline { rootPageNode, dokkaContext ->
+        val query = """
+            |/jvmSrc/javadoc/test/Test.kt
+            |package javadoc.test
+            |class Test {}
+            |
+            |/jvmSrc/another/javadoc/example/Referenced.kt
+            |package javadoc.example.another
+            |
+            |/**
+            |  * Referencing element from another package: [javadoc.test.Test]
+            | */
+            |class Referenced {}
+        """.trimIndent()
+
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
             val testClassNode = rootPageNode.firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.example.another" }
                 .firstChildOfType<JavadocClasslikePageNode> { it.name == "Referenced" }
@@ -162,7 +161,18 @@ class JavadocLocationTest : BaseAbstractTest() {
 
     @Test
     fun `should resolve typealias function parameter`() {
-        locationTestInline { rootPageNode, dokkaContext ->
+        val query = """
+            |/jvmSrc/javadoc/test/FunctionParameters.kt
+            |package javadoc.test.functionparams
+            |
+            |typealias FunctionParamTypealias = String
+            |
+            |class FunctionParameters {
+            |    fun foo(param: FunctionParamTypealias) {}
+            |}
+        """.trimIndent()
+
+        locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
             val containingClass = rootPageNode
                 .firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test.functionparams" }
@@ -177,6 +187,28 @@ class JavadocLocationTest : BaseAbstractTest() {
 
             assertEquals(expectedSignatureHtml, methodSignatureHtml)
         }
+    }
+
+    private fun locationTestInline(query: String, testHandler: (RootPageNode, DokkaContext) -> Unit) {
+        val config = dokkaConfiguration {
+            format = "javadoc"
+            sourceSets {
+                sourceSet {
+                    sourceRoots = listOf("jvmSrc/")
+                    externalDocumentationLinks = listOf(
+                        DokkaConfiguration.ExternalDocumentationLink.jdk(8),
+                        DokkaConfiguration.ExternalDocumentationLink.kotlinStdlib()
+                    )
+                    analysisPlatform = "jvm"
+                }
+            }
+        }
+        testInline(
+            query = query,
+            configuration = config,
+            cleanupOutput = false,
+            pluginOverrides = listOf(JavadocPlugin())
+        ) { renderingStage = testHandler }
     }
 
     private fun htmlTranslator(rootPageNode: RootPageNode, dokkaContext: DokkaContext): JavadocContentToHtmlTranslator {
--- a/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/location/JavadocLocationTest.kt
+++ b/plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/location/JavadocLocationTest.kt
@@ -11,6 +11,7 @@ import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.plugability.plugin
 import org.jetbrains.dokka.plugability.querySingle
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
+import org.jetbrains.dokka.javadoc.pages.JavadocFunctionNode
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.Assertions.assertEquals
 
@@ -165,28 +166,68 @@ class JavadocLocationTest : BaseAbstractTest() {
             |/jvmSrc/javadoc/test/FunctionParameters.kt
             |package javadoc.test.functionparams
             |
-            |typealias FunctionParamTypealias = String
+            |typealias StringTypealias = String
             |
             |class FunctionParameters {
-            |    fun foo(param: FunctionParamTypealias) {}
+            |    fun withTypealias(typeAliasParam: StringTypealias) {}
             |}
         """.trimIndent()
 
         locationTestInline(query) { rootPageNode, dokkaContext ->
             val transformer = htmlTranslator(rootPageNode, dokkaContext)
-            val containingClass = rootPageNode
-                .firstChildOfType<JavadocPackagePageNode> { it.name == "javadoc.test.functionparams" }
-                .firstChildOfType<JavadocClasslikePageNode> { it.name == "FunctionParameters" }
-
-            val methodWithTypealiasParam = containingClass.methods.single { it.name == "foo" }
+            val methodWithTypealiasParam = rootPageNode.findFunctionNodeWithin(
+                packageName = "javadoc.test.functionparams",
+                className = "FunctionParameters",
+                methodName = "withTypealias"
+            )
             val methodSignatureHtml = transformer.htmlForContentNode(methodWithTypealiasParam.signature, null)
 
             val expectedSignatureHtml = "final <a href=https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/index.html>Unit</a> " +
-                    "<a href=javadoc/test/functionparams/FunctionParameters.html#foo(javadoc.test.functionparams.FunctionParamTypealias)>foo</a>" +
-                    "(<a href=https://docs.oracle.com/javase/8/docs/api/java/lang/String.html>String</a> param)"
+                    "<a href=javadoc/test/functionparams/FunctionParameters.html#withTypealias(javadoc.test.functionparams.StringTypealias)>withTypealias</a>" +
+                    "(<a href=https://docs.oracle.com/javase/8/docs/api/java/lang/String.html>String</a> typeAliasParam)"
 
             assertEquals(expectedSignatureHtml, methodSignatureHtml)
         }
+    }
+
+    @Test
+    fun `should resolve definitely non nullable function parameter`() {
+        val query = """
+            |/jvmSrc/javadoc/test/FunctionParameters.kt
+            |package javadoc.test.functionparams
+            |
+            |class FunctionParameters {
+            |    fun <T> withDefinitelyNonNullableType(definitelyNonNullable: T & Any) {}
+            |}
+        """.trimIndent()
+
+        locationTestInline(query) { rootPageNode, dokkaContext ->
+            val transformer = htmlTranslator(rootPageNode, dokkaContext)
+            val methodWithVoidParam = rootPageNode.findFunctionNodeWithin(
+                packageName = "javadoc.test.functionparams",
+                className = "FunctionParameters",
+                methodName = "withDefinitelyNonNullableType"
+            )
+            val methodSignatureHtml = transformer.htmlForContentNode(methodWithVoidParam.signature, null)
+
+            val expectedSignatureHtml = "final &lt;T extends <a href=https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/index.html>Any</a>&gt; " +
+                    "<a href=https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/index.html>Unit</a> " +
+                    "<a href=javadoc/test/functionparams/FunctionParameters.html#withDefinitelyNonNullableType(T)>withDefinitelyNonNullableType</a>" +
+                    "(<a href=javadoc/test/functionparams/FunctionParameters.html#withDefinitelyNonNullableType(T)>T</a> definitelyNonNullable)"
+
+            assertEquals(expectedSignatureHtml, methodSignatureHtml)
+        }
+    }
+
+    private fun RootPageNode.findFunctionNodeWithin(
+        packageName: String,
+        className: String,
+        methodName: String
+    ): JavadocFunctionNode {
+        return this
+            .firstChildOfType<JavadocPackagePageNode> { it.name == packageName }
+            .firstChildOfType<JavadocClasslikePageNode> { it.name == className }
+            .methods.single { it.name == methodName }
     }
 
     private fun locationTestInline(query: String, testHandler: (RootPageNode, DokkaContext) -> Unit) {


plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
```diff
@@ -2,14 +2,20 @@ package org.jetbrains.dokka.javadoc.pages
...
+}
+
+private fun Bound.asString(): String = when (this) {
+    is Nullable -> this.inner.asString()
+    is DefinitelyNonNullable -> this.inner.asString()
+    is TypeConstructor -> listOf(this.dri.packageName, this.dri.classNames).joinToString(".")
+    is TypeParameter -> this.name
+    is PrimitiveJavaType -> this.name
+    is UnresolvedBound -> this.name
+    is TypeAliased -> this.typeAlias.asString()
```
IgnatBeresnev, line 17: `TypeAliased` was missing, so it defaulted to `toString()`, which leaked all configuration details.

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
```diff
@@ -1,15 +1,22 @@
...
-            is PrimitiveJavaType -> bound.name
-            is UnresolvedBound -> bound.name
-            is JavaObject -> "Object"
-            else -> bound.toString()
-        }
-    }})"
\ No newline at end of file
+internal fun JavadocFunctionNode.getAnchor(): String {
+    val parameters = parameters.joinToString(",") { it.typeBound.asString() }
+    return "$name($parameters)".htmlEscape()
```
vmishenev, line 8: Why is `htmlEscape` needed here? `UrlEncode`?

IgnatBeresnev: > Why is htmlEscape needed here? 

The returned value is used in the HTML template directly, so if `toString()` returned something like `trololo" something else`, it would result in `<a href="trololo" something else>`, allowing for easy html injections and in general invalid HTML. Added a comment with a brief explanation

Look at the screenshots in #2803 for more examples/context .

> UrlEncode?

I'm afraid `UrlEncode` cannot be used on the whole anchor because it has stricter rules and will not allow `(`, `)`, `<` and `>` - all of which are used by the Javadoc to generate links. See [AbstractList.html#%3Cinit%3E()](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/AbstractList.html#%3Cinit%3E()) for instance

I could use `UrlEncode` selectively instead of using it on the resulting string, but I don't think it's worth the effort as there's no constant threat, it's more or less "just in case" that doesn't cost us much

vmishenev: For your example:
`<init>`.htmlEscape() returns `&lt;<init>&gt;` (it brokes a link).
but `URLEncode` returns `%3Cinit%3E`. 
On [AbstractList.html#%3Cinit%3E()](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/AbstractList.html#%3Cinit%3E())  we have `<a href="#%3Cinit%3E()" class="member-name-link">AbstractList</a>`.

TWiStErRob: Agree here with @vmishenev.

Sorry for the long-winded text below, hope this helps in some way to move this along.

Note that the output code is more a part of URL than part of HTML. HTML escape is (by the name) is meant for outputting text as HTML so HTML language-special characters need to be escaped. That said URL escape is probably not good either because that's meant for **referencing** an ID. So if you have `id="<init>()"` you need `href="#%3Cinit%3E()`, but if you URL escape the ID: `id="%3Cinit%3E()`, you'll need `href="#%253Cinit%253E()"`.

So this pretty much means that you cannot just use the same string for id/name and href, as they need different escapes.

Note that HTML was very restrictive on IDs: https://www.w3.org/TR/html4/types.html#:~:text=ID%20and%20NAME%20tokens%20must,periods%20(%22.%22), while HTML 5 is permissive: https://stackoverflow.com/a/31773673/253468

That said, the overall HTML code has to be valid (parseable). As far as I know, to achieve this, you really only need to escape the `"` with `&quot;` (as to not accidentally terminate the attribute), and the rest can be verbatim.

Hrefs are another matter, because it doesn't only need to be valid HTML, but the parsed value must be a valid URL (RFC for it is quite complex) and it actually should point to a real location. So in this case one would need to URL encode the link (percent encoding), and then on top of that make sure that it can be safely put into an HTML attribute (use correct XML/HTML entities). Note that url-encoding too much is usually not a problem, the browser will just decode each percent-escape; however double encoding is a usual problem that the browser cannot resolve, so us devs have to be careful.

This complexity will stay regardless of how you render HTML, but different templating engines make it easier or harder to do this.

IgnatBeresnev: As I've mentioned, I don't think I can encode the whole anchor. The example with `<init>()` demonstrates that Javadoc applies encoding selectively. It encodes `<init>`, but does not encode `()`. Moreover, when signature links contain multiple arguments, it does not encode commas (for example, `addAll(int,java.util.Collection)`).

If I encode the whole anchor, it would be `%3Cinit%3E%28%29` and `addAll%28int%2Cjava.util.Collection%29`, which might work, but won't look as good and might lead to problems - I'm not sure.

As I said, I could apply encoding selectively (i.e encode every param separately), but you can see all possible param values just underneath - I don't see any that need encoding, and Dokka does not actually use `<init>()` and instead uses `ClassName()` for referencing constructors.

> HTML escape is (by the name) is meant for outputting text as HTML so HTML language-special characters need to be escaped. 

Unfortunately, that's what it's doing - whatever string `getAnchor()` generates, it'll be put into the HTML as is, without any prior escaping: `<a id="{{ method.anchorLink }}">`. 

That's the primary reason I put the escape there - to avoid possible invalid HTML, and to warn future generations and make them think about their changes to this function.

> So this pretty much means that you cannot just use the same string for id/name and href, as they need different escapes.

That's actually a really good point that I didn't even think about - this function is indeed used for generating both the id and the href value, which is not correct. Javadoc doesn't actually encode IDs: `<section .. id="<init>()">`

____

To sum up,

* URL encoding the anchor seems to break IDs and/or links
* HTML escaping the anchor doesn't seem to break anything yet, but it will if/when `<init>` gets introduced
* `getAnchor` needs to be refactored into two separate functions, one for generating the id and another one for the link, each with different escaping
* At the moment, unless names permit, nothing actually needs encoding/escaping

This issue/PR is about embarrassing configuration details that leak in the generated HTML, not about encoding/escaping, and I'd like to keep it short and just about that. This part has been fixed by removing the `else` branch for the sealed `Bound`, and by properly handling typealiases, which caused the problem.

The `htmlEscape` part is just extra that I added to bring attention to the problem and really just in case. I'm happy with removing it and just leaving a detailed comment with a link to this discussion, or we can leave it be and I'll still add the comment.

Thoughts?

vmishenev: > If I encode the whole anchor, it would be %3Cinit%3E%28%29 and addAll%28int%2Cjava.util.Collection%29, which might work, but won't look as good and might lead to problems - I'm not sure.

What problems? 

> Javadoc doesn't actually encode IDs: <section .. id="<init>()">

It encodes
`<section class="detail" id="&lt;init&gt;()">`

> getAnchor needs to be refactored into two separate functions, one for generating the id and another one for the link, each with different escaping

`getAnchor` should not depend on the output format or the place of use.
Anyway, it does not make sense to have `htmlEscape` here.

TWiStErRob: I vote for removing the htmlEscape from this PR and merging it. And we can open another issue where the escaping is worked out without hacking around. Just in case adding a function where it doesn't belong just creates tech debt and breaks things without knowing.

The model should never be aware of where the output goes. Just generate the preferred name and that's it. It is then the up to whoever builds the HTML to get it right:
```html
<a id="{{ HTMLEscape(method.anchorLink) }}">
<a href="{{ HTMLEscape(URLEncode(method.anchorLink)) }}">
```
if you use a half-decent templating library this should be done for you... `{{}}` looks like Mustache, but the word doesn't exist in the codebase, what is used to render HTML?

> won't look as good

I think it's better to be functional and "just works" than looking good :)
Did you consider the case that someone will copy those links and if it's too eagerly not-escaped it might not work in GitHub / Slack / GMail etc. I'm thinking about https://host.name/path/to/resource#whops a-space_in(the,url) vs https://host.name/path/to/resource#whops%20a-space_in(the,url) . The first one is borked, the second one is ugly, but at least it's clickable.

IgnatBeresnev: > It encodes <section class="detail" id="&lt;init&gt;()">

It html escapes the IDs, it doesn't URLEncode them, as far as I can see. Otherwise you would not have `()` in there, it would be `%28%29`. As @TWiStErRob has pointed out, URLEncoding the id would also require a different href.

> getAnchor should not depend on the output format or the place of use.

> The model should never be aware of where the output goes. Just generate the preferred name and that's it. It is then the up to whoever builds the HTML to get it right:
> 
> <a id="{{ HTMLEscape(method.anchorLink) }}">
> <a href="{{ HTMLEscape(URLEncode(method.anchorLink)) }}">

> I think it's better to be functional and "just works" than looking good :)

In an ideal world where we develop our own format - I would totally agree with you. Here, however, Dokka tries to clone the Javadoc format, which does not encode parentheses and commas - you seem to be ignoring this fact :) 

I guarantee you, if we don't follow it, someone will be comparing the links and will create a bug report, asking "Why, for the same input, does the Javadoc anchor link have plain symbols in `addAll(int,java.util.Collection)`, and Dokka's Javadoc anchor link contains gibberish in `addAll%28int%2Cjava.util.Collection%29`". 

I'm also not sure how the tools that accept Javadoc HTML as input will react (like IntelliJ for `javadoc.jar`), if they fully decode the URL or not - the kind of possible problems I was referring to.

If we want to honor the Javadoc format rules, right now I don't see any other way but to have two different implementations that apply `URLEncode`  selectively, ignoring parantheses and commas in the generated anchor itself.

Out of curiosity I had a look at how the actual Javadoc tool does it, and they just have their own implementation for `encodeURL` that is less restrictive: https://github.com/openjdk/jdk/blob/b88ee1ee22a4ea859f2a7bdf80a12c1d56fe6fd2/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/markup/HtmlTree.java#L299-L325

^ This would work and be neat, we won't need two different implementations then. But if we want to use the standard `URLEncoder` for the **whole** anchor, it'll lead to problems, imho

___

I've removed `htmlEscape` and left a comment. Hopefully, our discussion will help whoever fixes it in the future.

TWiStErRob: > Dokka tries to clone the Javadoc format, which does not encode parentheses and commas - you seem to be ignoring this fact :)

I wasn't aware this is a strict requirement. Also keep in mind that you cannot do that 100%, because Kotlin has free-form names, while Java is pretty strict with naming.

> they just have their own implementation

They are aware of what the possible characters are, so they wrote their own. You'll need to do the same, but beware of ``fun `this (is) "a" test!#%20$`() { }`` and similar valid Kotlin names. I'm pretty certain that if I document this name, and generate Dokka Javadoc, it just won't work.

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
TWiStErRob: You also *changed* the behavior of Dynamic and Void, those need tests to lock in the "correct" behavior?
```
object Void : Bound()
object Dynamic : Bound()
```
`bound.toString()` was `Void@hashcode`, but now it's `void`.

IgnatBeresnev: Currently, under the normal circumstances, it is not possible to get `Dynamic` or `Void` bounds in the Javadoc format. 

`Dynamic` is JS specific, and the Javadoc format does not support multiplatform projects (it has a source set checker), while the `Void` Bound is only used as the return type. If you accept `Void` as a function parameter, it'll be `TypeConstructor` instead (so `java.lang.Void`). For that reason, it's not possible to add proper tests

Yet it feels a bit scary to be throwing the exception in these cases, as the checker can be disabled without much trouble, and we or someone else might add support for multiplatform projects or another mapping for the `Void` bound, so they're still handled as they should. I've added comments to explain it though

I keep a list of similar "assumptions" that Dokka makes to play it safe, and one day I'll get to using asserts that only work in debug/dev builds of Dokka, so that we can catch these things. That is, for `Void` it would be something like `if (isInDevMode) error("Add proper support") else "void"`

Added a test for definitely non nullable types :+1:

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
```diff
@@ -1,15 +1,39 @@
 package org.jetbrains.dokka.javadoc.pages
 
 import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.utilities.htmlEscape
```
vmishenev, line 4: Unused import

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
```diff
@@ -1,15 +1,39 @@
...
-            is UnresolvedBound -> bound.name
-            is JavaObject -> "Object"
-            else -> bound.toString()
-        }
-    }})"
\ No newline at end of file
+/**
+ * Returns an unencoded, unescaped function anchor.
+ *
+ * Should be URL encoded / HTML escaped at call site,
```
vmishenev, line 9: But you do not encode it in the PR.

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
IgnatBeresnev: Removed it

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
IgnatBeresnev: Yes, what's the problem? This method does no encoding/escaping, so the result should be encoded/escaped in the places where it is called if there's need for it ("depending on usage" from the line below)

vmishenev: Sorry, this comment should be on the next line. You don't encode/escape a returned value at call site,
e.g. https://github.com/Kotlin/dokka/blob/53a57b0e4dd6abcb7b6d68be0c5983050906e0d8/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/renderer/JavadocContentToTemplateMapTranslator.kt#L125  and https://github.com/Kotlin/dokka/blob/53a57b0e4dd6abcb7b6d68be0c5983050906e0d8/plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/renderer/JavadocContentToTemplateMapTranslator.kt#L170
Have a look at templates before changing.

IgnatBeresnev: Yeah, for sure, I don't. We discussed why it's not as trivial in the [conversation above](https://github.com/Kotlin/dokka/pull/2813#discussion_r1072293998). 

TLDR: this PR is not about encoding/escaping, it's about fixing leaking details due to `.toString()`. The KDoc for the function has little to do with where/how it's actually used, I documented the behaviour of the function.

vmishenev: So can you open an issue for this?

vmishenev (approved):

plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/pages/utils.kt
IgnatBeresnev: #2854

PR #2820, opened 2023-01-19
Add og-image variable for documentation

IgnatBeresnev:
At this moment, nested documentation modules (like Dokka) do not inherit settings from the base repository, so Dokka's docs don't have a preview image when the links are shared on social media -- it's required for the announcement tweet. 

This should add the preview image until it's implemented/fixed.

### DIFF
--- a/docs/cfg/buildprofiles.xml
+++ b/docs/cfg/buildprofiles.xml
@@ -5,6 +5,11 @@
         <enable-browser-edits>true</enable-browser-edits>
         <browser-edits-url>https://github.com/Kotlin/dokka/edit/master/docs/</browser-edits-url>
         <allow-indexable-eaps>true</allow-indexable-eaps>
+
+        <!-- At this moment, nested documentation modules (like Dokka) do not inherit settings from the base -->
+        <!-- repository, so Dokka's docs don't have a preview image when the links are shared on social media. -->
+        <!-- This variable adds the preview image until it's  implemented/fixed. -->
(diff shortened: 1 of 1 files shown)


docs/cfg/buildprofiles.xml
```diff
@@ -5,6 +5,11 @@
         <enable-browser-edits>true</enable-browser-edits>
         <browser-edits-url>https://github.com/Kotlin/dokka/edit/master/docs/</browser-edits-url>
         <allow-indexable-eaps>true</allow-indexable-eaps>
+
+        <!-- At this moment, nested documentation modules (like Dokka) do not inherit settings from the base -->
+        <!-- repository, so Dokka's docs don't have a preview image when the links are shared on social media. -->
+        <!-- This variable adds the preview image until it's  implemented/fixed. -->
+        <og-image>https://kotlinlang.org/assets/images/open-graph/docs.png</og-image>
```
IgnatBeresnev, line 12: @nikpachoo will this one tag be enough?

Also, I tried to find a ticket related to this, but couldn't - does it exist? If it does, can you please link it so that I can put it into the comment?

IgnatBeresnev:
The PR shall remain open until we actually need to use the preview image. It might be fixed from the webhelp side earlier than that, in which case there would be no point in merging it and then rolling it back.

IgnatBeresnev:
The problem was fixed upstream, so no need to add anything on our side.

The tweets that reference Dokka's docs have a preview image now: https://twitter.com/kotlin/status/1618179117194657793

Issue #2621, opened 2022-08-11
Inherited docs can have unresolvable links


owengray-google:
**Describe the bug**
```
package a
class Foo
```
```
import a.Foo

package b
/** does [Foo] things, with a `Baz` twist */
class FooBaz : Foo() {
    /** does stuff like a `Baz`, not a [Foo] */
    fun likeABaz()
}
```
```
import b.FooBaz

package c
class Bar : FooBaz() {
    override fun likeABaz()
}
```

`Bar` and `Bar.likeABaz` inherit documentation from `FooBaz`. However, the `[Foo]` links in their docs cannot resolve, because `a` is not imported and `Foo` is not in scope.

This will result in broken doc links, and should result in command line warnings.
However, importing `Foo` in `Bar` will both force a direct dependency on package `a` (when e.g. the indirect dependency via `b` could have shrinking, jarjaring, or explicit dependency versioning) and cause IDEs/linters to complain about unused imports.

**Expected behaviour**
Overriding a function and inheriting docs does not cause those inherited docs to break, and all unresolved links can be warned on, and all such warnings can be made actionable.

**How to fix**
I really don't know.
One option would be doing the link resolution _before_ docs are inherited, but I suspect that might have its own issues.
Another option could be somehow associating the things in scope for a set of DocTags _with_ those DocTags, so as to be accessible once inherited.
Another option could be doing global searches through the classpath when trying to resolve links, but that could have issues with both performance and name duplication.
Another option could be trying to crawl up to supertypes when resolving links, but that would be complicated if it were even feasible and effective.

Issue #2369, opened 2022-02-17
Apply documentation from property to explicit getters and setters


CharlesG-Branch:
**Is your feature request related to a problem? Please describe**

When the javadocs are generated for this code:
```
class Foobar {
    /** This is my description */
    var foo: Int
        get() = TODO()
        set(value) = TODO()

    /** This is my description */
    val bar: Int
        get() = TODO()
}
```
Then the docstrings won't appear for **any** of the methods. You have to apply the docstring directly to the `get()` and `set()` method. E.g.
```
class Foobar {
    val bar: Int
        /** This is my description */
        get() = TODO()
}
```

**Describe the solution you'd like**
If there is no documentation for the getter / setter then it should use the documentation from the property for the getter / setter.

**Are you willing to provide a PR?**
Yes, if it's not too large

PR #2824, opened 2023-01-20
Remove old test data

IgnatBeresnev:
Looks like some pre-1.4 test data that is just collecting dust. I see that some of the tests have been ported, i.e https://github.com/Kotlin/dokka/commit/cacf1e0c6cda4e42fe6581946cad53a377c71ec7


vmishenev (commented):
Can we reuse some tests?

IgnatBeresnev:
> Can we reuse some tests?

Which tests? It's just some old test data which is not used anymore

vmishenev:
> Which tests? It's just some old test data which is not used anymore

I mean we can reuse these test cases.

IgnatBeresnev:
> I mean we can reuse these test cases.

Do you realistically want to go over 440 files that were created in 2014-2018 and that for the most part have the following contents

```
/**
 * This is link to [Foo]
 */
class Foo {

}
```

and spend time writing asserts for it? Why, what for? 

If you want to increase test coverage, let's run kover, see the uncovered code and write tests for it. If you want to look at the old test cases - you'll have them in the git history, they're not going anywhere.

vmishenev:
> Do you realistically want to go over 440 files that were created in 2014-2018 and that for the most part have the following contents

I see some special non-trivial cases.  
Simple test data is not interesting.

vmishenev:
I have discovered some test cases, that our tests do not check, at random.

- https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/format/htmlEscaping.kt 
It would be great to have tests for `@` as well 
- https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/typealias/deprecated.kt deprecation for a typealias - Do we check `deprecation` in  declaration lists?
 https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/typealias/chain.kt  We do not check links between typealias. I noticed we have some tests cases for typealias from  https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/typealias/
-  https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/format/throwsTag.kt We do not test `@throws` tags at all
- https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/format/brokenLink.kt Do we have this case? I have not found it in our tests.

 Broken test cases: 
 
- https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/format/accessor.kt
We do not render the documentation for a getter and setter, see #2369

- https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/format/inheritedLink.kt
See #2621
- https://github.com/Kotlin/dokka/blob/b6aa42c4a8b589f9004c70e98b3f75889297408e/core/testdata/format/multipleTypeParameterConstraints.md
Currently, Dokka does not display `where` clause and renders `fun <T : A, B> f()`. It needs to create an issue. 

What do you think about adding these test cases (except broken ones)?

IgnatBeresnev:
Thanks for taking a look, but I don't think it changes anything, sorry. Unless you want to add the tests yourself, of course - I don't mind. But I'm definitely not going to cover these cases in this PR or any time in the foreseeable future, I'd rather just close it.

I can create an issue with a vague description like "Add missing test cases", without specifying what exactly to test (IR? HTML? Documentables? All at once?) and what the expected output should be (how should the "broken" link behave?), but it's most likely going to be rotting forever and no one is going to delete it, because it's too difficult to check which tests were implemented and which weren't (and, frankly, not that many people want to write tests for fun).

If we leave these files be and close this PR, the outcome will most likely be the same: it will be rotting there unnoticed, until someone else deletes it.

If you or someone else has a few hours to spare and wants to write tests, I'm sure we can run kover and find some parts which are more important, yet still uncovered - the time will be better spent improving those areas as opposed to some corner cases that have not been reported until now, and might actually be covered transitively through other tests (for instance, I don't think there's any need to write tests for having the deprecated annotation on every single declaration type - one is enough, as long as there are tests for applying annotations themselves).

> We do not test @throws tags at all

I actually remember Andrei adding the tests for it in a recent PR https://github.com/Kotlin/dokka/pull/2749/files#diff-e6fee1a5bc7132ab77d3ed5784461b5cbec2e37a863f22262b1db291fdbca0c0

Which is a good demonstration that, as time goes by, we'll improve coverage, so these old test cases become even less useful by the day.

qwwdfsad:
Is it correct that PR is removing obsolete testdata that is not used anywhere, but has samples that are not covered by existing unit-tests?

If so, my vote is to accept and merge this PR: old testdata is unlikely to be taken care of, and cherry-picking specific testdata is much less productive than any systematic approach (kover, dedicated effort to write specific groups of tests etc.).
Anything else seems to be a waste of time

vmishenev (approved):

IgnatBeresnev:
> Is it correct that PR is removing obsolete testdata that is not used anywhere, but has samples that are not covered by existing unit-tests?

JFYI: Yeah, I think there used to be a test framework that worked through having separate `.kt` files with the test data, somewhat similar to how it works in IntelliJ. Since Dokka 1.4 was a complete re-write, the test framework got replaced by something else, but the old test data was never deleted.

PR #2826, opened 2023-01-26
Disable caching of Dokka's integration tests

IgnatBeresnev:
Because integration tests are cached by default, it produces no output for showcasing documentation through S3 or GitHub artifacts.

Tried to explain what needs to be done in the comment. I tried, but it takes too much time to implement and test it properly, and I'd like to restore the demo functionality sooner, so proposing to just disable it for now.

Regression from #2711

### DIFF
--- a/integration-tests/build.gradle.kts
+++ b/integration-tests/build.gradle.kts
@@ -23,9 +23,11 @@ subprojects {
 
     /**
      * Dokka's integration test task is not cacheable because the HTML outputs
-     * it produces when running the tests are used for showcasing resulting documentation.
+     * it produces when running the tests are used for showcasing resulting documentation,
+     * which does not work well with caching.
      *
      * At the moment there are two problems that do not allow to make it cacheable:
+     *
      * 1. The task's inputs are such that changes in Dokka's code do not invalidate the cache,
      *    because it is run with the same version of Dokka ("DOKKA_VERSION") on the same
      *    test project inputs.
@@ -37,7 +39,7 @@ subprojects {
      *
      * @see [org.jetbrains.dokka.it.TestOutputCopier] for more details on showcasing documentation
      */
-    @DisableCachingByDefault
+    @DisableCachingByDefault(because = "Contains incorrect inputs/outputs configuration, see the KDoc for details")
     abstract class NonCacheableIntegrationTest : Test()
 
     val integrationTest by tasks.register<NonCacheableIntegrationTest>("integrationTest") {
--- a/integration-tests/build.gradle.kts
+++ b/integration-tests/build.gradle.kts
@@ -21,7 +21,26 @@ subprojects {
         implementation(project(":integration-tests"))
     }
 
-    val integrationTest by tasks.register<Test>("integrationTest") {
+    /**
+     * Dokka's integration test task is not cacheable because the HTML outputs
+     * it produces when running the tests are used for showcasing resulting documentation.
+     *
+     * At the moment there are two problems that do not allow to make it cacheable:
+     * 1. The task's inputs are such that changes in Dokka's code do not invalidate the cache,
+     *    because it is run with the same version of Dokka ("DOKKA_VERSION") on the same
+     *    test project inputs.
+     * 2. The tests generate HTML output which is then used to showcase documentation.
+     *    The outputs are usually copied to a location from which it will be served.
+     *    However, if the test is cacheable, it produces no outputs, so no documentation
+     *    to showcase. It needs to be broken into two separate tasks: one cacheable for running
+     *    the tests and producing HTML output, and another non-cacheable for copying the output.
+     *
+     * @see [org.jetbrains.dokka.it.TestOutputCopier] for more details on showcasing documentation
+     */
+    @DisableCachingByDefault
+    abstract class NonCacheableIntegrationTest : Test()
+
+    val integrationTest by tasks.register<NonCacheableIntegrationTest>("integrationTest") {
         maxHeapSize = "2G"
         description = "Runs integration tests."
         group = "verification"


integration-tests/build.gradle.kts
```diff
@@ -21,7 +21,26 @@ subprojects {
...
+     *
+     * @see [org.jetbrains.dokka.it.TestOutputCopier] for more details on showcasing documentation
+     */
+    @DisableCachingByDefault
+    abstract class NonCacheableIntegrationTest : Test()
+
+    val integrationTest by tasks.register<NonCacheableIntegrationTest>("integrationTest") {
         maxHeapSize = "2G"
         description = "Runs integration tests."
         group = "verification"
```
aSemy, line 46:
```suggestion
    val integrationTest by tasks.register<Test>("integrationTest") {
        maxHeapSize = "2G"
        description = "Runs integration tests."
        group = "verification"

        System.getenv("DOKKA_TEST_OUTPUT_PATH")?.let { testOutputDir ->
            outputs.dir(testOutputDir)
        }
```

Hi 👋  If you want the `integrationTest` task to run whenever the HTML directory is not up-to-date, then I think you just need to register the directory as a task output. I presume `DOKKA_TEST_OUTPUT_PATH` holds the the output directory?

IgnatBeresnev: Hey! Not at all. 

`DOKKA_TEST_OUTPUT_PATH` is not the output path, but it's the path to which the output needs to be copied. The actual test output path is a random temp directory. The problem is that `DOKKA_TEST_OUTPUT_PATH` is almost static (see where it's set in GH workflows), but the temp directory is not. The temp directory is created in the AbstractIntegrationTest somewhere, not in the Gradle task.

To fix it properly, the following result should be achieved:

1. If `DOKKA_TEST_OUTPUT_PATH` is set, the test results need to be copied to that path. To do that, we could cache the test outputs, but the problem here is that on the task level we don't know what the output directory is.
2. If any Dokka code is changed, the task needs to be invalidated. Now it works through publishing Dokka to `mavenLocal`, and the version string is always the same, so it's viewed as the same input and the task is not invalidated properly - it knows nothing of changes in other modules. I'm also not sure if it'll invalidate if any of the test project code changes, as it's also hardcoded in the test itself, I don't think it's set up to be a task input.

All in all, I have a feeling that the current setup is not following best practices, that's why it's causing these problems. Maybe it'll be easier to just refactor it altogether.

You can have a go if you want, but I'd like to merge this in the meantime to make sure the tests are run

aSemy: > Hey! Not at all.
> 
> `DOKKA_TEST_OUTPUT_PATH` is not the output path, but it's the path to which the output needs to be copied. The actual test output path is a random temp directory. The problem is that `DOKKA_TEST_OUTPUT_PATH` is almost static (see where it's set in GH workflows), but the temp directory is not. The temp directory is created in the AbstractIntegrationTest somewhere, not in the Gradle task.

Okay clear, that makes sense. So the order of events is

1. the GitHub action sets `DOKKA_TEST_OUTPUT_PATH`, and triggers the integration test
2. Gradle starts the integration test task...
  1. The Kotlin code runs the tests using `ProcessBuilder()`
  2. HTML is produced to a random temp dir
  3. Finally, HTML is copied from the temp dir to `DOKKA_TEST_OUTPUT_PATH` by `TestOutputCopier`
5. The Gradle task concludes

The problems are on step 2 

- because the integration test projects don't directly depend on the other projects, but instead uses Maven Local, Gradle thinks the source code hasn't changed
- Gradle doesn't realise that the integration tests should output, and so Gradle doesn't realise the outputs aren't up-to-date

So ideally

1. Dokka should use the '[sharing outputs between projects](https://docs.gradle.org/current/userguide/cross_project_publications.html)' config to share the project files, rather than publishing to MavenLocal, so Gradle knows if the source code has changed
2. Somehow make the HTML output stable (there are lots of ways to do this - split up the integration tests, set stable directories via System Properties so Gradle is aware of them, use a single directory...)


The one other related improvement I can think of is to convert the AWS sync shell script `aws_sync.sh` into a Gradle task, with properly registered task inputs/outputs.



> You can have a go if you want, but I'd like to merge this in the meantime to make sure the tests are run

Yes that make sense - skipping the build cache for the integration tests is a good fix.

IgnatBeresnev: Yeah, I think you got it right :+1: I'm not sure what the best solution/implementation for these tests should be, it needs to be researched more, but I do have a strong feeling it's not done right at the moment :sweat_smile:

integration-tests/build.gradle.kts
```diff
@@ -21,7 +21,26 @@ subprojects {
...
+     *    However, if the test is cacheable, it produces no outputs, so no documentation
+     *    to showcase. It needs to be broken into two separate tasks: one cacheable for running
+     *    the tests and producing HTML output, and another non-cacheable for copying the output.
+     *
+     * @see [org.jetbrains.dokka.it.TestOutputCopier] for more details on showcasing documentation
+     */
+    @DisableCachingByDefault
+    abstract class NonCacheableIntegrationTest : Test()
+
+    val integrationTest by tasks.register<NonCacheableIntegrationTest>("integrationTest") {
```
TWiStErRob, line 43: If this implementation stays, I don't think it's necessary to create a separate task for this:
```suggestion
    val integrationTest by tasks.register<Test>("integrationTest") {
        outputs.doNotCacheIf { true }
```
see https://docs.gradle.org/current/javadoc/org/gradle/work/DisableCachingByDefault.html

IgnatBeresnev: I tried it, but it didn't work - the tests still used caches and completed in an instant.

I think it's related to the following behaviour:

> (from the Javadoc for [doNotCacheIf](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskOutputs.html#doNotCacheIf-java.lang.String-org.gradle.api.specs.Spec-))
> The spec will be evaluated at task execution time, not during configuration.

It might work if all of the current caches are reset on CI and locally, but even then I'm not sure if it's not going to cache the task itself (not the output) after running it once, because the inputs for the task are the same between runs.

aSemy: I'm surprised that `doNotCacheIf { true }` doesn't work, but the annotation does... oh well, Gradle is weird.


I suspect that registering the environment variable as an input property would help

```kotlin
inputs.property("dokkaTestOutputPath", proiders.environmentVariable("DOKKA_TEST_OUTPUT_PATH"))
```

integration-tests/build.gradle.kts
aSemy, line 46:
```suggestion
        group = "verification"
        inputs.property("dokkaTestOutputPath", proiders.environmentVariable("DOKKA_TEST_OUTPUT_PATH"))
```

aSemy: This should help Gradle to determine if the test task is up-to-date and requires running again. It's not a complete solution though.

integration-tests/build.gradle.kts
TWiStErRob: ah, that's weird, ok

integration-tests/build.gradle.kts
TWiStErRob: I'm a bit confused on this, from @IgnatBeresnev's big comment in the code it sounds like this is an output dir, and an input param. Would adding
```
outputs.dir(providers.environmentVariable("DOKKA_TEST_OUTPUT_PATH"))
```
make the caching work? (i.e. even if it's from cache, it would put the right files in the right place)

TWiStErRob: Note that "up-to-date"-ness is defined as "for a given input hash the output hash matches". Remember `outputs.upToDateWhen { true }`.

Depending on how much I misunderstand the context either the above, or
```
inputs.dir(providers.environmentVariable("DOKKA_TEST_OUTPUT_PATH"))
```
might be also necessary, so that the contents are also considered, not just where it is.

TWiStErRob: Anyway, sorry for the drive-by, I realize I don't have enough context on Dokka workflow to help here.

aSemy: I explained a little more here: https://github.com/Kotlin/dokka/pull/2826#discussion_r1089959966

Setting the value of `DOKKA_TEST_OUTPUT_PATH` might help, but the task would still need to figure out when the Dokka source code changes. Because the Dokka source code is published to Maven Local, Gradle can't figure out when the source code has changed.

I suspect a more complete solution would involve

* setting up a 'Maven Publication Test' Gradle convention plugin, that will 
   1. publish the Dokka project to a project-local directory instead of Maven Local
   2. use Gradle Configurations to share the result of publications
* and using the JVM Test Suite plugin, or using `includeBuild()`, to better define and manage the integration tests

Which is complicated, and needs some prep-work first #2703, so yeah, for now it's best just to tell Gradle not to cache the integration tests.

TWiStErRob: Might be a silly thing, but using the `-sources.jar` as input?

aSemy: Possibly...? I'm not sure, I've not shared sources.jars before. I think that might also require some custom Gradle Configurations? And then you might as well just have Configurations with the Maven Publication, which would be more representative.

IgnatBeresnev: > This should help Gradle to determine if the test task is up-to-date and requires running again. It's not a complete solution though.

I'm not sure this would help - doesn't the value of the property have to change for the task to be invalidated/cached? There's only 3 possible values for `DOKKA_TEST_OUTPUT_PATH`, you can see them [here](https://github.com/Kotlin/dokka/blob/master/.github/workflows/gh-actions-artifacts-snapshots.yml#L28). I would expect that Gradle would have 3 different cache entries for all 3 values, it doesn't just store the last run, right? 

And `DOKKA_TEST_OUTPUT_PATH` is GitHub action specific, so it wouldn't help with the problem for local runs and for TeamCity runs if it starts using Gradle caches.

Anyhow, I think it's best to disable the test for now as it's not that critical, and spend time re-writing this part completely instead of trying to patch it up. Just need to get all of the other Gradle-related things in order first :sweat_smile:

integration-tests/build.gradle.kts
```diff
@@ -21,7 +21,26 @@ subprojects {
...
+     *    test project inputs.
+     * 2. The tests generate HTML output which is then used to showcase documentation.
+     *    The outputs are usually copied to a location from which it will be served.
+     *    However, if the test is cacheable, it produces no outputs, so no documentation
+     *    to showcase. It needs to be broken into two separate tasks: one cacheable for running
+     *    the tests and producing HTML output, and another non-cacheable for copying the output.
+     *
+     * @see [org.jetbrains.dokka.it.TestOutputCopier] for more details on showcasing documentation
+     */
+    @DisableCachingByDefault
```
vmishenev, line 40: Can you use `because` property?

vmishenev (commented):
Does the `--no-build-cache` option work for this?

integration-tests/build.gradle.kts
IgnatBeresnev: Sure, added a short description

IgnatBeresnev:
> Does the --no-build-cache option work for this?

Most likely yes, although I haven't checked.

vmishenev (approved):
I would try to use ` --no-build-cache` for our actions but you have implemented with `@DisableCachingByDefault`.

TWiStErRob:
The argument applies to the whole build, the annotation only to one task.

IgnatBeresnev:
Not only that, but integration tests are run on TeamCity (I don't think it uses caches, but still) and are also run locally, so we'd have to be using `--no-build-cache` all the time and probably add it to some of the instructions.

PR #2827, opened 2023-01-26
Make implementations of DokkaLogger thread-safe

IgnatBeresnev:
It was reported that `TestLogger` sometimes throws `ArrayIndexOutOfBoundsException`. While at it, made sure other implementations are OK as well.

### DIFF
--- a/core/src/main/kotlin/DokkaBootstrapImpl.kt
+++ b/core/src/main/kotlin/DokkaBootstrapImpl.kt
@@ -1,7 +1,7 @@
 package org.jetbrains.dokka
 
 import org.jetbrains.dokka.utilities.DokkaLogger
-import java.util.concurrent.atomic.LongAdder
+import java.util.concurrent.atomic.AtomicInteger
 
 import java.util.function.BiConsumer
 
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/DokkaBootstrapImpl.kt
```diff
@@ -11,8 +12,16 @@ import java.util.function.BiConsumer
...
+        private val warningsCounter = AtomicInteger()
+        private val errorsCounter = AtomicInteger()
+
+        override var warningsCount: Int
+            get() = warningsCounter.get()
+            set(value) = warningsCounter.set(value)
+
+        override var errorsCount: Int
+            get() = errorsCounter.get()
+            set(value) = errorsCounter.set(value)
```
IgnatBeresnev, line 24: I thought about extracting this common logic into a separate AtomicCounterDokkaLogger class that encapsulates the counter increments, but I'm not sure it's worth it (new API + possible compatibility problems).

Can do if you like it more, though

vmishenev: Up to you. I have no preferences.

vmishenev (approved):

PR #2829, opened 2023-01-26
Tweaks for stdlib docs build

ilya-g:


vmishenev (approved):

plugins/base/src/main/kotlin/renderers/html/shouldRenderSourceSetBubbles.kt
```diff
@@ -5,7 +5,7 @@ import org.jetbrains.dokka.pages.ContentPage
 import org.jetbrains.dokka.pages.RootPageNode
 
 internal fun shouldRenderSourceSetBubbles(page: RootPageNode): Boolean {
-    return page.withDescendants()
+    return shouldRenderSingleSourceSetBubble() || page.withDescendants()
```
vmishenev, line 8: It needs to check that we have a single source set (count() == 1). 
Some pages (e.g. a multimodule page) do not have source sets.

plugins/base/src/main/kotlin/renderers/html/shouldRenderSourceSetBubbles.kt
```diff
@@ -14,3 +14,7 @@ internal fun shouldRenderSourceSetBubbles(page: RootPageNode): Boolean {
         .distinct()
         .count() > 1
 }
+
+private const val SHOULD_RENDER_SINGLE_SOURCE_SET_BUBBLE = "dokka.shouldRenderSingleSourceSetBubble"
+private fun shouldRenderSingleSourceSetBubble() =
+    System.getProperty(SHOULD_RENDER_SINGLE_SOURCE_SET_BUBBLE).toBoolean()
```
IgnatBeresnev, line 20: Discussed this in DMs and addressed in #2848

IgnatBeresnev (approved):

PR #2833, opened 2023-01-29
Gradle Plugin: update custom Gradle utils to use Gradle Kotlin DSL

aSemy:
Part of #2700 

This PR removes some of the custom Gradle utility functions that are now provided by the Gradle Kotlin DSL. 

* `getSafe()` ➡️ `get()`, or `orNull`
* `safeProperty()` ➡️ `org.gradle.kotlin.dsl.property()`
* `safeConvention()` ➡️ `convention()`

The benefit is closer alignment with the Gradle API, and more idiomatic code that helps with working with the Gradle plugin, as the code can be understood without needing to know about the custom utilities specific to Dokka.

This PR is similar to #2706.

Depends on 

* #2702
* #2705 
* #2706

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/GradleExternalDocumentationLinkBuilder.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/GradleExternalDocumentationLinkBuilder.kt
@@ -5,6 +5,7 @@ import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
+import org.gradle.kotlin.dsl.property
 import org.jetbrains.dokka.DokkaConfigurationBuilder
 import org.jetbrains.dokka.ExternalDocumentationLink
 import org.jetbrains.dokka.ExternalDocumentationLinkImpl
@@ -52,7 +53,7 @@ class GradleExternalDocumentationLinkBuilder(
      * ```
      */
     @Input
-    val url: Property<URL?> = project.objects.safeProperty()
+    val url: Property<URL> = project.objects.property()
 
     /**
      * Specifies the exact location of a `package-list` instead of relying on Dokka
@@ -66,10 +67,10 @@ class GradleExternalDocumentationLinkBuilder(
      */
     @Optional
     @Input
-    val packageListUrl: Property<URL?> = project.objects.safeProperty()
+    val packageListUrl: Property<URL> = project.objects.property()
 
     override fun build(): ExternalDocumentationLinkImpl = ExternalDocumentationLink(
-        url = checkNotNull(url.getSafe()) { "url not specified " },
-        packageListUrl = packageListUrl.getSafe()
+        url = checkNotNull(url.get()) { "url not specified " },
+        packageListUrl = packageListUrl.orNull,
     )
 }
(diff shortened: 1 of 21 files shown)


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/GradleExternalDocumentationLinkBuilder.kt
```diff
@@ -66,10 +67,10 @@ class GradleExternalDocumentationLinkBuilder(
      */
     @Optional
     @Input
-    val packageListUrl: Property<URL?> = project.objects.safeProperty()
+    val packageListUrl: Property<URL> = project.objects.property()
```
IgnatBeresnev, line 70: I wonder what the difference is between `Property<URL>` (now) and `Property<URL?>` (before)

It seems that the property with `Property<URL>` is still optional, nothing fails if I set `null`. How do you then determine if something is nullable or not? 

Trying to understand the logic for calling `.get()` vs `.orNull` and whether it's possible that this change can introduce regression where a previously optional (nullable) property becomes required because `get()`  throws an error on `null`. 

I'm trying to break something locally, but it doesn't want to fail no matter what I do, and I don't understand why :sweat_smile:

IgnatBeresnev (approved):
TIL, thanks :)  I've triggered the lengthy integration tests just in case, will merge the PR once they pass

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/GradleExternalDocumentationLinkBuilder.kt
aSemy: I think that since `Property<T>` is defined in Java, and it's not annotated as non-null, `T` is a platform type so it's always considered nullable.

My guess is that it's supposed to be used in tandem with the custom helper functions (like `safeConvention()`), which are supposed to 'extend' the existing Gradle API to try and make it impossible to set `null` into a `Property<String>`, but I'm not sure.

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/DokkaCollectorTask.kt
```diff
@@ -15,15 +15,15 @@ abstract class DokkaCollectorTask : AbstractDokkaParentTask() {
...
     override fun buildDokkaConfiguration(): DokkaConfigurationImpl {
         val initialDokkaConfiguration = DokkaConfigurationImpl(
-            moduleName = moduleName.getSafe(),
-            outputDir = outputDirectory.getSafe(),
-            cacheRoot = cacheRoot.getSafe(),
-            failOnWarning = failOnWarning.getSafe(),
-            offlineMode = offlineMode.getSafe(),
+            moduleName = moduleName.get(),
+            outputDir = outputDirectory.get(),
+            cacheRoot = cacheRoot.get(),
```
IgnatBeresnev, line 20: Looks like this needs to be `.orNull`, there's a failing test `Collector0IntegrationTest`:

```
Caused by: org.gradle.api.internal.provider.MissingValueException: Cannot query the value of task ':moduleA:dokkaHtmlCollector' property 'cacheRoot' because it has no value available.
  at org.gradle.api.internal.provider.AbstractMinimalProvider.get(AbstractMinimalProvider.java:86)
  at org.jetbrains.dokka.gradle.DokkaCollectorTask.buildDokkaConfiguration$gradle_plugin(DokkaCollectorTask.kt:20)
```

The rest looks good to me :+1:

aSemy:
```suggestion
            cacheRoot = cacheRoot.orNull,
```

Good catch!

aSemy: It would be nice to have a local test for this - I'll see if I can create one.

Goooler (commented):
🚀 LGTM

PR #2834, opened 2023-01-29
Gradle Plugin: Remove 'by' property setter helper function

aSemy:
(a new version of https://github.com/Kotlin/dokka/pull/2706)

Part of #2700 

I found these extension functions very confusing. I'm not sure what their purpose is. Do they only rename an existing function? 

It's confusing because 'by' is already a Kotlin keyword for delegation, but these extension functions don't do anything with delegation. I think using the normal `.set()` function makes the code more readable and understandable because `.set()` is a common Gradle function.


IgnatBeresnev (approved):
Looks good, thanks!

I've started TeamCity-based integration tests, and will merge the PR once they pass

PR #2835, opened 2023-01-30
Remove usage of internal gradle classes in task dependencies 

### LINKED ISSUE #2822: Remove the dependency on internal Gradle API
Dokka depends on classes from `org.gradle.api.internal` for which there are no guarantees about backward compatibility and proper deprecation as demonstrated by #2816

It looks like `TaskDependencyInternalWithAdditions` can be replaced with Gradle's `dependsOn` for setting up task dependencies.

atyrin:
Intermediate fix #2822

Switch in using storage of child tasks from an internal set of paths to Gradle set `dependsOn`.

Changes:
* add check for a type inside `addChildTask(path: String)`
  * it will produce a warning in the import process if we can resolve the task by the path and it is not an `AbstractDokkaTask`. That part is changed compared with 1.7.20 (previously reported only on executing of `:dokkaHtmlMultiModule`). 
  * it will not report anything during the import if we cannot resolve the task by the path. Otherwise, we will report the errors during the import if some submodules don't have applied the Dokka plugin.
* fix formatting for an error message if there is no submodule with applied Dokka.


atyrin:
Also, I noticed that `addSubprojectChildTasks`, `removeSubprojectChildTasks`, `removeChildTasks(projects: Iterable<Project>, childTasksName: String)`, `addChildTasks(projects: Iterable<Project>, childTasksName: String)` don't work in runtime (in project build files) but works during configuration (during pairing subproject files and parent)

```
tasks.dokkaHtmlMultiModule.configure {
    addChildTask(":kn-moodule:dokkasub") // works
    addChildTasks(setOf(project(":kn-moodule")), ":dokkasub") // doesn't work
}
```

image

atyrin:
We have one more reference for internal package: `import org.gradle.api.internal.plugins.DslObject` inside `/dokka/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaLeafTask.kt`

IgnatBeresnev (commented):
Overall LGTM :+1:

IgnatBeresnev (approved):

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaParentTask.kt
```diff
@@ -2,81 +2,104 @@ package org.jetbrains.dokka.gradle
...
 import org.gradle.api.Task
-import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Nested
 import org.gradle.work.DisableCachingByDefault
 
+private const val DEPRECATION_MESSAGE = """
+    It is an anti-pattern to declare cross-project dependencies as it leads to various build problems. 
+    For this reason, this API wil be removed with the introduction of project isolation. 
+    As an alternative, don't apply the Dokka plugin in subprojects that you don't want to document, 
+    or disable individual tasks using the Gradle API.
```
IgnatBeresnev, line 12: After some internal feedback, it looks like it can be helpful to mention that it's safe to use this API until project isolation comes along, and that we will provide a migration guide when it happens.

```suggestion
    It is an anti-pattern to declare cross-project dependencies as it leads to various build problems. 
    For this reason, this API wil be removed with the introduction of project isolation. 
    When it happens, we will provide a migration guide. In the meantime, you can keep using this API
    if you have to, but please don't rely on it if possible. If you don't want to document a certain project,
    don't apply the Dokka plugin for it, or disable individual project tasks using the Gradle API .
```

vmishenev (approved):
Can you rebase onto master?

IgnatBeresnev:
Well, at least now we know! @atyrin thanks for the research

Proposing to do the following:

1. Since we're working on project isolation and we're aiming to include it in the stable release, I don't think spending any more time trying to make the `dependsOn` change work makes sense. We tried it as an easy drop-in solution, but it turned out to be more difficult than I imagined. Hopefully, the internal Gradle API will not change dramatically until the stable release. So proposing to revert the `dependsOn` part.
2. We're going to remove public API methods such as `addChildTasks` and `removeChildTasks` anyway, so I'm proposing to keep the deprecation
3. This research was very valuable, maybe it was the whole reason why it was implemented through the internal API and not via `dependsOn`, so it makes sense to leave a short comment with a link to this PR

atyrin:
I've created a separate PR for further updates regarding deprecations only: #2857

PR #2837, opened 2023-01-31
Fill `packageName` and `classNames` for constructors from parent

atyrin:
In addition to #2795

The generated default constructor has an empty package name. Meanwhile, the explicit one has the proper package (as the parent).
In the change, we directly replace `parentName` and `classNames` properties in the DRI of the constructor with values from the parent.


IgnatBeresnev (approved):
:+1: 

Hopefully it won't backfire in some other place :sweat_smile:

PR #2840, opened 2023-02-01
Remove an unused function

IgnatBeresnev:
The creation of source set buttons was moved to the `source_set_selector.ftl` template, but never removed from the code

Found it while trying to find the logic for creating the source set filter buttons. It was very confusing to see and it made me think that it was invoked through reflection or that IDEA wasn't indexing things right.

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -143,29 +143,6 @@ open class HtmlRenderer(
         }
     }
 
-    private fun FlowContent.filterButtons(page: PageNode) {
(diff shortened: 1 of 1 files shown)


vmishenev (commented):
Is there only one unused function in Dokka?

vmishenev:
> I don't know, but I described how this particular one is causing confusion, so it's best to remove it.

Can you run an inspection to discover it?

IgnatBeresnev:
> Can you run an inspection to discover it?

I can, but it'll take additional time to investigate each case, as not every "unused" function can be removed without thinking - it can still be invoked from other places indirectly (reflection/inheritance/etc) or it might be highlighted due to an indexing issue. I don't have that time now.

I had to find how source set bubbles are created for another task, so I'm sure that this one is not used.

vmishenev (approved):

PR #2841, opened 2023-02-01
Remove outdated stdlib integration test

IgnatBeresnev:
~(for after 1.8.10)~

The stdlib integration test is quite problematic:

* It's using an outdated version of Kotlin 1.4, which we won't be able to support forever.
* It's using an outdated version of Gradle 5.6, which is already causing compatibility problems (see #2702)
* The [project used for the integration](https://github.com/JetBrains/kotlin-dokka-stdlib) (added as a git submodule) is not supported anymore (no updates since 2021), and contains non-trivial build configuration, which makes it very costly to update.
* The test has not helped catch any bugs that weren't detected by coroutines/serialization tests, whereas other KMP libraries such as ktor have caught non-trivial and specific problems.

The test has mostly been used as a live demo / preview of changes in Dokka (see [Github Action configuration for S3](https://github.com/Kotlin/dokka/blob/master/.github/workflows/s3-snapshots.yml#L10)), which is quite useful, but not needed for every commit.

Stdlib is currently moving towards using modern versions of Dokka, and there already are TeamCity builds that can generate documentation for a modern version of stdlib with any version of Dokka, with the output saved to artifacts. So it will be possible to check compatibility between uo-to-date Dokka and Stdlib, and it will be possible to preview Dokka changes whenever needed and before releases.


IgnatBeresnev:
Rebased onto master and marked ready for review. This can be merged now.

The release has a separate branch 1.8.10, so it won't affect it.

Issue #2845, opened 2023-02-05
Extensions/implementations from another modules in multi-module documentation


ilya-g:
When combining documentation of multiple modules, Dokka resolves links between them, so for example the types of a core module referenced in the signatures of an extension module get proper links to their respective documentation pages.

However, there's no information in the core module documentation that dependent modules could have some extensions for its API.
I'd expect to see at least the following references to the dependent modules:
- extension functions/properties for the types in the core module. For example, `asObservable` extensions for `Flow` from [rx2](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx2/kotlinx.coroutines.rx2/as-observable.html) and [rx3](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx3/kotlinx.coroutines.rx3/as-observable.html) modules are not shown in the [Flow index page](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/). Beware that since they come from different modules/packages, they shouldn't be merged in the same overload group in the index page.
- inheritors of types declared in the core modules. For example, the class [HandlerDispatcher](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-android/kotlinx.coroutines.android/-handler-dispatcher/) should be shown in the list of inheritors of its base class [MainCoroutineDispatcher](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-main-coroutine-dispatcher/)

IgnatBeresnev:
Similar request about inheritors: #1960

PR #2846, opened 2023-02-06
Sort divergent elements deterministically

### LINKED ISSUE #2784: Non deterministic sorting of extension functions
Merged extension function that have the same name and the same signature are sorted nondeterministically, leading to non-reproducible builds.

This makes it difficult to compare outputs of different versions/runs of Dokka since it adds noise to the comparison.

# Screenshots

image

image

IgnatBeresnev:
Fixes #2784

### DIFF
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -9,6 +9,7 @@ import org.jetbrains.dokka.base.transformers.documentables.ClashingDriIdentifier
 import org.jetbrains.dokka.base.transformers.pages.comments.CommentsToContentConverter
 import org.jetbrains.dokka.base.transformers.pages.tags.CustomTagContentProvider
 import org.jetbrains.dokka.base.translators.documentables.PageContentBuilder.DocumentableContentBuilder
+import org.jetbrains.dokka.links.Callable
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.doc.*
@@ -526,28 +527,29 @@ open class DefaultPageCreator(
                     .mapValues { if (it.value.any { it is DClasslike }) it.value.filter { it !is DTypeAlias } else it.value }
                     .toSortedMap(compareBy(nullsLast(String.CASE_INSENSITIVE_ORDER)) { it })
                     .forEach { (elementName, elements) -> // This groupBy should probably use LocationProvider
+                        val sortedElements = sortDivergentElementsDeterministically(elements)
                         row(
-                            dri = elements.map { it.dri }.toSet(),
-                            sourceSets = elements.flatMap { it.sourceSets }.toSet(),
+                            dri = sortedElements.map { it.dri }.toSet(),
+                            sourceSets = sortedElements.flatMap { it.sourceSets }.toSet(),
                             kind = kind,
                             styles = emptySet(),
                             extra = elementName?.let { name -> extra + SymbolAnchorHint(name, kind) } ?: extra
                         ) {
                             link(
                                 text = elementName.orEmpty(),
-                                address = elements.first().dri,
+                                address = sortedElements.first().dri,
                                 kind = kind,
                                 styles = setOf(ContentStyle.RowTitle),
-                                sourceSets = elements.sourceSets.toSet(),
+                                sourceSets = sortedElements.sourceSets.toSet(),
                                 extra = extra
                             )
                             divergentGroup(
                                 ContentDivergentGroup.GroupID(name),
-                                elements.map { it.dri }.toSet(),
+                                sortedElements.map { it.dri }.toSet(),
                                 kind = kind,
                                 extra = extra
                             ) {
-                                elements.map {
+                                sortedElements.map {
                                     instance(
                                         setOf(it.dri),
                                         it.sourceSets.toSet(),
@@ -570,6 +572,23 @@ open class DefaultPageCreator(
             }
         }
     }
+
+    /**
+     * Divergent elements, such as extensions for the same receiver, can have identical signatures
+     * if they are declared in different places. If such elements are shown on the same page together,
+     * they need to be rendered deterministically to have reproducible builds.
+     *
+     * For example, you can have three identical extensions, if they are declared as:
+     * 1) top-level in package A
+     * 2) top-level in package B
+     * 3) inside a companion object in package A/B
+     *
+     * @see divergentBlock
+     */
+    private fun sortDivergentElementsDeterministically(elements: List<Documentable>): List<Documentable> =
+        elements.takeIf { it.size > 1 }
+            ?.sortedWith(divergentDocumentableComparator)
+            ?: elements
 
     private fun DocumentableContentBuilder.contentForCustomTagsBrief(documentable: Documentable) {
         val customTags = documentable.customTags
@@ -610,6 +629,19 @@ internal val Documentable.customTags: Map<String, SourceSetDependent<CustomTagWr
 
 private val Documentable.hasSeparatePage: Boolean
     get() = this !is DTypeAlias
+
+/**
+ * @see DefaultPageCreator.sortDivergentElementsDeterministically for usage
+ */
+private val divergentDocumentableComparator =
+    compareBy<Documentable, String?>(nullsLast()) { it.dri.packageName }
+        .thenBy(nullsFirst()) { it.dri.classNames } // nullsFirst for top level to be first
+        .thenBy(
+            nullsLast(
+                compareBy<Callable> { it.params.size }
+                    .thenBy { it.signature() }
+            )
+        ) { it.dri.callable }
 
 @Suppress("UNCHECKED_CAST")
 private fun <T : Documentable> T.nameAfterClash(): String =
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -525,7 +525,7 @@ open class DefaultPageCreator(
                     .groupBy { it.name } // This groupBy should probably use LocationProvider
                     // This hacks displaying actual typealias signatures along classlike ones
                     .mapValues { if (it.value.any { it is DClasslike }) it.value.filter { it !is DTypeAlias } else it.value }
-                    .toSortedMap(compareBy(nullsLast(String.CASE_INSENSITIVE_ORDER)) { it })
+                    .entries.sortedBy { it.key }
                     .forEach { (elementName, elements) -> // This groupBy should probably use LocationProvider
                         val sortedElements = sortDivergentElementsDeterministically(elements)
                         row(
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -584,9 +584,12 @@ open class DefaultPageCreator(
      * 3) inside a companion object in package A/B
      *
      * @see divergentBlock
+     *
+     * @param elements can contain types (annotation/class/interface/object/typealias), functions and properties
+     * @return the original list if it has one or zero elements
      */
     private fun sortDivergentElementsDeterministically(elements: List<Documentable>): List<Documentable> =
-        elements.takeIf { it.size > 1 }
+        elements.takeIf { it.size > 1 } // the majority are single-element lists, but no real benchmarks done
             ?.sortedWith(divergentDocumentableComparator)
             ?: elements
 
--- a/plugins/base/src/test/kotlin/pageMerger/PageNodeMergerTest.kt
+++ b/plugins/base/src/test/kotlin/pageMerger/PageNodeMergerTest.kt
@@ -1,52 +1,28 @@
 package pageMerger
 
-import org.jetbrains.dokka.pages.ClasslikePageNode
-import org.jetbrains.dokka.pages.ContentPage
-import org.jetbrains.dokka.pages.PageNode
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
+import org.jetbrains.dokka.model.dfs
+import org.jetbrains.dokka.model.withDescendants
+import org.jetbrains.dokka.pages.*
+import org.junit.jupiter.api.RepeatedTest
+import java.lang.IllegalArgumentException
+import kotlin.test.assertEquals
 
 class PageNodeMergerTest : BaseAbstractTest() {
 
-    /* object SameNameStrategy : DokkaPlugin() {
-        val strategy by extending { CoreExtensions.pageMergerStrategy with SameMethodNamePageMergerStrategy }
-    }
-
-    class DefaultStrategy(val strList: MutableList<String> = mutableListOf()) : DokkaPlugin(), DokkaLogger {
-        val strategy by extending { CoreExtensions.pageMergerStrategy with DefaultPageMergerStrategy(this@DefaultStrategy) }
-
-        override var warningsCount: Int = 0
-        override var errorsCount: Int = 0
-
-        override fun debug(message: String) = TODO()
-
-        override fun info(message: String) = TODO()
-
-        override fun progress(message: String) = TODO()
-
-        override fun warn(message: String) {
-            strList += message
+    private val defaultConfiguration = dokkaConfiguration {
+        sourceSets {
+            sourceSet {
+                sourceRoots = listOf("src/main/kotlin")
+            }
         }
-
-        override fun error(message: String) = TODO()
-
-        override fun report() = TODO()
     }
-     */
 
     @Test
     fun sameNameStrategyTest() {
-
-        val configuration = dokkaConfiguration {
-            sourceSets {
-                sourceSet {
-                    sourceRoots = listOf("src/main/kotlin/pageMerger/Test.kt")
-                }
-            }
-        }
-
         testInline(
             """
             |/src/main/kotlin/pageMerger/Test.kt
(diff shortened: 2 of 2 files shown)


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -611,6 +630,19 @@ internal val Documentable.customTags: Map<String, SourceSetDependent<CustomTagWr
...
+ */
+private val divergentDocumentableComparator =
+    compareBy<Documentable, String?>(nullsLast()) { it.dri.packageName }
+        .thenBy(nullsFirst()) { it.dri.classNames } // nullsFirst for top level to be first
+        .thenBy(
+            nullsLast(
+                compareBy<Callable> { it.params.size }
+                    .thenBy { it.signature() }
+            )
+        ) { it.dri.callable }
```
IgnatBeresnev, line 644: There's definitely room for improvement in terms of sorting rules, but it's a start. To me it makes sense to display top level declarations to be on top as they have a wider scope, and to sort vararg-like overloads ([like these](https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/UpdateQuery.html#addValues(org.jooq.Row2,org.jooq.Row2))) in a strairs-like manner - both are implemented.

Tell me if you can think of anything else right off the bat

plugins/base/src/test/kotlin/pageMerger/PageNodeMergerTest.kt
```diff
@@ -187,4 +153,220 @@ class PageNodeMergerTest : BaseAbstractTest() {
             }
         }
     }
+
+    @RepeatedTest(3)
+    fun `should deterministically render same name property extensions`() {
```
IgnatBeresnev, line 158: I covered all the cases I could think of, but if you can think of a corner case that isn't covered - I'd be happy to add tests for it

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -611,6 +630,19 @@ internal val Documentable.customTags: Map<String, SourceSetDependent<CustomTagWr
 private val Documentable.hasSeparatePage: Boolean
     get() = this !is DTypeAlias
 
+/**
+ * @see DefaultPageCreator.sortDivergentElementsDeterministically for usage
+ */
+private val divergentDocumentableComparator =
+    compareBy<Documentable, String?>(nullsLast()) { it.dri.packageName }
```
ilya-g, line 637: It can be confusing that extensions from different packages are placed in the same group. Extensions from different packages have different pages, each own one, and the group title is a link that leads to the page of the first one.

It doesn't block this PR, but a consideration for the further improvement.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -526,28 +527,29 @@ open class DefaultPageCreator(
                     .mapValues { if (it.value.any { it is DClasslike }) it.value.filter { it !is DTypeAlias } else it.value }
                     .toSortedMap(compareBy(nullsLast(String.CASE_INSENSITIVE_ORDER)) { it })
```
ilya-g, line 528: Another problem with sorting map entries this way: a sorted map can have only one entry per key, and key uniqueness is determined by the comparator. Here the comparator is ignoring case so it will consider two group names with differing case equal and thus one group will overwrite the other in the sorted map.
Consider using `entries.sortedBy`

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -611,6 +630,19 @@ internal val Documentable.customTags: Map<String, SourceSetDependent<CustomTagWr
 private val Documentable.hasSeparatePage: Boolean
     get() = this !is DTypeAlias
 
+/**
+ * @see DefaultPageCreator.sortDivergentElementsDeterministically for usage
+ */
+private val divergentDocumentableComparator =
+    compareBy<Documentable, String?>(nullsLast()) { it.dri.packageName }
+        .thenBy(nullsFirst()) { it.dri.classNames } // nullsFirst for top level to be first
```
vmishenev, line 638: I would like to have the function `topLevelExtensionFirst` instead of the comment.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -571,6 +573,23 @@ open class DefaultPageCreator(
...
+     *
+     * For example, you can have three identical extensions, if they are declared as:
+     * 1) top-level in package A
+     * 2) top-level in package B
+     * 3) inside a companion object in package A/B
+     *
+     * @see divergentBlock
+     */
+    private fun sortDivergentElementsDeterministically(elements: List<Documentable>): List<Documentable> =
+        elements.takeIf { it.size > 1 }
```
vmishenev, line 589: This is redundant. The same check is inside `sortedWith`.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -571,6 +573,23 @@ open class DefaultPageCreator(
...
+     * they need to be rendered deterministically to have reproducible builds.
+     *
+     * For example, you can have three identical extensions, if they are declared as:
+     * 1) top-level in package A
+     * 2) top-level in package B
+     * 3) inside a companion object in package A/B
+     *
+     * @see divergentBlock
+     */
+    private fun sortDivergentElementsDeterministically(elements: List<Documentable>): List<Documentable> =
```
vmishenev, line 588: Minor comment: it sorts deterministically only by signature.

vmishenev (changes requested):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Are you sure? As far as I can see, `sortedWith` returns a brand new list for single-element lists

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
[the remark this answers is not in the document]
IgnatBeresnev: `packageName` is nullable, so it has to be either first or last :) By default comparison rules, all nulls are last - the explicit  use just locks it into place

> should the "null" package be first?

If you have a reason for why it should, I don't mind

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Not sure if you're proposing something

Also not sure what you mean by `only by signature` - are containing packages and classes part of the signature? :thinking:

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Like this?

```kotlin
fun Comparator<Documentable>.topLevelExtensionFirst() : Comparator<Documentable> = 
    thenBy(nullsFirst()) { it.dri.classNames }

compareBy<Documentable, String?>(nullsLast()) { it.dri.packageName }
    .topLevelExtensionFirst()
    .thenBy(..)
```

If yes, why?

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Yeah, I also noticed it, but couldn't come up with how it can be fixed, other than not grouping such extensions under one block. 

Hopefully, one day we will get rid of the left column altogether as planned, but in the meantime I can create an issue for fixing this, if you think it's significant enough.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: It costs almost nothing.
It is consistent to have a new one.
e.g. 
```kt
val list = mutableListOf(...)
val doWeHaveANewListOrNot = sortDivergentElementsDeterministically(list)
list == doWeHaveANewListOrNot
// modify list
```

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: I mean if some elements have the same signature then the sort is not deterministic anymore. My proposal is to rename it to e.g. `sortDivergentElementsBySignature`.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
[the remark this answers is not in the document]
vmishenev: Null for package means the top level 
https://github.com/Kotlin/dokka/blob/683f6bb418bada9c90d2ffbd5040b3968e1c3522/core/src/main/kotlin/links/DRI.kt#L24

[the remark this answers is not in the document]
vmishenev: But I have checked we have an empty string for the root package

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Changed it to `entries.sortedBy`, all tests are green, YOLO

Now `fun bAr()` and `fun BaR()` will be in different "groups", whereas before they were under the same one

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: Something like:

```kt
fun Comparator<Documentable>.thenTopLevelExtensionFirst(comparatorOfClassNames: Comparator<String>) = thenBy (nullsFirst(comparatorOfClassNames)) { it.dri.classNames }
private val divergentDocumentableComparator =
    compareBy<Documentable, String?>(nullsLast()) { it.dri.packageName }
    .thenTopLevelExtensionFirst(comparatorOfClassNames = naturalOrder())
```
It needs no code comment.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: > if some elements have the same DRI

I thought DRI is supposed to be a unique identifier :thinking: 

Can you give an example of such Kotlin code that has 2 functions/properties, it compiles, and both declarations have the same DRI?

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
ilya-g: > whereas before they were under the same one

There were two groups before because `groupBy` used `it.name`, but one of them was dropped by placing them under the same key in the `SortedMap`

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
ilya-g: I haven't seen such case of extensions from different packages in stdlib so far, but it can happen in kotlinx.coroutines once #2845 is implemented.

ilya-g (approved):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: Sorry for my bad explanation)
I confused myself.
- This sort is deterministic since  `sortedWith` is a stable one so something like `sortDivergentElements` can be better.
 - But a comparator can be deterministic as well, but I am not sure about DRI and other fields.
 As for me, it is too formal to have `deterministic` postfix for the comparator. 
 The naming is up to you. 
  
> I thought DRI is supposed to be a unique identifier thinking

Only in compilable projects.  Some users call Dokka with arbitrary source sets.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Yep, by "group" I meant `DivergentGroup`, not the result of `groupBy` - in other words, one line in the table

Now there will be two different lines, so you'll have two entries for `bitstream` and `bitStream`

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Unless it's critical for you, I'd like to leave it as is, even if it costs almost nothing. This will be called for every type (annotation/class/interface/object/typealias), function and property, the overwhelming majority of which will be 1 element lists that don't need sorting. The check is simple enough to understand and shouldn't hurt.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: Sorry, but in this particular case I don't think having a separate function makes it any more readable - if anything, it makes it a bit more complicated than it needs to be. I don't think the function will be re-usable or that it encapsulates some logic that will be easier to refactor/add to.

Note on the naming: it's a comparison of divergent elements, and types and members are considered to be that as well, so `thenTopLevelExtensionFirst` would have to be renamed to something like `thenByClassNameWithTopLevelMemersFirst`

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
IgnatBeresnev: > As for me, it is too formal to have deterministic postfix for the comparator.

The comparator doesn't have the `deterministic` part, it's called `divergentDocumentableComparator`.

But it is indeed very formal to have `deterministic` in the name of the function - that's the whole purpose of it. It's not just some sorting, it solves the problem of deterministic rendering, which, if someone starts refactoring things or fixing nearby bugs, must be respected

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: >  which will be 1 element lists that don't need sorting.

They will not be sorted. I have not understood what the problem with short-liver single-element lists is.

> The check is simple enough to understand and shouldn't hurt. 

I gave the example above when it can hurt.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: > The comparator doesn't have the deterministic part, it's called divergentDocumentableComparator.

Sorry, it would be too formal if it even had `deterministic`.

> It's not just some sorting

Without context it seems about sorting.

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: Up to you

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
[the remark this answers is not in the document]
qwwdfsad: It seems like the last question here is about `null` but it's unclear whether nulls are possible at all.

The reasonable outcomes might be the following:
* Just pick `nullsFirst()` or `nullsLast()`, file an issue and investigate whether `null`s are possible as `packageName`
* Figure out it here and fix accordingly.

Taking into account release pressure, my vote is for the first option

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
ilya-g: Originally, I also thought that this shortcut was redundant, but given the no-copy optimization consideration in case of single-element lists and the fact that this function is private and used only in one place where we know for sure that its input is not mutable, I don't think it can cause confusion. 

And to avoid it further, the function can be provided a kdoc describing its contract in case it is needed elsewhere, e.g. "returns the original list if it has one or zero elements"

IgnatBeresnev: Given how this is a client app that often runs on free-tier CI nodes with limited resources, I believe these things can add up to something noticeable over time. Even if not, I don't think it hurts to the point where it needs to be deleted now. If it gets in the way after some refactoring - sure, I don't mind, we can remove it.

With or without the "optimization", I'd still like to have a separate function to have documentation, so we might as well check for size while we're at it.

Added documentation as proposed

vmishenev: By the way, speaking about micro-optimization, you can use `MutableList<Documentable>` and in-place `sortWith`. It will allow to get more performance and avoid this heuristic.
It needs to replace `.groupBy{ it.name }` with `.groupByTo(mutableMapOf()) { it.name }`

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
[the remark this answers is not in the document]
vmishenev: We have some stubs (as an empty string) for an unnamed package 
https://github.com/Kotlin/dokka/blob/a5a20cdd23db75b65cbf784854c3f26e6584953b/kotlin-analysis/src/main/kotlin/org/jetbrains/dokka/analysis/DRIFactory.kt#L37
If we do not break it, the sort should work (I hope).

So the first option is reasonable for me to merge the PR sooner.

>  I picked last for consistency with callable

But it is not consistent with `.thenBy(nullsFirst()) { it.dri.classNames }` so I dropped attention to this.

vmishenev (approved):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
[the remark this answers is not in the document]
IgnatBeresnev: > But it is inconsistent with 

Yes, that's why `nullsFirst()` has a comment to explain the decision

Issue #1868, opened 2021-04-21
HTML (and possibly others) module order should be consistent and definable


rnett:
**Is your feature request related to a problem? Please describe**
When I generate multi-module docs, the order of the modules on the index page doesn't match the order on the sidebar, and neither are the order I want.  Example:

![image](https://user-images.githubusercontent.com/32281897/115483989-bf6c9580-a206-11eb-847d-477e942a9485.png)

**Describe the solution you'd like**
I would like a way to set the order of modules in mutli-module projects in both the sidebar and index screen.  The default should be consistent across locations.

rnett:
I'm working on a PR for this.  However there's a few ways to go about doing it and I'm not sure which would be preferred:

1. Have a order setting in `DokkaBaseConfiguration`, use it in `NavigationPageInstaller` and `MultimodulePageCreator`.  This is the "proper" way I would think, but it has poor discovery and only applies to HTML.  Would setting the order in `AllModulesPageGeneration.processSubmodules` apply to the navigation sidebar?  Because that would be a bit better.
2. Add the setting to `DokkaMultiModuleTask` and set the module order there.  This is the only entrypoint for multi-module docs, so we can just set the module order here and everything else that uses modules will pick it up (as long as they don't sort themselves, but that's fine).
3. Add the order to `DokkaConfigurationImpl`'s constructor, sort there.  Add a config setting to `DokkaMultiModuleTask`.

2 and 3 are much more resilient to changes and work with other formats better.  3 would be my preference.  Thoughts?

rnett:
I also don't have the slightest idea where to add tests.

PR #2849, opened 2023-02-06
Allow to specify order of modules in the multi-module documentation page in DokkaMultiModuleTask

ilya-g:
Allow to specify order of modules in the multi-module documentation page in DokkaMultiModuleTask and then preserve that order of modules in other places.

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleTask.kt
@@ -52,6 +52,10 @@ abstract class DokkaMultiModuleTask : AbstractDokkaParentTask() {
     val fileLayout: Property<DokkaMultiModuleFileLayout> = project.objects.safeProperty<DokkaMultiModuleFileLayout>()
         .safeConvention(DokkaMultiModuleFileLayout.CompactInParent)
 
+    @Internal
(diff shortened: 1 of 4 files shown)


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleTask.kt
```diff
@@ -52,6 +52,10 @@ abstract class DokkaMultiModuleTask : AbstractDokkaParentTask() {
     val fileLayout: Property<DokkaMultiModuleFileLayout> = project.objects.safeProperty<DokkaMultiModuleFileLayout>()
         .safeConvention(DokkaMultiModuleFileLayout.CompactInParent)
 
+    @Internal
+    val moduleOrder: Property<DokkaMultiModuleOrder> = project.objects.safeProperty<DokkaMultiModuleOrder>()
+        .safeConvention(DokkaMultiModuleOrder.Default)
```
ilya-g, line 57: This has a problem that such property is not included in the cache key, therefore changing just the order won't rebuild the documentation.

ilya-g:
Perhaps another approach for ordering would be to introduce some new property, Int or String, in the Dokka partial task, so that the modules will be sorted first by the specified property value (nulls last), and then by their relative path as they are sorted now.

IgnatBeresnev:
This has the potential to fix #1868

PR #2857, opened 2023-02-13
Deprecate methods for adding/removing task dependencies

atyrin:
Deprecate usage of custom functions to add dependencies for parent tasks(Multimodule and Collector).
It is still possible to use them but better to get rid of them. In the future, by implementing project isolation, they will be removed.

Relevant PR: #2835


IgnatBeresnev (approved):
:+1: :+1:

PR #2858, opened 2023-02-13
Do not render filter-section for empty source sets

IgnatBeresnev:
Regression from #2848

Dokka was rendering an empty `filter-section` div on the all-modules page, which made the JS script think that everything was filtered out.

image


atyrin (approved):

vmishenev (approved):
Should we have a test to avoid it further?

vmishenev:
The same problem was in https://github.com/Kotlin/dokka/pull/2829

IgnatBeresnev:
> Should we have a test to avoid it further?

I'll definitely try, although I've never written a multimodule test and don't know if it's at all possible

I've written it down and I'll add it a bit later separately, as this is blocking Ilya

PR #2860, opened 2023-02-14
Gradle 8.0

Goooler:
https://docs.gradle.org/8.0/release-notes.html
https://docs.gradle.org/8.0/userguide/upgrading_version_7.html#changes_8.0


Goooler:
https://github.com/Kotlin/dokka/actions/runs/4171267730/jobs/7221031783

```
* Where:
Build file '/home/runner/work/dokka/dokka/kotlin-analysis/build.gradle.kts' line: 3

* What went wrong:
An exception occurred applying plugin request [id: 'com.github.johnrengelman.shadow', version: '7.1.2']
> Failed to apply plugin class 'com.github.jengelman.gradle.plugins.shadow.ShadowJavaPlugin'.
   > You can't map a property that does not exist: propertyName=classifier
```

This issue of Shadow has been filed at https://github.com/johnrengelman/shadow/issues/820.

Goooler:
`Jar:classifier` had been removed from Gradle 8, need to migrate to `archiveClassifier`, just bumping Shadow plugin in buildSrc.

https://docs.gradle.org/7.6/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:classifier

Goooler:
```
* Exception is:
java.lang.NoSuchMethodError: org.gradle.api.internal.tasks.AbstractTaskDependency: method 'void <init>()' not found
    at org.jetbrains.dokka.gradle.TaskDependencyInternalWithAdditions.<init>(TaskDependencyInternalWithAdditions.kt:14)
    at org.jetbrains.dokka.gradle.TaskDependencyInternalWithAdditionsKt.plus(TaskDependencyInternalWithAdditions.kt:9)
    at org.jetbrains.dokka.gradle.DokkaMultiModuleTask.getTaskDependencies(DokkaMultiModuleTask.kt:45)
    at org.jetbrains.dokka.gradle.DokkaMultiModuleTask_Decorated.getTaskDependencies(Unknown Source)
    at org.jetbrains.dokka.gradle.DokkaMultiModuleTask.getTaskDependencies(DokkaMultiModuleTask.kt:18)
    at org.gradle.execution.plan.LocalTaskNode.getDependencies(LocalTaskNode.java:147)
    at org.gradle.execution.plan.LocalTaskNode.resolveDependencies(LocalTaskNode.java:121)
    at org.gradle.execution.plan.DefaultExecutionPlan.discoverNodeRelationships(DefaultExecutionPlan.java:184)
```

Goooler:
Thanks, I'll continue some work after 1.8.10 released.

examples/gradle/dokka-customFormat-example/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
```
3flex, line 3:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.1-bin.zip
```

3flex: Thanks - it actually needs to be updated in a few places!

Goooler: Yeah, I'll update them after Dokka 1.8.10 released.

Goooler:
@aSemy `apiVersion` overrides not working after applying `kotlin-dsl`?

runners/gradle-plugin/build.gradle.kts
```diff
@@ -22,6 +23,13 @@ dependencies {
     testImplementation("com.android.tools.build:gradle:4.0.1")
 }
 
+tasks.withType<KotlinCompile>().configureEach {
+    compilerOptions {
+        apiVersion.set(KotlinVersion.fromVersion("1.4"))
+        languageVersion.set(KotlinVersion.fromVersion("1.4"))
+    }
+}
+
```
aSemy, line 32: hey @Goooler, thanks for the ping

[Gradle 8 bumped the Kotlin language level to 1.8](https://docs.gradle.org/current/userguide/compatibility.html#kotlin)

Also, the `kotlin-dsl` plugin will automatically configure the language level ([as well as other stuff](https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:kotlin-dsl_plugin)), so there's no need to set it manually.

aSemy:
```suggestion
```

Goooler: Cause there are some outdated APIs used, seems we can't elevate API level to 1.8 for now.

https://github.com/Kotlin/dokka/actions/runs/4344123199/jobs/7587028523
https://github.com/Kotlin/dokka/actions/runs/4344122470/jobs/7587025547

aSemy: Ah right, okay.

The deprecations look like minor things to fix

```
e: warnings found and -Werror specified
w: file:///home/runner/work/dokka/dokka/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaPlugin.kt:109:63 'decapitalize(): String' is deprecated. Use replaceFirstChar instead.
w: file:///home/runner/work/dokka/dokka/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/dokkaDefaultOutputDirectory.kt:11:59 'decapitalize(): String' is deprecated. Use replaceFirstChar instead.
w: file:///home/runner/work/dokka/dokka/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/isMainSourceSet.kt:26:34 'toLowerCase(): String' is deprecated. Use lowercase() instead.
```

Goooler: Yeah, I did, but more things in https://github.com/Kotlin/dokka/actions/runs/4343604835/jobs/7585857751

aSemy: Hmm okay... 

My guess is that the Android plugin needs to be updated

```
org.jetbrains.dokka.gradle.AndroidAutoConfigurationTest > at least one dokka task created FAILED
    java.lang.NoClassDefFoundError at AndroidAutoConfigurationTest.kt:15
        Caused by: java.lang.ClassNotFoundException at AndroidAutoConfigurationTest.kt:15
```

https://developer.android.com/studio/releases/gradle-plugin#updating-gradle

https://github.com/Kotlin/dokka/blob/ded804e5772399f1495016d598573cb20b673b58/runners/gradle-plugin/build.gradle.kts#L17

Goooler: Yeah, that's another story, we need that some migrations first or just keep the compatibilities. WDYT @IgnatBeresnev

3flex: I've been bitten by using kotlin-dsl in a plugin before, because the plugin was being used in versions of Gradle that were older than the version used to compile the plugin. I'd suggest reverting 2702 and avoiding use of `kotlin-dsl` in the plugin itself.

IgnatBeresnev: We generally want to have the same compatibility guarantees as the Kotlin Gradle plugin itself, whose [compatible versions are now this](https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin):



Since we're past 1.8.0, we can bump the AGP version to 4.1.3, but I don't know if it'll help much :(

> I've been bitten by using kotlin-dsl in a plugin before, because the plugin was being used in versions of Gradle that were older than the version used to compile the plugin

Was it Gradle 5 by any chance? We couldn't run under it with `kotlin-dsl`, but since KGP doesn't support it (we depend on it for autoconfiguration), there's no reason that we should. We also [test against older versions of Gradle](https://github.com/Kotlin/dokka/blob/f09b149b5bf6834609dec1eb70789539c5e61896/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt#L14), so it should hopefully catch bugs related to that

IgnatBeresnev: Do I understand correctly that the old version of AGP is holding us from updating? That is, AGP 4.0.1 is not compatible with Gradle 8?

3flex: > was it Gradle 5 by any chance?

I can't recall the details, I just know that we made a similar change in the detekt Gradle plugin, it mostly worked, but some users running older Gradle versions had problems so we reverted it.

Goooler: Here https://github.com/detekt/detekt/issues/1286

aSemy: > I've been bitten by using kotlin-dsl in a plugin before, because the plugin was being used in versions of Gradle that were older than the version used to compile the plugin. I'd suggest reverting 2702 and avoiding use of `kotlin-dsl` in the plugin itself.


The `kotlin-dsl` plugin just adds some [sensible defaults](https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:kotlin-dsl_plugin), removing it won't help - those defaults would have to be added back in and it would make the build config more complicated and brittle. So my guess is the problems you're referring to, that https://github.com/detekt/detekt/issues/1286 solved, would be caused by mis-configuration ([e.g. adding a runtime dependency on `gradleApi()`](https://github.com/3flex/detekt/blob/2495df812ddc8d6426e09888899eebe632f33edb/detekt-gradle-plugin/build.gradle.kts#L44)), or [it using the Gradle embedded Kotlin](https://github.com/3flex/detekt/blob/2495df812ddc8d6426e09888899eebe632f33edb/detekt-gradle-plugin/build.gradle.kts#L26). 

In any case, it's not relevant to this PR.


> Do I understand correctly that the old version of AGP is holding us from updating? That is, AGP 4.0.1 is not compatible with Gradle 8?

It's unclear. [That's what the test failure seemed to suggest](https://github.com/Kotlin/dokka/pull/2860#discussion_r1126523308), but it doesn't make sense to me. The Android dependency shouldn't be 'aware' of the Gradle version†, so why should there be a missing class just because the Gradle version is different?

I'll do some digging today. It might be that [we make DGP variants that are aware of the Gradle version](https://docs.gradle.org/current/userguide/implementing_gradle_plugins.html#plugin-with-variants), which on one hand is cool but on the other it will be a pain.

† While it is possible to [publish plugins variants that are aware of the Gradle version](https://docs.gradle.org/current/userguide/implementing_gradle_plugins.html#using_version_specific_variants_of_multi_variant_plugins) that say "I'm only compatible with Gradle 7.6", so Gradle won't select them. But that's not the error (it would be a dependency resolution error), and AGP is older than the Gradle version that introduced this feature.

IgnatBeresnev: Thanks for the context on `kotlin-dsl`! We'll keep an eye on it. So far it's not causing problems and I hope the integration tests cover it well

> It's unclear. https://github.com/Kotlin/dokka/pull/2860#discussion_r1126523308, but it doesn't make sense to me. 

Yeah, that's what I can't understand as well, the error is quite strange, considering the changes :( Unless AGP was using some Gradle API whch was removed/reworked

Thank you for taking a look!

aSemy: The actual cause of the test failure was hidden in the GitHub logs. I reproduced it locally, and here's the key parts of the stacktrace:

```
org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'android-auto-configuration-test'.

Caused by: org.gradle.internal.event.ListenerNotificationException: Failed to notify project evaluation listener.

Caused by: java.lang.NoClassDefFoundError: org/gradle/api/plugins/MavenPlugin
	at com.android.build.gradle.internal.variant.VariantHelper.setupArchivesConfig(VariantHelper.java:47)
	at com.android.build.gradle.internal.LibraryTaskManager.createBundleTask(LibraryTaskManager.java:396)

Caused by: java.lang.ClassNotFoundException: org.gradle.api.plugins.MavenPlugin
```

The `org.gradle.api.plugins.MavenPlugin` class was moved and marked as deprecated in Gradle 7, and now it's removed in Gradle 8 https://github.com/gradle/gradle/pull/22807


So yeah, Android-GP 4.0.1 isn't compatible with Gradle 8 (I also checked with 4.1.3 - same problem).

However, I think the test failure is caused by the test not creating a realistic scenario. It uses `org.gradle.testfixtures.ProjectBuilder`, which is intrinsically linked to the Gradle version used to build Dokka. 

https://github.com/Kotlin/dokka/blob/14c05d70b52814fe48e930b3f61fed5e8586718c/runners/gradle-plugin/src/test/kotlin/org/jetbrains/dokka/gradle/AndroidAutoConfigurationTest.kt#L13-L20


In real life that's not going to happen - Android-GP would have to use Gradle 7 or lower.

As I see it, the workaround is to re-write `AndroidAutoConfigurationTest` to use Gradle TestKit, which Dokka already has set up, and to specify the Gradle version to be 7 or lower for `AndroidAutoConfigurationTest`.

https://github.com/Kotlin/dokka/blob/14c05d70b52814fe48e930b3f61fed5e8586718c/integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt#L29-L49

aSemy:
An additional task for this PR: the tested Gradle versions should be updated to include Gradle 8, and possible drop older versions?

https://github.com/Kotlin/dokka/blob/14c05d70b52814fe48e930b3f61fed5e8586718c/integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt#L3-L50

runners/gradle-plugin/build.gradle.kts
IgnatBeresnev: Nice find! It indeed doesn't look like it's Dokka's problem in this case, since we do support AGP 4.X, but AGP 4.X itself is not compatible with Gradle 8 - not Dokka's fault

Refactoring the test to utilize TestKit would be good, preferably in a separate PR. Although I'm not sure if it'll be possible to check the same things as in `AndroidAutoConfigurationTest` - it requires access to the configured Dokka tasks. Maybe it could be done as verify tasks inside of the test project? We actually already have an integration test for Android - `Android0GradleIntegrationTest`)

Goooler:
Addressing to #2960.

PR #2861, opened 2023-02-15
Sort groups of divergent elements by their key first ignoring case, then preserving it

ilya-g:
Currently, sorting groups just by name leads to the following order:
- `getElementNS`
- `getElementName`

which is inconsistent with the sorting order in the navigation bar.

After this change, the order will be:
- `getElementName`
- `getElementNS`

but if we add `getElementNs` to that, it would be strictly after `getElementNS`

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
@@ -6,6 +6,7 @@ import org.jetbrains.dokka.base.transformers.documentables.isDeprecated
 import org.jetbrains.dokka.base.transformers.documentables.isException
 import org.jetbrains.dokka.base.translators.documentables.DocumentableLanguage
 import org.jetbrains.dokka.base.translators.documentables.documentableLanguage
+import org.jetbrains.dokka.base.utils.canonicalAlphabeticalOrder
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.pages.*
 
@@ -90,6 +91,9 @@ abstract class NavigationDataProvider {
         }
     }
 
+    private val navigationNodeOrder: Comparator<NavigationNode> =
+        compareBy(canonicalAlphabeticalOrder) { it.name }
+
     private fun ContentPage.navigableChildren() =
         if (this is ClasslikePage) {
             this.navigableChildren()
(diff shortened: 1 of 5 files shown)


plugins/base/src/main/kotlin/utils/alphabeticalOrder.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.dokka.base.utils
+
+
+/**
+ * Canonical alphabetical order to sort named elements
+ */
+internal val canonicalAlphabeticalOrder: Comparator<String?> = nullsFirst(String.CASE_INSENSITIVE_ORDER).thenBy { it }
```
ilya-g, line 7: or perhaps it should be `Comparator<String>` and whether to put nulls first or last should be decided on usage

plugins/base/src/main/kotlin/renderers/html/NavigationDataProvider.kt
```diff
@@ -90,14 +91,17 @@ abstract class NavigationDataProvider {
...
+
     private fun ContentPage.navigableChildren() =
         if (this is ClasslikePage) {
             this.navigableChildren()
         } else {
             children
                 .filterIsInstance<ContentPage>()
                 .map { visit(it) }
-                .sortedBy { it.name.toLowerCase() }
+                .sortedWith(navigationNodeOrder)
```
IgnatBeresnev, line 104: Could you please add a unit test for navigation to check the use case you described, in case the sorting ever changes

The test class is `NavigationTest`, it should be pretty straightforward

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -525,7 +529,7 @@ open class DefaultPageCreator(
                     .groupBy { it.name } // This groupBy should probably use LocationProvider
                     // This hacks displaying actual typealias signatures along classlike ones
                     .mapValues { if (it.value.any { it is DClasslike }) it.value.filter { it !is DTypeAlias } else it.value }
-                    .entries.sortedBy { it.key }
+                    .entries.sortedWith(groupKeyComparator)
```
IgnatBeresnev, line 532: Could you please add a unit test to check the described use case as well?

Here it might be a bit more difficult because of the deeply nested groupings. The test class is `PageNodeMergerTest`. You can have a look at the test named `should not ignore case when grouping by name` - you'll probably need something similar

IgnatBeresnev (commented):
Nice find! Didn't even think of that

plugins/base/src/main/kotlin/utils/alphabeticalOrder.kt
IgnatBeresnev: To me it's fine as it is, especially since it's `internal`, but `Comparator<String>` and extracting null comparators to the usages could add a little more transparency. Whatever you think is best

IgnatBeresnev (approved):
:+1: :+1:

Issue #2866, opened 2023-02-15
Improve how modules are merged - `relativePathToOutputDirectory` can't handle a module that is not a child


aSemy:
### Describe the bug

[`DokkaModuleDescription.relativePathToOutputDirectory`](https://github.com/Kotlin/dokka/blob/daed35f92b3b482688856d139da6849c8e6b4ab1/core/src/main/kotlin/configuration.kt#L243) is used by Dokka when aggregating Dokka Modules into a single directory. 

The value is used to create subdirectories in the Dokka output directory. For example, if there are two modules, where `relativePathToOutputDirectory` is `childProjectA` one and `childProjectB` in another, there will be equivalent subdirectories in the result.

 ```
 build/
 ├── htmlMultiModule/
 │   ├── childProjectA/
 │   │   ├── index.html
 │   │   └── ...
 │   └── childProjectB/
 │       ├── index.html
 │       └── ...
 ├── index.html
 └── ...
 ```

The value of `relativePathToOutputDirectory` is determined automatically by the relative path between the aggregating project and the module.

* https://github.com/Kotlin/dokka/blob/daed35f92b3b482688856d139da6849c8e6b4ab1/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleTask.kt#L93-L98
* https://github.com/Kotlin/dokka/blob/daed35f92b3b482688856d139da6849c8e6b4ab1/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/DokkaMultiModuleFileLayout.kt#L36-L43

This is problematic because it cannot handle modules that are _not_ in child directories. For example, if the aggregating project is not a direct parent of the children

```
project/
├── docs/
│   └── build.gradle.kts <- Dokka site is created here
├── parentProject/
│   ├── childProjectA/
│   │   └── build.gradle.kts
│   └── childProjectB/
│       └── build.gradle.kts
└── settings.gradle.kts
```

Additionally:

* The variable name `relativePathToOutputDirectory` is confusing - it's more like "the relative path INSIDE the output directory" or "moduleDocPath". 
* Is it necessary? Can't the directory default to the module's [name](https://github.com/Kotlin/dokka/blob/daed35f92b3b482688856d139da6849c8e6b4ab1/core/src/main/kotlin/configuration.kt#L242)?
[...]

PR #2867, opened 2023-02-15
TemplateProcessor: replace `mkdir()` with `mkdirs()`

aSemy:
Replace `mkdir()` with `mkdirs()`, so that `relativePathToOutputDirectory` can be set as a nested. This improves the handling of aggregating modules, as they can be placed into distinct directories.

I am not sure how to test this...

Part of #2866

### DIFF
--- a/plugins/templating/src/main/kotlin/templates/TemplateProcessor.kt
+++ b/plugins/templating/src/main/kotlin/templates/TemplateProcessor.kt
@@ -53,7 +53,7 @@ class DefaultSubmoduleTemplateProcessor(
         coroutineScope {
             val source = this@visit
             if (source.isDirectory) {
-                target.mkdir()
+                target.mkdirs()
                 val files = source.list().orEmpty()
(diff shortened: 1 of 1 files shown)


IgnatBeresnev (approved):
Thanks!

IgnatBeresnev:
> I am not sure how to test this...

I don't think it's critical to test this part now - if it backfires, other tests should hopefully fail (e.g it should be covered by the integration tests). Especially since we'll most like have to change `relativePathToOutputDirectory` to accept a non-relative path

If that was the only reason to keep in draft, I'm ready to merge it :)

aSemy:
I've done a bit of manual testing with Dokkatoo and I think that it more or less works 👍 

> Especially since we'll most like have to change `relativePathToOutputDirectory` to accept a non-relative path

Coincidentally I did a bit of research into `relativePathToOutputDirectory`, so it doesn't require any weird workarounds like I thought, I was just confused by the name - more details in #2866

PR #2868, opened 2023-02-16
Exclude user guides from mkdocs-built documentation

IgnatBeresnev:
User guides were completely re-written and it's now hosted on [kotlinlang.org](https://kotlinlang.org/docs/dokka-introduction.html)


mkdocs/src/doc/mkdocs.yml
```diff
@@ -51,27 +51,17 @@ markdown_extensions:
...
-      - Applying Dokka:
-          - Gradle: user_guide/applying/gradle.md
-          - Maven: user_guide/applying/maven.md
-          - Command line: user_guide/applying/cli.md
-      - Output formats:
-          - HTML: user_guide/output-formats/html.md
-      - Plugins:
-          - Versioning plugin: user_guide/plugins/versioning-plugin.md
-          - Android plugin: user_guide/plugins/android-plugin.md
+      - https://kotlinlang.org/docs/dokka-introduction.html
```
IgnatBeresnev, line 64: There will still be the "User guides" button in the header, but it will lead to the documentation on kotlinlang

vmishenev (approved):

PR #2884, opened 2023-02-22
build config: Gradle Version Catalog + type-safe project dependencies

aSemy:
* Moves versions out of `gradle.properties` and into `libs.versions.toml` [version catalog](https://docs.gradle.org/current/userguide/platforms.html)
* Uses [type-safe project dependencies](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:type-safe-project-accessors) for internal projects
* refactor the maven-cli downloader convention plugin (added in #2911) to use the version catalog, and for clarity
* Additionally updates JUnit to 5.9.2

Depends on #2652

### DIFF
--- a/build-logic/src/main/kotlin/org/jetbrains/publication.kt
+++ b/build-logic/src/main/kotlin/org/jetbrains/publication.kt
@@ -2,13 +2,9 @@ package org.jetbrains
 
 import com.github.jengelman.gradle.plugins.shadow.ShadowExtension
 import org.gradle.api.Project
-import org.gradle.api.plugins.JavaBasePlugin
 import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
-import org.gradle.api.tasks.TaskContainer
-import org.gradle.api.tasks.TaskProvider
-import org.gradle.api.tasks.bundling.Jar
 import org.gradle.kotlin.dsl.*
 import org.gradle.plugins.signing.SigningExtension
 import org.jetbrains.DokkaPublicationChannel.*
@@ -24,7 +20,10 @@ class DokkaPublicationBuilder {
 }
 
 
-fun Project.registerDokkaArtifactPublication(publicationName: String, configure: DokkaPublicationBuilder.() -> Unit) {
+fun Project.registerDokkaArtifactPublication(
+    publicationName: String,
+    configure: DokkaPublicationBuilder.() -> Unit
+) {
     configure<PublishingExtension> {
         publications {
             register<MavenPublication>(publicationName) {
@@ -143,9 +142,10 @@ private fun Project.signPublicationsIfKeyPresent(vararg publications: String) {
                 useInMemoryPgpKeys(signingKey, signingKeyPassphrase)
             }
             publications.forEach { publicationName ->
-                extensions.findByType(PublishingExtension::class)!!.publications.findByName(publicationName)?.let {
-                    sign(it)
-                }
+                extensions.getByType<PublishingExtension>()
+                    .publications
+                    .findByName(publicationName)
+                    ?.let { sign(it) }
             }
         }
     }
--- a/examples/gradle/dokka-customFormat-example/build.gradle.kts
+++ b/examples/gradle/dokka-customFormat-example/build.gradle.kts
@@ -31,6 +31,5 @@ tasks.dokkaHtml {
 }
 
 dependencies {
-    implementation(kotlin("stdlib"))
     testImplementation(kotlin("test-junit"))
 }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -47,7 +47,7 @@ jetbrainsIntelliJ-core = { module = "com.jetbrains.intellij.idea:intellij-core",
 jetbrainsIntelliJ-jpsStandalone = { module = "com.jetbrains.intellij.idea:jps-standalone", version.ref = "idea" }
 jetbrainsMarkdown = { module = "org.jetbrains:markdown", version.ref = "jetbrainsMarkdown" }
 jsoup = { module = "org.jsoup:jsoup", version.ref = "jsoup" }
-soywiz-korte = { module = "com.soywiz.korlibs.korte:korte", version.ref = "soywiz-korte" }
+soywiz-korte = { module = "com.soywiz.korlibs.korte:korte-jvm", version.ref = "soywiz-korte" }
 
 ## Kotlin libs ##
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
@@ -58,8 +58,8 @@ kotlin-common = { module = "org.jetbrains.kotlin:common", version.ref = "kotlin"
 kotlin-core = { module = "org.jetbrains.kotlin:core", version.ref = "kotlin" }
 kotlin-native = { module = "org.jetbrains.kotlin:native", version.ref = "kotlin" }
 
-kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli", version = "0.3.4" }
-kotlinx-html = { module = "org.jetbrains.kotlinx:kotlinx-html", version.ref = "kotlinx-html" }
+kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli-jvm", version = "0.3.4" }
+kotlinx-html = { module = "org.jetbrains.kotlinx:kotlinx-html-jvm", version.ref = "kotlinx-html" }
 
 kotlinx-coroutines-bom = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-bom", version.ref = "kotlinx-coroutines" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "kotlinx-coroutines" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,15 +4,17 @@ kotlin = "1.8.10"
 kotlin-plugin = "213-1.8.10-release-430-IJ6777.52"
 kotlinx-coroutines = "1.6.3"
 kotlinx-html = "0.7.5"
-jsoup = "1.15.3"
+kotlinx-cli = "0.3.4"
+
 idea = "213.6777.52"
+jetbrainsMarkdown = "0.3.1"
+
+jsoup = "1.15.3"
 
 jackson = "2.12.7" # jackson 2.13.X does not support kotlin language version 1.4, check before updating
 jacksonDatabind = "2.12.7.1" # fixes CVE-2022-42003
 
 freemarker = "2.3.31"
-
-jetbrainsMarkdown = "0.3.1"
 
 soywiz-korte = "2.7.0"
 
@@ -58,7 +60,7 @@ kotlin-common = { module = "org.jetbrains.kotlin:common", version.ref = "kotlin"
 kotlin-core = { module = "org.jetbrains.kotlin:core", version.ref = "kotlin" }
 kotlin-native = { module = "org.jetbrains.kotlin:native", version.ref = "kotlin" }
 
-kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli-jvm", version = "0.3.4" }
+kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli-jvm", version.ref = "kotlinx-cli" }
 kotlinx-html = { module = "org.jetbrains.kotlinx:kotlinx-html-jvm", version.ref = "kotlinx-html" }
 
 kotlinx-coroutines-bom = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-bom", version.ref = "kotlinx-coroutines" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,12 +1,6 @@
 [versions]
 
-## compilation ##
-jvm = "11"
 kotlin = "1.8.10"
-kotlin-languageLevel = "1.4"
-
-
-## libs ##
 kotlin-plugin = "213-1.8.10-release-430-IJ6777.52"
 kotlinx-coroutines = "1.6.3"
 kotlinx-html = "0.7.5"
@@ -20,6 +14,7 @@ freemarker = "2.3.31"
 
 jetbrainsMarkdown = "0.3.1"
 
+soywiz-korte = "2.7.0"
 
 apache-maven = "3.5.0"
 apache-mavenArtifact = "3.8.5"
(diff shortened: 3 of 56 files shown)


gradle/libs.versions.toml
```diff
@@ -0,0 +1,86 @@
...
+jvm = "11"
+kotlin = "1.8.10"
+kotlin-languageLevel = "1.4"
+
+
+## libs ##
+kotlin-plugin = "213-1.8.10-release-430-IJ6777.52"
+kotlinx-coroutines = "1.6.3"
+kotlinx-html = "0.7.5"
+jsoup = "1.15.3"
```
Goooler, line 13: We can inline these once-use versions.

integration-tests/build.gradle.kts
```diff
@@ -1,88 +1,13 @@
...
+    org.jetbrains.conventions.`kotlin-jvm`
 }
 
 dependencies {
     implementation(kotlin("stdlib"))
     api(project(":test-utils"))
     val coroutines_version: String by project
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version")
+    implementation(libs.kotlinxCoroutines.core)
     val jsoup_version: String by project
```
Goooler, line 10:
```suggestion
```

kotlin-analysis/intellij-dependency/build.gradle.kts
```diff
@@ -30,19 +21,19 @@ fun jpsModel() = zipTree(jpsStandalone.singleFile).matching {
...
 dependencies {
-    val kotlin_plugin_version: String by project
+    val kotlin_plugin_version: String = libs.versions.kotlin.plugin.get()
+
     api("org.jetbrains.kotlin:common:$kotlin_plugin_version")
     api("org.jetbrains.kotlin:idea:$kotlin_plugin_version") {
         isTransitive = false
     }
     api("org.jetbrains.kotlin:core:$kotlin_plugin_version")
     api("org.jetbrains.kotlin:native:$kotlin_plugin_version")
```
Goooler, line 31: Why not put these into toml?

plugins/base/base-test-utils/build.gradle.kts
```diff
@@ -1,11 +1,26 @@
...
+
 dependencies {
+    compileOnly(project(":core"))
+    implementation(kotlin("stdlib-jdk8"))
+    implementation(kotlin("stdlib"))
+    implementation(kotlin("reflect"))
+
     compileOnly(project(":plugins:base"))
     implementation(project(":core:test-api"))
     val jsoup_version: String by project
```
Goooler, line 16:
```suggestion
```

Goooler (commented):
Good to see! We can migrate to type-safe project accessors in next PR.

IgnatBeresnev (commented):
Had a glance - looking good, thanks!

I'll have a closer look once #2704 gets merged and this PR is rebased onto it

kotlin-analysis/intellij-dependency/build.gradle.kts
aSemy: this PR is still WIP (but I forgot to mark it). I got started with the Version Catalog in #2704, but decided to split it out and I made this PR so I could put a link next to the TODO https://github.com/Kotlin/dokka/pull/2704/files#diff-7b5c40eb5522b5216cfe8c75612e579a8db0daee9ded80df9ce6a56e8361a962R11

aSemy:
> Since there are a bunch of conflicts, proposing to merge #2652 first, it's been waiting for a while :) Hopefully shouldn't take as long

Agreed, sounds good!

IgnatBeresnev:
The PR is ready to be rebased :) Looking forward to it!

Please, if you can, try to address copy-pasted versions that I mentioned in https://github.com/Kotlin/dokka/pull/2912#discussion_r1133791517, I assumed that we'd have the actual version strings in the toml only

build-logic/src/main/kotlin/org/jetbrains/publication.kt
```diff
@@ -143,9 +142,10 @@ private fun Project.signPublicationsIfKeyPresent(vararg publications: String) {
...
             }
             publications.forEach { publicationName ->
-                extensions.findByType(PublishingExtension::class)!!.publications.findByName(publicationName)?.let {
-                    sign(it)
-                }
+                extensions.getByType<PublishingExtension>()
+                    .publications
+                    .findByName(publicationName)
+                    ?.let { sign(it) }
             }
```
aSemy, line 149: this is just some unrelated tidying & formatting

examples/gradle/dokka-customFormat-example/build.gradle.kts
```diff
@@ -31,6 +31,5 @@ tasks.dokkaHtml {
 }
 
 dependencies {
-    implementation(kotlin("stdlib"))
```
aSemy, line 34: > You no longer need to declare a dependency on the `stdlib` library in any Kotlin Gradle project, including a multiplatform one. The dependency is added by default.

https://kotlinlang.org/docs/whatsnew14.html#dependency-on-the-standard-library-added-by-default
 
Therefore I removed it from all subprojects and examples.

settings.gradle.kts
```diff
@@ -1,5 +1,58 @@
+@file:Suppress("UnstableApiUsage")
+
 rootProject.name = "dokka"
 
+pluginManagement {
+    includeBuild("build-logic")
```
aSemy, line 6: I moved `includeBuild` into the plugins block [as recommended](https://docs.gradle.org/current/userguide/composite_builds.html#included_plugin_builds)

> You may also use the `includeBuild` mechanism outside `pluginManagement` to include plugin builds. However, this does not support all use cases and including plugin builds like that might be deprecated in a future Gradle version.

I also moved the repository definitions to be above plugins block, since logically the repos should be defined before plugins are resolved - but that's not important and I can revert if you want.

gradle/libs.versions.toml
aSemy: I think it's best to define all the versions at the top, because it's consistent, it's easier to scan the top to see the versions, and if any versions do need to be re-used then then it's less work (personally I find editing `libs.versions.toml` a bit fiddly, so I try to avoid it).

settings.gradle.kts
Goooler: I moved it out cause I noticed https://github.com/detekt/detekt/commit/0245b0f24b032649908efa37f78b6abd2374efee.

gradle/libs.versions.toml
Goooler: We can easy to know which versions are reused after inlining unnecessary ones, either way.

Goooler (commented):
Good to see this, both version catalog and type-safe project accessors!
We can rename this PR's title to something like "Migrate to type-safe dependency accessors"?

runners/gradle-plugin/build.gradle.kts
```diff
@@ -8,15 +8,14 @@ plugins {
...
-    api(project(":core"))
-
-    compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin")
-    compileOnly("com.android.tools.build:gradle:4.0.1")
-    compileOnly(gradleKotlinDsl())
-    testImplementation(project(":test-utils"))
-    testImplementation(gradleKotlinDsl())
-    testImplementation("org.jetbrains.kotlin:kotlin-gradle-plugin")
-    testImplementation("com.android.tools.build:gradle:4.0.1")
+    implementation(projects.core)
```
IgnatBeresnev, line 11: Previously it was declared as `api(..)` - does it make any difference in the case of a Gradle plugin?

kotlin-analysis/intellij-dependency/build.gradle.kts
```diff
@@ -35,19 +35,17 @@ fun jpsModel() = zipTree(jpsStandalone.singleFile).matching {
...
+    api(libs.kotlinPlugin.core)
+    api(libs.kotlinPlugin.native)
 
-    val idea_version: String by project
-    intellijCore("com.jetbrains.intellij.idea:intellij-core:$idea_version")
+    intellijCore(libs.jetbrainsIntelliJ.core)
     implementation(intellijCoreAnalysis())
 
-    jpsStandalone("com.jetbrains.intellij.idea:jps-standalone:$idea_version")
+    jpsStandalone(libs.jetbrainsIntelliJ.jpsStandalone)
```
IgnatBeresnev, line 48: Qodana is complaining about unstable API usage

IgnatBeresnev (approved):
Great to see, as always!

[The main integration tests](https://github.com/Kotlin/dokka/actions/runs/4419616345/jobs/7775239862) are failing, could you please fix it? I'll run longer integration tests after that

settings.gradle.kts
IgnatBeresnev: Hm. We don't use configuration cache in this project, do we? So we don't have the warning yet, but it might resurface one day

I don't mind leaving it as is, we can deal with the warning when we get it, but if you do move it out - please add a comment why

gradle/libs.versions.toml
IgnatBeresnev: I think it should be consistent - either all versions are at the top, or all of the versions are inlined, otherwise it might eventually lead to chaos. At the moment, I personally don't have any preference, so I'm fine with either approach :+1:

settings.gradle.kts
Goooler: Make sense.

kotlin-analysis/intellij-dependency/build.gradle.kts
aSemy: I'll suppress it for now, but tbh I think this bit of code probably needs a refactor - the JARs can be fetched via existing configurations (possibly). 

I think these dependencies are related to https://github.com/Kotlin/dokka/issues/2889 too? So might they be removed in the future?

settings.gradle.kts
aSemy: I don't think the warning in https://github.com/detekt/detekt/pull/5690 is about configuration cache, it's about the (configuration of) build cache.

https://github.com/detekt/detekt/blob/b07acae2b5ce2d5c8e8216a4e3e022f89c5fa67f/settings.gradle.kts#L71-L84

But Dokka doesn't have that (JetBrains really should set up a Gradle Build Cache server though!)

runners/gradle-plugin/build.gradle.kts
aSemy: Good spot, thanks

runners/gradle-plugin/build.gradle.kts
```diff
@@ -8,15 +8,14 @@ plugins {
 }
 
 dependencies {
-    api(project(":core"))
-
-    compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin")
-    compileOnly("com.android.tools.build:gradle:4.0.1")
-    compileOnly(gradleKotlinDsl())
```
aSemy, line 15: I removed `compileOnly(gradleKotlinDsl())` because it's added by the `kotlin-dsl` plugin

integration-tests/gradle/projects/serialization/kotlinx-serialization
aSemy, line 1: I think I accidentally updated the git commit of KxS in my branch. How can I revert it back to the commit that dokka/master uses...?

kotlin-analysis/intellij-dependency/build.gradle.kts
IgnatBeresnev: Yep, suppressing it is totally fine. These should be removed/refactored quite soon, it's a blocked for 1.8.20

integration-tests/gradle/projects/serialization/kotlinx-serialization
IgnatBeresnev: Looks like you were able to figure it out :)

I usually cd into thee KxS project dir within Dokka, and git checkout the needed commit id, and then commit the change

aSemy: might be fixed now? 

For future reference, in terminal:

* `cd` into the KxS directory
* `git checkout daa95c79ffadc0eedbbb4a481a00556b78212e43` (or whatever the hash is, which can be found by clicking the link to the KxS project in this dir https://github.com/Kotlin/dokka/tree/master/integration-tests/gradle/projects/serialization)
* `cd ..` back up to Dokka root dir
* commit the KxS directory

aSemy: btw, it's possible to use Gradle to download GitHub projects as if they were dependencies. This might be a nice replacement because the version can be set explicitly, but the current approach also works.

https://stackoverflow.com/a/34327202/4161471

settings.gradle.kts
IgnatBeresnev: > JetBrains really should set up a Gradle Build Cache server though!

JFYI we actually have a JetBrains instance for Gradle Enterprise (ge.jetbrains.com), but it didn't look like it using it would benefit us in any way compared to what we have now. If there's some really cool things that can be done with it that will work with Dokka, we can switch to it, no problem

IgnatBeresnev:
Triggered teamcity integration tests, can be merged once they pass

gradle/libs.versions.toml
```diff
@@ -0,0 +1,113 @@
...
+kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
+
+kotlin-compiler = { module = "org.jetbrains.kotlin:kotlin-compiler", version.ref = "kotlin" }
+kotlin-idea = { module = "org.jetbrains.kotlin:idea", version.ref = "kotlin" }
+kotlin-common = { module = "org.jetbrains.kotlin:common", version.ref = "kotlin" }
+kotlin-core = { module = "org.jetbrains.kotlin:core", version.ref = "kotlin" }
+kotlin-native = { module = "org.jetbrains.kotlin:native", version.ref = "kotlin" }
+
+kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli", version = "0.3.4" }
+kotlinx-html = { module = "org.jetbrains.kotlinx:kotlinx-html", version.ref = "kotlinx-html" }
```
IgnatBeresnev, line 62: Doesn't seem to fail the tests (yet), but maybe it's worth using `*-jvm` as before, to avoid possible problems. If a problem with it surfaces, I don't want to have to roll back the whole PR, so imo it'd be nice to keep the artifacts the same as before, and any additional improvements separately

btw, the cli version 0.3.4 could be extracted into a variable, probably forgotten?

```suggestion
kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli-jvm", version = "0.3.4" }
kotlinx-html = { module = "org.jetbrains.kotlinx:kotlinx-html-jvm", version.ref = "kotlinx-html" }
```

settings.gradle.kts
aSemy: If there's public, read-only access to the JetBrains Gradle build cache server then that would definitely help! It means that everyone's builds would be faster.

https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_configure_remote

PR #2892, opened 2023-02-27
Stop testing on Java 15

3flex:
This version is unsupported so there's little value in testing with it.


IgnatBeresnev (approved):
Quite reasonable, thanks!

Issue #2272, opened 2021-12-14
Unable to correctly link to classes in a multi module project with non-unique package names


prempalsingh:
**Describe the bug**
When using `dokkaHtmlMultiModule` in a multi-module project with non-unique package names (two different modules can contain the same package name), dokka incorrectly links classes in some other module.

Example - 

https://img.ly/docs/vesdk/apidocs/android/v9/pesdk-backend-video-core/ly.img.android.pesdk.backend.model.state/-video-editor-save-settings/index.html

Here, `VideoEditorSaveSettings` (in module `pesdk-backend-video-core`) extends from the class `SaveSettings` (in module `pesdk-backend-core`). However, if you click on the link for `SaveSettings` from `VideoEditorSaveSettings`'s documentation (link above), it redirects to https://img.ly/docs/vesdk/apidocs/android/v9/pesdk-backend-adjustment/ly.img.android.pesdk.backend.model.state/-save-settings/index.html 

So, basically, it is trying to find `SaveSettings` in the `pesdk-backend-adjustment` module instead of `pesdk-backend-core`. I believe that this might be because the package `ly.img.android.pesdk.backend.model.state` exists in both `pesdk-backend-adjustment` and `pesdk-backend-core` which is somehow confusing dokka?

**Expected behaviour**
dokka should be correctly able to link to classes in a multi-module project with non-unique package names.

In the example above, clicking on `SaveSettings` should redirect to https://img.ly/docs/pesdk/apidocs/android/v9/pesdk-backend-core/ly.img.android.pesdk.backend.model.state/-save-settings/index.html

**To Reproduce**
Mentioned the conditions above. I can create a sample project where the issue is reproducible if needed.

**Dokka configuration**
Configuration of dokka used to reproduce the bug

```Kotlin
apply plugin: 'org.jetbrains.dokka'
./gradlew dokkaHtmlMultiModule
```

**Installation**
- Operating system: macOS/Linux
- Build tool: Gradle v6.1.1
- Dokka version: 1.4.30

**Are you willing to provide a PR?**
Yes, if someone can guide me which part of the code I should look at.

vmishenev:
This is a known bug https://github.com/Kotlin/dokka/issues/2037 
> It is a situation where 2 gradle modules have the same package and dokka doesn’t know where to link, so it links to first one. Unfortunately in this case it leads to unresolved link

IgnatBeresnev:
FYI

[...]

Issue #2037, opened 2021-07-26
Consider supporting split packages


MarcinAman:
Dokka right now doesn't handle split packages by design. At some point in the future this behaviour should probably change since in some edge cases it leads to unresolved links.

Example of such behaviour is present here, in `ApolloResponse`: https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3/-apollo-client/query.html

vmishenev:
I closed it in favor of #3425.

PR #2901, opened 2023-03-02
all-modules-page: Validate cross-module links during resolution

EddieRingle:
I scanned the list of open issues but didn't spot one that matched my experience directly, but the closest might be #2037 or #2272. I'm not sure if this is necessarily the correct or ideal fix, and the underlying wider issue of supporting split packages probably requires further changes, but I did want to at least share this as it makes Dokka usable for our use-cases.

From commit message:

>The existing behavior was just taking the first module that includes the package name of a given reference, so this resulted in broken links when a package is split across multiple modules.
>
>This change checks to ensure the resolved file exists before choosing it as the target for the resolved link.

### DIFF
--- a/plugins/all-modules-page/src/main/kotlin/ExternalModuleLinkResolver.kt
+++ b/plugins/all-modules-page/src/main/kotlin/ExternalModuleLinkResolver.kt
@@ -5,7 +5,6 @@ import org.jetbrains.dokka.base.DokkaBase
 import org.jetbrains.dokka.base.resolvers.shared.ExternalDocumentation
 import org.jetbrains.dokka.base.resolvers.shared.PackageList
 import org.jetbrains.dokka.base.resolvers.shared.PackageList.Companion.PACKAGE_LIST_NAME
-import org.jetbrains.dokka.base.resolvers.shared.RecognizedLinkFormat
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.plugability.plugin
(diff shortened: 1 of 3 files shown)


plugins/all-modules-page/src/main/kotlin/ExternalModuleLinkResolver.kt
```diff
@@ -52,15 +51,19 @@ class DefaultExternalModuleLinkResolver(val context: DokkaContext) : ExternalMod
...
     override fun resolve(dri: DRI, fileContext: File): String? {
-        val absoluteLink = elps.mapNotNull { it.resolve(dri) }.firstOrNull() ?: return null
-        val modulePath = context.configuration.outputDir.absolutePath.split(File.separator)
-        val contextPath = fileContext.absolutePath.split(File.separator)
-        val commonPathElements = modulePath.zip(contextPath)
+        val resolvedLinks = elps.mapNotNull { it.resolve(dri)?.removePrefix("file:") }
+        val modulePath = context.configuration.outputDir.absolutePath
+        val validLink = resolvedLinks.firstOrNull {
+            val absolutePath = File(modulePath + it)
+            absolutePath.isFile
```
pbajurko, line 58: This checks assumes that dependee submodules have be already processed but that is upto process order  
in `all-modules-page` plugin. For our project it lead to situation where same DRI was sometimes resolved and sometimes not.
I [workaround](https://github.com/pbajurko/dokka/commit/b420868da055ae9fd0e6f583ebbaab6a4eafcbc4) it by checking paths against Dokka's output in submodules since partial tasks will be done before multimodule starts but that is not clean solution.

plugins/all-modules-page/src/main/kotlin/ExternalModuleLinkResolver.kt
```diff
@@ -52,15 +51,19 @@ class DefaultExternalModuleLinkResolver(val context: DokkaContext) : ExternalMod
...
+        } ?: return null
```
vmishenev, line 59: Will it not work for the non-local external `package-list` at all? See the example with the `serialization` library from [External documentation links configuration](https://kotlinlang.org/docs/dokka-gradle.html#external-documentation-links-configuration). Can you fix it, please?

It is nice to fix the issue at least for cross-module links.
A check of existing external URLs online can be excessive and Dokka can be run without access to the Internet. 

Also, I think the fix should check links only if `resolvedLinks` has more than one link. What do you think? It can help avoid unexpected behaviour with resolving cross-module links.
This fix is local and temporary. We have a long-term plan #3368.

plugins/all-modules-page/src/test/kotlin/templates/ResolveLinkCommandResolutionTest.kt
```diff
@@ -56,7 +56,7 @@ class ResolveLinkCommandResolutionTest : MultiModuleAbstractTest() {
             }
         }
 
-        val contentFile = setup(link)
+        val contentFile = setup(link, testedDri)
```
vmishenev, line 59: Could you add a test case for [the case](https://github.com/Kotlin/dokka/issues/2272#issuecomment-997076375) from #2272?

vmishenev (commented):
Thank you so much for your contribution and sorry for the delay. The issue is very painful for Dokka. 

I left the comments in the review.
Could you rebase on `master`? I'm afraid some files have been moved.

plugins/all-modules-page/src/main/kotlin/ExternalModuleLinkResolver.kt
vmishenev: I agree. Thank you for noticing it. The [workaround](https://github.com/pbajurko/dokka/commit/b420868da055ae9fd0e6f583ebbaab6a4eafcbc4) is correct, but it should not depend on particular paths like `build/dokka/htmlMultiModule` or `build/dokka/htmlPartial/`. It would be great to use [`module.sourceOutputDirectory`](https://github.com/Kotlin/dokka/blob/8e5c63d035ef44a269b8c43430f43f5c8eebfb63/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/configuration.kt#L240)

PR #2904, opened 2023-03-07
Introduce InternalDokkaApi annotation

qwwdfsad:
Rationale:

dokka-core has a long history of bloating its API shape with utilities that were never intended to be public and that may expose unwanted implementation details, as well as unwanted compatibility burden.

Eventually, we would like to get rid of them (i.e. by making them internal), but first it would be nice to provide users with an explicit message about it


qwwdfsad:
If you are okay with that, I'll create a follow-up PR that marks all `utilities` and will ask Ignat to take a look

vmishenev (approved):

PR #2905, opened 2023-03-07
Improve JavadocParser and fix case-sensitivity

### LINKED ISSUE #2907: Dokka Javadoc's tag parser is case-insensitive
According to [Javadoc specification](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html), tags are case-sensitive, meaning that `@foo` and `@Foo` are different tags.

Currently, Dokka treats tags in a case-insensitive manner, resolving `@param`, `@PARAM` and `@pArAm` into the standardized `@param` tag. It would be nice to follow the spec here to avoid surprising behavior.

Additionally, I've verified that IDEA's Javadoc parser also treats tags in a case-sensitive manner

qwwdfsad:
* Get rid of safeEnumValueOf that was an unnecessary public API burden and constantly allocating
* Restructure JavadocParser.parseDocTag so it has one lever of nesting less
* Make tag parsing case-sensitive

Fixes #2907

### DIFF
--- a/plugins/base/src/test/kotlin/parsers/JavadocParserTest.kt
+++ b/plugins/base/src/test/kotlin/parsers/JavadocParserTest.kt
@@ -2,6 +2,7 @@ package parsers
 
 import com.jetbrains.rd.util.first
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
+import org.jetbrains.dokka.base.translators.psi.parsers.*
 import org.jetbrains.dokka.links.Callable
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.links.JavaClassReference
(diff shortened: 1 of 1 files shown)


qwwdfsad:
Note that the changeset contains breaking change.
Let's discuss it in slack. Either I will revert it or will file the corresponding issue to address and mention in changelog.

Please review the PR anyway, this one is technicality

vmishenev (approved):
It would be great to have a test (e.g. in `JavadocParserTest.kt`).
By the way, for KDoc there is the [different behaviour](https://github.com/JetBrains/kotlin/blob/92d200e093c693b3c06e53a39e0b0973b84c7ec5/compiler/psi/src/org/jetbrains/kotlin/kdoc/parser/KDocKnownTag.kt#L42)

qwwdfsad:
Nice one, filed https://github.com/Kotlin/dokka/issues/2929

PR #2908, opened 2023-03-08
Move the Properties block to be above Functions

IgnatBeresnev:
* It is more consistent with the declaration order described in our [coding conventions](https://kotlinlang.org/docs/coding-conventions.html#class-layout)
* There are usually less properties than functions, so it should help with not having to scroll as much to get to them

Current (that is, before this change) behaviour could be seen here: [kotlinx-datetime#Instant](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-instant/)

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -90,8 +90,8 @@ open class HtmlRenderer(
                     listOf(
                         BasicTabbedContentType.CONSTRUCTOR,
                         BasicTabbedContentType.TYPE,
-                        BasicTabbedContentType.FUNCTION,
-                        BasicTabbedContentType.PROPERTY
+                        BasicTabbedContentType.PROPERTY,
+                        BasicTabbedContentType.FUNCTION
                     )
                 ),
             if (extensions.isEmpty()) null else ContentTab(
@@ -99,8 +99,8 @@ open class HtmlRenderer(
                 listOf(
                     BasicTabbedContentType.CONSTRUCTOR,
                     BasicTabbedContentType.TYPE,
-                    BasicTabbedContentType.FUNCTION,
                     BasicTabbedContentType.PROPERTY,
+                    BasicTabbedContentType.FUNCTION,
                     BasicTabbedContentType.EXTENSION_PROPERTY,
                     BasicTabbedContentType.EXTENSION_FUNCTION
                 )
--- a/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
+++ b/plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
@@ -318,10 +318,10 @@ open class DefaultPageCreator(
                 "Inherited functions", inheritedFunctions + inheritedExtensionFunctions
             )
         } else {
-            functionsBlock("Functions", functions + extensionFuns)
             propertiesBlock(
                 "Properties", properties + extensionProps
             )
+            functionsBlock("Functions", functions + extensionFuns)
         }
     }
 
--- a/plugins/base/src/test/kotlin/renderers/html/TabbedContentTest.kt
+++ b/plugins/base/src/test/kotlin/renderers/html/TabbedContentTest.kt
@@ -142,4 +142,40 @@ class TabbedContentTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Test
+    fun `should have expected order of content types within a members tab`() {
+        val source = """
+            |/src/main/kotlin/test/Result.kt
+            |package example
+            |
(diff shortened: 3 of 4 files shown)


qwwdfsad:
Could you please show how documentation (any lib will work, not necessarily datetime) looks after the change?

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -90,17 +90,17 @@ open class HtmlRenderer(
                     listOf(
                         BasicTabbedContentType.CONSTRUCTOR,
                         BasicTabbedContentType.TYPE,
-                        BasicTabbedContentType.FUNCTION,
-                        BasicTabbedContentType.PROPERTY
+                        BasicTabbedContentType.PROPERTY,
+                        BasicTabbedContentType.FUNCTION
```
IgnatBeresnev, line 94: Changing the order here doesn't actually do anything, so it's just for consistency

IgnatBeresnev:
> Could you please show how documentation (any lib will work, not necessarily datetime) looks after the change?

If you want a live demo to be able to click around:

1. Open up the GitHub actions dropdown of this PR, you'll see one of `Build example projects and publish them to s3`. Go to details
2. Open up `Print link`:



But the link is dynamic based on the latest commit, and one branch can only have one live preview, so it's not permanent. These previews are also published to GitHub artifacts as zip archives.

If the goal is to preserve context / improve description, I can certainly make a screenshot or upload the live preview to a more permanent hosting, just let me know

___

Had to change the order of the actual blocks, updated and force-pushed the branch

IgnatBeresnev:
With Kotlin as Java applied, it's rather strange to first see the functions and then the properties

vmishenev (approved):

qwwdfsad:
I've skimmed through the generated documentation, few things caught my eye (I'm not sure they are relevant to this PR, though, so feel free to proceed):

* https://dokka-snapshots.s3.eu-central-1.amazonaws.com/move-properties-above-functions/coroutines/08f77ef/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/index.html shared flow link is broken (I have to press on `common` selector)
* https://dokka-snapshots.s3.eu-central-1.amazonaws.com/move-properties-above-functions/coroutines/08f77ef/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html a bit weird empty "properties" tab

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -90,17 +90,17 @@ open class HtmlRenderer(
...
                     )
                 ),
             if (extensions.isEmpty()) null else ContentTab(
                 "Members & Extensions",
                 listOf(
                     BasicTabbedContentType.CONSTRUCTOR,
                     BasicTabbedContentType.TYPE,
-                    BasicTabbedContentType.FUNCTION,
                     BasicTabbedContentType.PROPERTY,
+                    BasicTabbedContentType.FUNCTION,
```
qwwdfsad, line 103: Would be nice to drop a comment about the order's meaning and why the order is the way it is

qwwdfsad (approved):
Semantically, the change looks good (better and more natural than it was 👍).

It would be nice to have some tests on it or documentation though, so any accidental order change will lead to a build failure

IgnatBeresnev:
> I've skimmed through the generated documentation, few things caught my eye (I'm not sure they are relevant to this PR, though, so feel free to proceed):
>
> https://dokka-snapshots.s3.eu-central-1.amazonaws.com/move-properties-above-functions/coroutines/08f77ef/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/index.html shared flow link is broken (I have to press on common selector)
> https://dokka-snapshots.s3.eu-central-1.amazonaws.com/move-properties-above-functions/coroutines/08f77ef/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/index.html a bit weird empty "properties" tab

I've looked at the links and I don't see what's wrong :( Maybe there's something wrong with the cached styles or something?

The "properties" tab is not empty for me

<details>
<summary>screenshot</summary>



</details>

and all links work for me, not sure which one exactly you mean and how pressing the "common" selector can help with it :(

I'll try to remember to bring it up during one of the syncs, but I might forget

plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
IgnatBeresnev: Unfortunately, here the order doesn't make any difference, even though I expected it would :( Here it's purely cosmetic and I made the change for consistency. Something tells me leaving a comment like that here could be confusing down the road..

I'll write the tests for the true order though

PR #2909, opened 2023-03-08
Remove unused Gradle tasks

aSemy:
This PR removes some defined, but unused, Gradle tasks

These custom `sourceJar` tasks are not used. A `sourcesJar` is auto-created using `withSourcesJar()`

https://github.com/Kotlin/dokka/blob/6b46d95bf31a15952c07e8072bc134bbbf02ca5f/buildSrc/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts#L20-L22


IgnatBeresnev (approved):
:+1: :+1:

PR #2910, opened 2023-03-08
tidy up .gitignore files

aSemy:
* Removed unnecessary ignores for
  * Maven
  * Android
  * `.idea/` subdirectories (`.idea/` is already ignored)
  * old Gradle files (there's no Gradle UI any more, or task-cache file)
* Removed ignores for project files that don't exist any more

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,21 +1,6 @@
-*.ipr
-*.iml
 
-### Maven template
-target/
-pom.xml.tag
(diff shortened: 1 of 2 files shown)


IgnatBeresnev (approved):
Must've been a template, with things added to the end as time went on. I agree that the less unnecessary/confusing things, the better :+1:  Thanks!

PR #2911, opened 2023-03-08
Refactor Maven Runner build config

aSemy:
Part of #2703 

Updates the Maven runner build logic to work better with the Gradle API

* Removes the explicit cross-project dependency that was used to download the Maven exe twice
* Creates a convention plugin for downloading the Maven exe
* Removes eager downloading of the Maven exe
* Implements normalization (removes a timestamp) of a Maven property file
* Some minor build config tweaks to be a little more succint/Gradley/use newer functionality.


buildSrc/src/main/kotlin/org/jetbrains/SetupMaven.kt
```diff
@@ -1,45 +0,0 @@
...
-import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.Internal
-import org.gradle.api.tasks.Sync
-import org.gradle.kotlin.dsl.creating
-import org.gradle.kotlin.dsl.dependencies
-import org.gradle.kotlin.dsl.getValue
-import java.io.File
-
-@Suppress("LeakingThis")
-open class SetupMaven : Sync() {
```
aSemy, line 13: The properties in this file were moved to an extension in the new `maven-cli-setup.gradle.kts` convention plugin

buildSrc/src/main/kotlin/org/jetbrains/SetupMaven.kt
```diff
@@ -1,45 +0,0 @@
...
-                group = "org.apache.maven",
-                name = "apache-maven",
-                version = mavenVersion,
-                classifier = "bin", ext = "zip"
-            )
-        }
-    }
-
-    init {
-        from(mavenBinaryConfiguration.map { file -> project.zipTree(file) })
```
aSemy, line 41: This line caused eager configuration, so the binary was always downloaded whether it was necessary or not.

integration-tests/maven/build.gradle.kts
```diff
@@ -16,10 +14,13 @@ dependencies {
 tasks.integrationTest {
     dependsOnMavenLocalPublication()
 
-    val setupMavenTask = project(":runners:maven-plugin").tasks.withType<SetupMaven>().single()
-    dependsOn(setupMavenTask)
```
aSemy, line 20: Instead of using a cross-project dependency to re-use the Maven exec, just download it again.

Gradle will cache the downloaded dependency, so it's fine to set it up twice in different subprojects.

runners/maven-plugin/build.gradle.kts
```diff
@@ -1,95 +1,101 @@
...
+        rename { it.replace(".tpl.xml", ".xml") }
+
+        expand(
+            "mavenVersion" to setupMavenProperties.mavenVersion,
+            "dokka_version" to dokka_version,
+            "mavenPluginToolsVersion" to setupMavenProperties.mavenPluginToolsVersion,
+        )
     }
+
+    into(temporaryDir)
```
aSemy, line 41: Instead of worrying about Sync/Copy tasks overwriting, and having to use `presever {}`, I think it's best to always use Sync into the tasks temporary dir, and use final Sync task (`prepareMavenPluginBuildDir`) to sync all the files in one go.

runners/maven-plugin/build.gradle.kts
```diff
@@ -1,95 +1,101 @@
...
+                val lines = pluginHelpProperties.readText().lines().iterator()
+                // the first line is a descriptive comment
+                appendReproducibleNewLine(lines.next())
+                // the second line is the timestamp, which should be ignored
+                lines.next()
+                // the remaining lines are properties
+                lines.forEach { appendReproducibleNewLine(it) }
+            }
+        )
+    }
```
aSemy, line 81: Unfortunately the generated `maven-plugin-help.properties` contains a timestamp, so Gradle will always consider the output out-of-date.

I wrote this code to fix it - but maybe there's a Maven flag that will disable the timestamp?

This is perhaps a bit OTT/pre-optimising.

runners/maven-plugin/build.gradle.kts
```diff
@@ -1,95 +1,101 @@
...
+    val pomTemplateFile = layout.projectDirectory.file("pom.tpl.xml")
+
+    from(pomTemplateFile) {
+        rename { it.replace(".tpl.xml", ".xml") }
+
+        expand(
+            "mavenVersion" to setupMavenProperties.mavenVersion,
+            "dokka_version" to dokka_version,
+            "mavenPluginToolsVersion" to setupMavenProperties.mavenPluginToolsVersion,
+        )
```
aSemy, line 38: using the Gradle `expand ()` is a little nicer to use and look at than the previous manual find and replace.

To achieve this, I needed to update the properties in `pom.tpl.xml` to be of the format `${mavenVersion}`

runners/maven-plugin/build.gradle.kts
IgnatBeresnev: I actually don't understand what helpMojo is supposed to achieve, and I can't google anything sensible. It looks like it should either generate the `help` goal (which Dokka's Maven plugin doesn't have) or maybe add some help documentation? Were you able to figure it out?

Nothing looks broken to me, so I think it's fine to ignore the timestamp. Perhaps, at some point our Maven plugin should be revisited as well :sweat_smile:

buildSrc/src/main/kotlin/org/jetbrains/conventions/maven-cli-setup.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+ * Utility for downloading and installing a Maven binary.
+ */
+
+plugins {
+    base
+}
+
+abstract class SetupMavenProperties {
+    val mavenVersion = "3.5.0"
+    val mavenPluginToolsVersion = "3.5.2"
```
IgnatBeresnev, line 15: Very much optional, while we're at it: I think we actually forget to check/update the Maven version because it's buried in a place we don't visit that much. Extracting it to `gradle.properties` (or into the `toml` in the other PR) would not hurt

IgnatBeresnev (approved):
Looks very neat, and thank you for the explanation comments! Only have optional nitpicks

I'll start the integration tests and, if everything's good, will merge it

runners/maven-plugin/build.gradle.kts
aSemy: I think it auto generates a 'help' command based on the properties in the Maven plugin

The file it generates is in `runners/maven-plugin/build/maven/generated-sources`

But, like you, I'm not sure! I've never developed a Maven plugin.

```java

package org.jetbrains.dokka.maven;

/**
 * Display help information on dokka-maven-plugin.<br>
 * Call <code>mvn dokka:help -Ddetail=true -Dgoal=&lt;goal-name&gt;</code> to display parameter details.
 * @author maven-plugin-tools
 * @goal help
 * @requiresProject false
 * @threadSafe
 */
public class HelpMojo
    extends AbstractMojo
{
    /**
     * If <code>true</code>, display all settable properties for each goal.
     *
     * @parameter property="detail" default-value="false"
     */
    private boolean detail;

    /**
     * The name of the goal for which to show help. If unspecified, all goals will be displayed.
     *
     * @parameter property="goal"
     */
    private java.lang.String goal;

    /**
     * The maximum length of a display line, should be positive.
     *
     * @parameter property="lineLength" default-value="80"
     */
    private int lineLength;

...
```

I was looking at other options, and I found this Gradle plugin: https://github.com/britter/maven-plugin-development. It might help with developing a Maven plugin, but on the other hand, the current approach seems to be working.

runners/maven-plugin/build.gradle.kts
```diff
@@ -1,95 +1,101 @@
...
-        filter { line ->
-            line.replace("<version>dokka_version</version>", "<version>$dokka_version</version>")
-        }
-        filter { line ->
-            line.replace(
-                "<version>maven-plugin-plugin</version>",
-                "<version>${setupMaven.mavenPluginToolsVersion}</version>"
-            )
-        }
+    val pomTemplateFile = layout.projectDirectory.file("pom.tpl.xml")
```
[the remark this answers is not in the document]
aSemy: Good idea, done.

integration-tests/maven/build.gradle.kts
```diff
@@ -16,10 +14,13 @@ dependencies {
...
+    dependsOn(tasks.installMavenBinary)
+    val mvn = setupMavenProperties.mvn
+    inputs.file(mvn)
 
     val dokka_version: String by project
     environment("DOKKA_VERSION", dokka_version)
-    environment("MVN_BINARY_PATH", setupMavenTask.mvn.absolutePath)
+    doFirst {
```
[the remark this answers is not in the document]
aSemy: Ah yes, I forgot to put a comment here. I wanted to find the Gradle issue and put a link in there...


I updated the `mvn` exec file location property to be a [Gradle file property](https://docs.gradle.org/current/userguide/lazy_configuration.html#working_with_files_in_lazy_properties) (which implements the `Provider<T>` API). 

`Exec.environment()` hasn't been updated to be compatible with the [Gradle Provider API](https://docs.gradle.org/current/userguide/lazy_configuration.html#lazy_properties). Providers shouldn't be computed (using `.get()`) until the task is running.

As a workaround, I put the `.get()` in a `doFirst {}`. 

In this case, I am perhaps being a bit too strict. The `mvn` property is static and won't change, so it doesn't have to be a Gradle file property (it could just be a `java.io.File`), or it's fine if in this instance it's computed early by calling `.get()` in the task-config block.

```kotlin
tasks.integrationTest {
    // call .get() in the configuration-phase
    environment("MVN_BINARY_PATH", mvn.get().asFile.invariantSeparatorsPath)
}
```

buildSrc/src/main/kotlin/org/jetbrains/conventions/maven-cli-setup.gradle.kts
aSemy: Good point, I think so too. For now I'll put it in `gradle.properties`, and I think that later they can be moved to `libs.versions.toml` as part of #2884

aSemy: It was a little more work than I thought to refactor this, but I think it was worth it.

It revealed that the `CrossPlatformExec` task wasn't really needed, and could be replaced with a Provider that would determine the current operating system and use `mvn.cmd` on Windows, else use `mvn`.

IgnatBeresnev: Oof, sorry for asking you to do that then, and thanks for doing! I do think it was worth it though!

IgnatBeresnev:
After `master` was merged in, the `maven-cli` convention was needed to be moved into `build-logic`. I hope you don't mind I pushed it into your branch

I'll start integration tests once again, will merge it once they pass :+1: :+1:

aSemy:
> After `master` was merged in, the `maven-cli` convention was needed to be moved into `build-logic`. I hope you don't mind I pushed it into your branch

Not at all! Good job noticing it, and thanks for fixing it.

buildSrc/src/main/kotlin/org/jetbrains/conventions/maven-cli-setup.gradle.kts
aSemy: no apology needed :)

PR #2912, opened 2023-03-09
Migrate buildSrc to composite build

Goooler:
Addressed from #2652.


Goooler:
- `dokka-gradle-plugin` project applies convention plugins, need to convert it to a composite build too.
- `dokka-gradle-plugin` refs some sub-projects in root project, need to declare them in `dokka-gradle-plugin`'s `settings.gradle.kts`, it also refs some versions.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
```
IgnatBeresnev, line 3: I was under the impression that there were problems with upgrading to Gradle 8? Does it work OK now? If it doesn't and the update requires some additional changes, I'd prefer to have it in the separate PR (#2860) to not drag this one out

runners/gradle-plugin/settings.gradle.kts
```diff
@@ -0,0 +1,35 @@
+rootProject.name = "dokka-gradle-plugin"
+
+include(
```
IgnatBeresnev, line 3: I wonder why this is needed - shouldn't the Gradle plugin be treated the same way as any other module/project, or is it different for some reason?

IgnatBeresnev:
@aSemy could I ask you to take a quick look, please? Maybe you know a way to fix the gradle plugin without copying configuration

aSemy:
I think converting the Dokka-GP to an included build would be a good step in the future (as well as the integration tests), but it might require a lot of work (e.g. the Worker API), and isn't necessary right now (unless I've missed something @Goooler?). 

I think this PR should just set up the bare minimum needed for using included build, and focus on improving the Dokka-GP in another PR. So I think just keep these changes:

* Rename `buildSrc` to `build-logic`
* Rename the convention plugins to use the `id("org.jetbrains.base")` format
* add `includeBuild()` into the root `settings.gradle.kts`

And don't update to Gradle 8, and keep Dokka-GP as regular ol' subproject.

WDYT?

gradle/wrapper/gradle-wrapper.properties
Goooler: Yep, but 5f66e0ddd1ef741333a73d389e5fe03d4d0fa7bd need this.

runners/gradle-plugin/settings.gradle.kts
Goooler: Cause `dokka-gradle-plugin` depends on `build-logic`, but root project depends on `dokka-gradle-plugin`.

Goooler:
@aSemy Cause `dokka-gradle-plugin` depends on `build-logic`, but root project depends on `dokka-gradle-plugin`.

aSemy:
> @aSemy Cause `dokka-gradle-plugin` depends on `build-logic`, but root project depends on `dokka-gradle-plugin`.
> 
> We'd better to use online version of Dokka in root?
> 
> https://github.com/Kotlin/dokka/blob/6fcd66a956e489ee42045956164945f080ae772d/build.gradle.kts#L4-L6

Ahh right, I see what you mean now.

Yes, I think it's better to use the online version of Dokka. If there's problem in DokkaGP then it could prevent the whole project from compiling during development, which would be annoying.

build.gradle.kts
```diff
@@ -2,11 +2,11 @@ import org.jetbrains.ValidatePublications
...
-    org.jetbrains.conventions.base
-    id("org.jetbrains.dokka")
-    id("io.github.gradle-nexus.publish-plugin")
-
-    id("org.jetbrains.kotlinx.binary-compatibility-validator")
+    id("org.jetbrains.conventions.base")
+    id("org.jetbrains.dokka") version "1.8.10"
+    id("io.github.gradle-nexus.publish-plugin") version "1.1.0"
+    id("com.gradle.plugin-publish") version "0.20.0"
+    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.12.1"
```
IgnatBeresnev, line 9: It looks like the versions are duplicated here and in `build-logic/build.gradle.kts` now? Is it possible to have them in one place only?

IgnatBeresnev (approved):
Well done, thanks! I've triggered integration tests, let's wait for them to pass

build.gradle.kts
Goooler: Seems can't, these plugins are used in `rootProject`.

IgnatBeresnev:
Everything looks good :tada:

PR #2915, opened 2023-03-12
Use info level for progress in Dokka Gradle plugin

### LINKED ISSUE #1894: Silence logging when running dokkaHtml
**Describe the bug**

`dokkaHtml` produces the following messages with the default logging levels (e.g. lifecycle)

```
> Task :...:dokkaJavadoc
Initializing plugins
Dokka is performing: documentation for allure-base-plugin
Validity check
Creating documentation models
Merging documentation models
Creating pages
Transforming pages
Rendering
```

**Expected behaviour**

The debugging messages should come with `debug` level or something.
Frankly speaking, it is hard to understand what should the end-users do with the above messages.

**Dokka configuration**

```Kotlin
plugins {
    id("org.jetbrains.dokka") version "1.4.32"
}

    outputDirectory.set(buildDir.resolve("javadoc/$name"))
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v6.8.3
- Dokka version: 1.4.32

eirnym:
fixes #1894

Level for progress set to `info`. See the issue for more details

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
@@ -220,7 +220,7 @@ abstract class AbstractDokkaTask : DefaultTask() {
         when (level) {
             "debug" -> logger.debug(message)
             "info" -> logger.info(message)
-            "progress" -> logger.lifecycle(message)
+            "progress" -> logger.info(message)
             "warn" -> logger.warn(message)
             "error" -> logger.error(message)
         }


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
```diff
@@ -220,7 +220,7 @@ abstract class AbstractDokkaTask : DefaultTask() {
         when (level) {
             "debug" -> logger.debug(message)
             "info" -> logger.info(message)
-            "progress" -> logger.lifecycle(message)
+            "progress" -> logger.info(message)
```
IgnatBeresnev, line 223: Gradle's `lifecycle` level [(described as `Progress information messages`)](https://docs.gradle.org/current/userguide/logging.html#logLevels) is the ideal mapping option for Dokka's `progress` level, so I don't think this change is right, sorry

IgnatBeresnev:
Unfortunately, the mentioned problem is not as easy to fix. I might be wrong, but it looks like the proposed solution would silence Dokka overall, no messages will be emitted other than warnings and errors, so for the users it'll look like Dokka is broken all of a sudden.

A proper solution would require revisiting all messages that are logged by Dokka on all levels, and it's also not as easy to decide which messages need to be re-written, which need another level and so on. 

If you really want to help with it, it'll take time and we'll need to agree on the messages beforehand so that you don't waste time implementing something we wouldn't be able to accept. Either way, I'll close this PR as it's not really addressing the problem, and we can contibue the discussion on what needs to be done in the issue itself

Thank you for wanting to contribute though, we appreciate it

eirnym:
I haven't touch warnings and errors, also `--info` flag will output all messages

PR #2916, opened 2023-03-12
nitpick: Gradle examples tidy up

aSemy:
remove unnecessary brackets from Dokka plugin version, so the Gradle examples are more idiomatic.


IgnatBeresnev (approved):
:+1: :+1:

PR #2918, opened 2023-03-13
Fix test runtime Java versions

### LINKED ISSUE #2917: Regression: All versions of Java are not tested in CI/CD
> If we apply `toolchain` for all modules, they'll run tests on Java 8 instead of the JDK matrix on CI, we need to apply more toolchains in `Test` which will slow  down the build, I believe we should not use `toolchain` for now. Check
> 
> - https://github.com/detekt/detekt/pull/5735
> - https://github.com/Kotlin/dokka/blob/f5581c51d0de23d2b038e00b1a152faebaef0920/.github/workflows/gradle-test.pr.yml#L12-L13
> 
> _Originally posted by @Goooler in https://github.com/Kotlin/dokka/pull/2704#discussion_r1133571275_

aSemy:
fix #2917

This PR sets the [Gradle JVM Toolchains](https://docs.gradle.org/current/userguide/toolchains.html) used to run the tests.

Additionally it introduces a workaround for the JUnit test runner so all JUnit 4 and 5 tests will run - to be resolved in #2924 

### New properties.

It introduces two new Gradle [Project properties](https://docs.gradle.org/current/userguide/build_environment.html#sec:project_properties)

* `org.jetbrains.dokka.javaToolchain.mainCompiler` - the version of Java used to compile Dokka
* `org.jetbrains.dokka.javaToolchain.testLauncher` - the version of Java used to run tests

It also renames the existing [`language_level` property](https://github.com/Kotlin/dokka/blob/6fcd66a956e489ee42045956164945f080ae772d/gradle.properties#L11) to `org.jetbrains.dokka.kotlinLanguageLevel`, for consistency.

When developing locally, these values have defaults set in the root `gradle.properties`. The `testLauncher` property is overridden in CI/CD.

### Changes

- Create `DokkaBuildProperties` extension for convenient sharing values used to configure the Dokka build (only contains 3 values now, but this will be expanded later). 
  
   This will be very convenient for sharing properties when the build is updated to be a composite #2912.
- update `base-java` convention to [set the test Java launcher](https://docs.gradle.org/current/userguide/toolchains.html#toolchains_for_tasks)
- update GitHub workflow to set the java-version via `DokkaBuildProperties` and the [`ORG_GRADLE_PROJECT_`](https://docs.gradle.org/current/userguide/build_environment.html#sec:project_properties) override

### TODO

- [x] ~~Update TeamCity to use the new variables (if necessary?)~~ Not necessary, as per https://github.com/Kotlin/dokka/pull/2918#issuecomment-1471193510
   - [x] `org.jetbrains.dokka.javaToolchain.mainCompiler`
   - [x] `org.jetbrains.dokka.javaToolchain.testLauncher`
   - [x] `org.jetbrains.dokka.kotlinLanguageLevel`


aSemy:
@Goooler WDYT?

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
```diff
@@ -0,0 +1,53 @@
...
+    val mainJavaVersion: Provider<JavaLanguageVersion> =
+        dokkaProperty("javaToolchain.mainCompiler") { JavaLanguageVersion.of(it) }
+
+    /**
+     * The version of Java that should be used to run Dokka tests.
+     *
+     * This value is set in CI/CD environments to make sure that Dokka still works with different
+     * versions of Java.
+     */
+    val testJavaLauncherVErsion: Provider<JavaLanguageVersion> =
```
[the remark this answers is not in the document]
Goooler: And we should use Java 8 as the default value here if no property exist.

buildSrc/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts
```diff
@@ -8,15 +8,22 @@ package org.jetbrains.conventions
...
     }
 }
 
 java {
     withSourcesJar()
 }
+
+tasks.withType<Test>().configureEach {
+    javaLauncher.set(javaToolchains.launcherFor {
+        languageVersion.set(dokkaBuild.testJavaLauncherVErsion)
```
aSemy, line 27:
```suggestion
        languageVersion.set(dokkaBuild.testJavaLauncherVersion)
```

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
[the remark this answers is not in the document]
aSemy: Great idea, done. Thanks!

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
```diff
@@ -0,0 +1,53 @@
...
+     * The Kotlin language level that Dokka artifacts are compiled to support.
+     *
+     * Updating the language level is a breaking change.
+     */
+    val kotlinLanguageLevel: Provider<KotlinVersion> =
+        dokkaProperty("kotlinLanguageLevel") { KotlinVersion.fromVersion(it) }
+
+
+    private fun <T : Any> dokkaProperty(name: String, convert: (String) -> T) =
+        providers.gradleProperty("org.jetbrains.dokka.$name").map(convert)
```
Goooler, line 48:
```suggestion
    private fun <T : Any> dokkaProperty(name: String, transformer: (String?) -> T) =
        providers.gradleProperty("org.jetbrains.dokka.$name").map(transformer)
```

Goooler: `String` is nullable here?

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
```diff
@@ -0,0 +1,53 @@
...
+        dokkaProperty("javaToolchain.mainCompiler") { JavaLanguageVersion.of(it) }
+
+    /**
+     * The version of Java that should be used to run Dokka tests.
+     *
+     * This value is set in CI/CD environments to make sure that Dokka still works with different
+     * versions of Java.
+     */
+    val testJavaLauncherVersion: Provider<JavaLanguageVersion> =
+        dokkaProperty("javaToolchain.testLauncher") { JavaLanguageVersion.of(it) }
```
Goooler, line 36:
```suggestion
    val testJavaLauncherVersion: Provider<JavaLanguageVersion> =
        dokkaProperty("javaToolchain.testLauncher") { JavaLanguageVersion.of(it ?: "8") }
```

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
aSemy: No, it's not. If the Gradle property doesn't exist, then the map doesn't get called. 

For example, if `does-not-exist` is not defined, then the `map {}` does not get called.

```kotlin
val prop = providers.gradleProperty("does-not-exist").map { error("failed to map $it") } 

println("prop is ${prop.orNull}") // Output: prop is null
println("prop is ${prop.get()}")  // Error: Cannot query the value of this provider because it has no value available.
```

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
aSemy: I think it's best to have one place where the default value is defined, and that's in the root `gradle.properties` file. It could be annoying if someone updates it in one place, and doesn't realise it needs to also be updated in another.

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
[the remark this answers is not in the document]
aSemy: Oops, yes I forgot to push.

https://github.com/Kotlin/dokka/blob/f27fa8c2d467aa1040e3196bbc540aac01153f73/buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt#L35-L37

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
Goooler: Gotcha!

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
Goooler: Make sense.

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val providers: ProviderFactory,
+) {
+
+    /**
+     * The main version of Java that should be used to build Dokka source code.
+     *
+     * Updating the Java target is a breaking change.
+     */
+    val mainJavaVersion: Provider<JavaLanguageVersion> =
+        dokkaProperty("javaToolchain.mainCompiler") { JavaLanguageVersion.of(it) }
```
Goooler, line 27:
```suggestion
        dokkaProperty("javaToolchain.mainCompiler", JavaLanguageVersion::of)
```

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * The version of Java that should be used to run Dokka tests.
+     *
+     * This value is set in CI/CD environments to make sure that Dokka still works with different
+     * versions of Java.
+     */
+    val testJavaLauncherVersion: Provider<JavaLanguageVersion> =
+        dokkaProperty("javaToolchain.testLauncher") { JavaLanguageVersion.of(it) }
```
Goooler, line 36:
```suggestion
        dokkaProperty("javaToolchain.testLauncher", JavaLanguageVersion::of)
```

buildSrc/src/main/kotlin/org/jetbrains/DokkaBuildProperties.kt
```diff
@@ -0,0 +1,54 @@
...
+            .orElse(mainJavaVersion)
+
+    /**
+     * The Kotlin language level that Dokka artifacts are compiled to support.
+     *
+     * Updating the language level is a breaking change.
+     */
+    val kotlinLanguageLevel: Provider<KotlinVersion> =
+        dokkaProperty("kotlinLanguageLevel") { KotlinVersion.fromVersion(it) }
```
Goooler, line 45:
```suggestion
        dokkaProperty("kotlinLanguageLevel", KotlinVersion::fromVersion)
```

buildSrc/src/main/kotlin/org/jetbrains/conventions/base.gradle.kts
```diff
@@ -1,10 +1,20 @@
...
+ */
+
 plugins {
     base
 }
 
-// common Gradle configuration that should be applied to all projects
+val dokkaBuildProperties =
+    extensions.create<DokkaBuildProperties>(DokkaBuildProperties.EXTENSION_NAME)
+
```
Goooler, line 17:
```suggestion
val dokkaBuildProperties: DokkaBuildProperties = extensions.create(DokkaBuildProperties.EXTENSION_NAME)

```

buildSrc/src/main/kotlin/org/jetbrains/internal/gradleKotlinDslAccessors.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Retrieves the [dokkaBuild][org.jetbrains.DokkaBuildProperties] extension.
+ */
+internal val Project.dokkaBuild: DokkaBuildProperties
+    get() = extensions.getByType<DokkaBuildProperties>()
+
+/**
+ * Configures the [dokkaBuild][org.jetbrains.DokkaBuildProperties] extension.
+ */
+internal fun Project.dokkaBuild(configure: DokkaBuildProperties.() -> Unit) =
+    extensions.configure<DokkaBuildProperties>(configure)
```
Goooler, line 27:
```suggestion
/**
 * Retrieves the [dokkaBuild][org.jetbrains.DokkaBuildProperties] extension.
 */
internal val Project.dokkaBuild: DokkaBuildProperties
    get() = extensions.getByType()

/**
 * Configures the [dokkaBuild][org.jetbrains.DokkaBuildProperties] extension.
 */
internal fun Project.dokkaBuild(configure: DokkaBuildProperties.() -> Unit) =
    extensions.configure(configure)
```

IgnatBeresnev (approved):
Nice solution! Having build properties should come in handy :+1: And thanks to @Goooler for taking a look as well

I'll start teamcity-based integration tests, and merge it tomorrow during the day

IgnatBeresnev:
Non-blocking nitpick, will merge without it, but do let me know if you want to address it in this PR: maybe [the fix](https://github.com/Kotlin/dokka/blob/3ea414251d32108dab8b14d41a672e9834824690/integration-tests/gradle/build.gradle.kts#L23) we put in for the integration tests could be revisited? Is there a better option after current changes, like setting the new build property value to `11` and letting the convention plugin do the configuration part?

> Update TeamCity to use the new variables (if necessary?)

Doesn't look like it needs to be updated - they only run the `test` task, and have the same JVM version as the default one, so it should be good. The matrix thing is GitHub Actions only

IgnatBeresnev:
Oops, it looks like the majority of unit tests are not run at all :sweat_smile: 

Most likely [this configuration](https://github.com/Kotlin/dokka/commit/62c980707d23d2d451c75330f7ed1ba613777cf3#diff-c7867af7c6e16986ceedd11c74313cdb1f6c2a6976b7c01c5cfd9968d0f2c8feL20) was forgotten to be moved to the convention plugins.

Could you fix it in this or in a separate PR, please?

aSemy:
> Oops, it looks like the majority of unit tests are not run at all 😅
> 
> Most likely [this configuration](https://github.com/Kotlin/dokka/commit/62c980707d23d2d451c75330f7ed1ba613777cf3#diff-c7867af7c6e16986ceedd11c74313cdb1f6c2a6976b7c01c5cfd9968d0f2c8feL20) was forgotten to be moved to the convention plugins.

I think that link doesn't work, it just links to the whole PR. Which bit needs to be fixed?

IgnatBeresnev:
> I think that link doesn't work, it just links to the whole PR. Which bit needs to be fixed?

Oh, sorry. In the now deleted `plugins/build.gradle.kts` there used to be configuration for `tasks.test` with `useJUnitPlatform()` - I think that's what made the tests run

The only thing I didn't understand was how tests were run in other modules like core - found no configuration for tests whatsoever

integration-tests/gradle/build.gradle.kts
```diff
@@ -22,6 +22,8 @@ tasks.integrationTest {
 
     javaLauncher.set(javaToolchains.launcherFor {
         // kotlinx.coroutines requires Java 11+
-        languageVersion.set(JavaLanguageVersion.of(11))
+        languageVersion.set(dokkaBuild.testJavaLauncherVersion.map {
+            maxOf(it, JavaLanguageVersion.of(11))
+        })
     })
 }
```
aSemy, line 29: Following on from https://github.com/Kotlin/dokka/pull/2918#issuecomment-1471193510 @IgnatBeresnev 

> Non-blocking nitpick, will merge without it, but do let me know if you want to address it in this PR: maybe [the fix](https://github.com/Kotlin/dokka/blob/3ea414251d32108dab8b14d41a672e9834824690/integration-tests/gradle/build.gradle.kts#L23) we put in for the integration tests could be revisited? Is there a better option after current changes, like setting the new build property value to `11` and letting the convention plugin do the configuration part? GitHub actions configuration for them could be corrected as well it seems
> 
> > Update TeamCity to use the new variables (if necessary?)
> 
> Doesn't look like it needs to be updated - they only run the `test` task, and have the same JVM version as the default one, so it should be good. The matrix thing is GitHub Actions only

Good hint! I had an idea, so hopefully at least Java 11 and 17 will be integration tested.

Unfortunately because all the integration tests are lumped together, they need to use the same Java launcher, which isn't ideal.

Potential solutions

* update the Coroutines build config to support toolchains https://github.com/Kotlin/kotlinx.coroutines/issues/3665
* add unit-test tags for each test project, and create separate Gradle test tasks for each tag, so the launcher can be specified per project
* use JUnit conditionals to only run KxS on Java 11+ (JUnit 5 docs: https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-jre)

build-logic/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts
```diff
@@ -8,15 +8,24 @@ package org.jetbrains.conventions
...
+        languageVersion.set(dokkaBuild.mainJavaVersion)
     }
 }
 
 java {
     withSourcesJar()
 }
+
+tasks.withType<Test>().configureEach {
+    useJUnitPlatform()
```
aSemy, line 26: https://github.com/Kotlin/dokka/pull/2918#issuecomment-1471893538

> > I think that link doesn't work, it just links to the whole PR. Which bit needs to be fixed?
> 
> Oh, sorry. In the now deleted `plugins/build.gradle.kts` there used to be configuration for `tasks.test` with `useJUnitPlatform()` - I think that's what made the tests run

Very good catch! `useJUnitPlatform()` should have been moved to a convention plugin in #2704, but it was missed.

Hopefully this fix 'just works'. Some subprojects might need to override this with `useJUnit()`.


> The only thing I didn't understand was how tests were run in other modules like core - found no configuration for tests whatsoever

I _guess_ that `useJUnit()` is the default, and most projects just use plain kotlin.test.

I think all tests should be updated to use JUnit 5.

aSemy: nm, looks like `useJUnitPlatform()` isn't a good fit, a lot of subprojects need it, but should be revisited later. I'll set `useJUnit()` as the default for now.

integration-tests/gradle/build.gradle.kts
IgnatBeresnev: Setting the min version to 11 with the possibility to go above is good enough for now, I think :+1: 

We can keep this info in mind for when the integration test runner is refactored, maybe by that time kotlinx.coroutines will support toolchains or the integration tests themselves won't be as coupled

I should probably create an issue for it to not lose context, and quote your message and proposals

build-logic/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts
IgnatBeresnev: Created an issue for longer term refactoring: #2924

For now it'd be good to just run all the tests as before

aSemy: I've added the necessary JUnit runtime dependencies so JUnit Platform can run both JUnit 4 and 5 tests https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle-engines-configure

Locally I get more than 1130 test running

IgnatBeresnev:
Looks like unit tests have been fixed, thanks :tada: All 1097 of them

.github/workflows/gradle-test.pr.yml
```diff
@@ -10,32 +10,36 @@ jobs:
...
-        version: [ 8, 11, 17 ]
+        javaVersion: [ 8, 11, 17 ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: ${{ matrix.version }}
+          java-version: ${{ matrix.javaVersion }}
```
Goooler, line 20: Seems we can stick with Java 17 here.

Goooler: Addressing the fix to #2918.

PR #2921, opened 2023-03-14
Replace BiConsumer usages

Goooler:
Use Kotlin functions instead.

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -39,22 +39,22 @@ public final class org/jetbrains/dokka/DefaultExternalLinksKt {
 }
 
 public abstract interface class org/jetbrains/dokka/DokkaBootstrap {
-	public abstract fun configure (Ljava/lang/String;Ljava/util/function/BiConsumer;)V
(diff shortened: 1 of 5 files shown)


IgnatBeresnev:
I wish we could accept the PR, but as you can see the tests are failing with

```
    Caused by: java.lang.NoSuchMethodException: org.jetbrains.dokka.DokkaBootstrapImpl.configure(java.lang.String, kotlin.jvm.functions.Function2)
    	at org.jetbrains.dokka.gradle.DelegatedInvocationHandler.invoke(automagicTypedProxy.kt:37)
    	at com.sun.proxy.$Proxy150.configure(Unknown Source)
```

Unfortunately, there's no way to fix it :( I wrote out a longer explanation in https://github.com/Kotlin/dokka/pull/2875#issuecomment-1438827861

There might be ways to make it work, but I wouldn't waste much time on it, it's going to be removed in #2740 anyway

Sorry

Goooler:
Alright, thanks for your explination!

PR #2930, opened 2023-03-18
remove old Maven CLI setup tasks

aSemy:
Delete two files that were removed as part of #2911, but were accidentally re-introduced while merging in changes from #2912


IgnatBeresnev (approved):
:+1: :+1:

PR #2935, opened 2023-03-21
Capture throwable inside generateDocumentation thread

### LINKED ISSUE #2934: Gradle build successful even though there are Dokka failures
**Describe the bug**
Gradle build contains the following output:
```
> Task :dokkaHtml
Initializing plugins
Dokka is performing: documentation for Dokka Gradle Example
Validity check
Creating documentation models
Merging documentation models
Undocumented: demo/main/#kotlin.Array[kotlin.String]/ (jvm)
Undocumented: demo/Greeter/undocumentedFun/#/ (jvm)
Creating pages
Transforming pages
Rendering
Running post-actions



BUILD SUCCESSFUL in 3s
1 actionable task: 1 executed
Exception in thread "Thread-539" org.jetbrains.dokka.DokkaException: Failed with warningCount=2 and errorCount=0
	at org.jetbrains.dokka.base.generation.SingleModuleGeneration.reportAfterRendering(SingleModuleGeneration.kt:106)
	at org.jetbrains.dokka.base.generation.SingleModuleGeneration.generate(SingleModuleGeneration.kt:55)
	at org.jetbrains.dokka.DokkaGenerator$generate$1.invoke(DokkaGenerator.kt:29)
	at org.jetbrains.dokka.DokkaGenerator$generate$1.invoke(DokkaGenerator.kt:17)
	at org.jetbrains.dokka.DokkaGeneratorKt.timed(DokkaGenerator.kt:76)
	at org.jetbrains.dokka.DokkaGeneratorKt.access$timed(DokkaGenerator.kt:1)
	at org.jetbrains.dokka.DokkaGenerator.generate(DokkaGenerator.kt:22)
	at org.jetbrains.dokka.DokkaBootstrapImpl.generate(DokkaBootstrapImpl.kt:58)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.jetbrains.dokka.gradle.DelegatedInvocationHandler.invoke(automagicTypedProxy.kt:40)
	at com.sun.proxy.$Proxy191.generate(Unknown Source)
	at org.jetbrains.dokka.gradle.AbstractDokkaTask$generateDocumentation$1$1.run(AbstractDokkaTask.kt:208)
	at java.base/java.lang.Thread.run(Thread.java:829)
9.19.47: Execution finished 'dokkaHtml'.
```

**Expected behaviour**
Build should fail with following output:
```
> Task :dokkaHtml FAILED
1 actionable task: 1 executed

* What went wrong:
Execution failed for task ':dokkaHtml'.
> Failed with warningCount=2 and errorCount=0
```

**Screenshots**
N/A

**To Reproduce**
Take [dokka-gradle-example](https://github.com/Kotlin/dokka/tree/master/examples/gradle/dokka-gradle-example) and:
1. add a new undocumented function to class `HelloWorld`
2. configure dokka task to report undocumented and fail on warning

See required changes in the following diff:
```diff
diff --git a/build.gradle.kts b/build.gradle.kts
index 30cb4d7..6a47044 100644
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -32,6 +32,8 @@ tasks.withType<DokkaTask>().configureEach {
                 ))
                 remoteLineSuffix.set("#L")
             }
+            reportUndocumented.set(true)
+            failOnWarning.set(true)
         }
     }
 }
diff --git a/src/main/kotlin/demo/HelloWorld.kt b/src/main/kotlin/demo/HelloWorld.kt
index 172e18f..5e39c5d 100644
--- a/src/main/kotlin/demo/HelloWorld.kt
+++ b/src/main/kotlin/demo/HelloWorld.kt
@@ -13,6 +13,10 @@ class Greeter(val name: String) {
     fun greet() {
         println("Hello $name!")
     }
+
+    fun undocumentedFun() {
+        TODO()
+    }
 }

 fun main(args: Array<String>) {
```

**Dokka configuration**

```Kotlin
    dokkaSourceSets {
        named("main") {
            reportUndocumented.set(true)
            failOnWarning.set(true)
        }
    }
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v7.6
- Dokka version: 1.8.10

**Additional context**
The problem was most likely introduced in PR [#2678 file AbstractDokkaTask](https://github.com/Kotlin/dokka/pull/2678/files#diff-39c65c219caffa4e2317c82d1adbfbd566c5762e553d4162acba360c9d39041aR96-R99).
A separate thread is created but exceptions are not captured and propagated.

Yes

jush:
Capture any throwable thrown by the `generate()` function running in a separate thread to later on throw it in the context of the `generateDocumentation()` calling thread.

Fixes #2934

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
@@ -203,14 +203,22 @@ abstract class AbstractDokkaTask : DefaultTask() {
     internal open fun generateDocumentation() {
         DokkaBootstrap(runtime, DokkaBootstrapImpl::class).apply {
             configure(buildDokkaConfiguration().toCompactJsonString(), createProxyLogger())
+            var throwableInGenerate: Throwable? = null
             /**
              * Run in a new thread to avoid memory leaks that are related to ThreadLocal (that keeps `URLCLassLoader`)
              * Currently, all `ThreadLocal`s leaking are in the compiler/IDE codebase.
              */
-            Thread { generate() }.apply {
+            Thread {
+                try {
+                    generate()
+                } catch (t: Throwable) {
+                    throwableInGenerate = t
+                }
+            }.apply {
                 start()
                 join()
             }
+            throwableInGenerate?.let { throw it }
         }
     }
 


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
```diff
@@ -203,14 +203,22 @@ abstract class AbstractDokkaTask : DefaultTask() {
...
+                try {
+                    generate()
+                } catch (t: Throwable) {
+                    throwableInGenerate = t
+                }
+            }.apply {
                 start()
                 join()
             }
+            throwableInGenerate?.let { throw it }
```
IgnatBeresnev, line 221: I'm not sure why Kotlin allows writing such code and what exactly happens "under the hood" when this is run on a JVM. Is it thread safe? I can't say for sure...

I do know that you cannot write such code in Java, it requires the used variable to be final or either effectively final for good reasons.

<details>
<summary>screenshots</summary>




</details>

To be on the safe side, I think we can use `AtomicReference` to hold the throwable, then it should be thread safe. 

While at it, we could use `UncaughtExceptionHandler` (which actually catches `Throwable`) instead of the try-catch, but it's optional :)3

To put the words into code:

```kotlin
val uncaughtExceptionHolder = AtomicReference<Throwable?>()
/**
 * Run in a new thread to avoid memory leaks that are related to ThreadLocal (that keeps `URLCLassLoader`)
 * Currently, all `ThreadLocal`s leaking are in the compiler/IDE codebase.
 */
Thread { generate() }.apply {
    setUncaughtExceptionHandler { _, throwable ->
        uncaughtExceptionHolder.set(throwable)
    }
    start()
    join()
}
uncaughtExceptionHolder.get()?.let { throw it }
```

IgnatBeresnev (commented):
Wow, thanks for noticing and fixing it!

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
jush: Thanks a lot for looking into it and providing such a detailed guidance.

> I'm not sure why Kotlin allows writing such code and what exactly happens "under the hood" when this is run on a JVM. Is it thread safe? I can't say for sure...

I took a look at the bytecode generated (see https://godbolt.org/z/6d47nPs4a) and under the hood is using [`ObjectRef`](https://github.com/JetBrains/kotlin/blob/b65a68ca0db564f4680fde19abea480425eaba0b/libraries/stdlib/jvm/runtime/kotlin/jvm/internal/Ref.java#L13), which is a `static final class` that holds an `element`.
That allows the use of `var throwableInGenerate` even though it's not really final.

Anyway, that doesn't guarantee thread safety. So your proposed solution using `AtomicReference` does indeed help with ensuring that the `throwable` is visible to all threads and it is set atomically.

IgnatBeresnev (approved):
Perfect, thank you very much for the fix, and sorry for the long review!

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
IgnatBeresnev: oh, it's nice to know what it's being compiled into, thanks for looking into it and sharing :)

taroenock:
[jush:jush/fix-build-success-exception]() [https://github.com/Kotlin/dokka/pull/2935#issue-1633364699](jush:jush/fix-build-success-exception) 

> Capture any throwable thrown by the `generate()` function running in a separate thread to later on throw it in the context of the `generateDocumentation()` calling thread.
> 
> Fixes #2934

Maffi

taroenock:
Good 😊😊😊😌

PR #2937, opened 2023-03-21
Mark utilities.* API as Dokka-internal

qwwdfsad:
Deprecate unused declaration, remove inline declaration (as it's binary compatible), opt-in into internal API at project level

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import org.jetbrains.ValidatePublications
+import org.jetbrains.kotlin.gradle.tasks.*
 import org.jetbrains.publicationChannels
 
 @Suppress("DSL_SCOPE_VIOLATION") // fixed in Gradle 8.1 https://github.com/gradle/gradle/pull/23639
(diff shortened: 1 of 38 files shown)


build.gradle.kts
```diff
@@ -35,6 +36,15 @@ val dokkaPublish by tasks.registering {
...
 
+// opt-in into internal API on the project level
+subprojects {
+    tasks.withType(KotlinCompile::class).all {
+        kotlinOptions {
+            freeCompilerArgs = freeCompilerArgs + listOf("-opt-in=org.jetbrains.dokka.InternalDokkaApi")
+        }
+    }
+}
+
```
aSemy, line 47: [Gradle recommends avoiding using `subprojects {}`](https://docs.gradle.org/8.0.1/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration) (and also `allprojects {}`).

This could be moved to the Kotlin/JVM convention plugin, which already has some opt-ins.

https://github.com/Kotlin/dokka/blob/ff5aa6721063599d8b68a245c482621b0d110fc6/build-logic/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts#L14-L29

IgnatBeresnev (approved):
:+1: :+1:

PR #2938, opened 2023-03-22
Stick with Java 17 on CI

Goooler:
Follow up #2918.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 11
+          java-version: 17
           cache: 'maven'
       - uses: gradle/gradle-build-action@v2
         with:
--- a/.github/workflows/dokka-examples.yml
+++ b/.github/workflows/dokka-examples.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 11
+          java-version: 17
           cache: 'maven'
       - uses: gradle/gradle-build-action@v2
       - run: ./gradlew build --no-daemon --stacktrace
@@ -59,7 +59,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 11
+          java-version: 17
           cache: 'maven'
       - uses: gradle/gradle-build-action@v2
       - run: ./gradlew dokkaHtml --no-daemon --stacktrace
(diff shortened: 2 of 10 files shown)


Goooler:
We enabled Java 8 as the default in tests via toolchain in #2918, and Java 8 as the compatibility before, now we can bump the running JDK to 17.

IgnatBeresnev:
Oh, we have the toolchain now, right

So basically Java 17 will be used for setting things up and some non-build java executions, and then the toolchain versions will be used for running the actual tests and whatnot?

Goooler:
Yep.

aSemy:
The Java toolchain used to build Dokka projects is set in `base-java.gradle.kts`

https://github.com/Kotlin/dokka/blob/3c3d2e47abb732a006b899d40162758568c41e34/build-logic/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts#L15-L19

The actual value is set using the Gradle property `org.jetbrains.dokka.javaToolchain.mainCompiler`, and the default is 8

https://github.com/Kotlin/dokka/blob/3c3d2e47abb732a006b899d40162758568c41e34/gradle.properties#L3

When Gradle tries to build Dokka it will realise it needs JDK 8, and try and [auto-detect](https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection) an installed version. 

It's similar for the tests: they are configured to run using JDK 8, 11, or 17 via a toolchain, and so Gradle will try and auto-detect those versions too.

Gradle doesn't automatically download JDKs, so they need to be pre-installed, and that's what the setup-java step is for. However, it looks like [the latest GitHub runners have JDK 8, 11, and 17 pre-installed](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#java), so technically the setup-java step isn't needed for building/testing all (I suspect that the setup-java step was needed in older GitHub runners). 

However, Gradle itself is a JVM tool, so it needs Java installed and it the `JAVA_HOME` environment variable must be set - I guess that's why the setup-java step is needed? I couldn't find any reference to a default `JAVA_HOME` in the GitHub runners.

The setup-java action does some caching, but I'm not sure how much impact that has (if any).

If I've understood correctly, perhaps it would be good to add a small bit of docs in `CONTRIBUTING.md` (feel free to copy/paste/adapt this comment) and add a link next to the setup-java steps?

Goooler:
May need https://github.com/gradle/foojay-toolchains

IgnatBeresnev:
@aSemy thank you for the detailed explanation, I'll indeed adapt it and add some other things to CONTRIBUTING.md in a separate PR, will tag you

@Goooler thanks for the contribution!

Goooler:
@IgnatBeresnev There are JDK 8 used in TeamCity runners, can we bump them also?

PR #2951, opened 2023-03-30
Clean up GitHub workflows

IgnatBeresnev:
We have quite a number of actions that do nothing or are not very helpful, yet they waste time and have a possibility of randomly failing for uncontrollable reasons, not to say they make it impossible to understand which checks are important and which aren't. This PR addresses that problem, along with some minor improvements like naming.

The minimal number of jobs per PR went down from 29 to just 7 

### Unit test changes

We currently (as in before this PR) run unit tests with a matrix of linux/windows and java 8/11/17. Running the tests on both windows and linux is helpful, it has caught significant problems before. 

However, running under different Java versions for every single commit is not that helpful. On the contrary, Java 17 tends to be very flaky, so we have to restart it all the time.

I've changed it so that we run: 
* fast (smoke) tests for every commit in a PR, on linux/windows, but only under Java 11 (can change it to Java 8). Should be enough to catch the tests failing (especially since they are also additionally run in TeamCity)
* long and thorough tests for every commit in master, with the following matrix: linux/windows/macos and Java 8/11/17. They'll be run right after a PR has been merged, so hopefully we'll notice if it introduced any problems, and address it in time.

Long and thorough tests can be triggered manually for a specific branch, so we should have the possibility of validating big changes in advance (i.e if we see that it might break java 8 or 17).

Later on it'd be nice to randomize some aspects of the tests using something like [github-actions-random-matrix](https://github.com/vlsi/github-actions-random-matrix). We'd be able to run random Java versions for PR tests, or shuffle platforms. This way, it'll still be quick, yet it should cover more cases overall as there are typically multiple commits in a PR.

### Common changes

* Do not trigger a workflow if it doesn't need to be run. For instance, we only need to build/publish the examples if the `examples/**` directory was touched. Otherwise, they are using a stable version of Dokka and no other changes in code should affect them.
* Change workflow/job names to be shorter and more understandable, will help with visually inspecting it from the UI.
* Change workflow file names to be grouped together and convey the intention more.

I'll leave more detailed/context comments for individual changes/files


.github/workflows/check-gradle.yml
IgnatBeresnev, line 1: Doesn't seem like we need to validate the wrapper unless it's been changed. 

* Extracted it into a separate file so that it can be run with a `paths` filter. 
* Additionally running it on every commit in master in case someone with write access pushes it without a PR or does something funny to it indirectly.

.github/workflows/rebase.yml
IgnatBeresnev, line 1: I don't think anyone needs it? I noticed it a while ago and thought it was useful, but I still never used it..

.github/workflows/tests-smoke.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          gradle-home-cache-cleanup: true
+      - name: Run tests under Ubuntu/Macos
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
+        run: ./gradlew clean test --stacktrace
+      - name: Run tests under Windows
+        if: matrix.os == 'windows-latest'
+        # Running tests without additional flags on windows agents leads to some very strange
+        # JVM crashes for some reason. Most likely a problem of Gradle/GitHub/Windows server
+        run: ./gradlew clean test --stacktrace --no-daemon --parallel --max-workers=1
```
IgnatBeresnev, line 35: Copy-pasted it from the old workflow code, not sure what the problem is or was exactly...

.github/workflows/preview-publish-ga.yml
```diff
@@ -34,9 +37,11 @@ jobs:
           path: /home/runner/work/dokka/coroutines
           retention-days: 7
 
-  serialization:
+  kotlinx-serialization:
     runs-on: ubuntu-latest
-    if: github.repository == 'Kotlin/dokka'
+    if: >  # Run this job only for pull requests from forks because members with write access have web preview
+      github.event_name != 'pull_request' || 
+      (github.event_name == pull_request && github.event.pull_request.head.repo.full_name != 'Kotlin/dokka')
```
Goooler, line 44:
```suggestion
      (github.event_name == pull_request && github.repository != 'Kotlin/dokka')
```

.github/workflows/preview-publish-ga.yml
```diff
@@ -63,7 +68,9 @@ jobs:
 
   biojava:
     runs-on: ubuntu-latest
-    if: github.repository == 'Kotlin/dokka'
+    if: >  # Run this job only for pull requests from forks because members with write access have web preview
+      github.event_name != 'pull_request' || 
+      (github.event_name == pull_request && github.event.pull_request.head.repo.full_name != 'Kotlin/dokka')
```
Goooler, line 73:
```suggestion
      (github.event_name == pull_request && github.repository != 'Kotlin/dokka')
```

.github/workflows/preview-publish-web-s3.yml
```diff
@@ -1,13 +1,13 @@
-name: Build example projects and publish them to s3
+name: Publish preview to web (S3)
 
-on: push
+on: [push, workflow_dispatch]
```
Goooler, line 3:
```suggestion
on: [ push, workflow_dispatch ]
```

.github/workflows/tests-smoke.yml
```diff
@@ -0,0 +1,35 @@
...
+          distribution: 'zulu'
+          java-version: 17
+          cache: 'maven'
+      - uses: gradle/gradle-build-action@v2
+        env:
+          ORG_GRADLE_PROJECT_org.jetbrains.dokka.javaToolchain.test: 11
+        with:
+          gradle-home-cache-cleanup: true
+      - name: Run tests under Ubuntu/Macos
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
```
Goooler, line 29: Only `ubuntu-latest, windows-latest` in test matrix.

.github/workflows/tests-smoke.yml
Goooler, line 35: `--parallel` is the default, you want `--no-parallel`?

.github/workflows/tests-smoke.yml
Goooler: And we can just use `matrix.os != 'windows-latest'`.

.github/workflows/check-gradle.yml
```diff
@@ -0,0 +1,17 @@
+name: Gradle check
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - 'gradle/**'
```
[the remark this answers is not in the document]
IgnatBeresnev: I pointed out in my [comment](https://github.com/Kotlin/dokka/pull/2951#discussion_r1153612825) that I intentionally want to run it for every commit in master in case anything is done to it indirectly, or the change is force pushed and the workflow result deleted, or the filenames/paths are changed without updating the paths here, and so on - it then has at least **some** chance of catching a bad actor.

.github/workflows/tests-smoke.yml
IgnatBeresnev: No idea :( As I [pointed out](https://github.com/Kotlin/dokka/pull/2951#discussion_r1153718954), this was copy-pasted from the previous configuration. Without it, the windows tests are consistently failing due to JVM crashes. Unfortunately, there was no explanation in the commit/PR that introduced these changes :man_shrugging:

.github/workflows/preview-publish-ga.yml
```diff
@@ -1,15 +1,18 @@
...
+  workflow_dispatch:
 
 jobs:
-  coroutines:
+  kotlinx-coroutines:
     runs-on: ubuntu-latest
-    if: github.repository == 'Kotlin/dokka'
+    if: >  # Run this job only for pull requests from forks because members with write access have web preview
+      github.event_name != 'pull_request' || 
+      (github.event_name == pull_request && github.event.pull_request.head.repo.full_name != 'Kotlin/dokka')
```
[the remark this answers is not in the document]
IgnatBeresnev: With the proposed change, I don't think it works the way I want it to work and the way it's described in the comment. The check for `github.repository` was actually present before if you look at the original code, and it was executed for all PRs, from Dokka's branches and from forks, which is not what's desired here

.github/workflows/preview-publish-ga.yml
IgnatBeresnev: Addressed in https://github.com/Kotlin/dokka/pull/2951#discussion_r1160151905

.github/workflows/preview-publish-ga.yml
IgnatBeresnev: Addressed in https://github.com/Kotlin/dokka/pull/2951#discussion_r1160151905, let's keep the conversation there, and I'll change this if it needs to be changed

.github/workflows/gh-pages-deploy-mkdocs.yml
```diff
@@ -1,9 +1,11 @@
-name: Build and deploy MkDocs to GitHub Pages
+name: Deploy MkDocs to GitHub Pages
 
 on:
   push:
     branches:
       - master
+    paths:
+      - 'mkdocs/**'
```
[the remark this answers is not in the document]
IgnatBeresnev: If I accept this change, it looks like it's not going to be consistent with the other similar cases - is there any reason for why this particular case should be different? `release` is still part of `on`

.github/workflows/check-gradle.yml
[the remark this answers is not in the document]
Goooler: Indeed.

.github/workflows/gh-pages-deploy-mkdocs.yml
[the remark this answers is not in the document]
Goooler: Nope, just a style change to divide `on` with `jobs`.

.github/workflows/preview-publish-ga.yml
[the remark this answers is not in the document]
Goooler: Gotcha.

.github/workflows/tests-smoke.yml
Goooler: Need `--no-daemon` to disable daemon for Windows, I patched this for Ktlint before, and `--parallel` is the default, could be removed for now.

.github/workflows/gh-pages-deploy-mkdocs.yml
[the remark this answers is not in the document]
IgnatBeresnev: oh, but the next line is not jobs, it's "on: release"

.github/workflows/tests-smoke.yml
IgnatBeresnev: Looks like the tests passed, thank you!

Goooler: There are more flaky test failures with Daemon enabled on Windows.

vmishenev (approved):

PR #2953, opened 2023-03-31
Increase test processes to start in parallel

Goooler:
More context see https://github.com/detekt/detekt/pull/5944.

https://docs.gradle.org/7.6/userguide/performance.html#execute_tests_in_parallel


build-logic/src/main/kotlin/org/jetbrains/conventions/base-java.gradle.kts
```diff
@@ -22,6 +22,12 @@ java {
 tasks.withType<Test>().configureEach {
     useJUnitPlatform()
 
+    maxParallelForks = if (System.getenv("CI") != null) {
```
Goooler, line 25: Or we can read `dokka_integration_test_parallelism` like

https://github.com/Kotlin/dokka/blob/3ea414251d32108dab8b14d41a672e9834824690/build-logic/src/main/kotlin/org/jetbrains/conventions/dokka-integration-test.gradle.kts#L54-L56

Goooler: 5facf7eee5f29f672345b1cae95d1fc98b9ad388

IgnatBeresnev:
Finally set out some time to test it.

I think it makes perfect sense to do this for unit tests, absolutely no questions here. But I wasn't sure what sort of an impact it would have on our integration tests.

I have a totally overspec'd workstation (16C, 32T, 128GB RAM) to draw any conclusions, but running the `integrationTest` task multiple times with caching disabled:
* With `maxParallelForks = 2` it takes ~17 minutes, max RAM consumed throughout each run is ~20GB
* With changes from this PR, so with `maxParallelForks = 16` it takes ~10 minutes, max RAM consumed throughout each run is ~50GB.

I actually wonder what would happen if I didn't have enough memory - would the integration tests fail, or would the time increase drastically because of GC problems? 

I'm trying to trigger integration tests on teamcity for this PR, but teamcity is not picking up the branch changes like it did before... Our admins might've introduced some security policies, so I'll have to resolve that first, which might take time...

I'd be happy to merge this change for unit tests only though, if you open a separate PR or change this one

IgnatBeresnev:
(looks like TeamCity was simply lagging yesterday, integration tests work now)

IgnatBeresnev (approved):
Thank you!

I'm still interested in having (or at least testing) this change for integration tests, so @Goooler if you want and have the time, feel free to open a separate PR :)

PR #2961, opened 2023-04-16
disable publication of shadowRuntimeElements variant

aSemy:
Fixes the issue described in https://github.com/Kotlin/dokka/pull/2704#issuecomment-1499517930

### Test

To test this I ran `./gradlew publishAllPublicationsToMavenProjectLocalRepository` and searched for `shadowRuntimeElements` in the directory `$rootDir/build/maven-project-local`

On master branch there was a match. On the PR branch, there was no match.

### DIFF
--- a/build-logic/src/main/kotlin/org/jetbrains/conventions/maven-publish.gradle.kts
+++ b/build-logic/src/main/kotlin/org/jetbrains/conventions/maven-publish.gradle.kts
@@ -61,6 +61,8 @@ publishing {
 
 plugins.withType<ShadowPlugin>().configureEach {
     // manually disable publication of Shadow elements https://github.com/johnrengelman/shadow/issues/651#issue-839148311
+    // This is done to preserve compatibility and have the same behaviour as previous versions of Dokka.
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/org/jetbrains/conventions/maven-publish.gradle.kts
```diff
@@ -56,3 +58,9 @@ publishing {
         }
     }
 }
+
+plugins.withType<ShadowPlugin>().configureEach {
+    // manually disable publication of Shadow elements https://github.com/johnrengelman/shadow/issues/651#issue-839148311
```
IgnatBeresnev, line 63: Can we leave some more information for not only what it does, but why this is needed? So that the future maintainers are not afraid to remove/change it later on if they need to

```suggestion
    // manually disable publication of Shadow elements https://github.com/johnrengelman/shadow/issues/651#issue-839148311
    // This is done to preserve compatibility and have the same behaviour as previous versions of Dokka.
    // For more details, see https://github.com/Kotlin/dokka/pull/2704#issuecomment-1499517930
```

IgnatBeresnev (approved):
Thanks for addressing it!

PR #2962, opened 2023-04-17
migrate test utils to use Test Fixtures plugin

aSemy:
Migrate test-util project to use Test Fixtures, to improve project organisation and structure.

This is a speculative PR that I have made because I think it a demonstration is more clear than trying to write it up in an issue. It introduces some breaking changes for 3rd party consumers, which might be undesirable.

Part of #2703 

### Summary

This PR migrates the existing test-utility projects to instead be contained inside the project in an independent source set, thanks to the [Java Test Fixtures](https://docs.gradle.org/8.1/userguide/java_testing.html#sec:java_test_fixtures) plugin.

* code and dependencies for the test-util subprojects (`:core:test-api`, `:core:content-matcher-test-utils`, `:plugins:base:base-test-utils`) have been merged into their 'parent' subprojects
* Local and remote projects do not need to apply the Test Fixtures plugin, and can depend on the test util code using the `testFixtures()` helper function embedded into Gradle.

### Breaking changes

There are two breaking changes

#### Updated dependency coordinates for the test fixtures.

Currently the test utility projects are published as Maven artifacts (e.g. [dokka-base-test-utils](https://search.maven.org/artifact/org.jetbrains.dokka/dokka-base-test-utils/1.8.10/jar)). These artifacts will no longer exist. Instead, the artifacts that contain the test fixtures (e.g. `dokka-base`) will have two [variants](https://docs.gradle.org/current/userguide/feature_variants.html) published, which can be depended on as follows:

```
dependencies {
  testImplementation("org.jetbrains.dokka:dokka-base-test-utils:1.8.10") // old

  testImplementation(testFixtures("org.jetbrains.dokka:dokka-base:1.8.10")) // new
}
```

There are two potential workarounds to make this change a little less grating:

* Keep the `:plugins:base:base-test-utils` projects (and others), which would expose the test fixtures as an API dependency, but would not contain any source code
   
   ```
   // plugins/base/base-test-utils/build.gradle.kts
   
   dependencies {
     api(testFixtures(projects.plugins.base))
   }
   ```

* [publish Maven relocation information](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:relocation)

#### Binary Compatibility Validator support

BCV does not currently support the Java Test Fixtures plugin https://github.com/Kotlin/binary-compatibility-validator/issues/117

I've created a fork of BCV that does support them https://github.com/adamko-dev/kotlin-binary-compatibility-validator-mu


build-logic/build.gradle.kts
```diff
@@ -4,14 +4,15 @@ plugins {
 
 kotlin {
     jvmToolchain {
-        languageVersion.set(JavaLanguageVersion.of(8))
+        languageVersion.set(JavaLanguageVersion.of(11))
```
aSemy, line 7: bcv-mu requires Java 11. Perhaps there's a better way around this, but it shouldn't matter. This line of config might not even be necessary any more.

build.gradle.kts
```diff
@@ -35,21 +33,3 @@ val dokkaPublish by tasks.registering {
...
-        "kotlin-analysis",     // :kotlin-analysis
-        "compiler-dependency", // :kotlin-analysis:compiler-dependency
-        "intellij-dependency", // :kotlin-analysis:intellij-dependency
-
-        "integration-tests",   // :integration-tests
-        "gradle",              // :integration-tests:gradle
-        "cli",                 // :integration-tests:cli
-        "maven",               // integration-tests:maven
-    )
-}
```
aSemy, line 55: bcv-mu is applied on individual subprojects, so this configuration can be removed.

core/api/testFixtures/core.api
aSemy, line 1: this file is the merged result of the `:core:test-api` and `:core:content-matcher-test-utils` projects' api dumps

kotlin-analysis/build.gradle.kts
```diff
@@ -15,3 +15,7 @@ dependencies {
 registerDokkaArtifactPublication("dokkaAnalysis") {
     artifactId = "dokka-analysis"
 }
+
+binaryCompatibilityValidator {
+    enabled.set(false)
+}
```
aSemy, line 21: the kotlin-analysis projects have API verification disabled, but perhaps they should indeed be validated? After all, they are published artifacts.

build-logic/src/main/kotlin/org/jetbrains/conventions/maven-publish.gradle.kts
```diff
@@ -1,10 +1,13 @@
 package org.jetbrains.conventions
 
+import org.gradle.internal.component.external.model.TestFixturesSupport.TEST_FIXTURE_SOURCESET_NAME
+
 plugins {
     id("org.jetbrains.conventions.base")
     `maven-publish`
     signing
     id("org.jetbrains.conventions.dokka")
+    id("dev.adamko.kotlin.binary-compatibility-validator")
```
aSemy, line 10: Rather than BCV-classic's approach of validating _all_ projects by default, and opting out on a case-by-case basis, I thought it made sense to apply BCV-MU in a slightly more targeted way and apply it to all published subprojects.

whyoleg:
Hey, @adam-enko, is it possible to finish this PR? (or reopen new one, as there were a lot of changes in the meanwhile)
* https://github.com/Kotlin/binary-compatibility-validator/issues/117 - fixed
* `keep the :plugins:base:base-test-utils projects (and others), which would expose the test fixtures as an API dependency` - this looks fine, but better to test it

adam-enko:
I think it's probably best to just close this PR. It's not entirely necessary to use test-fixtures, and if they would help, then it would make sense to start again from scratch with the latest changes from master.

PR #2963, opened 2023-04-17
Add Gradle integration tests for configuration

IgnatBeresnev:
* Added an integration test project that can be used to test Dokka's configuration parameters specifically. We used to do it in other existing tests like `it-basic`, but it's better if it's isolated and each test project tests one thing or area
* Added the first integration test for `failOnWarning`, it got broken, and we didn't notice it until it was fixed in #2935


integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/ConfigurationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+            "-Pfail_on_warning=true",
+            "dokkaHtml"
+        ).buildAndFail()
+
+        assertEquals(TaskOutcome.FAILED, assertNotNull(result.task(":dokkaHtml")).outcome)
+
+        result.output.contains("> Task :dokkaHtml FAILED")
+        result.output.contains(
+            """
+               FAILURE: Build failed with an exception.
```
vmishenev, line 60: Note: Should `.` (a dot) be escaped?

vmishenev (approved):

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/ConfigurationTest.kt
IgnatBeresnev: Definitely won't hurt :+1:  Added it

PR #2966, opened 2023-04-18
Add source links to the dokka-multimodule-example project

IgnatBeresnev:
The configuration of source links in multimodule projects has been asked about multiple times, and we don't seem to have an example for it.

A community member asked about it yesterday in the Kotlin Community Slack, so I decided to add an example both to check that it works and to point people to it for configuration tips

### DIFF
--- a/examples/gradle/dokka-multimodule-example/parentProject/build.gradle.kts
+++ b/examples/gradle/dokka-multimodule-example/parentProject/build.gradle.kts
@@ -1,5 +1,6 @@
 import org.jetbrains.dokka.DokkaConfiguration.Visibility
 import org.jetbrains.dokka.gradle.DokkaTaskPartial
+import java.net.URL
 
 plugins {
     kotlin("jvm")
@@ -17,6 +18,19 @@ subprojects {
                 Visibility.PUBLIC,
                 Visibility.PROTECTED
             ))
+
+            // In multi-project builds, `remoteUrl` must point to that project's dir specifically, so if you
+            // want to configure sourceLinks at once in `subprojects {}`, you have to find the relative path.
+            // Alternatively, you can move this configuration up into subproject build scripts,
+            // and just hardcode the exact paths as demonstrated in the basic dokka-gradle-example.
+            sourceLink {
+                val exampleDir = "https://github.com/Kotlin/dokka/tree/master/examples/gradle/dokka-multimodule-example"
+                val projectRelativePath = rootProject.projectDir.toPath().relativize(projectDir.toPath())
+
+                localDirectory.set(file("src"))
+                remoteUrl.set(URL("$exampleDir/$projectRelativePath/src"))
+                remoteLineSuffix.set("#L")
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


examples/gradle/dokka-multimodule-example/parentProject/build.gradle.kts
```diff
@@ -17,6 +18,19 @@ subprojects {
...
+
+            // In multi-project builds, `remoteUrl` must point to that project's dir specifically, so if you
+            // want to configure sourceLinks at once in `subprojects {}`, you have to find the relative path.
+            // Alternatively, you can move this configuration up into subproject build scripts,
+            // and just hardcode the exact paths as demonstrated in the basic dokka-gradle-example.
+            sourceLink {
+                val exampleDir = "https://github.com/Kotlin/dokka/tree/master/examples/gradle/dokka-multimodule-example"
+                val projectRelativePath = rootProject.projectDir.toPath().relativize(projectDir.toPath())
+
+                localDirectory.set(projectDir.resolve("src"))
```
vmishenev, line 30: I have not checked, but there is the shortest way (in this case): 
```
localDirectory.set(projectDir)
remoteUrl.set(URL("$exampleDir"))
remoteLineSuffix.set("#L")
```

examples/gradle/dokka-multimodule-example/parentProject/build.gradle.kts
```diff
@@ -17,6 +18,19 @@ subprojects {
                 Visibility.PUBLIC,
                 Visibility.PROTECTED
             ))
+
+            // In multi-project builds, `remoteUrl` must point to that project's dir specifically, so if you
+            // want to configure sourceLinks at once in `subprojects {}`, you have to find the relative path.
+            // Alternatively, you can move this configuration up into subproject build scripts,
+            // and just hardcode the exact paths as demonstrated in the basic dokka-gradle-example.
```
vmishenev, line 25: Could you add the link (https://kotlinlang.org/docs/dokka-gradle.html#source-link-configuration)?

vmishenev (approved):

examples/gradle/dokka-multimodule-example/parentProject/build.gradle.kts
IgnatBeresnev: It probably would work for single-module projects, but it doesn't work with multimodule ones as the generated link (remoteUrl) needs to have full path to the submodule - pretty much what I tried to describe in the comment and why I decided to add the example

So with the proposed code, Dokka generates the following link for `ChildProjectAClass`: 
* https://github.com/Kotlin/dokka/tree/master/examples/gradle/dokka-multimodule-example/src/main/kotlin/demo/ChildProjectAClass.kt#L8

And it needs to be:
* https://github.com/Kotlin/dokka/blob/master/examples/gradle/dokka-multimodule-example/parentProject/childProjectA/src/main/kotlin/demo/ChildProjectAClass.kt#L8

vmishenev: My bad, it should be
```
localDirectory.set(rootProject.projectDir) // instead of projectDir
```

IgnatBeresnev: Oh, indeed! I didn't think it would work tbh...

Addressed in #2974

PR #2968, opened 2023-04-18
Reorder dependency versions for update convenience

IgnatBeresnev:
Needed to update the Kotlin version, and got kinda lost in the toml

It's best to just check out the branch locally and see if it's convenient from the maintenance perspective. If some things are confusing or you know in advance that you will struggle with it - do suggest an alternative

___

* Reordered dependency versions and redid the grouping so that it's (subjectively) more convenient to update:
    * The most important and frequently updated ones are at the top (KGP, AGP, analysis, etc)
    * Analysis is grouped together and decoupled, so we can update KGP and the compiler versions independently. This should be pretty useful when migrating to K2, and should be useful in general when migrating to new/dev versions of Kotlin, and especially useful for troubleshooting.
* Renamed a few properties to better represent what they are: `kotlin` -> `gradlePlugin-kotlin`, `kotlinPlugin` -> `kotlinIdePlugin`, idea` -> `intellij`, etc.
* Removed some unused library declarations, mostly boms - they are not applied anywhere, we can add them once we start using it. I hope they weren't applied automatically, at least I couldn't google anything like that.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -19,17 +19,17 @@ soywiz-korte = "2.7.0"
 kotlinx-html = "0.7.5"
 
 ## Markdown
-jetbrainsMarkdown = "0.3.1"
+jetbrains-markdown = "0.3.1"
 
 ## JSON
 jackson = "2.12.7" # jackson 2.13.X does not support kotlin language version 1.4, check before updating
 jacksonDatabind = "2.12.7.1" # fixes CVE-2022-42003
 
 ## Maven
-apache-maven = "3.5.0"
-apache-mavenArtifact = "3.8.5"
-apache-mavenArchiver = "2.5"
-apache-mavenPluginTools = "3.5.2"
+apacheMaven-core = "3.5.0"
+apacheMaven-artifact = "3.8.5"
+apacheMaven-archiver = "2.5"
+apacheMaven-pluginTools = "3.5.2"
 
 ## CLI
 kotlinx-cli = "0.3.4"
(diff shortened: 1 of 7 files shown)


gradle/libs.versions.toml
```diff
@@ -1,115 +1,111 @@
...
+kotlinx-html = { module = "org.jetbrains.kotlinx:kotlinx-html-jvm", version.ref = "kotlinx-html" }
+soywiz-korte = { module = "com.soywiz.korlibs.korte:korte-jvm", version.ref = "soywiz-korte" }
 
-## Kotlin Plugins ##
-kotlinPlugin-common = { module = "org.jetbrains.kotlin:common", version.ref = "kotlin.plugin" }
-kotlinPlugin-idea = { module = "org.jetbrains.kotlin:idea", version.ref = "kotlin.plugin" }
-kotlinPlugin-core = { module = "org.jetbrains.kotlin:core", version.ref = "kotlin.plugin" }
-kotlinPlugin-native = { module = "org.jetbrains.kotlin:native", version.ref = "kotlin.plugin" }
+#### Markdown ####
+jetbrainsMarkdown = { module = "org.jetbrains:markdown", version.ref = "jetbrainsMarkdown" }
```
vmishenev, line 79: Nitpick:  I do not understand why this is called `jetbrainsMarkdown` rather than `jetbrains-markdown`(like `soywiz-korte`). Do we have a naming convention here?

vmishenev (approved):

gradle/libs.versions.toml
IgnatBeresnev: > Do we have a naming convention here?
I don't think do.. But the name here is responsible for the generation of the type-safe accessors.

* `jetbrainsMarkdown` would generate `libs.jetbrainsMarkdown`
* `jetbrains-markdown` would generate `libs.jetbrains.markdown`

I've tried to add some more consistency, can you have another look?

vmishenev (approved):

PR #2970, opened 2023-04-18
improve copying base-frontend files between subprojects

aSemy:
This is an https://github.com/Kotlin/dokka/labels/infrastructure issue, part of #2703 

Improve Dokka's Gradle compatibility by using '[sharing files between subprojects](https://docs.gradle.org/current/userguide/cross_project_publications.html)' config.

I expect that this has no impact on the project or published dependencies. I also expect that this should improve build-time developer-experience (Gradle syncing/configuring should be faster, as NPM will only run when needed, instead of _every_ time).

### Summary

* bump Gradle Node Plugin version
* add Gradle convention plugin, with configurations, for sharing base-frontend files between subprojects
* refactor how files are copied into base-plugin project (use generated resource dir and Sync task - this will better ensure copied files are up-to-date)
* remove empty & unused subproject `:plugins:base:search-component`
* remove .gitignore, no longer needed (instead, files are synced to build dir)


build-logic/src/main/kotlin/org/jetbrains/conventions/dokka-base-frontend-files.gradle.kts
aSemy, line 1: This convention plugin adds shared config based on [Variant-aware sharing of artifacts between projects](https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:variant-aware-sharing) docs

plugins/base/build.gradle.kts
```diff
@@ -33,45 +34,40 @@ dependencies {
...
     }
-    dependsOn(generateFrontendFiles)
-    destinationDir =
-        File(sourceSets.main.get().resources.sourceDirectories.singleFile, "dokka/styles")
-}
 
-val copyFrontend by tasks.registering {
-    dependsOn(copyJsFiles, copyCssFiles)
+    into(layout.buildDirectory.dir("generated/src/main/resources"))
 }
```
aSemy, line 63: Currently the frontend files are copied directly into `src/main/resources`. This works fine, but it might cause false-positives if there's an issue where the frontend files aren't generated - any previous files will remain and continue to work, until you run `clean`.

[Sync tasks are better for mirroring files](https://docs.gradle.org/current/userguide/working_with_files.html#sec:sync_task). However, they will delete any other files in the target directory...

So instead I synced the files into a new directory, `build/generated/src/main/...` and converted the task to be a file provider, and added that file to the resources dir.

Which IntelliJ also recognises

image

settings.gradle.kts
```diff
@@ -68,7 +68,6 @@ include(
 
     ":plugins:base",
     ":plugins:base:frontend",
-    ":plugins:base:search-component",
```
aSemy, line 71: I'm pretty sure that this subproject is unused? I couldn't find any usages, and it was empty.

plugins/base/frontend/build.gradle.kts
```diff
@@ -11,19 +15,34 @@ node {
...
+
+    inputs.dir("src/main")
+    inputs.files(
+        "package.json",
+        "webpack.config.js",
+    )
+
+    outputs.dir(distributionDirectory)
     outputs.cacheIf { true }
 }
```
aSemy, line 33: some minor tidying here, which hopefully makes the task more understandable.

I removed [the task-rule based creator](https://github.com/node-gradle/gradle-node-plugin/blob/642f30439f186c66e14d45bbc0c8518892e19bc5/src/main/kotlin/com/github/gradle/node/NodePlugin.kt#L67-L80) because I think it's a little less 'magic'.

settings.gradle.kts
IgnatBeresnev: It might've existed for what now resides in `frontend/src/main/components/search`, but it indeed looks to be unused as of this moment :+1:

build-logic/src/main/kotlin/org/jetbrains/conventions/dokka-base-frontend-files.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+ * Utility for sharing the Dokka Base Plugin frontend files between subprojects in a safe, cacheable way.
+ */
+
+plugins {
+    id("org.jetbrains.conventions.base")
+}
+
+/** Apply a distinct attribute to the incoming/outgoing configuration */
+fun AttributeContainer.dokkaBaseFrontendFilesAttribute() =
+    attribute(USAGE_ATTRIBUTE, objects.named("org.jetbrains.dokka.base-frontend-files"))
```
IgnatBeresnev, line 15: I would propose to rename all mentions of `base-frontend-files` to `html-frontend-files` while we're here.

The HTML plugin (and thus the format itself) will one day be extracted out of `base` and into it's own module, it's already been attempted in #2752, but we switched to more urgent matters

This convention plugin also has great potential for reusability, and in theory the frontend files could reside in multiple other modules that have nothing to do with `base`, so to me it makes sense to rename it to something broader

IgnatBeresnev:
Great job! Much cleaner and with reusability potential. I wish I understood what those frontend files were used for though :smiling_face_with_tear: 

Could you resolve build errors, please? There's a problem with the toml catalog of some sort.

Once all checks are green, I'll run integration tests and build coroutines and serialization sites - if their HTML frontend looks the same way as before, we can consider it a win and merge it :)

build-logic/src/main/kotlin/org/jetbrains/conventions/dokka-base-frontend-files.gradle.kts
aSemy: > I would propose to rename all mentions of `base-frontend-files` to `html-frontend-files` while we're here.

Good idea, done!

By the way, now that the Gradle config of Dokka is more Gradle-idiomatic (e.g. there's no `allprojects {}` Gradle configuration, subprojects are mostly isolated), it might help to think about restructuring the subprojects while you're doing this.

Personally I think the subproject nesting and terse subproject names can be confusing, so I prefer a more flat layout so that all of the subprojects are visible at once, while the subproject names are expanded so they are ordered. So something like this:

```
.
├── dokka-core/
│   ├── dokka-generator
│   ├── dokka-generator-configuration
│   ├── dokka-content-matcher-test-utils
│   └── dokka-test-api
└── dokka-plugins/
    ├── dokka-html
    ├── dokka-html-frontend
    ├── dokka-gfm
    ├── dokka-gfm-template-processing
    └── ...
```

(this example includes theoretical splitting of subprojects

The [Kotest](https://github.com/kotest/kotest/) and [Gradle](https://github.com/gradle/gradle/tree/master/subprojects) projects are examples of this layout that I like.

Just sharing an idea :)

aSemy:
> Great job! Much cleaner and with reusability potential. I wish I understood what those frontend files were used for though 🥲

Thanks!

So the subproject contains TypeScript and SCSS files that are compiled (by the Gradle Node Plugin) to plain JS and CSS files (`src/main/resources/dokka/scripts/main.js` and `/src/main/resources/dokka/styles/main.css`), and then these are added in the HTML files

https://github.com/Kotlin/dokka/blob/d9c8c6c0035ad59dc0a5495f3247205bb8553295/plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt#L47-L55

https://github.com/Kotlin/dokka/blob/d9c8c6c0035ad59dc0a5495f3247205bb8553295/plugins/base/src/main/kotlin/renderers/html/htmlPreprocessors.kt#L76-L82

> Could you resolve build errors, please? There's a problem with the toml catalog of some sort.

Fixed! I missed something during a merge so the `node` version was duplicated.

IgnatBeresnev:
Awesome, thanks! 

I've started the integration tests and the generation of preview websites, will check on it tomorrow and merge if everything's ok

build-logic/src/main/kotlin/org/jetbrains/conventions/dokka-base-frontend-files.gradle.kts
IgnatBeresnev: Good suggestion for sure, there's definitely lots of room for improvement! 

Although there's a bit too much refactoring going on at the moment with K2 and the extraction of the html format out of dokka-base, it'll introduce a lot nasty merge conflicts for several long-lived branches... Should be easier to do it closer to ~1.9.20 when as much as possible gets merged into master though

IgnatBeresnev:
Everything looks good, thanks once again!

Issue #2817, opened 2023-01-16
Dokka doesn't ignore some test sources for android projects


atyrin:
The issue is based on `AndroidAutoConfigurationTest`.
It fails if the dependency is updated from `4.0.1` to `4.2.2`/`7.3.1`. 

So, for now, sources that are located in `androidTestRelease` (aka folder for instrumented tests) are rendered in Dokka.

**Installation**
- Dokka version: 1.7.20

```
Expected :[androidTest, androidTestDebug, debug, main, release, test, testDebug, testRelease]
Actual   :[androidTest, androidTestDebug, androidTestRelease, debug, main, release, test, testDebug, testRelease]
```
and
```
Expected source set `androidTestRelease` to be suppressed by default
```

It would be nice to update the dependency version. If it does not affect the lower Gradle version compatibility.

atyrin:
Also, instrumented test sources under the folder with build type (debug/release) are marked as common sources. But the project has only `kotlin("android")` plugin
<img width="1219" alt="image" src="https://user-images.githubusercontent.com/7689049/221881459-b8d9e777-5ba2-4e49-b229-709c82c6661f.png">
<img width="1002" alt="image" src="https://user-images.githubusercontent.com/7689049/221881659-c74c30de-1430-400d-b064-f2750806b37c.png">

whyoleg:
@atyrin, I think that in light of Dokka 2.2.0 and #4472, it's not an issue anymore.

PR #2972, opened 2023-04-19
Update Kotlin Gradle Plugin to 1.8.20

IgnatBeresnev:
Based on #2968, it needs to be merged first, and then this branch rebased onto master

### DIFF
--- a/core/src/main/kotlin/utilities/json.kt
+++ b/core/src/main/kotlin/utilities/json.kt
@@ -20,10 +20,11 @@ private val objectMapper = run {
 }
 
 @PublishedApi
-internal class TypeReference<T> private constructor(
+internal class TypeReference<T> @PublishedApi internal constructor(
     internal val jackson: JacksonTypeReference<T>
 ) {
     companion object {
+        @PublishedApi
         internal inline operator fun <reified T> invoke(): TypeReference<T> = TypeReference(jacksonTypeRef())
     }
 }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 
-gradlePlugin-kotlin = "1.8.10"
+gradlePlugin-kotlin = "1.8.20"
 gradlePlugin-android = "4.0.1"
 gradlePlugin-dokka = "1.7.10"
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 
 gradlePlugin-kotlin = "1.8.20"
-gradlePlugin-android = "4.0.1"
+gradlePlugin-android = "4.1.3"
 gradlePlugin-dokka = "1.7.10"
 
 kotlinx-coroutines = "1.6.3"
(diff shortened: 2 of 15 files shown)


core/src/main/kotlin/utilities/json.kt
```diff
@@ -20,10 +20,11 @@ private val objectMapper = run {
 }
 
 @PublishedApi
-internal class TypeReference<T> private constructor(
+internal class TypeReference<T> @PublishedApi internal constructor(
     internal val jackson: JacksonTypeReference<T>
 ) {
     companion object {
+        @PublishedApi
         internal inline operator fun <reified T> invoke(): TypeReference<T> = TypeReference(jacksonTypeRef())
```
IgnatBeresnev, line 28: I think the rules for exposing internal declarations in inline functions have become more strict.

This is called in inline functions, so it needs to be `@PublishedApi` - not sure why it didn't before, seems logical. 

And it used to call a private constructor, which is not allowed now, so the constructor needs to be both `internal` and annotated with `@PublishedApi` now

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 
-gradlePlugin-kotlin = "1.8.10"
-gradlePlugin-android = "4.0.1"
+gradlePlugin-kotlin = "1.8.20"
+gradlePlugin-android = "4.1.3"
```
IgnatBeresnev, line 4: If I remember correctly, KGP 1.8.20 now requires AGP to be <= 4.1.3. Could not build the project without bumping the version here.

runners/gradle-plugin/src/test/kotlin/org/jetbrains/dokka/gradle/AndroidAutoConfigurationTest.kt
```diff
@@ -26,6 +26,7 @@ class AndroidAutoConfigurationTest {
     }
 
     @Test
+    @Ignore // TODO fails because of a new source set androidTestRelease, which is not in the list
```
IgnatBeresnev, line 29: This was reported in #2817. Need to figure out what to do with this source set.. Not even sure why it and how to handle it

plugins/base/src/main/kotlin/signatures/KotlinSignatureProvider.kt
```diff
@@ -103,7 +103,7 @@ class KotlinSignatureProvider(ctcc: CommentsToContentConverter, logger: DokkaLog
     }
 
     private fun <T : DClasslike> classlikeSignature(c: T): List<ContentNode> {
-        @Suppress("UNCHECKED_CAST")
+        @Suppress("UNCHECKED_CAST", "TYPE_MISMATCH_WARNING_FOR_INCORRECT_CAPTURE_APPROXIMATION")
```
IgnatBeresnev, line 106: More info on `TYPE_MISMATCH_WARNING_FOR_INCORRECT_CAPTURE_APPROXIMATION`: [KT-49404](https://youtrack.jetbrains.com/issue/KT-49404/Fix-type-unsoundness-for-contravariant-captured-type-based-on-Java-class)

This will be reported as error in 2.0, so we need to fix it either now or in 1.9. And I'm not sure if fixing it would result in breaking API changes. If so, it's quite sad as this is a frequently used function, and in this case I think we'd better do it sooner than later.

So weed to fix the warning for the `get()` operator override, and then remove all existing `TYPE_MISMATCH_WARNING_FOR_INCORRECT_CAPTURE_APPROXIMATION` suppresses in this PR

PR #2975, opened 2023-04-23
Tidy/remove unnecessary suppressions

aSemy:
* Remove unused imports
* Remove unnecessary `@Suppress("FunctionName")` - function names are allowed to be CamelCase if they are [factory functions](https://kotlinlang.org/docs/coding-conventions.html#function-names)
   factory functions used to create instances of classes can have the same name as the abstract return type
   In cases where a suppress was required, I moved the annotation to be on the actual function, so as not to suppress warnings in other functions that might be invalid.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,4 @@
 import org.jetbrains.ValidatePublications
-import org.jetbrains.kotlin.gradle.tasks.*
 import org.jetbrains.publicationChannels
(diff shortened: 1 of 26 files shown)


IgnatBeresnev (approved):
Thanks!!

(I also prefer more isolated/narrowed down suppresses, been bitten a few times by file-wide ones already...)

PR #2983, opened 2023-04-25
DGP: set cacheRoot using convention(), not set()

aSemy:
This is a fix for https://github.com/Kotlin/dokka/labels/runner%3A%20Gradle%20plugin

By using `set()` this can override any configuration done in user scripts, but _before_ the `configureEach{}` block is evaluated. `convention()` means that even if `configureEach{}` runs later, it won't override any previous `set()`.

Unfortunately there's no clean way to set the convention of a DirectoryProperty https://github.com/gradle/gradle/issues/23708, hence the weird syntax.


IgnatBeresnev (approved):
:+1: :+1:

PR #3002, opened 2023-05-19
Multi-param methods should not qualify for a setter

### LINKED ISSUE #2992: Presence of a private backing field breaks having multiple Java setters
**Describe the bug**
The second assert fails:
```
@Test
    fun `Can have two setters in Java`() {
        var classlike = """
            |public class HasTwoSettersAndNoPrivateBackingField {
            |   public String setFoo(String foo) { return "Setter One"; }
            |   public String setFoo(String foo, String anotherArg) { return "Setter Two"; }
            |}
        """.render(java = true).page("HasTwoSettersAndNoPrivateBackingField").data.content
        var setListenerMethods =
            classlike.data.publicFunctionsSummary.data.items.filter { it.name() == "setFoo" }
        assertThat(setListenerMethods.size).isEqualTo(2)

        classlike = """
            |public class HasTwoSettersAndPrivateBackingField {
            |   private String foo;
            |   public String setFoo(String foo) { return "Setter One"; }
            |   public String setFoo(String foo, String anotherArg) { return "Setter Two"; }
            |}
        """.render(java = true).page("HasTwoSettersAndPrivateBackingField").data.content
        setListenerMethods =
            classlike.data.publicFunctionsSummary.data.items.filter { it.name() == "setFoo" }
        assertThat(setListenerMethods.size).isEqualTo(2)
    }
```
**Expected behaviour**
The second assert passes

**Screenshots**
Screenshot 2023-05-03 at 7 15 37 AM
Screenshot 2023-05-03 at 7 16 02 AM

**Installation**
- Dokka version: 1.8.10-dev-203

This is very related to https://github.com/Kotlin/dokka/issues/2603

IgnatBeresnev:
Fixes #2992

Haven't tested it extensively, but I'm pretty sure that's where the problem is

### DIFF
--- a/plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
+++ b/plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
@@ -387,15 +387,15 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
     }
 
     @Test
-    fun `should preserve regular functions that look like accessors, but are not accessors`() {
+    fun `should preserve regular functions that are named like getters, but are not getters`() {
         testInline(
             """
             |/src/main/java/test/A.java
             |package test;
             |public class A {
-            |    public int a = 1;
-            |    public void setA() { } // no arg
+            |    private int a = 1;
             |    public String getA() { return "s"; } // wrong return type
+            |    public String getA(String param) { return "s"; } // shouldn't have params
             |}
         """.trimIndent(),
             configuration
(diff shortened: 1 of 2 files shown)


owengray-google:
I would suggest a test with
```
private field foo: Shape
public setFoo(new: Triangle)
public setFoo(new: Square)
```
to make sure that the issue isn't with having multiple setters.

IgnatBeresnev:
@owengray-google  good call, will add it 👍

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -10,17 +10,16 @@ import org.jetbrains.dokka.model.doc.*
...
 import org.junit.jupiter.api.Test
 import utils.assertNotNull
 
 class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
-    @Suppress("DEPRECATION") // for includeNonPublic
     val configuration = dokkaConfiguration {
         sourceSets {
             sourceSet {
                 sourceRoots = listOf("src/main/java")
-                includeNonPublic = true
```
IgnatBeresnev, line 23: It wasn't obvious that the default configuration had this set to true.. 

it bit me and it's not desired in the majority of the tests, so I made it public-only by default, and added separate local configurations for where private api is needed

plugins/base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -387,35 +386,176 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
     }
 
     @Test
-    fun `should preserve regular functions that look like accessors, but are not accessors`() {
```
IgnatBeresnev, line 390: I realized that the old test didn't make much sense as it was

If the field is public, no getters and setters are set for it (there's a separate test for that). If the field is private, but the getter doesn't qualify, the setter should stay a regular function (added a separate test for that).

So I refactored this one to check specifically for getter lookalikes as it wasn't covered

IgnatBeresnev:
@owengray-google could I ask you to review the PR, please? I added the test cases you suggested (if I got the idea right), as well as some others I could think of on the spot

asfalcone (commented):
Approved!

Issue #2999, opened 2023-05-17
Can I give order to multiple Markdown files?


eunwoo-park-nhn:
**Question**
During the build of DokkaHtmlMultiModule task,
I have multiple Markdown (.md) files, Dokka combining them into a single file when I write like this.
```
tasks.dokkaHtmlMultiModule {
    moduleName.set("Dokka test")
    includes.from("README.md", "DeprecatedAPIs.md")
}
```
However, I'm wondering is there a way to give order to these documents?

**Installation**
- Operating system: macOS
- Build tool: dokka-gradle-plugin
- Dokka version: 1.8.10

IgnatBeresnev:
Hi! I'm not sure what exactly mean or want to achieve by ordering the markdown documents.. 

Could you describe your use case? What's the content of the files, what is Dokka generating and how would you like to see it instead?

eunwoo-park-nhn:
Hello, 

When I write below, then Dokka combiles 2 markdown pages and creates a single page for intro.
```
tasks.dokkaHtmlMultiModule {
    includes.from("README.md", "DeprecatedAPIs.md") 
}
```


However, the current behavior is DeprecatedAPIs.md's content displayed first, followed by the content of README.md.
But what I want is, for the contents of README.md to appear first, followed by the content of DeprecatedAPIs.md.

I would like the content to be merged in the order I specified in includes.from().

IgnatBeresnev:
Could you please provide the contents of those files so that I can replicate it locally, or make a reproducer project / screenshots?

I don't think the order is defined by the order of the files at all, the content is most likely sorted alphabetically by the names of the modules/packages, but there might be something I'm missing or I'm not aware of.

eunwoop:
@IgnatBeresnev 
It seems content is sorted by hash values because it's Set.
I made a simple PR for this. Could you please review this?
I replace Set to LinkedHashSet for keep order what specified in includes.from().

eunwoop:
The files I sent you do not contain # Module or # Package headers because they are not for child modules, 
they are for top level first intro page. So they don't produce any exceptions in my environment. Did you set files for top level for tasks.dokkaHtmlMultiModule's include?
And I will share my screenshot.

Anyway I don't mean current state is problem, I just want if there is feature for ordering markdown files for intro(first) page for dokkaHtmlMultiModule. and I want to make this change with my PR

> What I want is when I write like this
```
includes.from("README.md", "DeprecatedAPIs.md") 
```
[...]

PR #3006, opened 2023-05-22
DokkaConfiguration: Use LinkedHashSet instead of Set to preserve order for includes when deserialize (#2999)

eunwoop:
* for markdown files for multi module description

### DIFF
--- a/core/src/main/kotlin/utilities/json.kt
+++ b/core/src/main/kotlin/utilities/json.kt
@@ -13,6 +13,7 @@ import com.fasterxml.jackson.core.type.TypeReference as JacksonTypeReference
 private val objectMapper = run {
     val module = SimpleModule().apply {
         addSerializer(FileSerializer)
+        addAbstractTypeMapping(Set::class.java, LinkedHashSet::class.java)
     }
     jacksonObjectMapper()
         .registerModule(module)
--- a/core/src/test/kotlin/utilities/JsonKtTest.kt
+++ b/core/src/test/kotlin/utilities/JsonKtTest.kt
@@ -1,5 +1,6 @@
 package utilities
 
+import org.jetbrains.dokka.utilities.parseJson
 import org.jetbrains.dokka.utilities.serializeAsCompactJson
 import org.jetbrains.dokka.utilities.serializeAsPrettyJson
 import kotlin.test.assertEquals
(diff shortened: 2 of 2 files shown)


IgnatBeresnev:
Hi! Does it resolve #2999?

eunwoop:
It’s because DokkaConfigurationImpl is serialized and then deserialized using JSON and the order is not preserved during that process.

When I print the log before serialization, the order of elements in the `includes` is preserved. However, after deserializing the JSON string and using a `Set` type for the `includes` field, the order is not preserved.
I guess Jackson does not guarantee the order of fields for `Set` types.
So when I changed the `includes` field's type to `LinkedHashSet`, it preserves the order of elements during deserialization.

The serialization and then deserialization happens here --> 

In AbstractDokkaTask, the `buildDokkaConfiguration()` method returns a `DokkaConfiguration` object, and then `toCompactJsonString()` method is called to convert it into a JSON string.
```
    @TaskAction
    internal open fun generateDocumentation() {
        DokkaBootstrap(runtime, DokkaBootstrapImpl::class).apply {
            configure(buildDokkaConfiguration().toCompactJsonString(), createProxyLogger())
```

And then inside `DokkaBootstrapImpl.configure()`, the `DokkaConfigurationImpl` is created with the serialized JSON configuration.
```
    override fun configure(serializedConfigurationJSON: String, logger: BiConsumer<String, String>) = configure(
        DokkaProxyLogger(logger),
        DokkaConfigurationImpl(serializedConfigurationJSON) // HERE
    )
```

IgnatBeresnev:
> However, after deserializing the JSON string

Oh yeah, I completely forgot about it.. what a strange bug... Thanks for diving deep and researching it!

Using `LinkedHashSet` and breaking the `Set` abstraction kind of feels bad, and it'll be a breaking API change for some of the clients (since this is public API).. 

Please, give me a few days to discuss this with my colleagues, maybe we can think of something. I'll try to get back to you on Friday

eunwoop:
@IgnatBeresnev
Hello, I updated my changes to avoid breaking API change.
And instead, I use addAbstractTypeMapping for objectMapper.

eunwoo-park-nhn:
@IgnatBeresnev Hello I updated the PR. Could you reveiw this?

IgnatBeresnev:
Hi! Yes, sorry for the delay and thank you for not giving up :)

IgnatBeresnev (approved):
Neat solution, and I think it makes sense!

If you have the time, could you please add a simple test to `core/src/test/kotlin/utilities/JsonKtTest.kt` for the `parseJson` function? Just to make sure it doesn't break in the future

IgnatBeresnev:
Thanks!

Issue #2873, opened 2023-02-16
Dump DokkaConfiguration as JSON when running with the debug logging level


IgnatBeresnev:
When Dokka is run with the debug logging level via Gradle/Maven/CLI runners, it should dump a pretty-printed JSON string of the used `DokkaConfiguration`.

This would be very helpful when debugging various user problems as a lot of the times bugs occur due to invalid configuration.

This would also help with comparing configurations across projects/versions and allow us to manually check the compatibility on configuration level.

aSemy:
Could this be done at a runner level? I have a couple of suggestions for dumping the config file in the Gradle plugin.

1. make [`fun buildDokkaConfiguration()`](https://github.com/Kotlin/dokka/blob/d9c8c6c0035ad59dc0a5495f3247205bb8553295/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt#L221) public, but guarded with `@InternalDokkaApi`
    
    ```kotlin
    @InternalDokkaApi
    abstract fun buildDokkaConfiguration(): DokkaConfigurationImpl
    ```
    
    This would allow users to add some custom config to dump the file:
    
    ```kotlin
    tasks.withType<AbstractDokkaTask>().configureEach {
        val configJsonFile = temporaryDir.resolve("dokka-config.json")
        outputs.file(configJsonFile).withPropertyName("configJsonFile")
        
        doLast {
            @OptIn(InternalDokkaApi::class)
            val dokkaConfig = buildDokkaConfiguration()
            configJsonFile.parentFile.mkdirs()
            configJsonFile.createNewFile()
            configJsonFile.writeText(dokkaConfig.toJsonString())
            logger.lifecycle("$path Dokka Config JSON: ${configJsonFile.toURI()}")
        }
    }
    ```
2. Formalise the above with a custom property in `AbstractDokkaTask`

    ```kotlin
    /**
     * For debugging purposes, the [DokkaConfiguration] JSON used to run [DokkaGenerator] can be
     * dumped to a file.
     *
     * To disable this behaviour, set the value of this property to `null`.
     */
    @get:Optional
    @get:OutputDirectory
    @InternalDokkaApi
    abstract val dokkaConfigurationJson: RegularFileProperty
    ```

    and then create a new function that will dump the JSON to file, if the property has a value

    ```kotlin
    @TaskAction
    internal open fun generateDocumentation() {
[...]

Issue #2936, opened 2023-03-21
Investigate possibility to avoid kotlinx.coroutines dependency


qwwdfsad:
Currently, Dokka actively depends (and exposes in public API via `AsyncSourceToDocumentableTranslator`) on `kotlinx.coroutines` for particular analysis tasks.

Historically, it was done as the natural way to speed up the rendering stage because every declaration (function, property, constant etc.) had its own page, and each page contained navigation data (thus the IO contention).
This no longer holds true, but coroutines are still here.

It has multiple downsides:

* Taking into account Dokka's internal coupling (=> inability to have a clear boundary on what should and shouldn't be thread-safe), it leads to classic concurrent bugs: #2926, #2698, #2308 that are hard to prevent and test 
* It contributes non-necessary dependency to the Dokka (whatever isolation mechanism we'll choose, it's still a bunch of classes to load and compile)
* It doesn't interfere well with workers' API. #2903 will bring not only workers but also natural Gradle-level parallelism, meaning that an additional layer of parallelism within individual tasks is likely to slow down the overall execution.

Ideally, we should prototype the solution (note that it couldn't be "coroutines in single-threaded mode", it has to be a complete removal) and measure the potential impact -- e.g. compare execution times of *individual* tasks on stdlib and kotlinx.coroutines, and decide whether slowdown is acceptable

IgnatBeresnev:
Might resolve: #2926

aSemy:
I have experimented with this locally, and anecdotally I find it improves performance in the Dokkatoo tests.

I prototyped it by find/replacing the coroutines functions with the `run {}` scope function

* `launch {` ➡ `run {`
* `async {` ➡ `run {`
* `coroutineScope {` ➡ `run {`
* etc...

and fixing any resulting issues.

I could share my work in a PR if it would help?

PR #3008, opened 2023-05-23
Add the ability to get serialized configuration of a task

IgnatBeresnev:
A proposal for an intermediate solution for #2873, which should help with debugging configuration and compatibility issues.

cc @aSemy

### DIFF
--- a/runners/gradle-plugin/api/gradle-plugin.api
+++ b/runners/gradle-plugin/api/gradle-plugin.api
@@ -174,7 +174,8 @@ public final class org/jetbrains/dokka/gradle/SourceSetKotlinGistConfigurationKt
 }
 
 public final class org/jetbrains/dokka/gradle/internal/AbstractDokkaTaskExtensionsKt {
-	public static final fun buildJsonConfiguration (Lorg/jetbrains/dokka/gradle/AbstractDokkaTask;)Ljava/lang/String;
+	public static final fun buildJsonConfiguration (Lorg/jetbrains/dokka/gradle/AbstractDokkaTask;Z)Ljava/lang/String;
+	public static synthetic fun buildJsonConfiguration$default (Lorg/jetbrains/dokka/gradle/AbstractDokkaTask;ZILjava/lang/Object;)Ljava/lang/String;
 }
 
 public final class org/jetbrains/dokka/gradle/kotlin/KotlinClasspathUtilsKt {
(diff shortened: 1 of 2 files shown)


runners/gradle-plugin/api/gradle-plugin.api
```diff
@@ -173,6 +173,10 @@ public final class org/jetbrains/dokka/gradle/SourceSetKotlinGistConfigurationKt
 	public static final fun configureWithKotlinSourceSet (Lorg/jetbrains/dokka/gradle/GradleDokkaSourceSetBuilder;Lorg/jetbrains/kotlin/gradle/plugin/KotlinSourceSet;)V
 }
 
+public final class org/jetbrains/dokka/gradle/internal/AbstractDokkaTaskExtensionsKt {
+	public static final fun buildJsonConfiguration (Lorg/jetbrains/dokka/gradle/AbstractDokkaTask;)Ljava/lang/String;
+}
+
```
aSemy, line 179: Shouldn't this be excluded from the API dump because it's marked as `InternalDokkaApi`?

aSemy:
Ahh clever solution, I like it.

I have a couple of observations.

(btw you linked the wrong task in the description, it should be #2873, not #2936)

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/internal/AbstractDokkaTaskExtensions.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Serializes [DokkaConfiguration] of this [AbstractDokkaTask] as pretty json
+ *
+ * Should be used for debugging only, no guarantees are given for the support of this API.
+ *
+ * Better alternative should be introduced as part of [#2873](https://github.com/Kotlin/dokka/issues/2873).
+ */
+@InternalDokkaApi
+fun AbstractDokkaTask.buildJsonConfiguration(): String {
+    return this.buildDokkaConfiguration().toPrettyJsonString()
+}
```
aSemy, line 18: Did you consider adding a parameter, `pretty: Boolean`, and then conditionally returning `toPrettyJsonString()` or `toCompactJsonString()`? 

It's fine as it is, and I don't see a strong argument for returning compact JSON. I'm just curious.

IgnatBeresnev: Good point, definitely won't hurt 👍 will add it

IgnatBeresnev:
@aSemy could I ask you to review the PR since you've already had a look? :) I've implemented the pretty print suggestion

aSemy (commented):
I approve this PR, but GitHub won't let me tick the 'Approve' button 

image

IgnatBeresnev:
I was actually wondering about that..

Thanks!

Issue #2779, opened 2022-12-15
Redesign platform / source set selector


IgnatBeresnev:
Current selector that can be found in the header causes a lot of questions, it's unclear how to use it and what it's supposed to achieve.

![platform-selector](https://user-images.githubusercontent.com/22685910/207959114-d6e10d37-b4a9-4590-961d-392a14661fce.png)

It should be redesigned with the following questions in mind:

1. Which problem / use case should it help solve?
2. Do platforms need to be colored? If so, how should colors be chosen in complex configurations with a lot of shared source sets, such as [in skikio](https://github.com/JetBrains/skiko/blob/a15632cd800ddeed2aa0b904c2e7690ee47863be/skiko/build.gradle.kts)

Notes:

* Currently, source set dependencies are not handled in any way, and it's difficult to understand which tabs are related and should be enabled/disabled together. For example, if I'm developing a `macos` application, I most likely want to see declarations from `macos` and other source sets that `macos` depends on. However, with the current selector, it's not always clear. I can assume I need `native` and `common`, but there's no confirmation. It would help if it was automatic (i.e `native` and `common` are selected as well if you click on `macos`) or if there was a graph/list of dependencies somewhere.

Related issue: #2696

IgnatBeresnev:
Two more things to consider / address:

* Users might use the selector to determine module/package platform, as in some cases it's the only indicator.

![2023-02-01_21-07-55](https://user-images.githubusercontent.com/22685910/216151830-e45f47ce-19c6-4287-8d01-8d8db1c3ddc9.png)

___

* What should be the order in which platforms/source sets are enumerated in one line? It seems that currently it is not sorted by name only, but even if it was - sorting by the name might not be logical.

![image](https://user-images.githubusercontent.com/22685910/216151107-61e48277-d492-4112-93dd-8a65333b882f.png)

ilya-g:
I've noticed that source set bubble sort order is different in header and in packages:
![image](https://user-images.githubusercontent.com/4257577/216421967-7e8a2057-cf54-43e8-9da3-ef61cda168cf.png)

IgnatBeresnev:
Could be useful: #2906

IgnatBeresnev:
Related: #1549

jeffdgr8:
I'll add some input for how my library behaves currently and what I would expect the behavior to be.

This is my library's source set hierarchy:
```
                     ┌────────┐
                     │ common │
                     └───┬────┘
[...]

PR #3019, opened 2023-05-30
Fix invisible packages for module names with space

### LINKED ISSUE #3011: Custom module name leads to hidden packages for maven runner
Configure the Dokka Maven plugin with custom module name:
```xml
<plugin>
                <groupId>org.jetbrains.dokka</groupId>
                <artifactId>dokka-maven-plugin</artifactId>
                <version>1.8.10</version>
<!--                <version>1.7.20</version>-->
                <executions>
                    <execution>
                        <phase>pre-site</phase>
                        <goals>
                            <goal>dokka</goal>
                        </goals>
                    </execution>
                </executions>

                <configuration>
                    <moduleName>Dokka Maven Project</moduleName>
                </configuration>
            </plugin>
```

As a result, rendered documentation will filter out all sources:
image

The issue appeared in `1.8.10`. For `1.7.20`, it works fine. Also, it does not reproduce for Gradle runner.

IgnatBeresnev:
Fixes #3011 

Read my comment to understand what exactly the problem was: https://github.com/Kotlin/dokka/issues/3011#issuecomment-1568620493

There are better ways to resolve this problem, but this is the only solution that I could think of that wouldn't lead to a lot of other regressions and problems (including for Gradle users), because the value of `scopeId` is used in other parts of the code as well, such as sorting and anchor links, so changing it would lead to regression in behaviour.

One could argue that module names should not contain spaces - I would agree, but the `moduleName` property is used as the parent project name, so the desire to change it to something human-readable is quite understandable, and if we fix it know it's likely going to lead to bug reports. It really needs to be a separate property..

Ideally, a better solution should be implemented as part of #2779 - the current logic is quite error-prone and difficult to navigate..

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -215,7 +215,7 @@ open class HtmlRenderer(
             node.isAnchorable -> buildAnchor(
                 node.anchor!!,
                 node.anchorLabel!!,
-                node.sourceSetsFilters
+                node.buildSourceSetFilterValues()
             ) { childrenCallback() }
             node.extra[InsertTemplateExtra] != null -> node.extra[InsertTemplateExtra]?.let { templateCommand(it.command) }
                 ?: Unit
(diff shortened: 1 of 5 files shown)


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -982,8 +987,5 @@ private val PageNode.isNavigable: Boolean
 private fun PropertyContainer<ContentNode>.extraHtmlAttributes() = allOfType<SimpleAttr>()
 private fun PropertyContainer<ContentNode>.extraTabbedContentType() = this[TabbedContentTypeExtra]
 
-private val ContentNode.sourceSetsFilters: String
-    get() = sourceSets.sourceSetIDs.joinToString(" ") { it.toString() }
```
IgnatBeresnev, line 986: Before every table-row, there's an invisible link - I think it's used by the anchor link to focus on the right row. This property was used for the filter value in it, but I couldn't understand if it was actually needed to be present - made no difference upon testing.

atyrin (approved):
👌

PR #3021, opened 2023-05-30
Incorrectly handles a click on a method and other elements in the mobile version

zoobestik:
https://youtrack.jetbrains.com/issue/KTL-1028

The first click on a method in the mobile version, the copy icon appears.


IgnatBeresnev (approved):
:+1: :+1:

PR #3030, opened 2023-06-04
Refactor publishing configuration

aSemy:
This is an https://github.com/Kotlin/dokka/labels/infrastructure issue, part of #2703, specifically [this comment](https://github.com/Kotlin/dokka/issues/2703#issuecomment-1575494149)

This PR is larger than I intended, I started picking things apart and I couldn't stop! I think it's worth having it all in one place so you can have a general overview, but just give me the word and I'll split out anything worth investigating into a separate PR.

I've done some preliminary tests using `./gradlew publishAllPublicationsToSpaceDokkaDevRepository --dry-run`, but of course this isn't a real-life scenario.

### Summary

- ⚠️ rename `dokka_publication_channel`, `dokka_publication_channels` to `org.jetbrains.dokka.publicationChannels`, so the properties are consistent
- add `publicationChannels` do `DokkaBuildProperties`, so it's more easily accessible
- refactor `registerDokkaArtifactPublication()` helper
- add docs to `ValidatePublications` task
- move `dokka_build` val to `DokkaBuildProperties`, but don't start updating usages yet
- refactor `projectsWithoutOptInDependency` to be more generic
- migrate `ValidatePublications` checks into `maven-publish.gradle.kts` (which is just a messy first draft and could use with a re-write)

### Additional changes:

- move integration test properties (parallelism and exhaustive) into DokkaBuildProperties
- re-write 'disable DokkaHtml for MavenLocal' check to use config similar to the [Restricting publications to specific repositories](https://docs.gradle.org/current/userguide/publishing_customization.html#sec:publishing_maven:conditional_publishing) docs.

PR #2919, opened 2023-03-13
Update com.gradle.plugin-publish 1.2.0

Goooler:

PR #3031, opened 2023-06-05
Fix publishing to Gradle Plugin Portal

IgnatBeresnev:
This seems to be a regression from #2912. It was blocking the release, but I [commited this fix](https://github.com/Kotlin/dokka/commit/46c10ac2b169a007aefdad3c9720031a8e05d784) directly into the release branch, so no rush now.

Without this change, running `./gradlew publishPlugins` results in:

```
> Task :publishPlugins FAILED

* What went wrong:
Execution failed for task ':publishPlugins'.
> No plugins defined. Please declare at least one plugin in a pluginBundle plugins block

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org
```

With this change, the artifacts are built and published as expected (well, time will tell).

As far as I understand, we need to define the plugin use in the root project so that it can be used later on in the Gradle runner: 

https://github.com/Kotlin/dokka/blob/2b73ddd36d226e5547398daaf0e4370e48f4bb46/runners/gradle-plugin/build.gradle.kts#L7

If the plugin application is removed from the root project, then the line above in the Gradle runner will fail to resolve the plugin ("plugin was not found in any of the following sources").

So I'm not sure that what I'm suggesting is the best fix, looks like it could be improved, but it works for now. Perhaps, it should be revisited in https://github.com/Kotlin/dokka/pull/2919 or https://github.com/Kotlin/dokka/pull/3030. 

If a better alternative is suggested, I can try to implement it as part of this PR, but as long as it doesn't take a lot of time or a lot of re-writing - then it's best addressed separately.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
     id("org.jetbrains.conventions.dokka")
 
     alias(libs.plugins.kotlinx.binaryCompatibilityValidator)
-    alias(libs.plugins.gradlePublish)
+    alias(libs.plugins.gradlePublish) apply false // see #xx for "apply false" reasoning
     alias(libs.plugins.nexusPublish)
 }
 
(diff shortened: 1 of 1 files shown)


IgnatBeresnev:
It seems like there's no better way as of now :(

PR #3046, opened 2023-06-14
Fix sources not being linked when using Maven

freya022:
This PR transforms source links path into canonical paths, as the paths being checked when getting source links are absolute

This fixes sources not being linked when using relative paths in Maven projects, that was caused by `SourceLinksTransformer#resolveSources` comparing absolute paths to relative paths, when Maven was used.


IgnatBeresnev (approved):
Good job, thanks for fixing it!

Sorry it took so long to get to - 4 months for a PR is not that uncommon in the open source world, eh? :sweat_smile:

PR #3051, opened 2023-06-26
Generate dedicated pages for typealiases

vmishenev:


plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
```diff
@@ -329,7 +348,7 @@ open class DefaultPageCreator(
         sortedWith(compareBy({ it.name }, { it.parameters.size }, { it.dri.toString() }))
 
     /**
-     * @param documentables a list of [DClasslike] and [DEnumEntry] with the same dri in different sourceSets
+     * @param documentables a list of [DClasslike] and [DEnumEntry] and [DTypeAlias] with the same dri in different sourceSets
      */
     protected open fun contentForClasslikesAndEntries(documentables: List<Documentable>): ContentGroup =
```
IgnatBeresnev, line 353: We definitely need to revisit these functions at some point, it gets very confusing with all the `and` and `or` functions.. Right now `pageForClasslikes**Or**TypeAlias` calls `pageForClasslikes**And**TypeAliases`, which calls `contentForClasslikesAnd**Entries**` - all public API

It would probably be quite a bit of work to refactor this, and it's old technical debt anyway, so I'm not pursuing you to refactor it in this PR, but if you have the time and desire to do it now - we can jump on a call and discuss what can be done on a high level, so that at least the public API is clear and concise, and make the `and` and `or` functions private.

IgnatBeresnev (approved):

plugins/base/src/main/kotlin/translators/documentables/DefaultPageCreator.kt
vmishenev: For now, I revert these names since renaming does not make sense in the current documentable model

PR #3053, opened 2023-06-28
Enhance typealias presentation

vmishenev:


plugins/base/src/main/kotlin/transformers/documentables/ActualTypealiasAdder.kt
```diff
@@ -6,6 +6,13 @@ import org.jetbrains.dokka.model.properties.WithExtraProperties
 import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.transformers.documentation.DocumentableTransformer
 
+/**
+ * Since we can not merge [DClasslike] and [DTypeAlias] by design,
+ * we have this transformer to add [ActualTypealias] extra in expect [DClasslike]
+ *
+ * The transformer should be applied after merging all documentables
+ */
+// TODO [DTypeAlias.expectPresentInSet] always = null, for actual typeAlias it should be no null.
```
IgnatBeresnev, line 15: 👍👍👍 

Any details for why it was done this way and/or why we can't merge `DClasslike` and `DTypeAlias`? That is, what's the design limitation / idea behind it?

It would also be nice to have a call to action in the TODO (with a verb of some sort), right now you've described the behaviour, but it's not very clear what exactly needs to be done. If I understand correctly, the TODO is to make `expectPresentInSet` not null in actual typealiases?

IgnatBeresnev (approved):

Issue #2289, opened 2021-12-22
Redundant dot in the beginning of the path for items in the root package in search block


atyrin:
Extra dot at the beginning of the pathname for items the root package

![image](https://user-images.githubusercontent.com/7689049/147120501-ca4bd309-96bf-4e48-b71e-1ab8cdffff06.png)

**Installation**
- Dokka version: 1.6.10

PR #3057, opened 2023-07-03
HTML: fix of redundant dot and related autotest (#2289)

d-ambatenne:


plugins/base/src/test/kotlin/renderers/html/SearchbarDataInstallerTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
+import org.junit.Test
+import org.junit.jupiter.api.Assertions
+import utils.TestOutputWriterPlugin
+import utils.pagesJson
+
+class SearchbarDataInstallerTest: BaseAbstractTest() {
+
+    @Test
+    fun `should be displayed without leading dot`() {
```
IgnatBeresnev, line 12: The detail that this bug is reproduced only with root-level declarations (those without a package) is not obvious, so I would add it to the name of the test and/or link to the bug

```suggestion
    @Test // see #2289
    fun `should display description of root declarations without a leading dot`() {
```

IgnatBeresnev (approved):
Well done 👍

PR #3062, opened 2023-07-07
Update Kotlin compiler to 1.8.20

vmishenev:


subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/configuration/AnalysisEnvironment.kt
```diff
@@ -528,7 +530,14 @@ class AnalysisEnvironment(
     }
 
     // Set up JDK classpath roots explicitly because of https://github.com/JetBrains/kotlin/commit/f89765eb33dd95c8de33a919cca83651b326b246
-    internal fun configureJdkClasspathRoots() = configuration.configureJdkClasspathRoots()
```
IgnatBeresnev, line 531: Was this function removed from the compiler with no replacement, or what was the motivation to copy paste the body?

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/configuration/resolve/DokkaJsResolverForModuleFactory.kt
```diff
@@ -42,6 +42,10 @@ internal class DokkaJsResolverForModuleFactory(
         )
     }
 
+    @Deprecated(
```
IgnatBeresnev, line 45: These modules are 100% internal, I think there's no real benefit in deprecating things, and they can just be removed. If you delete it, make sure to also delete the other deprecated `createResolverForModule`, for some reason there are two..

IgnatBeresnev (approved):

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/configuration/AnalysisEnvironment.kt
vmishenev: What do you mean? Whose body?
It copies nothing.
`configuration.configureJdkClasspathRoots()` still exists, it is used here too.

IgnatBeresnev: I don't understand where the rest of the new code in this function came from, and why it's needed :) As far as I can see, it [already checks](https://github.com/JetBrains/kotlin/blob/1.9.0/compiler/cli/cli-base/src/org/jetbrains/kotlin/cli/jvm/config/JvmContentRoots.kt#L94) for `java.home`

vmishenev: > I don't understand where the rest of the new code in this function came from, and why it's needed :)

It's needed for https://github.com/JetBrains/kotlin/blob/45b1c33ed8a33c92207d1bd9f2e83738995836ae/compiler/cli/cli-base/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment.kt#L208
The compiler does the same  [here](https://github.com/JetBrains/kotlin/blob/45b1c33ed8a33c92207d1bd9f2e83738995836ae/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/jvmArguments.kt#L149)

> As far as I can see, the called compiler function [already checks](https://github.com/JetBrains/kotlin/blob/1.9.0/compiler/cli/cli-base/src/org/jetbrains/kotlin/cli/jvm/config/JvmContentRoots.kt#L94) for java.home

It does not put `JVMConfigurationKeys.JDK_HOME` in the configuration.

PR #3064, opened 2023-07-07
Set the correct testLauncher property in workflows

IgnatBeresnev:
The `mainCompiler` version was updated to `11` in https://github.com/Kotlin/dokka/pull/3062/commits/80fc13b5906c9d500605bb414d0529f9141ff03f, which should've caused the [thorough tests](https://github.com/Kotlin/dokka/actions/runs/5483737953) to fail when run under Java 8, but all tests passed successfully.

Had a look at why and noticed that the property name was a bit off, so it should hopefully fix it.

### DIFF
--- a/.github/workflows/tests-smoke.yml
+++ b/.github/workflows/tests-smoke.yml
@@ -6,6 +6,9 @@ concurrency:
   group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
   cancel-in-progress: true
 
+env:
+  JAVA_TEST_VERSION: 11
+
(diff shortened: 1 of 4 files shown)


IgnatBeresnev:
There were two problems:

1. `javaToolchain.test` should've been `javaToolchain.testLauncher`
2. Environment variables don't play nice with dots in the key name, see https://github.com/gradle/gradle/issues/1295

On the second problem, I played around with it, but didn't find a nice-looking way that would work. To avoid renaming properties or introducing magic, I migrated the scripts to passing project properties via `-P`.

vmishenev (approved):

Issue #1217, opened 2020-07-23
Consider supporting Gradle's configuration cache


eskatos:
The configuration cache is a feature that significantly improves build performance by caching the result of the configuration phase and reusing this for subsequent builds.

Implementing support for the configuration cache in this plugin would improve users experience.

See https://docs.gradle.org/nightly/userguide/configuration_cache.html

The configuration cache is introduced in Gradle 6.6.
This can already be tested using `6.6-rc-2`.

eskatos:
Given this plugin is currently being deeply reworked I didn't bother reporting what exact issues the "old" dokka or the "new" dokka have. Running a build with `--configuration-cache` will produce a report of the found problems.

sellmair:
This is awesome. We try to implement this (probably even with #1194 )

eskatos:
FYI, Kotlin `master` branch should now work with the configuration cache without emitting problems. You can use dev builds >= `1.4.20-dev-3361` in your tests.

3flex:
There are still a couple of warnings when using the configuration cache with dokka plugin. The configuration cache report points to `plugins` and `runtime` properties in `AbstractDokkaTask`, because `Configuration` type is not serializable.

https://github.com/Kotlin/dokka/blob/ac4b818a8616e4de8f702167fe6ddbe911df5d24/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/AbstractDokkaTask.kt#L70-L74

I think the best thing is to setup the configurations when the tasks are created and set the classpath files there, but I'm not familiar with the dokka Gradle plugin internals and it's a bit more complex than I'm comfortable with otherwise I would attempt the change. But if `runtime` and `plugins` can be changed to `FileCollection` or `ConfigurableFileCollection` type then configuration cache should then be able to handle it.

JavierSegoviaCordoba:
Any new update?

emartynov:
It looks like there is no update on the configuration cache yet.

adam-enko:
Hi all, I have an update.

We're planning to release a new version of Dokka Gradle Plugin (DGPv2 for short) in the [Dokka 2.0.0](https://github.com/Kotlin/dokka/milestone/31) release. DGPv2 is based on [Dokkatoo](https://github.com/adamko-dev/dokkatoo/). The current Dokka Gradle Plugin (DGPv1) will become deprecated.

DGPv2 fully supports configuration cache and build cache!

[...]

PR #3070, opened 2023-07-11
Mark tasks as not compatible with the Gradle configuration cache

BoD:
Since, as for now, Dokka is not compatible with the Gradle configuration cache (see [#1217](https://github.com/Kotlin/dokka/issues/1217)), using it in a project that has it enabled will fail the build. 
Marking the tasks as incompatible avoids these fails.

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
@@ -232,6 +232,7 @@ abstract class AbstractDokkaTask : DefaultTask() {
 
     init {
         group = JavaBasePlugin.DOCUMENTATION_GROUP
+        super.notCompatibleWithConfigurationCache("Dokka tasks are not yet compatible with the Gradle configuration cache. See https://github.com/Kotlin/dokka/issues/1217")
     }
 
     internal fun buildPluginsConfiguration(): List<PluginConfigurationImpl> {
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
@@ -23,6 +23,7 @@ import org.jetbrains.dokka.plugability.DokkaPlugin
 import java.util.concurrent.atomic.AtomicReference
 import java.util.function.BiConsumer
 import kotlin.reflect.full.createInstance
+import kotlin.reflect.full.memberFunctions
 
 @DisableCachingByDefault(because = "Abstract super-class, not to be instantiated directly")
 abstract class AbstractDokkaTask : DefaultTask() {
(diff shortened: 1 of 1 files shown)


IgnatBeresnev (approved):
Thank you for the contribution, very much appreciated!

I'll start Gradle integration tests now, which should take a couple of hours, and if everything's alright - will merge and cherry pick it into the upcoming 1.9.0 release.

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
```diff
@@ -232,6 +232,7 @@ abstract class AbstractDokkaTask : DefaultTask() {
 
     init {
         group = JavaBasePlugin.DOCUMENTATION_GROUP
+        super.notCompatibleWithConfigurationCache("Dokka tasks are not yet compatible with the Gradle configuration cache. See https://github.com/Kotlin/dokka/issues/1217")
```
IgnatBeresnev, line 235: Unfortunately, the integration tests are failing :(

```
Caused by: java.lang.NoSuchMethodError: 'void org.gradle.api.DefaultTask.notCompatibleWithConfigurationCache(java.lang.String)
```

The used method (`notCompatibleWithConfigurationCache`) was introduced in Gradle 7.4, whereas we still have to keep compatibility with Gradle 6.9+ (for as long as Kotlin itself supports it).

We probably need to use something like [plugin variants](https://docs.gradle.org/8.3/userguide/implementing_gradle_plugins.html#plugin-with-variants) to fix this properly, but there's little point in investing time into the current Gradle plugin - it will be completely re-written anyway.

However, it would still be nice to have this change for the time being. I think we can add a dirty hack - check with reflection that this method exists, and only then call it. What do you think?

Something along the lines of

```kotlin
val containsNotCompatibleWithConfigurationCache = this::class.memberFunctions
    .any { it.name == "notCompatibleWithConfigurationCache" && it.parameters.firstOrNull()?.name == "reason" }

if (containsNotCompatibleWithConfigurationCache) {
    super.notCompatibleWithConfigurationCache("Dokka tasks are not yet compatible with the Gradle configuration cache. See https://github.com/Kotlin/dokka/issues/1217")
}
```

IgnatBeresnev (changes requested):

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/tasks/AbstractDokkaTask.kt
BoD: Yes I think using introspection is a fair workaround :) Feel free to amend the PR (or I can do it if you prefer). 👍

IgnatBeresnev: I don't want to rob you of the valuable contribution :sweat_smile: If you have the time and the desire, I'd be happy if you did it :) I also didn't test the suggestion - I mean, it should work if this function is returned in `memberFunctions`, but who knows

BoD: Thanks! I've just pushed the change. It seems to work on my side but let's see if the CI agrees :)

IgnatBeresnev (approved):
Looking good :+1: 

I had to run the integration tests locally because this branch is quite old. All of the tests are green, so it should be good to go

Once again, thank you for addressing the issue!

Issue #463, opened 2019-04-09
Function documentation not inherited for external super type


knolleCC:
If extending a class or interface from outside my own module, the documentation of overridden functions is not inherited.

Example:

    class MyList : List<String> {
        override fun contains(element: String): Boolean { ... }
        ...
    }

Calling `dokka` produces an empty function doc and the warning:

    No documentation for my.package.MyList$contains(kotlin.String)

IgnatBeresnev:
Seems to be still relevant in 1.7.20, but with a different message.

Configuration:

```kotlin
tasks.withType<DokkaTask>().configureEach {
    dokkaSourceSets.configureEach {
        reportUndocumented.set(true)
    }
}
```

Reproducer:

```kotlin
class MyList(override val size: Int) : List<String> {
    override fun contains(element: String): Boolean = TODO("Not yet implemented")
    override fun containsAll(elements: Collection<String>): Boolean = TODO("Not yet implemented")
    override fun get(index: Int): String = TODO("Not yet implemented")
    override fun isEmpty(): Boolean = TODO("Not yet implemented")
    override fun iterator(): Iterator<String> = TODO("Not yet implemented")
    override fun listIterator(): ListIterator<String> = TODO("Not yet implemented")
    override fun listIterator(index: Int): ListIterator<String> = TODO("Not yet implemented")
    override fun subList(fromIndex: Int, toIndex: Int): List<String> = TODO("Not yet implemented")
    override fun lastIndexOf(element: String): Int = TODO("Not yet implemented")
    override fun indexOf(element: String): Int = TODO("Not yet implemented")
}
```

Warning messages:

```
Undocumented: me.beresnev.dokka.debug/MyList/contains/#kotlin.String/ (jvm)
Undocumented: me.beresnev.dokka.debug/MyList/containsAll/#kotlin.collections.Collection[kotlin.String]/ (jvm)
Undocumented: me.beresnev.dokka.debug/MyList/get/#kotlin.Int/ (jvm)
Undocumented: me.beresnev.dokka.debug/MyList/isEmpty/#/ (jvm)
Undocumented: me.beresnev.dokka.debug/MyList/iterator/#/ (jvm)
Undocumented: me.beresnev.dokka.debug/MyList/listIterator/#/ (jvm)
Undocumented: me.beresnev.dokka.debug/MyList/listIterator/#kotlin.Int/ (jvm)
[...]

PR #3071, opened 2023-07-11
Hardcode documentation for the synthetic Enum.entries property

### LINKED ISSUE #2762: Support Enum.entries
Kotlin 1.8.20-1.9.0 adds a new feature - [Enum.entries](https://youtrack.jetbrains.com/issue/KT-48872)

This is how it works for `Enum.values()`

image

Dokka should render `entries` the same way it rendered `values`

IgnatBeresnev:
Fixes #2762

### DIFF
--- a/plugins/base/src/main/resources/dokka/docs/kdoc/EnumEntries.kt.template
+++ b/plugins/base/src/main/resources/dokka/docs/kdoc/EnumEntries.kt.template
@@ -1,0 +1,3 @@
+Returns a representation of an immutable list of all enum entries, in the order they're declared.
+
+This method may be used to iterate over the enum entries.
--- a/plugins/base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
+++ b/plugins/base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
@@ -918,6 +918,69 @@ val soapXml = node("soap-env:Envelope", soapAttrs,
     }
 
     @Test
+    fun `should have documentation for synthetic Enum entries property`() {
+        testInline(
+            """
+            |/src/main/kotlin/test/KotlinEnum.kt
+            |package test
+            |
+            |enum class KotlinEnum {
+            |    FOO, BAR;
+            |}
+            """.trimIndent(),
+            configuration
+        ) {
+            documentablesMergingStage = { module ->
+                val kotlinEnum = module.packages.find { it.name == "test" }
+                    ?.classlikes
+                    ?.single { it.name == "KotlinEnum" }
+
+                checkNotNull(kotlinEnum)
+
+                val entriesProperty = kotlinEnum.properties.single { it.name == "entries" }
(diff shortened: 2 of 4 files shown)


plugins/base/src/main/resources/dokka/docs/kdoc/EnumEntries.kt.template
```diff
@@ -0,0 +1,3 @@
+Returns a representation of an immutable list of all enum entries, in the order they're declared.
+
+This method may be used to iterate over the enum entries.
```
IgnatBeresnev, line 3: I didn't find the documentation text in the KEEP, and `entries` hasn't been documented anywhere else, so I decided to play it safe and stick to the short description similar to the `values()` function for now.

Should be discussed with @qwwdfsad and the libraries team, maybe there's a KDoc lying around somewhere, or the authors know how they want to document it.

IgnatBeresnev:
As a side note: I noticed that Dokka is unable to parse the KDocs for `ordinal` and `name` properties, even though they are present in [Enum's source code](https://github.com/JetBrains/kotlin/blob/5424c54fae7b4836506ec711edc0135392b445d6/core/builtins/native/kotlin/Enum.kt#L16-L26). I tried to address it in this PR, but realized it wasn't a simple fix, so I think it should be addressed separately.

TLDR is that both `name` and `ordinal` are not synthetic functions and are not fake overrides, but they don't have the source code attached in the descriptor, thus empty documentation. Not sure if it's a bug of the compiler or what.

Should try it with K2 - maybe it's been fixed there already.

vmishenev (approved):

vmishenev:
> even though they are present in [Enum's source code](https://github.com/JetBrains/kotlin/blob/5424c54fae7b4836506ec711edc0135392b445d6/core/builtins/native/kotlin/Enum.kt#L16-L26).

This is like https://github.com/Kotlin/dokka/issues/463. The compiler knows nothing about the source code of builtins.

vmishenev:
> TLDR is that both name and ordinal are not synthetic functions and are not fake overrides,

They are fake overrides for enums and entries.
image

qwwdfsad (approved):

IgnatBeresnev:
> They are fake overrides for enums and entries.

Oh yeah, indeed, thanks for taking a look!

I noticed that they weren't ignored, so I assumed that the result of the [isObvious](https://github.com/Kotlin/dokka/blob/9559158bfeeb274e9ccf1b4563f1b23b42afc493/subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt#L634) check was `false` (thus they must be neither fake override nor synthetic), but I've only now noticed that we handle `isObvious` for functions only, and these are properties, so they are always present in the documentation...

It looks like we need to rename `suppressObviousFunctions` to `suppressObviousMembers` or something similar, and hide `ordinal` and `name` by default (as "obvious") if they don't have any documentation - for consistency with functions.

PR #3074, opened 2023-07-14
Update Kotlin to 1.9.0

vmishenev:


subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -1147,14 +1147,14 @@ private class DokkaDescriptorVisitor(
             fullEnumEntryName(),
             DRI(enumClassId.packageFqName.asString(), fullEnumEntryName())
         )
-        is ConstantsKtClassValue -> when (value) {
-            is NormalClass -> (value as NormalClass).value.classId.let {
+        is ConstantsKtClassValue -> when (val value = value) {
```
vmishenev, line 1150: to  avoid https://youtrack.jetbrains.com/issue/KT-22517/Deprecate-smartcasts-for-local-delegated-properties

build-logic/src/main/kotlin/org/jetbrains/conventions/gradle-plugin.gradle.kts
```diff
@@ -0,0 +1,8 @@
+package org.jetbrains.conventions
+
+plugins {
+    id("org.gradle.kotlin.kotlin-dsl")
+    id("org.jetbrains.conventions.maven-publish")
+    id("org.jetbrains.conventions.kotlin-jvm")
+    id("com.gradle.plugin-publish")
+}
```
IgnatBeresnev, line 8: Was there a need to introduce this convention, did it interfere with the update somehow? 

If not, I have some concerns about why it is needed and whether it breaks something now or in the future (`kotlin-jvm` was previously not part of the gradle runner)

gradle/libs.versions.toml
```diff
@@ -46,6 +47,8 @@ gradlePlugin-shadow = "7.1.2"
 gradlePlugin-nexusPublish = "1.1.0"
 gradlePlugin-gradlePluginPublish = "0.20.0"
 gradlePlugin-gradleNode = "3.5.1"
+gradlePlugin-kotlinDsl= "2.4.1"
+gradlePlugin-pluginPublish = "0.20.0"
```
IgnatBeresnev, line 51: Not relevant if the `gradle-plugin` convention is removed, but it's [not recommended](https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:kotlin-dsl_plugin) to specify the version of `kotlin-dsl`, and it looks like the dependency for `pluginPublish` is [already defined](https://github.com/Kotlin/dokka/pull/3074/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR118) below under the `[plugins]` section (unless these are two different dependencies with the same version)

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
```diff
@@ -55,8 +55,11 @@ private fun KotlinCompilation.classpathOf(project: Project): FileCollection {
 
     val shouldKeepBackwardsCompatibility = (kgpVersion != null && kgpVersion < KotlinGradlePluginVersion(1, 7, 0))
     return if (shouldKeepBackwardsCompatibility) {
-        @Suppress("DEPRECATION_ERROR")
-        kotlinCompile.classpath // deprecated with error since 1.8.0, left for compatibility with < Kotlin 1.7
+        // removed since 1.9.0, left for compatibility with < Kotlin 1.7
+        @Suppress("UNCHECKED_CAST")
+        val classpathGetter= kotlinCompile::class.members
+            .first { it.name == "getClasspath" }
+        classpathGetter.call(kotlinCompile) as FileCollection
```
IgnatBeresnev, line 62: Looks like the suppress is for the wrong statement?
```suggestion
        val classpathGetter = kotlinCompile::class.members
            .first { it.name == "getClasspath" }
          
        @Suppress("UNCHECKED_CAST")
        classpathGetter.call(kotlinCompile) as FileCollection
```

runners/gradle-plugin/src/test/kotlin/org/jetbrains/dokka/gradle/KotlinSourceSetGistTest.kt
```diff
@@ -232,15 +232,10 @@ class KotlinSourceSetGistTest {
...
-            "Expected jvmMain to depend on commonMain by default"
-        )
-
         /* Why not? */
         jvmMainSourceSet.dependsOn(macosMainSourceSet)
         assertEquals(
-            setOf("commonMain", "macosMain"), jvmMainSourceSetGist.dependentSourceSetNames.get(),
+            setOf("macosMain"), jvmMainSourceSetGist.dependentSourceSetNames.get(),
             "Expected dependent source set changes to be reflected in gist"
         )
```
IgnatBeresnev, line 240: Any idea what happened here, is this expected? Do source sets not depend on commonMain anymore?

subprojects/analysis-kotlin-descriptors/build.gradle.kts
```diff
@@ -8,6 +8,10 @@ plugins {
 }
 
 dependencies {
+    // to override some interfaces (JvmAnnotationEnumFieldValue, JvmAnnotationConstantValue) from compiler since thet are empty there
+    api(libs.intellij.java.psi.api) {
+        isTransitive = false
+    }
```
IgnatBeresnev, line 14: Is it actually used anywhere or was there a problem related to it? Don't see any new code in the PR related to the mentioned classes. 

If it is actually needed, does it have to be `api` (i.e exposed to others) or can it be an `implementation`?

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/configuration/resolve/DokkaJsResolverForModuleFactory.kt
```diff
@@ -61,13 +63,14 @@ internal class DokkaJsResolverForModuleFactory(
...
-            targetEnvironment,
-            languageVersionSettings
+            moduleContext = moduleContext,
+            declarationProviderFactory = declarationProviderFactory,
+            bindingTrace = CodeAnalyzerInitializer.getInstance(moduleContext.project).createTrace(), // BindingTraceContext(/* allowSliceRewrite = */ true),
+            platform = moduleDescriptor.platform!!,
+            analyzerServices = JsPlatformAnalyzerServices,
+            targetEnvironment = targetEnvironment,
+            languageVersionSettings = languageVersionSettings,
+            absentDescriptorHandlerClass = null
```
IgnatBeresnev, line 73: Should the function parameter be passed here? Same thing in `DokkaNativeResolverForModuleFactory`

```suggestion
            absentDescriptorHandlerClass = absentDescriptorHandlerClass
```

gradle/libs.versions.toml
vmishenev: `kotlin-dsl` will be replaced with `expectedKotlinDslPluginsVersion`, see  https://github.com/gradle/gradle/issues/17016

> it looks like the dependency for pluginPublish is [already defined](https://github.com/Kotlin/dokka/pull/3074/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR118) below under the [plugins] section (unless these are two different dependencies with the same version)

Oh, I will remove it, for some reason the alias is not used in the runner project

runners/gradle-plugin/src/test/kotlin/org/jetbrains/dokka/gradle/KotlinSourceSetGistTest.kt
vmishenev: > Do source sets not depend on commonMain anymore?

They depend on. 
`project.evaluate()` is needed

build-logic/src/main/kotlin/org/jetbrains/conventions/gradle-plugin.gradle.kts
vmishenev: `kotlin-dsl` added `plugin.sam.with.receiver.gradle.plugin` and other plugins with default version 1.7.10 instead of 1.9.0, KGP should align their version to 1.9.0, but it did not work.

IgnatBeresnev (approved):

PR #3076, opened 2023-07-17
Add a test for parsing data objects

### LINKED ISSUE #3042: Support data objects
The language feature is to be released as stable in 1.9.0, the description can be found in [this KEEP](https://github.com/Kotlin/KEEP/blob/data-objects/proposals/data-objects.md).

Dokka should support rendering the new `data` keywords for objects.

IgnatBeresnev:
Fixes #3042

### DIFF
--- a/plugins/base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
+++ b/plugins/base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
@@ -1,13 +1,13 @@
 package translators
 
 import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.base.signatures.KotlinSignatureUtils.modifiers
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.links.PointingToDeclaration
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.doc.*
-import org.junit.jupiter.api.Assertions.assertEquals
-import org.junit.jupiter.api.Assertions.assertNull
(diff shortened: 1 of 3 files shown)


plugins/base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
```diff
@@ -986,6 +986,30 @@ val soapXml = node("soap-env:Envelope", soapAttrs,
             }
         }
     }
+
+    @Test
+    fun `should add data modifier to data objects`() {
```
vmishenev, line 991: I think it should be in `model.ClassesTest`

vmishenev (approved):

plugins/base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
IgnatBeresnev: It's a good idea to add a test to `mode.ClassesTest`, maybe it might be accidentally modified by one of the transformers or something, as model tests check `documentablesTransformationStage`

I will leave this test here as well, because I want to have tests specifically for parsing documentables before they are merged or transformed, to test the translator layer specifically

IgnatBeresnev: Done

vmishenev: > I will leave this test here as well, because I want to have tests specifically for parsing documentables before they are merged or transformed, to test the translator layer specifically

The problem is more general than specific only for data objects.
As for me, you can modify `ObjectTest` to check documentable before and after the transforming. (run the same action function at the different stages)

PR #3078, opened 2023-07-17
Remove "project dependencies are published" check

IgnatBeresnev:
The check was likely added because bintray did not perform any checks of its own, but Maven Central should actually check that before the staging repository is closed and all the artifacts are published, so it's not needed locally

It's in the way now because it fails with a false positive: analysis modules shadow all dependencies, and the final `.pom` has nothing in the `dependencies` block. But this local check (because it's quite primitive) does not understand it, and thinks we are trying to publish a project with unpublished dependencies.

### DIFF
--- a/build-logic/src/main/kotlin/org/jetbrains/ValidatePublications.kt
+++ b/build-logic/src/main/kotlin/org/jetbrains/ValidatePublications.kt
@@ -27,37 +27,9 @@ open class ValidatePublications : DefaultTask() {
                 .filterIsInstance<MavenPublication>()
                 .filter { it.version == project.dokkaVersion }
                 .forEach { _ ->
-                    checkProjectDependenciesArePublished(subProject)
                     subProject.assertPublicationVersion()
                 }
         }
-    }
-
-    private fun checkProjectDependenciesArePublished(project: Project) {
-        val implementationDependencies = project.findDependenciesByName("implementation")
-        val apiDependencies = project.findDependenciesByName("api")
-
-        val allDependencies = implementationDependencies + apiDependencies
-
-        allDependencies
-            .filterIsInstance<ProjectDependency>()
-            .forEach { projectDependency ->
-                val publishing = projectDependency.dependencyProject.extensions.findByType<PublishingExtension>()
-                    ?: throw UnpublishedProjectDependencyException(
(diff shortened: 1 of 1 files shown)


vmishenev:
> The check was likely added because bintray did not perform any checks of its own, but Maven Central should actually check that before the staging repository is closed and all the artifacts are published, so it's not needed locally

Can you have a link to doc describing it? Does it lead to failing closing?
For example, Sonatype OSSRH gives only a recommendation. [here](https://central.sonatype.org/publish/requirements/#sufficient-metadata) about the correct dependencies.

> It's in the way now because it fails with a false positive: analysis modules shadow all dependencies, 
Can we filter out such dependencies, e.g. that have `ShadowExtension`?

IgnatBeresnev:
> Do you have a link to doc describing it? Does it lead to failing closing?

I asked Vsevolod about it during our last sync, so I mostly trust his answer that it will be checked and it will fail the publication.

I did find these docs though that seem to confirm it: [Enforcing Standards for Deployment and Promotion with Rulesets](https://help.sonatype.com/repomanager2/staging-releases/enforcing-standards-for-deployment-and-promotion-with-rulesets)

More specifically, I think it's the rule called "POM must not contain system scoped dependencies".

> Can we filter out such dependencies, e.g. that have ShadowExtension?

We can, but I honestly don't see the point in adding such filters **if** it is checked by the sonatype staging repository. 

If it was indeed a very noticeable problem, I'm sure someone would've come up with a better solution than what we currently have, and it would be used by our other libraries, but it doesn't seem to be the case.

vmishenev:
> I did find these docs though that seem to confirm it: [Enforcing Standards for Deployment and Promotion with Rulesets](https://help.sonatype.com/repomanager2/staging-releases/enforcing-standards-for-deployment-and-promotion-with-rulesets)

It's a feature of Manager **Pro**. As far as I understand, it can be disabled by default. But if Vsevolod says...

> We can, but I honestly don't see the point in adding such filters if it is checked by the sonatype staging repository.

My point is to check it before releasing, during integration tests. It would be independent of the custom sonatype rules and useful for other repositories. 

>  If it was indeed a very noticeable problem, I'm sure someone would've come up with a better solution than what we currently have,

We have not researched it. It could be a reason to file an issue. 

Anyway, it is not a blocker to remove it.

PR #3081, opened 2023-07-20
potential fix for compatibility with KGP 1.9

### LINKED ISSUE #3038: Unable to run dokka with single target multiplatform project and kotlin 1.9.0
Dokka tasks for a multiplatform project with a single target fail with

```
* Exception is:
org.gradle.api.internal.tasks.TaskDependencyResolveException: Could not determine the dependencies of task ':single-target-mpp-module:dokkaHtml'
Caused by: java.lang.UnsupportedOperationException: Empty collection can't be reduced.
	at org.jetbrains.dokka.gradle.kotlin.KotlinClasspathUtilsKt.classpathOf(kotlinClasspathUtils.kt:83)
	at org.jetbrains.dokka.gradle.kotlin.KotlinSourceSetGistKt$gistOf$3.call(KotlinSourceSetGist.kt:22)
	at org.jetbrains.dokka.gradle.kotlin.KotlinSourceSetGistKt$gistOf$3.call(KotlinSourceSetGist.kt)
```

**Installation**
- Dokka version: 1.8.20
- Kotlin Gradle plugin version: 1.9.0-Beta

### LINKED ISSUE #3068: Failed to resolve inter-module dependencies: Cannot change dependencies of dependency configuration
**Describe the bug**
Dokka fails to correctly resolve task dependencies in a multi-module multiplatform project with inter-module dependencies.
This results in a build failure. Note that this is a true multiplatform project with JVM and iOS targets.

The full error message is:
> Cannot change dependencies of dependency configuration ':vclib-openid:iosMainImplementationDependenciesMetadata' after task dependencies have been resolved

However, this also happens when manually invoking `dokkaHtml` for individual modules.

**Expected behaviour**
Dokka correctly resolves task dependencies of all modules and executes them as required to produce documentation.

**To Reproduce**
```shell
git clone --recursive -b tmp/dokkaFailKGP1.9 git@github.com:a-sit-plus/kmm-vc-library.git
./gradlew dokkaHtmlMultiModule
```

**Dokka configuration**
Note: we introduced the the inter-module Dokka-related task dependencies when upgrading to Gradle 8.1, because these dependencies were reported to be implicit and needed to be made explicit for the Dokka tasks to successfully run. The only custom configuration we really depend on is manually setting the output directory of Dokka  to reference it for publishing HTML documentation on https://a-sit-plus.github.io/kmm-vc-library/. <br> (This is also the reason why we manually declared the publishing tasks' dependencies on the signing tasks.)

Note: Removing these manual declarations of Dokka-related task dependencies does not impact the behaviour.

I'm also satisfied with a workaround as a stop-gap solution so we can upgrade to Kotlin 1.9. Everything worked fine with KGP 1.8.20 (Except Gradle 8.1 promoting some warnings to errors compared to Gradle 7.6, requiring the previously mentioned manual specification of task dependencies.)

whyoleg:
attempt to fix #3068 which is blocking building documentation for https://github.com/whyoleg/cryptography-kotlin/tree/dev (`dev` branch, KGP 1.9.0)
for recent changes consult: https://github.com/Kotlin/dokka/pull/3081#issuecomment-1666818036

### DIFF
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
@@ -57,8 +57,14 @@ private fun KotlinCompilation.getKotlinCompileTask(kgpVersion: KotlinGradlePlugi
 }
 
 private fun KotlinCompilation.platformDependencyFiles(): FileCollection {
+    val excludePlatformDependencyFiles = project.classpathProperty("excludePlatformDependencyFiles", default = false)
+
+    if (excludePlatformDependencyFiles) return project.files()
     return (this as? AbstractKotlinNativeCompilation)
         ?.target?.project?.configurations
         ?.findByName(@Suppress("DEPRECATION") this.defaultSourceSet.implementationMetadataConfigurationName) // KT-58640
         ?: project.files()
 }
+
+private fun Project.classpathProperty(name: String, default: Boolean): Boolean =
+    (findProperty("org.jetbrains.dokka.classpath.$name") as? String)?.toBoolean() ?: default
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
@@ -40,7 +40,7 @@ private fun KotlinCompilation.compileClasspathOf(project: Project): FileCollecti
 
 private fun KotlinCompilation.newCompileClasspathOf(project: Project): FileCollection {
     val compilationClasspath = (compileTaskProvider.get() as? KotlinCompileTool)?.libraries ?: project.files()
-    return compilationClasspath + platformDependencyFiles()
+    return compilationClasspath + platformDependencyFiles(project)
 }
 
 private fun KotlinCompilation.oldCompileClasspathOf(project: Project): FileCollection {
@@ -48,7 +48,7 @@ private fun KotlinCompilation.oldCompileClasspathOf(project: Project): FileColle
         return this.classpathOf(project)
     }
 
-    return this.compileDependencyFiles + platformDependencyFiles() + this.classpathOf(project)
+    return this.compileDependencyFiles + platformDependencyFiles(project) + this.classpathOf(project)
 }
 
 private fun KotlinCompilation.classpathOf(project: Project): FileCollection {
@@ -76,7 +76,7 @@ private fun KotlinCompilation.getKotlinCompileTask(kgpVersion: KotlinGradlePlugi
     }
 }
 
-private fun KotlinCompilation.platformDependencyFiles(): FileCollection {
+private fun KotlinCompilation.platformDependencyFiles(project: Project): FileCollection {
     val excludePlatformDependencyFiles = project.classpathProperty("excludePlatformDependencyFiles", default = false)
 
     if (excludePlatformDependencyFiles) return project.files()
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
@@ -28,12 +28,7 @@ private fun KotlinCompilation.compileClasspathOf(project: Project): FileCollecti
         return this.classpathOf(project)
     }
 
-    val platformDependencyFiles: FileCollection = (this as? AbstractKotlinNativeCompilation)
-        ?.target?.project?.configurations
-        ?.findByName(@Suppress("DEPRECATION") this.defaultSourceSet.implementationMetadataConfigurationName) // KT-58640
-        ?: project.files()
-
-    return this.compileDependencyFiles + platformDependencyFiles + this.classpathOf(project)
+    return this.compileDependencyFiles + platformDependencyFiles() + this.classpathOf(project)
 }
 
 private fun KotlinCompilation.classpathOf(project: Project): FileCollection {
@@ -43,7 +38,7 @@ private fun KotlinCompilation.classpathOf(project: Project): FileCollection {
     val shouldKeepBackwardsCompatibility = (kgpVersion != null && kgpVersion < KotlinGradlePluginVersion(1, 7, 0))
     return if (shouldKeepBackwardsCompatibility) {
         // removed since 1.9.0, left for compatibility with < Kotlin 1.7
-        val classpathGetter= kotlinCompile::class.members
+        val classpathGetter = kotlinCompile::class.members
             .first { it.name == "getClasspath" }
         classpathGetter.call(kotlinCompile) as FileCollection
     } else {
@@ -59,4 +54,11 @@ private fun KotlinCompilation.getKotlinCompileTask(kgpVersion: KotlinGradlePlugi
     } else {
         this.compileTaskProvider.get() as? KotlinCompile // introduced in 1.8.0
     }
+}
+
+private fun KotlinCompilation.platformDependencyFiles(): FileCollection {
+    return (this as? AbstractKotlinNativeCompilation)
+        ?.target?.project?.configurations
+        ?.findByName(@Suppress("DEPRECATION") this.defaultSourceSet.implementationMetadataConfigurationName) // KT-58640
+        ?: project.files()
 }
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
@@ -6,6 +6,7 @@ import org.jetbrains.dokka.gradle.isAndroidTarget
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.gradle.plugin.mpp.AbstractKotlinNativeCompilation
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompileTool
 
 internal fun Project.classpathOf(sourceSet: KotlinSourceSet): FileCollection {
     val compilations = compilationsOf(sourceSet)
@@ -23,6 +24,26 @@ internal fun Project.classpathOf(sourceSet: KotlinSourceSet): FileCollection {
 }
 
 private fun KotlinCompilation.compileClasspathOf(project: Project): FileCollection {
+    val kgpVersion = project.getKgpVersion()
+
+    // if KGP version < 1.9 or org.jetbrains.dokka.classpath.useOldResolution=true
+    // we will use old (pre 1.9) resolution of classpath
+    if (kgpVersion == null ||
+        kgpVersion < KotlinGradlePluginVersion(1, 9, 0) ||
+        project.classpathProperty("useOldResolution", default = false)
+    ) {
+        return oldCompileClasspathOf(project)
+    }
+
+    return newCompileClasspathOf(project)
+}
+
+private fun KotlinCompilation.newCompileClasspathOf(project: Project): FileCollection {
+    val compilationClasspath = (compileTaskProvider.get() as? KotlinCompileTool)?.libraries ?: project.files()
+    return compilationClasspath + platformDependencyFiles()
+}
+
+private fun KotlinCompilation.oldCompileClasspathOf(project: Project): FileCollection {
     if (this.target.isAndroidTarget()) { // Workaround for https://youtrack.jetbrains.com/issue/KT-33893
         return this.classpathOf(project)
     }
--- a/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
+++ b/runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
@@ -12,14 +12,13 @@ internal fun Project.classpathOf(sourceSet: KotlinSourceSet): FileCollection {
     return if (compilations.isNotEmpty()) {
         compilations
             .map { compilation -> compilation.compileClasspathOf(project = this) }
-            .reduce { acc, fileCollection -> acc + fileCollection }
+            .reduce(FileCollection::plus)
     } else {
         // Dokka suppresses source sets that do no have compilations
         // since such configuration is invalid, it reports a warning or an error
         sourceSet.withAllDependentSourceSets()
-            .toList()
             .map { it.kotlin.sourceDirectories }
-            .reduce { acc, fileCollection -> acc + fileCollection }
+            .reduce(FileCollection::plus)
     }
 }
 


runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
```diff
@@ -41,12 +48,20 @@ private fun KotlinCompilation.compileClasspathOf(project: Project): FileCollecti
...
+        val platformDependencyFiles: FileCollection = (this as? AbstractKotlinNativeCompilation)
+            ?.target?.project?.configurations
+            ?.findByName(@Suppress("DEPRECATION") this.defaultSourceSet.implementationMetadataConfigurationName) // KT-58640
+            ?: project.files()
 
-    return this.compileDependencyFiles + platformDependencyFiles + this.classpathOf(project)
+        this.compileDependencyFiles + platformDependencyFiles + this.classpathOf(project)
+    } else {
+        // compileTaskProvider is available starting from kotlin 1.8.0
+        (compileTaskProvider.get() as? KotlinCompileTool)?.run { sources + libraries } ?: project.files()
```
whyoleg, line 63: this is fix for https://github.com/Kotlin/dokka/issues/3068
previous solution uses `compileDependencyFiles` which in itself has dependency on https://github.com/JetBrains/kotlin/blob/b4335c86c8378a43a63052ca00c2472f355d375c/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/sources/KotlinSourceSetMetadataTransformation.kt#L23 (deprecated for removal)

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
```diff
@@ -23,7 +28,9 @@ internal fun Project.classpathOf(sourceSet: KotlinSourceSet): FileCollection {
              * Dokka needs only HMPP variant
              * Ignore [org.jetbrains.kotlin.gradle.plugin.mpp.KotlinCommonCompilation]  for `commonMain`  sourceSet with name `main`
              */
-            .filterNot { compilation -> isHMPPEnabled && compilation is KotlinCommonCompilation && compilation.name == "main" }
+            .filterNot { compilation ->
+                shouldKeepBackwardsCompatibility && isHMPPEnabled && compilation is KotlinCommonCompilation && compilation.name == "main"
```
whyoleg, line 32: this is fix for #3038 (single target MPP projects)
Not sure about the real reason, but while debugging in my own project, I see that `isHMPPEnabled=true`, even while I have no `kotlin.mpp.enableGranularSourceSetsMetadata` specified anywhere in project.
BTW, comment is a little misleading also, as it mentions `Compatibility Metadata Variant` which is enabled by other flag `kotlin.mpp.enableCompatibilityMetadataVariant=true` - may be checking for correct flag will be enough to keep compatibility, but haven't tested it. Also, PR introduced this changes also don't have description (https://github.com/Kotlin/dokka/pull/2634) so I'm a little afraid on how those changes will affect building of other projects

IgnatBeresnev: (in reply to the [comment](https://github.com/Kotlin/dokka/pull/3081/files#r1269975175))

Yeah, this fix is quite janky - we badly needed a solution that worked (as it also blocked some projects), and this is what we ended up with after discussing it with the Kotlin Multiplatform team. I don't fully understand it though, and I certainly didn't expect it to backfire...

> I'm a little afraid on how those changes will affect building of other projects

This is where it becomes interesting :)) The reason we put it there to begin with is because we couldn't build ktor without it. I've ran your branch on latest ktor, and indeed it doesn't work :( [stacktrace here](https://gist.github.com/IgnatBeresnev/27c623e531ebc822b033094020900ca1)

However, this particular filter is the reason for why single-target multiplatform projects don't work: KGP 1.9.0 removed a redundant commonMain compilation, and we filter out the only thing that's left, so compilations end up being empty -> reduce on an empty collection.

I've already found a fix for this particular problem, and it seems to work in both cases, so I'll create a separate PR, and can tag you in it if you're curious :)

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
IgnatBeresnev: I played around with your changes to try and understand why it fixes the problem, as this is also legacy that we carry without fully understanding what it's doing and why.

It looks like the problem is not with `compileDependencyFiles`, but with `platformDependencyFiles`. I see that removing the whole `platformDependencyFiles` block fixes the problem both in your project and in the one from #3068, but I don't quite understand why :smiling_face_with_tear: Maybe you had a chance to research it and have an idea?

I've also noticed that both reproducer projects have quite unusual and non-standard Gradle configuration subprojects, so I wouldn't be surprised if this bug needs specific conditions to be reproduced. I tried to reproduce it on ktor, but they haven't migrated to Kotlin 1.9.0 yet due to some problems. Can't think of any other average multiplatform multi-module project to test this on...

I'll consult with the multiplatform team in the upcoming days, maybe they have an idea about what's going on..

IgnatBeresnev, line 63: I ended up with the following  changes that work:

```suggestion
        this.compileDependencyFiles + platformDependencyFiles + this.classpathOf(project)
    } else {
        this.compileDependencyFiles + this.classpathOf(project)
```

`(compileTaskProvider.get() as? KotlinCompileTool)` is already returned by `getKotlinCompileTask` (called transitively through `classpathOf`), and the `libraries` files are also returned by the `classpathOf` overload below, so this bit can be simplified.

Removing `this.compileDependencyFiles` feels scary as I don't know what it was introduced for, and I'd personally leave it if it doesn't get in the way.

Calling `sources` doesn't seem to make any difference, but I haven't tested it extensively - was there a particular reason you used it? I mean the `run { sources + libraries }` bit

IgnatBeresnev (commented):
I've finally had time to have a closer look at this PR and the issues with Kotlin 1.9.0 in general.

It seems like these are two separate problems:

* support for single-target multiplatform projects
* inter-module native dependency resolution thing

The first problem about single-target KMP projects is not completely fixed in this PR - I'll provide more details in one of the comments. We also decided that we need to include a fix for it to Dokka 1.9.0, which is already overdue, so we'd like to take this particular issue off your hands :)

The second issue, however, is interesting - you've potentially found the problem and a fix for it, so I'd like to discuss it further and merge it if the fix is OK. I'll also leave a more detailed comment below

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
whyoleg: > (compileTaskProvider.get() as? KotlinCompileTool) is already returned by getKotlinCompileTask

I think, it's not, because in `getKotlinCompileTask` task is casted to `KotlinCompile`, which AFAIK is JVM only, but in PR `KotlinCompileTool` is used - which is implemented by all platform compilation tasks (JVM/Native/JS). Still, may be it's enough to cast to `KotlinCompileTool` in `getKotlinCompileTask`, but Im not sure about the specific version when this interface was introducec

> Removing this.compileDependencyFiles feels scary

While Im not 100% sure, but recently introduced `libraries` should contain everything what should exists in `compileDependencyFiles` (as from what I see while diving in KGP source code)
The main reason for not to use it, as it's mutable variable, which could be not fully configured (for some unknown reason) when fetching it - I don't know if it's true or not. I would go with using just `libraries` (at least starting from 1.9.0), as it's looks like real public API, comparing to `compileDependencyFiles` + introducing fallback property (on task, or just some gradle property) in Dokka plugin to also add `compileDependencyFiles` - this is just because Im afraid that if using it, may be in next KGP version something will change again

> Calling sources doesn't seem to make any difference

Yeah, I think it could be removed, as Dokka process sources separately in `KotlinSourceSetGist`

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
whyoleg: Yeah, sound good, Thanks!

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
whyoleg: > but I don't quite understand why

me neither :)

> I've also noticed that both reproducer projects have quite unusual and non-standard Gradle configuration subprojects

Overall, at least with my project, while it could be unusual, it's mostly recommended by Gradle practice to use small separate convention plugins and apply them to subproject instead of configuration via `allprojects/subprojects` in root repository - this will be the only way to do it if/when project isolation support in Gradle will be implemented fully. Or are you talking about something else unusual?
Just in case, almost full configuration of Dokka for subproject is in one file - https://github.com/whyoleg/cryptography-kotlin/blob/dev/gradle/plugins/build-logic/src/main/kotlin/buildx-documentation.gradle.kts
Which is applied in scope of other convention plugin

IgnatBeresnev: To expand on the unusual project configuration 

I usually see a flat and simple structure, where you have `buildSrc` / `build-logic` in the root, and then convention plugins inside it. The version catalog is usually a toml file inside `gradle/libs.toml` that gets imported by the simple `buildSrc` / `build-logic`.

cryptography-kotlin and [kmm-vc-library](https://github.com/a-sit-plus/kmm-vc-library/tree/main/conventions-vclib) on the surface level have the same approach, but both seem to have some additional configurations/projects in places you wouldn't expect to find them (like separate projects for the version catalog and the kotlin version)

That's not to say that Dokka is doing everything correctly and there is no bug (there likely is), but knowing how fragile Gradle is and looking at the error message `Cannot change dependencies of dependency configuration .. after task dependencies have been resolved` - I wouldn't be surprised if these specific project configurations are playing a role here

I also don't think there were any other reports about this particular problem, but there for sure must be many multiplatform multi-module projects that use Dokka. So there must be something common between cryptography and kmm-vc-library. Or maybe people haven't updated to 1.9.0 yet and the storm is coming our way (god, I hope not) 😅

IgnatBeresnev: > I think, it's not, because in getKotlinCompileTask task is casted to KotlinCompile, which AFAIK is JVM only, but in PR KotlinCompileTool is used - which is implemented by all platform compilation tasks (JVM/Native/JS)

Oh, I didn't even think in that direction! I checked that `KotlinCompile` implements `KotlinCompileTool`, but judging by the name alone I didn't expect it to be JVM-only... 

Thank you for explaining your reasoning, it makes total sense, and it looks like it should be safe to proceed like you propose. Especially if we leave a gradle property to revert to the old behaviour with `platformDependencyFiles` and `compileDependencyFiles ` 👍 

Just in case, I'll discuss this and other matters with a Multiplatform/KGP expert tomorrow morning, in case there's something we're missing, and get back to you

whyoleg: Yeah, I understand that it's not the most straightforward approach, but still it's rather good supported from gradle side.
Minor off-top why it's done like this:
1. `build-parameters` generates type-safe accessors for custom gradle project properties. Overall it's alternative to similar concept in KGP: [PropertiesProvider](https://github.com/JetBrains/kotlin/blob/6943d03883eed6013d4f8f1557be28a6b324d8dc/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/PropertiesProvider.kt#L18). It's possible to write all those property accessors just with plain string, but it's not a good idea IMO. And it's in separate included build, because it should be used in other included builds :)
2. `kotlin-version-catalog` this is rather specific thing in my projects to simplify testing of new Kotlin releases (like 1.9.0, 1.9.20-dev). It allows to override kotlin version for all places in a project via a single command line argument (`-PuseKotlin=1.9.20`). The idea is similar to what could be archived by using recently created [community-project-gradle-plugin](https://github.com/Kotlin/community-project-gradle-plugin) but using different strait forward approach and integrated inside build, so Im able to easy test new kotlin versions (even dev versions) both locally, in IDEA and on CI. And because it overrides kotlin versions in a lot of places (both runtime dependencies and gradle plugin dependencies), the only way to achieve it is to use included build.

While all of it could be achieved via simpler ways it will increase complexity and possibility to misuse something

P.S. end of off-top :)

Returning to cause of an issue - it's really hard to understand for me what is the real cause on why it worked before and doesn't work now. May be the main difference is that Dokka in my project applied in specific subprojects and not how stated in readme (in root via `subprojects {}`).
Also I see, that some other projects start to adopt https://github.com/adamko-dev/dokkatoo because of better compatibility with new and shiny gradle features :)
just FYI, even `gradle` uses it instead of official Dokka gradle plugin: https://github.com/search?q=repo%3Agradle%2Fgradle%20dokkatoo&type=code

IgnatBeresnev:
Single-target projects were somewhat "fixed" in #3103 (sorry that you need to resolve merge conflicts now), so we can focus on addressing #3068 in this PR

If you don't mind, I'll resolve the long conversations above as there were mostly clarifications, and I'll sum them up here instead. I've also discussed the proposed solutions with Sebastian from the Kotlin Multiplatform team, and will include his thoughts below.

* `KotlinCompile` is indeed JVM only, so `KotlinCompileTool` could/should be used instead. Not sure how and why it worked before :sweat_smile: 
* It should be safe to remove `this.compileDependencyFiles` - Sebastian agreed that `kotlinCompileTool.libraries` should provide compile dependency files already, and in a better/safer way
* Removing `platformDependencyFiles` fixes the issue from #3068, but Sebastian mentioned that this configuration might contain some native-specific dependencies that are not present in `kotlinCompileTool.libraries`. Unfortunately, I don't have enough context to understand the details, but as far as I remember it's related to some dependencies that are either system-provided or just provided by default, so they're not added to `libraries`, or something along those  lines. He mentioned iOS x64 as something that might uncover the bugs with this.

___

Given the tight schedule for 1.9.0 (it's pretty much ready for final testing), and that this place in general is quite scary to change and has led to serious problems before, I would propose the following:

For < 1.9.0, use the current behaviour.

For >= 1.9.0:
* replace `KotlinCompile` with `KotlinCompileTool`
* remove `compileDependencyFiles`
* add a fallback flag (`false` by default), which, if enabled, will use pre-1.9.0 behaviour.

Additionally:
* use `platformDependencyFiles` by default the same way as now, but
* introduce a second flag (`false` by default) that removes `platformDependencyFiles` from classpath calculation, so that projects affected by #3068 have a workaround

The reason for not enabling the `platformDependencyFiles` flag by default: there are not that many projects that reported this problem, so we don't know what the exact reason is, or what they have in common. And there's a considerable chance of something breaking, judging by what Sebastian mentioned.

Right after merging the PR and cherry-picking the commit into the release branch:
* change the flag for `platformDependencyFiles` to `true` in Dokka's master branch

There should be about a month between 1.9.0 and 1.9.10, which gives us time to dogfood and find problems. If we find bugs - we'll just revert the platform flag to `false` by default.

Code-wise, to avoid a bunch of if-else branches, maybe it'll be easier to refactor the file a little, and extract pre-1.9.0 and >=1.9.0 solutions into separate classes / functions, it's up to you

What do you think? :)

whyoleg:
@IgnatBeresnev Nice plan! I like it. Also big thanks for summarising everything, good job!
Don't worry about merge conflicts, I expected them :)
I also really like the idea of splitting behaviour of pre and post 1.9 - while it could cause a little of duplications, we could try to stick for almost public API in post 1.9, this is good especially in scope of stabilisation of multi-platform support in KGP in 1.9.
I will actualise branch with what proposed, do checks on my project and promote PR from draft in coming days (likely by Monday)

IgnatBeresnev:
@whyoleg awesome, thank you!

whyoleg:
So, I've pushed changes. 2 flags introduced:
* org.jetbrains.dokka.classpath.excludePlatformDependencyFiles - if true, will exclude platformDependencyFiles
* org.jetbrains.dokka.classpath.useOldResolution - when KGP >= 1.9, if true will use old resolution with compileDependencyFiles and all other stuff, if false will only use `KotlinCompileTool.libraries` (+ platformDependencyFiles in both cases)

I've tested it on my project - everything works fine here with flag `org.jetbrains.dokka.classpath.excludePlatformDependencyFiles=true`.

After that, I've decided to run some integration tests with this flag set to true and `it-multiplatform-0` failed. It failed, because now it can not resolve `CPointer` and `CPointed` classes - this means, that they are coming from deprecated `implementationMetadataConfigurationName`.
I checked is there a way to provide those dependencies in other way and found that now they are added in [KotlinNativePlatformDependencies.kt](https://github.com/JetBrains/kotlin/blob/887dc18719a167da1dbbbb5cd475314a1a562065/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/targets/native/internal/KotlinNativePlatformDependencies.kt#L37) and I didn't found an easy way, on how we could access them in Dokka, as everything is super internal.

Then I've decided to check on replacement of deprecation of `implementationMetadataConfigurationName` and found [resolvableMetadataConfiguration](https://github.com/JetBrains/kotlin/blob/887dc18719a167da1dbbbb5cd475314a1a562065/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/resolvableMetadataConfiguration.kt#L39C38-L39C69) which is also used in IDE import in KGP, but it's also highly internal and not exposed anywhere else.

So, to summarise, excluding platformDependencyFiles is not an option for projects, which expose `kotlinx.cinterop` and other platform libraries in documentation :(

It's up to you to decide, should we leave this flag, for excluding platformDependencyFiles, or not, while for my project it's a solution (as there is no such declarations exposed), for other it will be not acceptable.

whyoleg:
Also, is there a good place to document those flags?

IgnatBeresnev:
Well done! I'll have a closer look towards the evening

> So, to summarise, excluding platformDependencyFiles is not an option for projects, which expose kotlinx.cinterop and other platform libraries in documentation :(

Oh, thank you for researching and confirming what Sebastian mentioned! I believe in these situations `kotlinx.cinterop` types will be rendered as `Error class`, while everything else should rendered as expected, so I think it's still a valid workaround for the time being - at least then affected users will be able to build documentation, even if some types are rendered as `Error class`

> Also, is there a good place to document those flags?

As there are no guarantees and we don't plan to support these flags for long, I think it's better to only post them in the comments of issues (which should be googlable by the error message), as opposed to some official documentation. I need to close the duplicates though, so that all of the related bugs lead to a single issue in which these flags will be suggested as a workaround - I'll do it this week. 

I guess I should also mention it in the release notes, that some projects with `kotlinx.cinterop` might experience issues, and link to the workaround comment.

But maybe you have some other suggestions?

whyoleg:
>  I believe in these situations kotlinx.cinterop types will be rendered as Error class, while everything else should rendered as expected

Yeah, that's what I've seen when checking produced by test results.

> I think it's better to only post them in the comments of issues

Yes, I agree, that will be enough, as those flags are mostly for workarounds

> and it seems like there are some compatibility problems:

Oops, haven't run all integration tests (too much resources needed), will recheck locally later today, thx!

whyoleg:
I've pushed fix for compatibility with older kotlin versions. Haven't spotted, that previously `project` was resolved from argument, but after extraction from `KotlinCompilation.project` which was introduced in recent KGP versions...

Also, run `SerializationGradleIntegrationTest`, `CoroutinesGradleIntegrationTest` and `Multiplatform0GradleIntegrationTest` locally - everything works (and now I have 11 K/N compilers installed locally :) )

IgnatBeresnev (approved):
Looking good, thank you!

Just in case, I'll trigger longer integration tests on TeamCity now, as well as thorough tests with more platforms, and will merge the PR tomorrow if there are no problems

(upd to not lose the link: [thorough tests run](https://github.com/Kotlin/dokka/actions/runs/5788937734))

runners/gradle-plugin/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinClasspathUtils.kt
```diff
@@ -6,34 +6,49 @@ import org.jetbrains.dokka.gradle.isAndroidTarget
...
+    // if KGP version < 1.9 or org.jetbrains.dokka.classpath.useOldResolution=true
+    // we will use old (pre 1.9) resolution of classpath
+    if (kgpVersion == null ||
+        kgpVersion < KotlinGradlePluginVersion(1, 9, 0) ||
+        project.classpathProperty("useOldResolution", default = false)
+    ) {
+        return oldCompileClasspathOf(project)
+    }
+
+    return newCompileClasspathOf(project)
```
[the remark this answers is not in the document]
whyoleg: yeah, I agree, first idea at my side for naming was like `pre19*` and `post19*` or just use `pre19` for old function and no prefix for new, but it also started to look a bit odd for me

PR #3092, opened 2023-07-25
Fix handling of Java contravariance

### LINKED ISSUE #3091: Java contravariant type is represented as Covariant
**Describe the bug**
The following are Java and Kotlin definitions of a function with a parameter with a contravariant type:
```
public void bar(java.util.List<? super String> param) {}
fun bar(param: List<in String>) {}
```

The Kotlin definition is correctly parsed as `Contravariant` with a bound of `String`, but the Java definition becomes `Covariant` with a bound of `Object`.

It seems like in the PSI model, a `PsiWildcardType` will always have an `extendsBound` of `java.lang.Object`, even if that is not explicitly written in the source. In [`DokkaPsiParser.getVariance`](https://github.com/Kotlin/dokka/blob/900fbccf3da483946227d99ce52b71afe0e62007/subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt#L569-L573), because the `extendsBound` is checked first, wildcard types will always turn into `Covariant`s, even when they have a contravariant `super` bound.

I think #3092 fixes the issue

juliamcclellan:
Previously, because `PsiWildcardType`s have an implicit bound of `extends Object`, a wildcard type with a `super` bound was being turned into a `Covariance` instead of a `Contravariance`.

Fixes #3091

### DIFF
--- a/plugins/base/src/test/kotlin/model/JavaTest.kt
+++ b/plugins/base/src/test/kotlin/model/JavaTest.kt
@@ -10,6 +10,7 @@ import org.jetbrains.dokka.model.doc.Text
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Test
 import utils.AbstractModelTest
+import utils.assertContains
 import utils.assertNotNull
 import utils.name
 import kotlin.test.assertEquals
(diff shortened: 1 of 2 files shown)


vmishenev (approved):

juliamcclellan:
Thanks for the review. Good point, I updated the test to include assertions that the three functions are present.

IgnatBeresnev (approved):
Thank you!

PR #3102, opened 2023-08-02
Fix and refactor Sample Transformer

vmishenev:
Also, the PR has some refactoring. I am not sure that the page transformer (`SampleTransformer`) should be in the analysis plugin.

### DIFF
--- a/plugins/base/src/main/kotlin/DokkaBase.kt
+++ b/plugins/base/src/main/kotlin/DokkaBase.kt
@@ -149,7 +149,6 @@ class DokkaBase : DokkaPlugin() {
 
     val pageMerger by extending {
         CoreExtensions.pageTransformer providing ::PageMerger order {
-            // TODO [beresnev] make last() or at least after samples transformer
         }
     }
 
--- a/plugins/base/src/main/kotlin/DokkaBase.kt
+++ b/plugins/base/src/main/kotlin/DokkaBase.kt
@@ -275,11 +275,6 @@ class DokkaBase : DokkaPlugin() {
     val defaultKotlinAnalysis: org.jetbrains.dokka.plugability.Extension<org.jetbrains.dokka.analysis.KotlinAnalysis, *, *>
         get() = throw org.jetbrains.dokka.base.deprecated.AnalysisApiDeprecatedError()
 
-    @Suppress("DeprecatedCallableAddReplaceWith")
-    @Deprecated(message = org.jetbrains.dokka.base.deprecated.ANALYSIS_API_DEPRECATION_MESSAGE, level = DeprecationLevel.ERROR)
-    val defaultSamplesTransformer: org.jetbrains.dokka.plugability.Extension<PageTransformer, *, *>
-        get() = throw org.jetbrains.dokka.base.deprecated.AnalysisApiDeprecatedError()
-
     @Suppress("DEPRECATION_ERROR", "DeprecatedCallableAddReplaceWith")
     @Deprecated(message = org.jetbrains.dokka.base.deprecated.ANALYSIS_API_DEPRECATION_MESSAGE, level = DeprecationLevel.ERROR)
     val defaultExternalDocumentablesProvider: org.jetbrains.dokka.plugability.Extension<org.jetbrains.dokka.base.translators.descriptors.ExternalDocumentablesProvider, *, *>
--- a/plugins/base/src/main/kotlin/DokkaBase.kt
+++ b/plugins/base/src/main/kotlin/DokkaBase.kt
@@ -1,5 +1,3 @@
-@file:Suppress("unused")
-
 package org.jetbrains.dokka.base
 
 import org.jetbrains.dokka.CoreExtensions
@@ -19,6 +17,7 @@ import org.jetbrains.dokka.base.signatures.KotlinSignatureProvider
 import org.jetbrains.dokka.base.signatures.SignatureProvider
 import org.jetbrains.dokka.base.templating.ImmediateHtmlCommandConsumer
 import org.jetbrains.dokka.base.transformers.documentables.*
+import org.jetbrains.dokka.base.transformers.pages.DefaultSamplesTransformer
 import org.jetbrains.dokka.base.transformers.pages.annotations.SinceKotlinTransformer
 import org.jetbrains.dokka.base.transformers.pages.comments.CommentsToContentConverter
 import org.jetbrains.dokka.base.transformers.pages.comments.DocTagToContentConverter
@@ -33,6 +32,7 @@ import org.jetbrains.dokka.plugability.PluginApiPreviewAcknowledgement
 import org.jetbrains.dokka.transformers.documentation.PreMergeDocumentableTransformer
 import org.jetbrains.dokka.transformers.pages.PageTransformer
 
+@Suppress("unused")
 class DokkaBase : DokkaPlugin() {
 
     val preMergeDocumentableTransformer by extensionPoint<PreMergeDocumentableTransformer>()
@@ -189,6 +189,12 @@ class DokkaBase : DokkaPlugin() {
 
     val rootCreator by extending {
         htmlPreprocessors with RootCreator applyIf { !delayTemplateSubstitution }
+    }
+
+    val defaultSamplesTransformer by extending {
+        CoreExtensions.pageTransformer providing ::DefaultSamplesTransformer order {
+            before(pageMerger)
+        }
     }
 
     val sourceLinksTransformer by extending {
--- a/plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
@@ -24,7 +24,6 @@ import org.jetbrains.dokka.pages.ContentPage
 import org.jetbrains.dokka.pages.PageNode
 import org.jetbrains.dokka.plugability.DokkaContext
 import org.jetbrains.dokka.plugability.configuration
-import org.jsoup.Jsoup
 import java.net.URI
 
 class DefaultTemplateModelFactory(val context: DokkaContext) : TemplateModelFactory {
@@ -93,17 +92,6 @@ class DefaultTemplateModelFactory(val context: DokkaContext) : TemplateModelFact
 
     private fun Appendable.resourcesForPage(pathToRoot: String, resources: List<String>): Unit =
         resources.forEach { resource ->
-            /**
-             * Workaround for resources that have no extension
-             * e.g. `<script src="`https://unpkg.com/kotlin-playground@1`"></script>`
-             */
-            if (resource.startsWith("<")) {
-                val tag = Jsoup.parse(resource).head().child(0)
-                if (tag.tagName() == "script" || tag.tagName() == "style") {
-                    append(resource)
-                }
-                return@forEach
-            }
 
             val resourceHtml = with(createHTML()) {
                 when {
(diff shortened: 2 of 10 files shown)


plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
```diff
@@ -91,29 +92,41 @@ class DefaultTemplateModelFactory(val context: DokkaContext) : TemplateModelFact
         get() = URI(this).isAbsolute
 
     private fun Appendable.resourcesForPage(pathToRoot: String, resources: List<String>): Unit =
-        resources.forEach {
+        resources.forEach { resource ->
+            /**
```
vmishenev, line 96: The root of the problem was here

subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/kotlin/analysis/kotlin/internal/SampleProvider.kt
```diff
@@ -0,0 +1,23 @@
+package org.jetbrains.kotlin.analysis.kotlin.internal
+
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.InternalDokkaApi
+
+@InternalDokkaApi
+interface SampleProviderFactory {
+    fun build(): SampleProvider
+}
```
IgnatBeresnev, line 9: It's a bit unclear why `SampleProviderFactory` is needed at this moment - will there be scenarios in the near future where it's possible to have multiple providers, or what was the idea behind it?

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/KotlinSampleProvider.kt
```diff
@@ -0,0 +1,154 @@
...
+class KotlinSampleProviderFactory(val context: DokkaContext): SampleProviderFactory {
+    override fun build(): SampleProvider {
+        return KotlinSampleProvider(context)
+    }
+
+}
+/**
+ * It's declared as open since StdLib has its own sample transformer
+ * with [processBody] and [processImports]
+ */
```
IgnatBeresnev, line 31: Oh, I didn't know stdlib actually used something like this... 

It can be used as a very short-term solution that unblocks stdlib, but it's definitely not OK for stdlib to depend on internal API, let alone on one of the analysis modules.

Would you be able to research what exactly the use case in stdlib is, and to think of a stable alternative to go into `kotlin-analysis-api`? As a separate issue / PR, of course

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/KotlinSampleProvider.kt
```diff
@@ -0,0 +1,154 @@
...
+     * calling this `close` method more than once may have some
+     * visible side effect, unlike `Closeable.close` which is
+     * required to have no effect if called more than once.
+     *
+     * However, implementers of this interface are strongly encouraged
+     * to make their `close` methods idempotent.
+     *
+     * @throws Exception if this resource cannot be closed
+     */
+    override fun close() {
```
IgnatBeresnev, line 134: Was there a reason to copy-paste this gigantic javadoc? :) It threw me off a little - I thought maybe you added some documentation to it that I missed.

plugins/base/src/main/kotlin/DokkaBase.kt
```diff
@@ -191,6 +191,12 @@ class DokkaBase : DokkaPlugin() {
         htmlPreprocessors with RootCreator applyIf { !delayTemplateSubstitution }
     }
 
+    val defaultSamplesTransformer by extending {
+        CoreExtensions.pageTransformer providing ::DefaultSamplesTransformer order {
+            before(pageMerger)
```
IgnatBeresnev, line 196: Could you please remove the `TODO` that I left about this above? This resolves it 

https://github.com/Kotlin/dokka/pull/3102/files#diff-50d822fb0953e3df82e04eb35a5362a37a47adf1390bd2c9fadf97f7f8a6e127R152

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
```diff
@@ -91,29 +92,41 @@ class DefaultTemplateModelFactory(val context: DokkaContext) : TemplateModelFact
...
+             * Workaround for resources that have no extension
+             * e.g. `<script src="`https://unpkg.com/kotlin-playground@1`"></script>`
```
IgnatBeresnev, line 98: If this resource is the only reason this workaround is needed, it looks like `https://unpkg.com/kotlin-playground@1/dist/playground.min.js` could be used instead - it only has the major version hardcoded (`@1`), but still redirects to the same page with `@1.28.0`.

plugins/base/src/main/kotlin/DokkaBase.kt
```diff
@@ -33,6 +32,7 @@ import org.jetbrains.dokka.plugability.PluginApiPreviewAcknowledgement
 import org.jetbrains.dokka.transformers.documentation.PreMergeDocumentableTransformer
 import org.jetbrains.dokka.transformers.pages.PageTransformer
 
+@Suppress("unused")
```
IgnatBeresnev, line 35: I see that the suppress migrated from file-level to class-level, which makes it more noticeable and questionable. Is it even needed? Maybe let's remove it completely? As far as I know, the compiler/ide shouldn't complain about unused public API

subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/kotlin/analysis/kotlin/internal/SampleProvider.kt
vmishenev: `SampleProvider` is a short-life instance (but it takes memory enough during its existence) that the transformer only needs.

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/KotlinSampleProvider.kt
vmishenev: IDE auto-generation...

subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/KotlinSampleProvider.kt
vmishenev: https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-stdlib-docs/plugins/dokka-samples-transformer-plugin/src/main/kotlin/org/jetbrains/dokka/kotlinlang/KotlinWebsiteSamplesTransformer.kt
It transforms `assert*` to `println`.
The easiest way to move the transformer to Dokka, but it needs discussion.

plugins/base/src/main/kotlin/DokkaBase.kt
vmishenev: > As far as I know, the compiler/ide shouldn't complain about unused public API

This inspection is enabled in IDE by default (in Kotlin plugin).

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
vmishenev: https://github.com/JetBrains/kotlin-playground or https://www.npmjs.com/package/kotlin-playground  uses `https://unpkg.com/kotlin-playground@1`
Potentially the file path can be changed, but the chance is negligible.

plugins/base/src/main/kotlin/DokkaBase.kt
IgnatBeresnev: It's enabled, and the unused properties are highlighted in gray, but neither IDEA nor the compiler seem to complain about it, i.e underline them with wiggly lines or fail the build due to `enableWarningsAsErrors`. So to me this suppress does nothing useful, but can backfire in the future. 

But it's up to you, I don't think it's important

subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/kotlin/analysis/kotlin/internal/SampleProvider.kt
IgnatBeresnev: oh, makes sense. The `Factory` pattern made me think that there could be multiple implementations of `SampleProvider`. I have an idea about how to make the API friendlier here without it, but no time to discuss it, unfortunately :( I guess I'll try to remember about it later

Could you please document that it's used for creating short-lived instances due to memory consumption?

plugins/base/src/main/kotlin/DokkaBase.kt
vmishenev: image

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
IgnatBeresnev: Sorry, but I really think this workaround can be easily avoided, and I think it should be, especially given that there are no tests and we have no plans to address it in the near future.

Is there a problem with using `https://unpkg.com/kotlin-playground@1/dist/playground.min.js`? As you said, the chance that the URL/path change is low, and I would honestly be more worried about us using a third party CDN in case of a supply chain attack. 

If you want this to be stable and work for years to come, I think the way forward is to ask the Kotlin Website team to provide a stable url instead, and we won't have to worry about it. In the meantime, it looks like we can just use `https://unpkg.com/kotlin-playground@1/dist/playground.min.js`.

subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/internal/SampleProvider.kt
```diff
@@ -0,0 +1,23 @@
+package org.jetbrains.dokka.analysis.kotlin.internal
+
+import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.InternalDokkaApi
+
+@InternalDokkaApi
+interface SampleProviderFactory {
+    fun build(): SampleProvider
+}
```
IgnatBeresnev, line 9: My previous comment about this was resolved, but I didn't notice if any documentation was added - perhaps, I missed it?

If not, could you please add a comment somewhere, stating that SampleProvider exists because its instances are supposed to be short-lived, with a reduced scope of use, otherwise it can lead to high memory consumption / leaks? Without it, it's easy to think the factory layer is useless and can be removed.

IgnatBeresnev (approved):
Thank you!

PR #3105, opened 2023-08-04
Refactor and document DisplaySourceSet

qwwdfsad:
* Decommission `SelfRepresentingSingletonSet` as harmful
* Document the purpose and the semantics of `DisplaySourceSet`
* Slightly refactor code around


core/src/main/kotlin/model/DisplaySourceSet.kt
```diff
@@ -1,32 +1,56 @@
...
+ * @property sourceSetIDs unique stable id of the display source set.
+ *  It is composite by definition, as it uniquely defines the source set and all nested source sets.
+ *  Apart from names, it also contains a substitute to a full source set path in order to differentiate
+ *  source sets with the same name in a stable manner.
+ * @property name corresponds to the name of the original [DokkaSourceSet]
+ * @property platform the platform of the source set. If the source set is a mix of multiple source sets
+ *  that correspond to multiple KMP platforms, then it is [Platform.common]
  */
-data class DisplaySourceSet(
+public data class DisplaySourceSet(
```
qwwdfsad, line 26: IMO it is worth making all constructors `internal` and leaving only `DokkaSourceSet.toDisplaySourceSet`. The API already kind of dictate that, but I'm not sure how drastic this change will be

qwwdfsad:
The integration test failure is related to pathsaver plugin in Knit -- I'll see if it's patchable locally

qwwdfsad:
I've patched the Knit, updated it in coroutines/serialization and checked it helps. Next step is to update the submodule

vmishenev (approved):

core/src/main/kotlin/model/DisplaySourceSet.kt
qwwdfsad: Probably worth figuring it out systematically. This one might be too strict, especially for 1.9.20

PR #3109, opened 2023-08-07
Get rid of bintray and jcenter mentions in our buildscript

qwwdfsad:
Otherwise, our integration tests try to resolve these repositories on failures


IgnatBeresnev (approved):
I thought we got rid of all jcenter and bintray mentions a long time ago, but apparently not 👍

PR #3116, opened 2023-08-12
fix(KTL-1177): Resize Observer has indefinite loop

zoobestik:
symbol wrapper script in dokka has Infinity cycling code based on [observe](https://github.com/Kotlin/dokka/blob/879d4f2652d528f20c95d5bb17899d6cb10c5e85/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js#L29)/[unobserve](https://github.com/Kotlin/dokka/blob/879d4f2652d528f20c95d5bb17899d6cb10c5e85/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js#L8) inside resize handler.

MRE:
```javascript
const observer = new ResizeObserver(entries => entries.forEach({ target } => {
  observer.unobserve(target);
  observer.observe(target); // fire re-call this ResizeObserver handler
}));

document.querySelectorAll('...').forEach(item => observer.observe(item));
```

### DIFF
--- a/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
+++ b/plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
@@ -16,16 +16,18 @@
 
     function wrapSymbolParameters(entry) {
         const symbol = entry.target;
-        const futureWidth = entry.borderBoxSize && entry.borderBoxSize[0] && entry.borderBoxSize[0].inlineSize;
+        const symbolBlockWidth = entry.borderBoxSize && entry.borderBoxSize[0] && entry.borderBoxSize[0].inlineSize;
 
         // Even though the script is marked as `defer` and we wait for `DOMContentLoaded` event,
-        // or if this block is a part of hidden tab, it can happen that `futureWidth` is 0,
+        // or if this block is a part of hidden tab, it can happen that `symbolBlockWidth` is 0,
         // indicating that something hasn't been loaded.
         // In this case, observer will be triggered onсe again when it will be ready.
-        if (futureWidth > 0) {
+        if (symbolBlockWidth > 0) {
             const node = symbol.querySelector(".parameters");
 
-            if (node) { // need to re-calculate if params need to be wrapped after resize
+            if (node) {
+                // if window resize happened and observer was triggered, reset previously wrapped
+                // parameters as they might not need wrapping anymore, and check again
                 node.classList.remove("wrapped");
                 node.querySelectorAll(".parameter .nbsp-indent")
                     .forEach(indent => indent.remove());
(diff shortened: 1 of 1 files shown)


plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
```diff
@@ -1,83 +1,57 @@
...
-    if (shouldWrapParams) {
-        parametersBlock.classList.add("wrapped")
-        parametersBlock.querySelectorAll("span.parameter").forEach(param => {
-            // has to be a physical indent so that it can be copied. styles like
-            // paddings and `::before { content: "    " }` do not work for that
-            param.prepend(createNbspIndent())
-        })
+    function wrapSymbolParameters(entry) {
+        const symbol = entry.target;
+        const futureWidth = entry.borderBoxSize && entry.borderBoxSize[0] && entry.borderBoxSize[0].inlineSize;
```
IgnatBeresnev, line 15: Not sure I understand this line fully. Why is it called `**future**Width`? Is it not the current width of the entry?

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
```diff
@@ -1,83 +1,57 @@
...
+        // or if this block is a part of hidden tab, it can happen that `futureWidth` is 0,
+        // indicating that something hasn't been loaded.
+        // In this case, observer will be triggered one again when it will be ready.
+        if (futureWidth > 0) {
+            const node = symbol.querySelector(".parameters")
+
+            if (node) { // need to re-calculate if params need to be wrapped after resize
+                node.classList.remove("wrapped")
+                node.querySelectorAll(".parameter .nbsp-indent")
+                    .forEach(indent => indent.remove());
```
IgnatBeresnev, line 27: The comment migrated from the old code, and it seems to be misleading now - it doesn't actually check if the params need to be re-wrapped after resize (like it did before with `window.onresize`), the reset will happen anyway, right?

Is the comment from the suggestion correct?

```suggestion
            if (node) {
            
                // if window resize happened and observer was triggered, reset previously wrapped
                // parameters as they might not need wrapping anymore, and check again
                node.classList.remove("wrapped")
                node.querySelectorAll(".parameter .nbsp-indent")
                    .forEach(indent => indent.remove());
```

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
zoobestik: As i know, resize observer handler happens on *layout* step of browser rendering. On this step all node sizes have been calculated but don't showed on screen, cause it may modified in handler. And, technically, it's future width. But in our case it isn't necessary information and over complicated. I returned old name `symbolBlockWidth`

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
zoobestik: ✅

plugins/base/src/main/resources/dokka/scripts/symbol-parameters-wrapper_deferred.js
IgnatBeresnev: ooh, ok, after you explained it - `futureWidth` makes total sense :)

IgnatBeresnev (approved):
:+1: :+1:

nikpachoo (approved):

PR #3118, opened 2023-08-15
Fix cast error to `ClassDescriptor`

vmishenev:
After [#3053 ](https://github.com/Kotlin/dokka/blob/cbd9733d3dd2f52992e98e7cebd072091a572529/plugins/base/src/main/kotlin/transformers/documentables/ActualTypealiasAdder.kt#L81) we have a cast error in StdLib


IgnatBeresnev (approved):
Adding a test would be nice

vmishenev:
We have the test already. To reproduce this issue, it needs another extensions order:
```
    val extensionsExtractor by extending {
        CoreExtensions.documentableTransformer with ExtensionExtractorTransformer() order {
            after(actualTypealiasAdder)
        }
    }
``` 
It is a general problem so I have opened #3125

PR #3119, opened 2023-08-15
add Gradle to Dokka users in README

aSemy:
Adds a link to the Gradle docs, generated with Dokka.

I thought it would be nice to add a link to a large library that most developers will have heard of.


IgnatBeresnev (approved):
Great idea, their API looks very sweet. Thanks!

PR #3139, opened 2023-08-24
Enable explicit API mode

IgnatBeresnev:
Depends on #3138, will rebase onto master once it gets merged.

This PR enables explicit API mode, which adds compiler checks for explicit visibility modifiers and return types.

You can see by the changes in the `.api` files that I changed some of the return types - I believe they were inferred incorrectly, as in most cases the supertype was deduced instead of the interface / base type. Other than that, I think everything should be fine in terms of backward compatibility.

Additionally, because visibility modifiers and return types made the lines longer, I did some small reformatting along the way, mostly adding line breaks or converting functions to block bodies.

### DIFF
--- a/build-logic/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts
+++ b/build-logic/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts
@@ -5,6 +5,7 @@
 package org.jetbrains.conventions
 
 import org.jetbrains.configureDokkaVersion
+import org.jetbrains.kotlin.gradle.dsl.ExplicitApiMode
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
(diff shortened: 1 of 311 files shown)


build-logic/src/main/kotlin/org/jetbrains/conventions/kotlin-jvm.gradle.kts
```diff
@@ -10,25 +11,44 @@ plugins {
 
 configureDokkaVersion()
 
-val projectsWithoutOptInDependency = setOf(
-    ":integration-tests", ":integration-tests:gradle", ":integration-tests:maven", ":integration-tests:cli")
+kotlin {
+    explicitApi = ExplicitApiMode.Strict
```
IgnatBeresnev, line 15: If `explicitApi` was set in `tasks.withType<KotlinCompile>()`, it would be applied to the test sources as well, which we don't want. But setting it in `kotlin {}` fixes it, so I did that and moved other compiler arguments into it as well.

core/content-matcher-test-utils/src/main/kotlin/matchers/content/ContentMatchersDsl.kt
```diff
@@ -20,120 +20,160 @@ fun ContentNode.assertNode(block: ContentMatcherBuilder<ContentComposite>.() ->
...
 
-fun ContentMatcherBuilder<*>.platformHinted(block: ContentMatcherBuilder<ContentGroup>.() -> Unit) =
+public fun ContentMatcherBuilder<*>.platformHinted(block: ContentMatcherBuilder<ContentGroup>.() -> Unit) {
     composite<PlatformHintedContent> { group(block) }
+}
 
-fun ContentMatcherBuilder<*>.list(block: ContentMatcherBuilder<ContentList>.() -> Unit) = composite(block)
+public fun ContentMatcherBuilder<*>.list(block: ContentMatcherBuilder<ContentList>.() -> Unit) {
+    composite(block)
+}
```
IgnatBeresnev, line 126: Instead of specifying the `: Unit` return type, I converted void functions to block bodies. Didn't think it would lead to so many changes lines, sorry :(

IgnatBeresnev:
Had to rebase and force push again after #3149

vmishenev (approved):

PR #3157, opened 2023-09-01
[K2] Fix functional type and improve logging for unresolved link

vmishenev:
For example
`typealias CompletionHandler = (cause: Throwable?) -> Unit` has a functional type with no type arguments in K2. In K1 we have a usual generic type


subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -111,7 +111,10 @@ internal class AnnotationTranslator {
                 DRI(packageName = "", classNames = ERROR_CLASS_NAME)
             )
 
-            KtUnsupportedAnnotationValue -> TODO()
+            KtUnsupportedAnnotationValue -> ClassValue(
```
vmishenev, line 114: Incorrect input source code should not lead to an exception.

IgnatBeresnev (approved):

PR #3164, opened 2023-09-05
Move Gradle runner

IgnatBeresnev:


runners/gradle-plugin/build.gradle.kts
IgnatBeresnev, line 1: This file was already copied in #3161, so I didn't move it

runners/gradle-plugin/gradle.properties
IgnatBeresnev, line 1: This file exists already in the target directory (created in #3161), so I deleted this one and moved the properties there.

PR #3165, opened 2023-09-05
Move integration tests

IgnatBeresnev:


dokka-integration-tests/gradle/projects/coroutines/kotlinx-coroutines
IgnatBeresnev, line 1: Need to make sure that git submodules are working in the end

PR #3176, opened 2023-09-05
[Structure refactor] move int test utils

aSemy:
Move the integration test utilities into a separate subproject

Depends on 

* https://github.com/Kotlin/dokka/pull/3175

PR #3177, opened 2023-09-05
[Structure refactor] add gradle properties to included builds

aSemy:
add `gradle.properties` to included builds

This is because the `dokka_version` value is mandatory, even for projects that aren't published.

This requirement should be removed later, and ideally a better way of sharing the project version should be implemented.

Depends on 

* #3175 
* #3176


dokka-integration-tests/gradle.properties
```diff
@@ -0,0 +1,11 @@
...
+# Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+#
+
+# Project Settings
+# TODO remove 'dokka_version' - it doesn't need to be mandatory for non-published subprojects
+dokka_version=1.9.10-SNAPSHOT
+org.jetbrains.dokka.javaToolchain.mainCompiler=8
+org.jetbrains.dokka.javaToolchain.testLauncher=8
+org.jetbrains.dokka.kotlinLanguageLevel=1.4
+dokka_integration_test_parallelism=2
```
IgnatBeresnev, line 11: Some notes:
* I remember we had some problems with running kotlinx.coroutines under Java 8, but because everything is coupled - a workaround was added to [override the launcher version to 11](https://github.com/Kotlin/dokka/blob/6a181a7a2b03ec263788d137610e86937a57d434/integration-tests/gradle/build.gradle.kts#L29-L34). It looks like now we can set it to 11 here and remove the hack?
* I think `kotlinLanguageLevel=1.4` isn't necessary for integration tests, they don't publish any API, so it should be OK remove it?
* `dokka_integration_test_parallelism=2` reminded me that I saw this [property set in `dokka-subprojects`](https://github.com/Kotlin/dokka/blob/project-structure-refactoring/dokka-subprojects/gradle.properties) as well, and I think it can be removed from there

Already starting to see benefits from separating the configurations :)

IgnatBeresnev (approved):
Already starting to see benefits from separating the configurations :) Ready to merge it 👍 

Left one optional comment, but it can be addressed later

dokka-integration-tests/gradle.properties
aSemy: > * I remember we had some problems with running kotlinx.coroutines under Java 8, but because everything is coupled - a workaround was added to [override the launcher version to 11](https://github.com/Kotlin/dokka/blob/6a181a7a2b03ec263788d137610e86937a57d434/integration-tests/gradle/build.gradle.kts#L29-L34). It looks like now we can set it to 11 here and remove the hack?

Unfortunately I think that must remain, because all Gradle integration tests still run together.

There are a few options to make it better, and I think one is to use [JVM Test Suites](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html), and make a separate suite for each project.

> * I think `kotlinLanguageLevel=1.4` isn't necessary for integration tests, they don't publish any API, so it should be OK remove it? It would somewhat help with writing tests as kotlinx.test has limited API in 1.4

Yes, I agree!

> * `dokka_integration_test_parallelism=2` reminded me that I saw this [property set in `dokka-subprojects`](https://github.com/Kotlin/dokka/blob/project-structure-refactoring/dokka-subprojects/gradle.properties) as well, and I think it can be removed from there

Agreed!

IgnatBeresnev:
Finally got back to it. I'll merge the PR into the branch and then address my comment above as a separate commit

Thanks once again!

PR #3180, opened 2023-09-18
[K2] Display enum entry members

vmishenev:


subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -453,28 +481,18 @@ internal class DokkaSymbolVisitor(
     }
 
     private fun KtAnalysisSession.visitEnumEntrySymbol(
-        enumEntrySymbol: KtEnumEntrySymbol
+        enumEntrySymbol: KtEnumEntrySymbol, scope: DokkaScope
     ): DEnumEntry = withExceptionCatcher(enumEntrySymbol) {
         val dri = getDRIFromEnumEntry(enumEntrySymbol)
         val isExpect = false
 
-        val scope = enumEntrySymbol.getMemberScope()
```
vmishenev, line 461: `getMemberScope` for an enum entry is removed in https://github.com/JetBrains/kotlin/commit/e72a38dc824b2d395157bc39b317102fbdb32265

plugins/base/src/test/kotlin/transformers/MergeImplicitExpectActualDeclarationsTest.kt
```diff
@@ -299,12 +297,16 @@ class MergeImplicitExpectActualDeclarationsTest : BaseAbstractTest() {
             configuration(true),
             cleanupOutput = true
         ) {
+            documentablesCreationStage = {
+                print(it)
+            }
```
IgnatBeresnev, line 302: Leftover from debugging?

plugins/base/src/test/kotlin/transformers/MergeImplicitExpectActualDeclarationsTest.kt
```diff
@@ -299,12 +297,16 @@ class MergeImplicitExpectActualDeclarationsTest : BaseAbstractTest() {
...
             pagesTransformationStage = { root ->
                 val classPage = root.dfs { it.name == "SMTH" } as? ClasslikePageNode
                 assertNotNull(classPage, "Tested class not found!")
 
                 val functions = classPage.findSectionWithName("Functions").assertNotNull("Functions")
-                val method1 = functions.children.singleOrNull().assertNotNull("method1")
+                val method1 = functions.children.single { it.sourceSets.size == 2 && it.dci.dri.singleOrNull()?.callable?.name == "method1" }
+                    .assertNotNull("method1")
```
IgnatBeresnev, line 309: Could you please explain why the filter needed to be changed in this test?

The test is called `should merge enum entries`, and the old assert somewhat makes sense - it checks that there's only 1 child even though 2 enums were given as input.

The new assert makes me think there are now 2 children, which I understand as they weren't merged and the test is now outdated

IgnatBeresnev: having gone through the code one more time, I think I have an idea - is something else in scope now, some other function that didn't exist before?

I assumed `functions.children` was of size 2 because the `method1` functions weren't merged, but I now think that one child is the merged `method1`, and the other child is something else?

vmishenev: I'm sorry for not explaining.
The test should check that `method1` is in two source sets.
`functions.children.singleOrNull()` has the disadvantages:
- in a potential case, Dokka can filter the second copy `method1` out rather than merge it.
- your idea: enum entry can have default functions. The test should not depend on it.
Currently, for K2 the enum has two functions: `pageMerger.classA.compareTo` and `method1`. In K1 `compareTo` is filtered out since it is `kotlin.enum.compareTo`. 
It is not relevant to enum by itself.

IgnatBeresnev: I see, thanks!

If enum entries didn't have `compareTo` functions before, but now they do and it's expected, could you please write a separate test for it? (along with other default functions that are now present, if any)

vmishenev: It is not expected, it is a bug in Analysis API. 
https://github.com/Kotlin/dokka/issues/3196

PR #3184, opened 2023-09-21
Implement analysis test API

IgnatBeresnev:
Adds convenient API for testing Kotlin analysis implementations.

### DIFF
--- a/core/src/main/kotlin/transformers/documentation/DefaultDocumentableMerger.kt
+++ b/core/src/main/kotlin/transformers/documentation/DefaultDocumentableMerger.kt
@@ -2,18 +2,27 @@
  * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-package org.jetbrains.dokka.base.transformers.documentables
+package org.jetbrains.dokka.transformers.documentation
 
 import org.jetbrains.dokka.DokkaConfiguration
-import org.jetbrains.dokka.base.utils.firstNotNullOfOrNull
+import org.jetbrains.dokka.InternalDokkaApi
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.properties.ExtraProperty
 import org.jetbrains.dokka.model.properties.MergeStrategy
 import org.jetbrains.dokka.model.properties.mergeExtras
 import org.jetbrains.dokka.plugability.DokkaContext
-import org.jetbrains.dokka.transformers.documentation.DocumentableMerger
+import org.jetbrains.dokka.CoreExtensions
 
-internal class DefaultDocumentableMerger(val context: DokkaContext) : DocumentableMerger {
+/**
+ * Should NOT be used outside of Dokka itself, there are no guarantees
+ * this class will continue to exist in future releases.
+ *
+ * This class resides in core because it is a non-trivial implementation
+ * for a core extension [CoreExtensions.documentableMerger], which is needed
+ * in modules that only have access to `dokka-core`.
+ */
+@InternalDokkaApi
+public class DefaultDocumentableMerger(context: DokkaContext) : DocumentableMerger {
     private val dependencyInfo = context.getDependencyInfo()
 
     override fun invoke(modules: Collection<DModule>): DModule? =
@@ -123,7 +132,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         }
     }
 
-    fun DPackage.mergeWith(other: DPackage): DPackage = copy(
+    public fun DPackage.mergeWith(other: DPackage): DPackage = copy(
         functions = mergeExpectActual(functions + other.functions) { f1, f2 -> f1.mergeWith(f2) },
         properties = mergeExpectActual(properties + other.properties) { p1, p2 -> p1.mergeWith(p2) },
         classlikes = mergeExpectActual(classlikes + other.classlikes) { c1, c2 -> c1.mergeWith(c2) },
@@ -133,7 +142,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         sourceSets = sourceSets + other.sourceSets
     ).mergeExtras(this, other)
 
-    fun DFunction.mergeWith(other: DFunction): DFunction = copy(
+    public fun DFunction.mergeWith(other: DFunction): DFunction = copy(
         parameters = merge(this.parameters + other.parameters) { p1, p2 -> p1.mergeWith(p2) },
         receiver = receiver?.let { r -> other.receiver?.let { r.mergeWith(it) } ?: r } ?: other.receiver,
         documentation = documentation + other.documentation,
@@ -145,7 +154,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         generics = merge(generics + other.generics) { tp1, tp2 -> tp1.mergeWith(tp2) },
     ).mergeExtras(this, other)
 
-    fun DProperty.mergeWith(other: DProperty): DProperty = copy(
+    public fun DProperty.mergeWith(other: DProperty): DProperty = copy(
         receiver = receiver?.let { r -> other.receiver?.let { r.mergeWith(it) } ?: r } ?: other.receiver,
         documentation = documentation + other.documentation,
         expectPresentInSet = expectPresentInSet ?: other.expectPresentInSet,
@@ -158,7 +167,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         generics = merge(generics + other.generics) { tp1, tp2 -> tp1.mergeWith(tp2) },
     ).mergeExtras(this, other)
 
-    fun DClasslike.mergeWith(other: DClasslike): DClasslike = when {
+    private fun DClasslike.mergeWith(other: DClasslike): DClasslike = when {
         this is DClass && other is DClass -> mergeWith(other)
         this is DEnum && other is DEnum -> mergeWith(other)
         this is DInterface && other is DInterface -> mergeWith(other)
@@ -167,7 +176,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         else -> throw IllegalStateException("${this::class.qualifiedName} ${this.name} cannot be merged with ${other::class.qualifiedName} ${other.name}")
     }
 
-    fun DClass.mergeWith(other: DClass): DClass = copy(
+    private fun DClass.mergeWith(other: DClass): DClass = copy(
         constructors = mergeExpectActual(
             constructors + other.constructors
         ) { f1, f2 -> f1.mergeWith(f2) },
@@ -185,7 +194,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         sourceSets = sourceSets + other.sourceSets
     ).mergeExtras(this, other)
 
-    fun DEnum.mergeWith(other: DEnum): DEnum = copy(
+    private fun DEnum.mergeWith(other: DEnum): DEnum = copy(
         entries = merge(entries + other.entries) { ee1, ee2 -> ee1.mergeWith(ee2) },
         constructors = mergeExpectActual(
             constructors + other.constructors
@@ -202,7 +211,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         sourceSets = sourceSets + other.sourceSets
     ).mergeExtras(this, other)
 
-    fun DEnumEntry.mergeWith(other: DEnumEntry): DEnumEntry = copy(
+    private fun DEnumEntry.mergeWith(other: DEnumEntry): DEnumEntry = copy(
         functions = mergeExpectActual(functions + other.functions) { f1, f2 -> f1.mergeWith(f2) },
         properties = mergeExpectActual(properties + other.properties) { p1, p2 -> p1.mergeWith(p2) },
         classlikes = mergeExpectActual(classlikes + other.classlikes) { c1, c2 -> c1.mergeWith(c2) },
@@ -211,7 +220,7 @@ internal class DefaultDocumentableMerger(val context: DokkaContext) : Documentab
         sourceSets = sourceSets + other.sourceSets
     ).mergeExtras(this, other)
 
-    fun DObject.mergeWith(other: DObject): DObject = copy(
+    private fun DObject.mergeWith(other: DObject): DObject = copy(
         functions = mergeExpectActual(functions + other.functions) { f1, f2 -> f1.mergeWith(f2) },
         properties = mergeExpectActual(properties + other.properties) { p1, p2 -> p1.mergeWith(p2) },
         classlikes = mergeExpectActual(classlikes + other.classlikes) { c1, c2 -> c1.mergeWith(c2) },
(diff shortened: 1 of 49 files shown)


core/src/main/kotlin/transformers/documentation/DefaultDocumentableMerger.kt
```diff
@@ -2,18 +2,27 @@
...
+/**
+ * Should NOT be used outside of Dokka itself, there are no guarantees
+ * this class will continue to exist in future releases.
+ *
+ * This class resides in core because it is a non-trivial implementation
+ * for a core extension [CoreExtensions.documentableMerger], which is needed
+ * in modules that only have access to `dokka-core`.
+ */
+@InternalDokkaApi
+public class DefaultDocumentableMerger(context: DokkaContext) : DocumentableMerger {
```
IgnatBeresnev, line 25: I moved this class to `core` since I need it for bootstrapping analysis tests.

Don't really understand why we need the `DocumentableMerger` extension point at all - I find it very difficult to believe someone will need to re-implement it or will even understand how it works (I don't). Should return to it some time in the future when we revisit all core extensions.

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/configuration/TestDokkaConfigurationBuilder.kt
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * Must be invoked manually during the verification of the
+     * higher-level [BaseTestDokkaConfigurationBuilder.verify].
+     */
+    open fun verify() {
+        additionalSourceRoots.forEach {
+            verifyFilePathStartsWithSlash("additionalSourceRoots", it)
+        }
+        additionalClasspath.forEach {
+            verifyFileExtension("additionalClasspath", it, ".jar") // should it include .klib?
```
vmishenev, line 105: > It can be a list of JARs, KLIBs, folders with .class files.

subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavaAnalysisTest.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.jetbrains.dokka.analysis.test.jvm.java
+
+import org.jetbrains.dokka.analysis.test.api.javaTestProject
+import org.jetbrains.dokka.analysis.test.api.parse
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JavaAnalysisTest {
```
vmishenev, line 12: Strict explicit API mode does not work here, in tests. It should have `public`

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Prepares this [TestProject] for analysis by creating
+     * the test files, setting up context and configuration.
+     */
+    private fun TestProject.initialize(outputDirectory: File): Pair<DokkaConfiguration, DokkaContext> {
+        analysisLogger.progress("Initializing and verifying project $this")
+        this.verify()
+        require(outputDirectory.exists() && outputDirectory.isDirectory) {
```
vmishenev, line 102: `outputDirectory.exists()` can be omitted

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
```diff
@@ -0,0 +1,223 @@
...
+        this.getTestData().getFiles().forEach {
+            val testDataFile = relativeToDir.resolve(it.pathFromProjectRoot.removePrefix("/"))
+            try {
+                testDataFile.parentFile.mkdirs()
+            } catch (e: Exception) {
+                // the IOException thrown from `mkdirs()` has no details and thus is more difficult to debug.
+                throw IllegalStateException("Unable to create dirs \"${testDataFile.parentFile}\"", e)
+            }
+
+            analysisLogger.debug("Creating \"${testDataFile.absoluteFile}\"")
```
vmishenev, line 131: Nitpicking: Do you mean  `testDataFile.absolutePath`?

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
```diff
@@ -0,0 +1,223 @@
...
+import org.jetbrains.dokka.transformers.documentation.DocumentableMerger
+import org.jetbrains.dokka.transformers.sources.SourceToDocumentableTranslator
+import org.jetbrains.dokka.utilities.DokkaConsoleLogger
+import org.jetbrains.dokka.utilities.LoggingLevel
+import java.io.File
+
+/**
+ * The main logger used for running Dokka and analyzing projects.
+ *
+ * Changing the level to [LoggingLevel.DEBUG] can help with debugging faulty tests
```
vmishenev, line 30: How a user can change a logging level?

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Analyzer of the test projects, it is essentially a very simple Dokka runner.
+ *
+ * Takes all virtual files of the given [TestProject], creates the real files for
+ * them in a temporary directory, and then runs Dokka with this temporary directory
+ * as the input user project. This allows us to simulate Dokka's behavior and results
+ * on a made-up project as if it were real and run via the CLI runner.
+ *
+ * Executes only a limited number of steps and uses a small subset of [CoreExtensions]
```
vmishenev, line 43: Is it used only one extension point `CoreExtensions.sourceToDocumentableTranslator`?

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
```diff
@@ -0,0 +1,223 @@
...
+ * that are necessary to test the analysis logic.
+ *
+ * Works only with single-module projects, where the source code of this project
+ * resides in the root `src` directory. Works with multiple source sets and targets,
+ * so both simple Kotlin/JVM and more complicated Kotlin Multiplatform project must work.
```
vmishenev, line 48: Currently, it works only with `Kotlin/JVM`, not KMP?

subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavaAnalysisTest.kt
qwwdfsad: Explicit API is deliberately disabled in tests, it only makes sense for main sources

subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavaAnalysisTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    @Test
+    fun javaTestProject() {
+        val testProject = javaTestProject {
+            dokkaConfiguration {
+                moduleName = "java-module-name-for-unit-test"
+            }
+            javaFile(pathFromSrc = "org/jetbrains/dokka/test/java/Bar.java") {
+                +"""
```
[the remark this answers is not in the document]
IgnatBeresnev: I understand it might lead to some issues down the road, but I couldn't think of any substantial problems to justify changing it as our use case is rather simple. 

This API is not public, so we'll have a chance to change it quickly if it's ever published externally.

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/TestProjectFactory.kt
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * Creates a project where a number of JVM language sources are allowed,
+ * like Java and Kotlin sources co-existing in the same source directory.
+ *
+ * This can be used to test interoperability between JVM languages.
+ *
+ * By default, it has two source sets:
+ * * Kotlin source set, residing in `/src/main/kotlin` - both Kotlin and Java sources are allowed
```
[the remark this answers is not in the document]
IgnatBeresnev: Hm, interesting, I didn't know Gradle was strict about it

Here, however, we're testing Dokka as an engine, not the Gradle runner, so as long as Dokka itself permits it and Kotlin doesn't prohibit it (does it?) - we need to be able to test it

Runners-wise, at the very least it's a valid case for the CLI runner, and judging by [documentation](https://kotlinlang.org/docs/maven.html#compile-kotlin-and-java-sources) the Maven runner might also support it (haven't checked though)

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/configuration/TestDokkaConfigurationBuilder.kt
IgnatBeresnev: I'll fix the comment :+1:. Let's change the actual check once we have the first related tests and are sure that they work.

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
IgnatBeresnev: For now I think yes. It would call the documentable merger extension point, but the only implementation is registered in dokka-base, so it's just created as a plain object for every invocation

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
IgnatBeresnev: At the moment there's API to test JVM projects only, but it doesn't mean it's not going to work for KMP projects :) Once KMP test project API is added - hopefully it'll just work, I don't see a reason why it shouldn't

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
IgnatBeresnev: At the moment, the only user is us (it's not public), so I thought we would just be able to change it temporarily.

But in #3195 I needed to assert some logging messages, so this logger became the default logger, and you can just pass your own one to the `parse` function.

subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavaAnalysisTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import org.jetbrains.dokka.analysis.test.api.javaTestProject
+import org.jetbrains.dokka.analysis.test.api.parse
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JavaAnalysisTest {
+
+    @Test
+    fun javaTestProject() {
```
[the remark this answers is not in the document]
IgnatBeresnev: Done: https://github.com/Kotlin/dokka/commit/9910be9453e2db74d49cc47bf1340513fba4a8a2

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/TestProjectFactory.kt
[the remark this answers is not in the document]
IgnatBeresnev: Oh, I think I understood where both my and your confusion is coming from.

What I really wanted from the user's perspective was to simulate the case when you have two source directories with their own files: `src/main/java` and `src/main/kotlin` (like [in spring](https://github.com/spring-projects/spring-framework/tree/main/spring-core/src/main)), and I didn't think of a better name than "source set" (a set of source files - here we have two), so I started using it. 

However, in Kotlin, "source sets" are used to describe a higher level abstraction which can contain multiple source directories, so I got confused while implementing it, so I just added these two source directories as different source sets :man_facepalming: 

In reality, it should be a single source set, but with two source directories and the ability to create files in both of them. Let me fix it.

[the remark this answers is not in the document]
IgnatBeresnev: Done, now `MixedJvmTestProject` has a single "jvm" source set, but two source directories (see `SampleMixedJvmAnalysisTest` for an example)

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt
IgnatBeresnev: Probably, yeah. Fixed it

subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/configuration/TestDokkaConfigurationBuilder.kt
IgnatBeresnev: Oh, sorry, missed the reply.

Not sure what you mean exactly. The only place where `kotlin/jvm/Strictfp.class` is used is to find the `.jar` file that contains it - passing the `.jar` file here works, yeah. But for now there's no need and there are no tests that pass individual class files

vmishenev: My bad. I thought `jvmStdlibPath` was an unpacked folder.

> But for now there's no need and there are no tests that pass individual class files

Indeed, the stdlib jar is only used in the classpath.

> // this check can be extended to accept .klib, .class or other files

I'm not sure Dokka's analysis can work with individual `.class` files. But it should work with directories containing `.class` files, e.g. `someProject/build/classes/`.

vmishenev (approved):

Issue #3131, opened 2023-08-17
Gradle Runner 2.0


IgnatBeresnev:
## Background

Dokka's current Gradle plugin has many problems:

* It is not idiomatic, not up to modern standard and does not follow best practices (see #2700).
* It's very difficult to use and confusing to configure, especially for beginners (confirmed by our UX researches and user interviews)
* It has a lot of internal technical debt, as multiple generations of maintainers worked on it without passing down the knowledge, and it wasn't looked after much this past year.
* There are things that do not cause trouble per se, but that we'd want to change given the chance (for example, Gradle integration tests).

Trying to improve the current plugin and resolve problems gradually proved to be difficult as it's gets overwhelming quick. Moreover, with this many changes needed, it would lead to multiple releases with noticeably broken backward compatibility.

At this point it seems like the best course of action is to just rip the bandaid off, and to redesign and reimplement the Gradle plugin from scratch, taking into consideration all of the experience and knowledge we have now. 

This issue exists to keep track of progress and to provide updates to the community.

## High-level goals

* Redesign the configuration DSL:
    * Use the knowledge we got from conducting UX researches and user interviews.
    * Comb through existing issues with the [configuration](https://github.com/Kotlin/dokka/issues?q=is%3Aissue+is%3Aopen+label%3Aconfiguration) tag and take them into account
    * Make sure the DSL respects project isolation and is configuration cache compatible from the get-go
    * Address our internal wants and wishes
* Make sure the new plugin is idiomatic and follows best practices, both on the surface and under the hood
* Significantly improve or even re-write user-facing documentation.

We won't be able to implement [all feature requests and fix all the bugs](https://github.com/Kotlin/dokka/issues?q=is%3Aissue+is%3Aopen+label%3A%22runner%3A+Gradle+plugin%22), but we'll make sure to lay a solid foundation to address them in the future.

## High-level plans

We plan to start with designing the new configuration DSL, and it'll take some thinking time and rounds of internal and external review. It's important to get it (mostly) right the first time around, as we want it to be very stable for at least the first couple of major releases after it's out.

[...]

PR #3188, opened 2023-09-22
add Dokkatoo

aSemy:
Initial draft PR for #3131 that adds Dokkatoo to Dokka

- replace package `dev.adamko.dokkatoo` with `org.jetbrains.dokka.dokkatoo`
- upgrade JVM version used to run Gradle to 11 (otherwise BCV-MU doesn't work, because it uses a version of Android Gradle Plugin that requires JVM11+)
- minor build script edits to get the build scripts to load


IgnatBeresnev (approved):
Enormous thanks for such a huge contribution and for investing time into making Dokka a better product in general! It's very much appreciated, both by maintainers and community members :)

I'll merge it as is, and we'll make it work in #3174

Issue #3120, opened 2023-08-15
java.net.URL is no longer a supported property type in Gradle


ZacSweers:
**Describe the bug**
Dokka currently uses URL for inputs for documentation urls, but in gradle 8.3 this warning is printed

```
Execution optimizations have been disabled for task ':sgp-monkeypatch-agp:dokkaHtmlPartial' to ensure correctness due to the following reasons:
  - In plugin 'org.jetbrains.dokka' type 'org.jetbrains.dokka.gradle.DokkaTaskPartial' property 'unsuppressedSourceSets.$0.externalDocumentationLinks.$0.packageListUrl' has @Input annotation used on type 'java.net.URL' or a property of this type. Reason: Type 'java.net.URL' is not supported on properties annotated with @Input because Java Serialization can be inconsistent for this type. For more information, please refer to https://docs.gradle.org/8.3-rc-4/userguide/validation_problems.html#unsupported_value_type in the Gradle documentation.
```

IgnatBeresnev:
Related: #2993

ZacSweers:
@IgnatBeresnev I saw this is marked as a 2.0 thing, but would you be open to deprecating this API for something better (probably just a String) and making the existing APIs route to it?

3flex:
It's also deprecated in Java 20, which will lead to more reports when Gradle 8.5 is released with support for running on Java 21 as projects update to latest LTS Java version.

whyoleg:
Hey!
We were able to overcome this issue don't breaking source/binary compatibility, and so the fix will be available in an upcoming release of Dokka. Proper removal of URL usage in Gradle will be implemented in the new Gradle plugin (#3131).
For more details, look [here](LINK_TO_COMMENT_ON_PR).

Additional note: to avoid receiving deprecation warnings when running Gradle with JDK 20 please use `URI("...").toURL()` instead of `URL("...")`.
Example:

```kotlin
sourceLink {
   remoteUrl.set(URI("https://github.com/kotlin/dokka/tree/master/src").toURL())
}
```

PR #3192, opened 2023-09-29
Don't expose java.net.URL

### LINKED ISSUE #2993: Replace `java.net.URL` with alternative, e.g. `java.net.URI`
### Is your feature request related to a problem? Please describe

Dokka Generator uses `java.net.URL` in the configuration.

`java.net.URL` has a number of drawbacks and unexpected behaviours.

* The hashcode is not serialized consistently, which affects Gradle caching https://github.com/gradle/gradle/issues/24979
* `URL.hashCode()` is blocking
* `.equals()` is not performance
* [In Java 20 the URL constructor is deprecated](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/net/URL.html#constructor-deprecation)
   > The `java.net.URL` constructors are deprecated. Developers are encouraged to use [`java.net.URI`](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/net/URI.html) to parse or construct a URL.

### Describe the solution you'd like

Here are some suggestions:

1. Replace all usages of `java.net.URL` with `java.net.URI`
2. Replace all usages of `java.net.URL` with `String`, and annotate it with `@Language("http-url-reference")` so IntelliJ recognises it as a URL
3. Create a custom class named something like `WebUrl` or `HttpLink`
    
    ```kotlin
    class Website(private val value: String)
    ```
4. Use another library's URL, e.g. [`ui.ktor.http.Url`](https://github.com/ktorio/ktor/blob/2.3.0/ktor-http/common/src/io/ktor/http/Url.kt)

### Describe alternatives you've considered

Dokkatoo can work around this and use an alternative, as it 'hides' the Dokka configuration from Gradle. 

Ideally Dokka would be updated, for consistency.

### Additional context

See also

* https://stackoverflow.com/questions/75908405/getting-around-the-deprecation-of-java-net-url
* https://stackoverflow.com/questions/2348399/why-does-java-net-urls-hashcode-resolve-the-host-to-an-ip
   > The IP resolution behavior of `URL.equals` and `URL.hashcode` is a trap for inexperienced Java programmers, even though it is clearly documented.
* https://stackoverflow.com/questions/1306290/i-know-that-java-net-url-tries-to-resolve-the-url-does-java-net-uri-do-the-same
   > 

Sure thing

(cc @martinbonnin)

boswelja:
When constructing a `URL` from Gradle, builds will spew warnings such as
```
In plugin 'org.jetbrains.dokka' type 'org.jetbrains.dokka.gradle.DokkaTaskPartial' property 'unsuppressedSourceSets.$0.externalDocumentationLinks.$0.packageListUrl' has @Input annotation used on type 'java.net.URL' or a property of this type. Reason: Type 'java.net.URL' is not supported on properties annotated with @Input because Java Serialization can be inconsistent for this type. For more information, please refer to https://docs.gradle.org/8.3/userguide/validation_problems.html#unsupported_value_type in the Gradle documentation.
```

This pull request is intended to remove problematic usages of `java.net.URL`, and thus resolve any compile-time warnings related to this.

This is a breaking change.

Resolves #2993 and #3120


core/src/main/kotlin/defaultConfiguration.kt
```diff
@@ -72,6 +72,10 @@ public data class SourceLinkDefinitionImpl(
     override val remoteLineSuffix: String?,
 ) : DokkaConfiguration.SourceLinkDefinition {
 
+    public constructor(
+        localDirectory: String, remoteUrl: String, remoteLineSuffix: String?
+    ) : this(localDirectory, URL(remoteUrl), remoteLineSuffix)
```
FWDekker, line 77: Since [the `URL` constructor is deprecated](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/net/URL.html#constructor-deprecation) as of Java 20, maybe use the recommended replacement `URI(remoteUrl).toURL()` instead? That has been available since at least Java 11.

FWDekker: Resolved in 61e0f3e47c0b9d726a51528fafdceac46dfe8673. Thank you!

pdvrieze:
Why not store the values as `URI` (to replace `URL`) rather than `String`? There is no problem with `URI`

boswelja:
Is there any special advantage a URI has over a String in this case?

pdvrieze:
> Is there any special advantage a URI has over a String in this case?

URI's allow various operations, they also restrict invalid inputs. Finally the implementation will still need to use `URL` (`URI.toURL`) for the connection so URI's would fit there.

But perhaps mostly, URI is a more specific type.

whyoleg:
Hey!
Thanks for the PR and all discussions here!
And sorry for the delay with review/answer (we are busy now with preparing the next version).

But I have some news now:

1. Unfortunately, we will not be able to accept this PR because of breaking changes (both binary and source). Breaking build scripts for users, even with good intentions, is not what we want to do now (more info on why in third bullet)
2. During refactoring happened in #3174 (rather big PR with a lot of changes) we also updated our build to Gradle 8.4 and so came across this issue. We were able to overcome it not breaking API/ABI. You can take a look at the approach [here](https://github.com/Kotlin/dokka/blob/40edb0b55f17458c7fa992650014d3fa00ea310e/dokka-runners/runner-gradle-plugin-classic/src/main/kotlin/org/jetbrains/dokka/gradle/GradleExternalDocumentationLinkBuilder.kt#L59-L63). In the end, in upcoming release (1.9.20) it will be fine to use `URL` there and no warning will be emitted.
3. We know that this solution may look like a hack, but we should keep backward compatibility. Though, during the implementation of the new Gradle plugin (#3131) we will make sure that `URL` is not exposed.

Additional note: to avoid receiving deprecation warnings when running Gradle with JDK 20 better to use `URI("...").toURL()` instead of `URL("...")`.
Example:

```kotlin
sourceLink {
   remoteUrl.set(URI("https://github.com/kotlin/dokka/tree/master/src").toURL())
}
```

Issue #3151, opened 2023-08-31
Make using of the compiler single-thread


vmishenev:
Dokka analyses Java and Kotlin code in multi-thread mode. ( see `AsyncSourceToDocumentableTranslator` and  https://github.com/Kotlin/dokka/issues/2936)

 Currently, the compiler is not thread-safe (see #1599, [KT-58327](https://youtrack.jetbrains.com/issue/KT-58327/JvmDependenciesIndexImpl-optimize-for-multi-thread-access)) so we have a patch for the compiler like https://github.com/Kotlin/dokka/pull/3143
 But we do not have a warranty that we patched everything or will not have new places in the future. It can lead to flaky analysis.
 
 **Possible solution**
 
 Make coroutines dispatcher single-thread.

IgnatBeresnev:
Update after an internal discussion: цe'll test if there's any noticeable performance difference between single-threaded and mult-threaded runs and proceed if there isn't. If there is, we'll think how to mitigate it or avoid it for K1 altogether.

vmishenev:
I measured Dokka performance in case of single-treaded calling compiler API/descriptors/PSI.
It was conducted on StdLib via CLI json (commit `f98451e38169a833f60b87618db4602133e02cf2`). The json configuration was got by a dump of Gradle runner.
The documentable model takes 16-20% more time to be generated in comparison with multi-threaded mode.
But it does not impact on total running time of Dokka since generating of documentable model  takes 10% of overall running time, i.e. total impact ±2%. 
So users will see no noticeable performance difference.

Some results: (the number of measurements=5)


Multi-thread |   |   | Single-thread |   |  
-- | -- | -- | -- | -- | --
  | Documentable model generating (sec) | Overall time (sec) |   | Documentable model generating (sec) | Overall time (sec)
_ | 56,005 | 515,93 | _ | 72,506 | 556,85
_ | 56,092 | 576,65 | _ | 69,625 | 549,12
_ | 56,385 | 587,15 | _ | 67,859 | 519,21
_ | 57,059 | 592,63 | _ | 67,504 | 558,76
_ | 55,216 | 537,96 | _ | 66,345 | 539,11
**Average** | 56,1514 | 562,064 |   | 68,7678 | 544,61

PR #3202, opened 2023-10-11
Make using of the compiler single-thread #3151

vmishenev:


plugins/base/src/main/kotlin/generation/SingleModuleGeneration.kt
```diff
@@ -61,11 +63,18 @@ public class SingleModuleGeneration(private val context: DokkaContext) : Generat
...
+    /**
+     * Implementation note: it runs in a separated single thread due to existing support of coroutines (see #2936)
+     */
+    @OptIn(DelicateCoroutinesApi::class)
+    public fun createDocumentationModels(): List<DModule> = newSingleThreadContext("Generating documentable model").use { coroutineContext -> // see https://github.com/Kotlin/dokka/issues/3151
+        runBlocking(coroutineContext) {
+            context.configuration.sourceSets.parallelMap { sourceSet -> translateSources(sourceSet, context) }.flatten()
+                .also { modules -> if (modules.isEmpty()) exitGenerationGracefully("Nothing to document") }
+        }
     }
```
IgnatBeresnev, line 75: There's a considerable chance it can lead to noticeably worse performance in projects with K1, and it might be especially evident with the CLI runner. Maybe you had time to do a few test runs to compare before and after?

It breaks the abstraction, but what do you think about using this single-threaded implementation for K2 only? Just to avoid raising too many questions for now with nothing to offer

IgnatBeresnev: Update on this question after some internal discussions:

Vadim did some benchmarking of the CLI runner to see if this has any impact on performance, and it looks like the documentable model takes 20% more time to be generated, but it's just a fraction of the overall time Dokka takes to generate documentation, so it's not really visible if you look at the whole run and not just an individual step.

Since we plan to get rid of coroutines either way, it's an acceptable solution for now.

IgnatBeresnev (approved):

PR #3203, opened 2023-10-11
[K2] Enable Java tests for symbols

vmishenev:
It is based on #3202. Otherwise, we will have the flaky tests.
Is is tested by https://github.com/Kotlin/dokka/actions/runs/6486414842

### DIFF
--- a/plugins/base/src/test/kotlin/content/params/ContentForParamsTest.kt
+++ b/plugins/base/src/test/kotlin/content/params/ContentForParamsTest.kt
@@ -247,7 +247,7 @@ class ContentForParamsTest : BaseAbstractTest() {
         }
     }
 
-    @JavaCode
+
     @Test
     fun `deprecated with multiple links inside`() {
         testInline(
@@ -352,7 +352,7 @@ class ContentForParamsTest : BaseAbstractTest() {
         }
     }
 
-    @JavaCode
+
     @Test
     fun `deprecated with an multiple inline links`() {
         testInline(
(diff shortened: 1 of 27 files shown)


plugins/base/src/test/kotlin/content/params/ContentForParamsTest.kt
```diff
@@ -805,7 +801,7 @@ class ContentForParamsTest : BaseAbstractTest() {
             | /**
             | * a normal comment
             | *
-            | * @return empty string when [Hash Map](java.util.HashMap.containsKey) doesn't contain value.
+            | * @return empty string when [Hash Map][java.util.HashMap.containsKey] doesn't contain value.
```
IgnatBeresnev, line 804: Out of curiosity: did you change it because the k2 tests failed or was it an unrelated fix while passing by?

vmishenev: The former. `[Hash Map](java.util.HashMap.containsKey)` is an incorrect link.

vmishenev: See https://kotlinlang.org/docs/kotlin-doc.html#links-to-elements

IgnatBeresnev: yeah, I understand, but it doesn't answer the question :) Does K1 support such incorrect links?

vmishenev: > Does K1 support such incorrect links?

Yes, that is why the test was passed.
But IDE and K2 do not support them. 
image

Issue #3132, opened 2023-08-18
Refactor artifact publishing


IgnatBeresnev:
## Background

The current solution for publishing artifacts is quite restrictive (adding a different repository requires hardcoding it), and the TeamCity configuration we use is neither convenient nor safe to use (it's a single build configuration with checkboxes and input fields that does everything), especially if used by people outside of the Dokka team.

There's an ongoing effort by @aSemy to refactor publishing in #3030, but it tries to keep compatibility with the current solution and TC build configuration, which can get in the way of some improvements. Now might be a good time to re-evaluate our approach to publishing, and to refactor not only the build scripts, but all of publishing in general (including TC configurations), which will allow us to implement long-awaited improvements.

## Publishing requirements

There are 3 types of repositories Dokka needs to publish to:

1. Gradle plugin portal.
2. Maven Central (sonatype)
3. Space Maven repositories.

### Gradle plugin portal

No specific requirements or complaints about the current solution, so this bit can most likely be reused.

Would be nice to have:

* Updating the version of the publishing plugin to the latest. They introduced some useful features such as [--validate-only](https://plugins.gradle.org/plugin/com.gradle.plugin-publish/1.2.1) - as far as I know, this is the only way to test publishing to Gradle Plugin Portal. The update was attempted before, but there were problems (#2919). Maybe with complete refactoring it'll be easier to fix.
* Due to #3131, it's likely that Dokka will have to support two Gradle plugins for a release or two. It would be nice to make the configuration either reusable or copy-pasteable.

### Maven Central

Currently, publishing to Maven Central happens in a single run, even though it really consists of 3 steps: 

1. The staging repository is initialized
2. All of the artifacts are uploaded
3. The [staging repository is closed](https://github.com/Kotlin/dokka/blob/f5393dd740eaf533fe9ff51003b175ef1c1418fe/build.gradle.kts#L33) (as part of the same run), effectively releasing the artifacts into the world.

It might be convenient in some cases, but I'd would like to separate these two steps. So publishing to Maven Central should initialize the staging repository and upload the artifacts there, and that's it. Then we can test the artifacts manually, make sure everything is correct, and close the staging repository ourselves. Otherwise (and how it works now), if something happens during publishing and one of the modules is built incorrectly, we can't do anything about it, because it's all one giant step.

[...]

PR #3204, opened 2023-10-11
Bump dependencies without visible API changes

IgnatBeresnev:
Bumps all dependencies that do not require any changes in Dokka's code, so hopefully there's no regression and it won't backfire :)

Ran the tests locally — looks good. Haven't seen the HTML output yet though, will do once GitHub generates it.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -5,9 +5,9 @@ gradlePlugin-kotlin = "1.9.10"
 gradlePlugin-android = "4.2.2"
 gradlePlugin-dokka = "1.9.0"
 
-kotlinx-coroutines = "1.6.3"
-kotlinx-collections-immutable = "0.3.4"
-kotlinx-bcv = "0.12.1"
+kotlinx-coroutines = "1.7.3"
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -39,7 +39,7 @@ apacheMaven-archiver = "2.5"
 apacheMaven-pluginTools = "3.5.2"
```
IgnatBeresnev, line 39: JFYI I didn't touch anything related to Maven, these versions are very, very old, and I think the chance of breaking something here is considerable. Let's do it when we start improving the Maven runner as a whole

gradle/libs.versions.toml
```diff
@@ -51,8 +51,8 @@ gradlePlugin-gradlePluginPublish = "0.20.0"
 gradlePlugin-gradleNode = "3.5.1"
```
IgnatBeresnev, line 51: JFYI I also didn't touch anything related to publishing to avoid breaking something. These versions and dependencies should be revisited as part of #3132

vmishenev (approved):

PR #3205, opened 2023-10-11
Bump korte aka korlibs-template to 4.0.10

IgnatBeresnev:


plugins/javadoc/src/main/kotlin/org/jetbrains/dokka/javadoc/renderer/KorteJavadocRenderer.kt
```diff
@@ -4,7 +4,7 @@
 
 package org.jetbrains.dokka.javadoc.renderer
 
-import com.soywiz.korte.*
+import korlibs.template.*
```
IgnatBeresnev, line 7: The korlibs packages were simply renamed, see https://github.com/korlibs/korge/issues/1319#issuecomment-1452103500

PR #3206, opened 2023-10-12
Bump kotlinx.html to 0.9.1

IgnatBeresnev:


plugins/base/api/base.api
```diff
@@ -549,7 +549,6 @@ public final class org/jetbrains/dokka/base/renderers/html/command/consumers/Imm
 	public fun onTagContentEntity (Lkotlinx/html/Entities;)V
 	public fun onTagContentUnsafe (Lkotlin/jvm/functions/Function1;)V
 	public fun onTagEnd (Lkotlinx/html/Tag;)V
-	public fun onTagError (Lkotlinx/html/Tag;Ljava/lang/Throwable;)V
```
IgnatBeresnev, line 552: I don't fully understand the reason why it was removed (see https://github.com/Kotlin/kotlinx.html/pull/224), but everything works, soo hopefully it wasn't used :)

vmishenev (approved):

plugins/base/api/base.api
qwwdfsad: `onTagError` implies that every `visit*` should do `try-finally` section which is kind of bytecode-heavy.

With heavy tag usages it, AFAIR, bloated bytecode

PR #3212, opened 2023-10-16
Fix active-tab selection for different page contexts

### LINKED ISSUE #2899: `active-tab` value reused in different set of tabs
For saving the last selected tab, Dokka stores it in the local storage in `active-tab` field.
Previously, the tabs on the package and classlike pages were similar, and the value was reused. But now the sets are different. And after selecting a tab in the classlike, the state of the selected tab in the package page is gone.

**Installation**
- Dokka version: 1.8.10

whyoleg:
fixes #2899

### DIFF
--- a/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
+++ b/plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
@@ -163,7 +163,6 @@ public open class HtmlRenderer(
                     contentTabs.forEachIndexed { index, contentTab ->
                         button(classes = "section-tab") {
                             if (index == 0) attributes["data-active"] = ""
-                            attributes["data-type"] = pageContext.getDocumentableType() ?: ""
                             attributes[TOGGLEABLE_CONTENT_TYPE_ATTR] =
                                 contentTab.tabbedContentTypes.joinToString(",") { it.toHtmlAttribute() }
                             text(contentTab.text)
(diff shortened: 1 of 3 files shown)


plugins/base/src/main/kotlin/renderers/html/HtmlRenderer.kt
```diff
@@ -163,6 +163,7 @@ public open class HtmlRenderer(
                     contentTabs.forEachIndexed { index, contentTab ->
                         button(classes = "section-tab") {
                             if (index == 0) attributes["data-active"] = ""
+                            attributes["data-type"] = pageContext.getDocumentableType() ?: ""
```
IgnatBeresnev, line 166: Oh btw, it might be worth adding a test that verifies that this attribute exists on the pages, and adding a comment to these tests that the value is used in js tab-handling scripts wouldn't hurt. Should be a simple and quick intro into writing html tests as well

We don't have any UI tests, so if this gets lost during refactoring or someone thinks it's not needed, it might lead to regression.

whyoleg: @vmishenev thanks! You are right, I was thinking, that it will be harder to reusing it, but in the end it simplified code a lot :)
Pushed changes with usage of it.

So I will now add test for checking that just `data-page-type` attribute is present on page

whyoleg:
Ready for review

IgnatBeresnev (approved):
:+1:

PR #3227, opened 2023-10-17
[K2] Support kotlin-as-java and javadoc plugins and update version of Analysis API

vmishenev:
Also, it contains some fixes to make unit tests of kotlin-as-java and javadoc plugins green.


subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolInheritanceBuilder.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.jetbrains.dokka.analysis.kotlin.internal.InternalKotlinAnalysisPlugin
+import org.jetbrains.dokka.model.*
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.plugability.plugin
+import org.jetbrains.dokka.plugability.querySingle
+
+/**
+ * This is copy-pasted from org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.impl.DescriptorInheritanceBuilder and adapted for symbols
+ */
+internal class SymbolInheritanceBuilder(context: DokkaContext) : InheritanceBuilder {
```
vmishenev, line 23: There is no way to verify it due to #3232

plugins/javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocIndexTest.kt
```diff
@@ -69,14 +70,30 @@ internal class JavadocIndexTest : AbstractJavadocTemplateMapTest() {
         }
     }
 
+    @Tag("onlyDescriptors") // K2: does not have getters: getName, getOrdinal
```
IgnatBeresnev, line 73: The comment explains the code (which is the more obvious part), but it doesn't explain the intention and reasoning behind it. Why doesn't K2 have getters for name and ordinal? Is this expected? Will it be fixed in the future? If this is a bug of K1, maybe the comment should be about that instead will be less confusing: "K1 mistakenly generates getName and getOrdinal functions"

It looks like this test can also benefit from what I proposed above about descriptor-only or symbol-only tests with different asserts, as right now this test is simply muted for K2 and it's unclear if it will ever be unmuted

subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/SymbolsAnalysisPlugin.kt
```diff
@@ -104,14 +105,14 @@ public class SymbolsAnalysisPlugin : DokkaPlugin() {
...
-    /* internal val kotlinToJavaMapper by extending {
-         plugin<InternalKotlinAnalysisPlugin>().kotlinToJavaService providing { DescriptorKotlinToJavaMapper() }
+    internal val kotlinToJavaMapper by extending {
+         plugin<InternalKotlinAnalysisPlugin>().kotlinToJavaService providing { SymbolKotlinToJavaMapper() }
      }
 
-     intern val descriptorInheritanceBuilder by extending {
-         plugin<InternalKotlinAnalysisPlugin>().inheritanceBuilder providing { DescriptorInheritanceBuilder() }
+    internal val descriptorInheritanceBuilder by extending {
+         plugin<InternalKotlinAnalysisPlugin>().inheritanceBuilder providing ::SymbolInheritanceBuilder
```
IgnatBeresnev, line 113:
```suggestion
    internal val symbolInheritanceBuilder by extending {
         plugin<InternalKotlinAnalysisPlugin>().inheritanceBuilder providing ::SymbolInheritanceBuilder
```

IgnatBeresnev (approved):

PR #3235, opened 2023-10-18
Add the link to GitHub repo to the header if there are source links defined

### LINKED ISSUE #2948: Dokka HTML: optionally add a custom link to source code homepage in navbar

I would like to be able to add a link to my source code homepage in the Dokka HTML nav bar

Example: a link to GitHub

image

https://adamko-dev.github.io/dokkatoo/

Update the navbar template to include an optional 'homepage' icon. If a URL is provided, the icon is shown. I would also like to be able to theme the icon using CSS. 

Ideally, Dokka would pre-package some icons so that I could select 'GitHub', 'GitLab', 'Gitea', etc as an icon.

It is not possible to do this via CSS `:before`/`:after`, as pseudo elements cannot have href links.

I have manually implemented this using find-replace in Dokkatoo

https://github.com/adamko-dev/dokkatoo/blob/5da6e4d1ed8654b9c6b5dae7ece6a440bea8230c/docs/build.gradle.kts#L27-L57

```kotlin
tasks.dokkatooGeneratePublicationHtml {
  doLast {
    outputDirectory.get().asFile.walk()
      .filter { it.isFile && it.extension == "html" }
      .forEach { file ->
        file.writeText(
          file.readText()
            .replace(
              """<html>""",
              """<html lang="en">""",
            )
            .replace(
              """
                <button id="theme-toggle-button">
              """.trimIndent(),
              """
                <div id="github-link"><a href="https://github.com/adamko-dev/dokkatoo/"></a></div>
                <button id="theme-toggle-button">
              """.trimIndent(),
            )
        )
      }
  }
}
```

**Additional context**

Not right now - the find and replace workaround is suitable.

whyoleg:
Partially fixes #2948 
For now only icon and link to GitHub is supported.

The logic is that, if `sourceLinks` configured and pointed to GitHub, we will add icon with link to GitHub repository - otherwise we show nothing.


plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
```diff
@@ -91,6 +93,30 @@ public class DefaultTemplateModelFactory(
             ?: DokkaBaseConfiguration.defaultFooterMessage)
     )
 
+    private fun calculateSourceUrlFromSourceLinks(): String? {
+        val githubLinkRegex = Regex("http(s)?://github\\.com/([\\w,\\-_]+)/([\\w,\\-_]+)/.*")
+
+        fun parseGithubInfo(link: String): Pair<String, String>? {
+            val (
+                _, // entire match
+                _, // optional 's' in http
```
vmishenev, line 102: What do you think about ```https?://github\\.com/([\\w,\\-_]+)/([\\w,\\-_]+)/.*``` where `s` is not a capturing group?

vmishenev (approved):

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
```diff
@@ -91,6 +93,30 @@ public class DefaultTemplateModelFactory(
             ?: DokkaBaseConfiguration.defaultFooterMessage)
     )
 
+    private fun calculateSourceUrlFromSourceLinks(): String? {
+        val githubLinkRegex = Regex("http(s)?://github\\.com/([\\w,\\-_]+)/([\\w,\\-_]+)/.*")
```
IgnatBeresnev, line 97: Optional as it's not that important atm: it looks like this function will be called for each page, so it could be optimized slightly:

1. Afaik, each instance of `Regex` will compile the pattern under the hood, which takes some time, and it's generally advised to reuse it if it makes sense and is possible. I think in our case the Regex can be extracted into a companion or a top-level val or something, to be reused.
2. We only need to calculate the return value once for the whole `DokkaConfiguration` (it will not change in-between invocations), and then we can reuse it. But I can't think of where to save this result off the top of my head. Maybe as a lateinit or a lazy property somewhere? Any ideas? :)

plugins/base/src/test/kotlin/renderers/html/HeaderTest.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Element
+import renderers.testPage
+import java.net.URL
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+class HeaderTest : HtmlRenderingOnlyTestBase() {
```
IgnatBeresnev, line 18: Oh, I even forgot we have `HtmlRenderingOnlyTestBase`... I personally try to stick to `testInline` (see `SignatureTest` for an example), it can verify all stages and levels (from documentable to html), and I think its tests are more correct and will be easier to refactor down the road. 

Once dokka-base is broken down into more manageable modules, I'll be advocating for cutting down on the number of test DSLs, so that there's one DSL per module or at least one per level, and it should be clear which one to use in which scenario (unlike now).

So these tests can stay the way that they are (as long as they work and find bugs) and we'll address it later on a larger scale, or you can give `testInline` a go, up to you :)

plugins/base/src/test/kotlin/renderers/html/HtmlRenderingOnlyTestBase.kt
```diff
@@ -48,13 +49,15 @@ abstract class HtmlRenderingOnlyTestBase : RenderingOnlyTestBase<Element>() {
         finalizeCoroutines = false
     )
 
-    override val context = MockContext(
+    override val context by lazy { context(configuration) }
+
+    fun context(testConfiguration: DokkaConfiguration): MockContext = MockContext(
```
IgnatBeresnev, line 54: (if you stick to `HtmlRenderingOnlyTestBase` and it's not reverted)

Was there some sort of a problem with it and this refactoring helped?

IgnatBeresnev (commented):
For some reason, I don't see the icon in the header :( [temporary preview link from GH actions](https://dokka-snapshots.s3.eu-central-1.amazonaws.com/source-link/coroutines/e38ff6b/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/index.html)

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
vmishenev: If `sourceUrl` does not depend on a page, you can use `buildSharedModel`.

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
whyoleg: yeah, agree, that should work, will fix it, thx!

plugins/base/src/main/kotlin/renderers/html/innerTemplating/DefaultTemplateModelFactory.kt
whyoleg: Thx!
1. extracted `Regex` to companion
2. moved calculation of link to `buildSharedModel`

plugins/base/src/test/kotlin/renderers/html/HeaderTest.kt
whyoleg: Okay, agree, I will refactor those new tests to use `BaseAbstractTest`

plugins/base/src/test/kotlin/renderers/html/HtmlRenderingOnlyTestBase.kt
whyoleg: The problem is, while overall all of extensions are the same for all tests in file, configuration should be different per different tests - otherwise it's not possible to change some specific variables here.
F.e. in this case, I need to test output based on `configuration.sourceSets[i].sourceLinks`, and so `configuration` is different in each test function

whyoleg:
> For some reason, I don't see the icon in the header :(

Most likely this is because kotlinx.coroutines has no configuration of sourceLinks for root `htmlMultiModule` task. And also at least now it has sourceLinks configured for `core` project (other modules don't have `source` in HTML generated, f.e. https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-android/kotlinx.coroutines.android/-handler-dispatcher/)

plugins/base/src/main/kotlin/DokkaBaseConfiguration.kt
```diff
@@ -12,6 +12,7 @@ public data class DokkaBaseConfiguration(
     var customStyleSheets: List<File> = defaultCustomStyleSheets,
     var customAssets: List<File> = defaultCustomAssets,
     var separateInheritedMembers: Boolean = separateInheritedMembersDefault,
+    var homepageLink: String? = null,
```
IgnatBeresnev, line 15: > Do we need to keep backward compatibility for DokkaBaseConfiguration?

I think it's worth moving this var to be the last parameter of the constructor, I think it'll help preserve source compatibility (in case someone imports the class in their build scripts and doesn't use named parameters), so it's at least something. 

Binary compatibility - I'm not sure how much sense it makes to try to preserve it here. Almost every Dokka release is breaking binary compatibility in some way, mostly due to data classes everywhere, so I wouldn't say it's expected that out-of-sync versions will work together flawlessly. And the HTML format will be configured differently once the new Gradle plugin is released, so I'd say we can ignore it for now

IgnatBeresnev (approved):
👍👍

IgnatBeresnev:
> it's a temporary icon I believe

We'll ask someone from the Kotlin Website team to help us with the icon and the styles, I've created a separate issue for that: #3300

plugins/base/src/main/kotlin/DokkaBaseConfiguration.kt
whyoleg: Yeah, agree, will move parameter to be the last (nor sure, why I haven't done it from the beginning :) )

integration-tests/gradle/src/integrationTest/kotlin/org/jetbrains/dokka/it/gradle/MultiModule0IntegrationTest.kt
```diff
@@ -73,6 +73,21 @@ class MultiModule0IntegrationTest : AbstractGradleIntegrationTest() {
             "Expected moduleC being mentioned in -modules.html"
         )
 
+        val indexHtmls = outputDir.walkTopDown().filter {
+            it.isFile && it.name == "index.html"
```
vmishenev, line 77: I think all html files except `navigation` might be checked.

plugins/base/src/main/resources/dokka/templates/includes/header.ftl
```diff
@@ -21,6 +21,9 @@
         <@source_set_selector.display/>
     </div>
     <div class="navigation-controls">
+        <#if homepageLink?has_content>
+            <div class="navigation-controls--btn navigation-controls--homepage" id="homepage-link" role="button"><a href="${homepageLink}"></a></div>
```
vmishenev, line 25: I have not checked but it seems to be `<a href="${homepageLink}"> <div class="navigation-controls--btn navigation-controls--homepage" id="homepage-link" role="button"/></a>`

vmishenev (approved):

plugins/base/src/main/resources/dokka/templates/includes/header.ftl
whyoleg: Im not an html-expert, but at least my variant works. If to switch tags, it doesn't work at my side (may be need to adapt more code)
I think it could be rechecked in scope of #3300

PR #3236, opened 2023-10-18
[K2] Reorganize project model for MPP

vmishenev:
We map [Dokka's source set ](https://kotlinlang.org/docs/dokka-gradle.html#source-set-configuration) directly to [a source module ](https://github.com/JetBrains/kotlin/blob/500de9f9bba5baad7e0f2c9260c298bb21cd5de7/analysis/project-structure/src/org/jetbrains/kotlin/analysis/project/structure/KtModule.kt#L25C27-L25C27) of Analysis API inside one [Analysis Standalone session](https://github.com/JetBrains/kotlin/blob/141333bdcd22c9d257d88dbce50fa8499206a0f8/analysis/analysis-api-standalone/src/org/jetbrains/kotlin/analysis/api/standalone/StandaloneAnalysisAPISession.kt#L18).

Analysis API session is created in `src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt` (see  `fun createAnalysisSession`)

Before the PR, one Dokka's source set corresponded to one Standalone API session with one source module that has source roots from dependent source sets.

The PR allows the enabling of some tests annotated with `OnlyDescriptorsMPP`.  
Also, tests with `OnlyDescriptorsMPP` that have unresolved common symbols are fixed by the new version of Analysis API.

### DIFF
--- a/plugins/base/src/test/kotlin/content/annotations/SinceKotlinTest.kt
+++ b/plugins/base/src/test/kotlin/content/annotations/SinceKotlinTest.kt
@@ -14,10 +14,7 @@ import org.jetbrains.dokka.model.doc.CustomTagWrapper
 import org.jetbrains.dokka.model.doc.Text
 import org.jetbrains.dokka.pages.ContentPage
 import signatures.AbstractRenderingTest
-import utils.ParamAttributes
-import utils.TestOutputWriterPlugin
-import utils.assertNotNull
-import utils.bareSignature
+import utils.*
 import kotlin.test.*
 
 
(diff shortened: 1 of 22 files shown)


plugins/base/src/test/kotlin/content/annotations/SinceKotlinTest.kt
```diff
@@ -125,30 +122,54 @@ class SinceKotlinTest : AbstractRenderingTest() {
         val configuration =   dokkaConfiguration {
             sourceSets {
                 sourceSet {
-                    sourceRoots = listOf("src/")
+                    sourceRoots = listOf("src/jvm/")
```
vmishenev, line 125: [#3239](https://github.com/Kotlin/dokka/issues/3239)

plugins/base/src/test/kotlin/linkableContent/LinkableContentTest.kt
```diff
@@ -286,15 +285,15 @@ class LinkableContentTest : BaseAbstractTest() {
             sourceSets {
                 sourceSet {
                     analysisPlatform = "js"
-                    sourceRoots = listOf("jsMain", "commonMain", "jvmAndJsSecondCommonMain").map {
+                    sourceRoots = listOf("jsMain").map {
```
vmishenev, line 288: #3239

plugins/base/src/test/kotlin/model/InheritorsTest.kt
```diff
@@ -60,15 +60,22 @@ class InheritorsTest : AbstractModelTest("/src/main/kotlin/inheritors/Test.kt",
     fun multiplatform() {
         val configuration = dokkaConfiguration {
             sourceSets {
+                val commonSourceSet = sourceSet {
```
vmishenev, line 63: #3239

IgnatBeresnev (approved):
Thank you for the explanation and the comments, it made the review much easier :+1:

yanex (approved):

PR #3242, opened 2023-10-20
Fix overflow of long names in HTML

### LINKED ISSUE #3241: Long names overflow their container in HTML
**Describe the bug**
While investigating #3093 I found some other places where long names overflow their container in HTML.

**Screenshots**
dokka-overflow-code-block
dokka-overflow-camelcase

**To Reproduce**
Any project containing a really long name.

**Installation**
- Operating system: Linux
- Build tool: Gradle v7.6
- Dokka version: master branch

Yes, but I'm currently stuck on failing test cases that check the structure of the resulting html.

seelchen:
This PR aims to fix the issue that long names sometimes overflow in the generated HTML pages.

The CSS property word-break: break-word is added to the root style to prevent any general overflows.
`ContentText` nodes with the `Monospace` style are made breakable since long names can occur in code blocks as well.

Resolves #3241.

### DIFF
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
@@ -899,8 +899,7 @@ public open class HtmlRenderer(
                     buildText(textNode, unappliedStyles - styleToApply)
                 }
             }
-            textNode.hasStyle(ContentStyle.RowTitle) || textNode.hasStyle(TextStyle.Cover) ||
-                    textNode.hasStyle(TextStyle.Monospace) ->
+            textNode.hasStyle(ContentStyle.RowTitle) || textNode.hasStyle(TextStyle.Cover) ->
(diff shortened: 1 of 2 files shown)


seelchen:
I am currently stuck because a lot of tests that check the structure of the generated HTML elements fail. Since new `span` tags are added to make the text breakable, the structure is different than expected.

I can rewrite all expected structures in the failing tests but I would like to get some feedback first if my approach to this issue is fine with you.

whyoleg:
Hey!
Thank you for PR and sorry for waiting.
We are now in process of preparing a UI Kit which will specify what to do in such situations. On current moment it's not really clear what to do with overflow/word-wrapping in code-blocks:
* in samples we provide horizontal scrolling
* on mobile(small) screens we could have rather strange behaviour when word-wrapping happens

Here is an example from `kotlinx.serialization` on small screen:
image
So I would say, that may be current wrapping in code-blocks exists there for historical reasons, rather than intentional.
And as this PR will change a lot of tests to adapt to changes, Im not sure that it's worth to do it right now until complete design will be ready.

Still, I think that adding `word-break` to `root` could be a good addition.
@berezinant do you have anything against it?
If not, It would be good to rebase PR and leave changes only regarding `word-break` style.

PR #3262, opened 2023-10-25
Update Kotlin compiler and KGP to 1.9.20-RC2

vmishenev:
The current changes are related to https://github.com/JetBrains/kotlin/commit/d797505f06d640b666829bbfb4b7d7c67f812026

### DIFF
--- a/core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
+++ b/core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
@@ -191,12 +191,11 @@ public abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : Dokk
             ?.replaceAfter(".jar", "")
     }
 
-    protected val commonStdlibPath: String? by lazy {
-        // TODO: feels hacky, find a better way to do it
-        ClassLoader.getSystemResource("kotlin/UInt.kotlin_metadata")
-            ?.file
-            ?.replace("file:", "")
-            ?.replaceAfter(".jar", "")
+    protected val commonStdlibPath: String?  by lazy {
+        // `kotlin-stdlib-common` is legacy
+        // we can use any platform dependency
+        // since common code should be resolved with all platform
+        jvmStdlibPath
     }
 
     protected val stdlibExternalDocumentationLink: ExternalDocumentationLinkImpl = ExternalDocumentationLinkImpl(
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -10,7 +10,7 @@ kotlinx-collections-immutable = "0.3.6"
 kotlinx-bcv = "0.13.2"
 
 ## Analysis
-kotlin-compiler = "1.9.10"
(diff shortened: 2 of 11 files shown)


core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
```diff
@@ -191,12 +191,11 @@ public abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : Dokk
...
-        // TODO: feels hacky, find a better way to do it
-        ClassLoader.getSystemResource("kotlin/UInt.kotlin_metadata")
-            ?.file
-            ?.replace("file:", "")
-            ?.replaceAfter(".jar", "")
+    protected val commonStdlibPath: String?  by lazy {
+        // `kotlin-stdlib-common` is legacy
+        // we can use any platform dependency
+        // since common code should be resolved with all platform
+        jvmStdlibPath
```
IgnatBeresnev, line 198: Can we remove/deprecate the existing paths (`jvmStdlibPath`, `jsStdlibPath`, `commonStdlibPath`), and just have a single `stdlibPath` instead?

IgnatBeresnev:
> The current changes are related to https://github.com/JetBrains/kotlin/commit/d797505f06d640b666829bbfb4b7d7c67f812026

Do you know if this has any user-visible effect? That is, can its removal lead to some bugs, inconsistencies or some other problems in the generated documentation/signatures?

core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
vmishenev: No, `jvmStdlibPath` and `jsStdlibPath` contain platform-specific declarations.

vmishenev:
> Do you know if this has any user-visible effect? That is, can its removal lead to some bugs, inconsistencies or some other problems in the generated documentation/signatures?

According to [KT-59643](https://youtrack.jetbrains.com/issue/KT-59643) it disables merging a forward declaration with an actual declaration. It is how the compiler works now. I am unsure it changes the view of signatures.

core/test-api/src/main/kotlin/testApi/testRunner/TestRunner.kt
IgnatBeresnev: Oh, I see what you mean. I misunderstood the comment as "you can use any platform stdlib in any scenario (even use jvm stdlib in js)", but you meant it as "any platform stdlib can be used as a 'common' stdlib"

IgnatBeresnev (approved):
Could you please update `TestedVersions` just in case?

IgnatBeresnev (approved):

gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 
-gradlePlugin-kotlin = "1.9.10"
+gradlePlugin-kotlin = "1.9.20-RC2"
```
whyoleg, line 3: kotlin 1.9.20 is released already :)

Issue #3112, opened 2023-08-11
Migrate Google's devsite-plugin to analysis-kotlin-api


IgnatBeresnev:
Dokka 1.9.0 introduces some [breaking changes](https://github.com/Kotlin/dokka/issues/3099) that affect the plugin that builds [Android's API reference](https://developer.android.com/reference) docs.

This plugin uses significantly more compiler internals than any other Dokka plugin, so we need to provide stable alternatives for the now-missing API, and help with migration where possible.

___

Definition of done: [dokka-devsite-plugin](https://android.googlesource.com/platform/tools/dokka-devsite-plugin) has migrated to `analysis-kotlin-api`, and no regression was noticed.

IgnatBeresnev:
API-wise, what needs to be done from Dokka's side:

* `ExternalDocumentablesProvider` - should be made public and stable. Used for resolving supertypes.
* `org.intellij.markdown.MarkdownElementTypes.MARKDOWN_FILE` - needs a public and stable substitution. The use case is the same as in Dokka: used as the default name for a doc tag, so it needs to be reflected in the name.
* `SampleTransformer` - the interface `SampleProvider` introduced in #3102 will cover the use cases. Should be made public and stable, and we should add more details to the documentation so that the body-processing guarantees (such as whitespace trimming) are not lost.

Not planned:
* `KotlinToJavaService` - not actually needed, can stay internal.
* Transitive utility functions (everything from `addToStdlib`,  `com.jetbrains.rd.util` and other packages) - no way for us to replace them.

IgnatBeresnev:
New request: there should be some API that helps determine the source language of a `Documentable`, to say that it's either coming from Kotlin code or Java code. Previously, the following check was possible: 

```kotlin
(documentable as? WithSources).sources.entries.singleOrNull.value is PsiDocumentableSource
```

We have an internal alternative in the form of `DocumentableSourceLanguageParser`, so it might be enough to "stabilize" it by providing a clear contract and making it public.

Notes on the implementation:
* Having an enum (like the existing `DocumentableLanguage`) might be undesirable because there's no room for extension in 3rd party plugins or analyses. Not sure if it's a capability worth providing though.
* This capability could be moved from an external interface/extension to be inside the `WithSources` interface, but `WithSources` is a bit misleading as it can exist even for declarations that do not have sources.

IgnatBeresnev:
Blocked by #3573

whyoleg:
[...]

PR #3312, opened 2023-10-31
Stabilize `ExternalDocumentableProvider`

IgnatBeresnev:
Part of #3112, depends on #3195

Creating this PR as a draft because I still need to write documentation for the new API, but the review can already start as I don't plan to introduce any significant code changes.


subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/documentable/ExternalDocumentableProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+package org.jetbrains.dokka.analysis.kotlin.documentable
+
+import org.jetbrains.dokka.DokkaConfiguration.DokkaSourceSet
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.DClasslike
+
+/**
+ * TODO [external-documentable-provider] add documentation
+ */
+public interface ExternalDocumentableProvider {
```
IgnatBeresnev, line 14: In a few words, this extension is used in dackka to create documentables for supertypes that come from dependencies, because, if I'm not mistaken, by default Dokka only creates the documentable model for the code it finds in the project itself, but not its classpath.

subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/ExternalDocumentableProviderTest.kt
```diff
@@ -0,0 +1,293 @@
...
+
+import org.jetbrains.dokka.analysis.test.api.javaTestProject
+import org.jetbrains.dokka.analysis.test.api.kotlinJvmTestProject
+import org.jetbrains.dokka.analysis.test.api.useServices
+import org.jetbrains.dokka.analysis.test.api.util.getResourceAbsolutePath
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.model.*
+import kotlin.test.*
+
+class ExternalDocumentableProviderTest {
```
vmishenev, line 15: Is this yet another set of tests for external documentable? 
What are the plans for `ExternalDocumentablesTest`? It also contains interesting corner cases.

vmishenev (commented):
> Creating this PR as a draft because I still need to write documentation for the new API, but the review can already start as I don't plan to introduce any significant code changes.

Ideal review is not only about code.

Is this just renaming `internal.ExternalDocumentablesProvider` into `ExternalDocumentableProvider`?

subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/documentable/ExternalDocumentableProvider.kt
whyoleg: BTW, what about typealiases? They are not `DClasslike` but could be supertypes, and so do we need to support them?

IgnatBeresnev: It could indeed be useful, but we'd have to think about the corner cases, write tests for it and make sure the implementations support it, and at this point it doesn't seem to be needed, at least not for #3112. If someone does come with such a request, we can certainly consider it

subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/ExternalDocumentableProviderTest.kt
IgnatBeresnev: Deleted the old tests, as the new tests should cover all sensible corner cases I could think of

IgnatBeresnev:
> Is this just renaming internal.ExternalDocumentablesProvider into ExternalDocumentableProvider?

Not only that, but adding comprehensible documentation and unit tests as well. The implementation didn't change though.

IgnatBeresnev:
Had to rebase the branch onto master and resolve some merge conflicts manually, so I squashed the commits and force pushed it.

It's ready for review now

whyoleg (approved):

vmishenev (approved):

subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/documentable/ExternalDocumentableProvider.kt
vmishenev: > BTW, what about typealiases? They are not DClasslike but could be supertypes, and so do we need to support them?

AFAIK, [dokka-devsite-plugin](https://android.googlesource.com/platform/tools/dokka-devsite-plugin) can use `TypeAliased` that has `inner`.

PR #3315, opened 2023-11-02
Add Gradle integration test for `wasm-js` and `wasm-wasi` targets

vmishenev:


integration-tests/gradle/projects/it-wasm-js-wasi-basic/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+        configureEach {
+            externalDocumentationLink {
+                url.set(URL("https://kotlinlang.org/api/kotlinx.coroutines/"))
+            }
+        }
+    }
+}
+
+// HACK: some dependencies (coroutines -wasm0 and atomicfu -wasm0) reference deleted *-dev libs
+configurations.all {
```
igoriakovlev, line 43: You do not need it becase new versions without deprecated dependencies were published.
Please check latest versions [here](https://youtrack.jetbrains.com/issue/KT-62871/K-Wasm-republish-libraries-for-wasm-js-with-1.9.20-RC2-or-newer-if-available).

integration-tests/gradle/projects/it-wasm-js-wasi-basic/gradle.properties
```diff
@@ -0,0 +1,5 @@
+#
+# Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+#
+
+dokka_it_kotlin_version=1.9.20-RC2
```
igoriakovlev, line 5: 1.9.20 was published. I suggest to use it.

igoriakovlev (approved):

whyoleg (approved):
LGTM
But, what do you think about replacing existing Wasm test instead of creating new one? Wasm is still experimental, and recently in 1.9.20 it started to using latest WasmGC opcodes, so now pre 1.9.20 built binaries will not work in recent browsers and nodejs. And so may be it doesn't make sense to have any tests for them, as eventually it will be deprecated and removed?

IgnatBeresnev (approved):
:+1: 

I think the decision to add it as a separate test project is good. The old test that checks that Dokka works with 1.9.10 Wasm doesn't cause any problems, and it will be easy to get rid of if it starts misbehaving or when it's time to drop 1.9.10 support

PR #3343, opened 2023-11-14
Remove modifier rendering if it empty

atyrin:
For testing K2 migration, I'm using the approach of comparing Dokka output with K1. There are a lot of places where the difference is not user-visible but just an empty block `<span class="token keyword"></span>` in signatures. This makes it challenging to analyze the results. The PR should help to omit such `span`s.

### DIFF
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
@@ -248,7 +248,7 @@ public class KotlinSignatureProvider(
                 p.modifier[sourceSet].takeIf { it !in ignoredModifiers }?.let {
                         if (it is JavaModifier.Empty) KotlinModifier.Open else it
                     }?.name?.let { keyword("$it ") }
-                p.modifiers()[sourceSet]?.toSignatureString()?.let { keyword(it) }
+                p.modifiers()[sourceSet]?.toSignatureString()?.takeIf { it.isNotEmpty() }?.let { keyword(it) }
                 if (p.isMutable()) keyword("var ") else keyword("val ")
                 list(p.generics, prefix = "<", suffix = "> ",
                     separatorStyles = mainStyles + TokenStyle.Punctuation,
@@ -303,7 +303,7 @@ public class KotlinSignatureProvider(
                     f.modifier[sourceSet]?.takeIf { it !in ignoredModifiers }?.let {
                         if (it is JavaModifier.Empty) KotlinModifier.Open else it
                     }?.name?.let { keyword("$it ") }
-                    f.modifiers()[sourceSet]?.toSignatureString()?.let { keyword(it) }
+                    f.modifiers()[sourceSet]?.toSignatureString()?.takeIf { it.isNotEmpty() }?.let { keyword(it) }
                     keyword("fun ")
                     list(
                         f.generics, prefix = "<", suffix = "> ",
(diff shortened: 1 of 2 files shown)


IgnatBeresnev (approved):
:+1: :+1:

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
```diff
@@ -248,7 +248,7 @@ public class KotlinSignatureProvider(
                 p.modifier[sourceSet].takeIf { it !in ignoredModifiers }?.let {
                         if (it is JavaModifier.Empty) KotlinModifier.Open else it
                     }?.name?.let { keyword("$it ") }
-                p.modifiers()[sourceSet]?.toSignatureString()?.let { keyword(it) }
+                p.modifiers()[sourceSet]?.toSignatureString()?.takeIf { it.isNotEmpty() }?.let { keyword(it) }
```
whyoleg, line 251: Maybe we can just return `null` from `toSignatureString` (used in 12 places with similar pattern) or even don't add any text in `PageContentBuilder.DocumentableContentBuilder#text` if the text is empty? So to fix all empty spans :)

atyrin: 🤷‍♂️ I've just added one more change in place where the `toSignatureString` is not used.

atyrin:
@IgnatBeresnev @vmishenev please look once again. I've changed one more place for a similar case and added a few tests. Also, there is a suggestion from Oleg. Maybe we can address it with another PR later.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
IgnatBeresnev: Ideally, it should indeed be fixed once and for all, with some generic tests that check we have no empty spans/text at all. 

However, this place and API is very rarely touched, there's a non-zero chance of new rendering bugs or corner cases appearing (especially if changes are made in `DocumentableContentBuilder#text`), and the empty span bugs are very minor and not noticeable by the users.

I'm afraid it might take way more time than what it's worth for now, so I'd focus on just removing obstacles for testing K2 and simplifying the diff.

IgnatBeresnev (approved):
I think it's good to go as it addresses one specific problem: simplifies the diffs for testing K2. 

It sucks to walk past the bad code you can easily improve, but I'd say fixing old technical debt falls outside the scope of this particular PR, as it wasn't even planned to begin with

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
whyoleg: I agree, better not to do it now. Thought, may be we should at least create issue to track this? Something like `Remove empty spans/tags in HTML`? (interesting, are other formatters affected?)

whyoleg (approved):

vmishenev (approved):

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
atyrin: I've created that one: https://github.com/Kotlin/dokka/issues/3344
Feel free to rewrite and add possible areas of improvement.

PR #3346, opened 2023-11-15
Do not find javadoc for non-JVM source set

vmishenev:


whyoleg (approved):
LGTM

dokka-subprojects/plugin-base/src/test/kotlin/model/MultiLanguageInheritanceTest.kt
```diff
@@ -26,6 +27,48 @@ class MultiLanguageInheritanceTest : BaseAbstractTest() {
...
+            |    */  
+            |    val withHintType: String
+            |}
+            |
+            |/src/main/kotlin/sample/Child.kt
+            |package sample
+            |public class Child : Parent {
+            |   override val withHintType: String 
+            |    get() = ""
+            |}
```
IgnatBeresnev, line 66: It looks like such a common case that I don't know how it didn't backfire before :fearful:

IgnatBeresnev (approved):
Nice :+1:

dokka-subprojects/plugin-base/src/test/kotlin/model/MultiLanguageInheritanceTest.kt
```diff
@@ -26,6 +27,48 @@ class MultiLanguageInheritanceTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    fun `should not try to find javadoc for non JVM source set`() {
```
whyoleg, line 31: BTW, may be can move this test to `analysis-kotlin-api`?

IgnatBeresnev: There's no API for multiplatform projects yet, unfortunately :( I didn't add it because in case something changed in the underlying framework (TestProject, TestFile, TestData, etc) it would take more time and effort to update all implementation, so for now it supports only JVM projects. If it proves to be convenient, we'll add multiplatform and move all tests from dokka-base into it

PR #3347, opened 2023-11-15
Support `inner` modifier for java non-`static` classes

### LINKED ISSUE #2793: Inner java class miss the `inner` modifier in signature
```Java
public class JavaClassWithNested {
    public class Inner{}
    public static class Nested{}
}
```

Will generate documentation like

image

The signature for class `Inner` should contain `inner` keyword.

whyoleg:
Fixes #2793

Note: almost all of combinations of parent/nested relations are `static` by default. Only nesting `class` inside `class` or `enum` could be `inner` in Kotlin.

<details>
<summary>P.S. Here is where IDEA renders redundant static modifiers: </summary>
image
image
image
image

</details>


vmishenev (approved):

IgnatBeresnev (approved):
Well done with thinking of corner cases and tests!

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/InnerModifierJavaAnalysisTest.kt
```diff
@@ -0,0 +1,293 @@
...
+        javaInterface.classlikes.forEach {
+            @Suppress("UNCHECKED_CAST")
+            assertTrue((it as WithExtraProperties<Documentable>).kotlinOnlyModifiers().isEmpty())
+        }
+    }
+
+    @Test
+    fun `java declarations nested inside annotation interface should not have kotlin inner modifier`() {
+        val testProject = javaTestProject {
+            javaFile(pathFromSrc = "JavaAnnotation.java") {
```
IgnatBeresnev, line 80: btw, for some reason I didn't expect Java to handle root packages correctly, but apparently it works :)

Issue #3196, opened 2023-10-06
[K2] Generic inherited members have own dedicated pages


vmishenev:
Generic inherited members should lead to pages of members from a superclass.

It happens due to [KT-62405]( https://youtrack.jetbrains.com/issue/KT-62405/Analysis-API-Symbols-SUBSTITUTIONOVERRIDE-have-no-overridden-symbols). 

Also, it leads to non-suppressed functions in enum: `compareTo`, `describeConstantable`, `getDeclaringClass`.

**Details**
KT-62405 results in an incorrect DRI that leads to creating a redundant page for generic inherited members.

After the fix, the following test should be added:
```kt
    @Test
    fun `generic inherited members should have correct dri`() {
        inlineModelTest(
            """
                |interface A<T> { fun x(): T }
                |interface B : A<Int> { }
                """
        ) {
            with((this / "function" / "B" / "x").cast<DFunction>()) {
                dri.classNames equals "A" 
                dri.packageName equals "function" 
                // callable ...
            }
        }
    }

```

__

User-facing: if B extends A and inherits a function, clicking on that function (inside B) leads to a separate page instead of to A (where it's declared).

atyrin:
One more side effect: There are visible "obvious functions" `compareTo`, `describeConstable` and `getDeclaringClass` on enum pages. It should be gone in the default configuration (`suppressObviousFunctions =true`) after the fix.

whyoleg:
In scope of #3349 there was added test which fails because of appearing of `getDeclaringClass`. Test is located in `ObviousAndInheritedFunctionsDocumentableFilterTest` in the end of file called `should not suppress toString, equals and hashcode of kotlin Enum`

PR #3349, opened 2023-11-15
Don't suppress obvious members in kotlin.Enum and kotlin.Any

### LINKED ISSUE #2863: Should not suppress obvious members declared in `Any`, `Enum` or `Object` themselves
If a class chooses to override obvious members, such as `equals()`, `hashCode()` or others, Dokka will NOT suppress it.

<details>
<summary>Example</summary>

Given the following code,

```kotlin
class TestFile {
    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (javaClass != other?.javaClass) return false
        return true
    }

    override fun hashCode(): Int {
        return javaClass.hashCode()
    }
}
```

Dokka generates:

</details>

However, when generating documentation for the standard library itself, `kotlin.Any` and `kotlin.Enum` don't have these member displayed, even though they are the declaring classes.

<details>
<summary>Screenshot</summary>

</details>

---

I think this is because we check for `containingDeclaration` to see if the obvious members were overridden, but in the case of `kotlin.Any` it returns itself, which leads to `obvious = true`. 

https://github.com/Kotlin/dokka/blob/35d15601f2d129a7d3db67dd9e2f4c41c87ef083/subprojects/analysis-kotlin-descriptors/compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt#L646-L649

A check needs to be added to account for documenting `kotlin.Any`, `java.lang.Object`, `kotlin.Enum` and `java.lang.Enum` themselves, in which no members should be obvious.

whyoleg:
Fixes #2863

This is draft because of 2 questions:
1. To allow Analysis API to analyse `kotlin` package, we need to provide flag `AnalysisFlags.allowKotlinPackage` - should we hide it under a flag at our side? And if so, what is the best way provide it to analysis? Or we can just set it `true` - Descriptor based analysis doesn't need any additional configuration
2. There is a test failing for `Enum`. During getting functions of class for `Enum` from test it fetches also methods from `Enum.java`: `clone`, `finalize`, `getDeclaringClass`. Somewhere later first 2 are filtered, but the last one is not, and so present in the documentation in the end. I was failed "to find why" and "what to do next" during debugging for 2 hours of FIR compiler internals... So asking for some help here :)

### DIFF
--- a/dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/TagsAnnotations.kt
+++ b/dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/TagsAnnotations.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.analysis.test
+
+import org.junit.jupiter.api.Tag
+
+// COPY OF dokka-subprojects/plugin-base/src/test/kotlin/utils/TagsAnnotations.kt
+
+/**
+ * Run a test only for descriptors, not symbols.
+ *
+ * In theory, these tests can be fixed
+ */
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY_GETTER,
+    AnnotationTarget.PROPERTY_SETTER
+)
+@Retention(
+    AnnotationRetention.RUNTIME
+)
+@Tag("onlyDescriptors")
+annotation class OnlyDescriptors(val reason: String = "")
+
+/**
+ * Run a test only for descriptors, not symbols.
+ *
+ * These tests cannot be fixed until Analysis API does not support MPP
+ */
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY_GETTER,
+    AnnotationTarget.PROPERTY_SETTER
+)
+@Retention(
+    AnnotationRetention.RUNTIME
+)
+@Tag("onlyDescriptorsMPP")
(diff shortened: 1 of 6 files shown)


whyoleg:
Behaviour is the same after rebase on master (with new Analysis API with fix for https://youtrack.jetbrains.com/issue/KT-63223/Analysis-API-reference-to-declarations-with-kotlin-package-are-not-resolved) - so it's unrelated

vmishenev:
> To allow Analysis API to analyse kotlin package, we need to provide flag AnalysisFlags.allowKotlinPackage - should we hide it under a flag at our side? And if so, what is the best way provide it to analysis? Or we can just set it true - Descriptor based analysis doesn't need any additional configuration

It is important only for stdlib.  `AnalysisFlags.allowKotlinPackage ` seems to be kind of `ignoreCommonBuiltIns`  in K1 (for more detail see https://github.com/JetBrains/kotlin/commit/3572a96bb94877b1b5ab11c91d4e06ad8fe40577).
But for now, Dokka K2 does not support stdlib and we have not planned it yet. I have filed https://github.com/Kotlin/dokka/issues/3354 to have an opportunity to refer to it. 

> There is a test failing for Enum. During getting functions of class for Enum from test it fetches also methods from Enum.java: clone, finalize, getDeclaringClass. Somewhere later first 2 are filtered, but the last one is not, and so present in the documentation in the end. I was failed "to find why" and "what to do next" during debugging for 2 hours of FIR compiler internals... So asking for some help here :)

#3196 https://github.com/Kotlin/dokka/issues/3196#issuecomment-1808061478


I think you can suppress these tests by the `OnlyDescriptors` annotation with the references to these issues.
It also would be nice to know about the tests in the issues.

whyoleg:
@vmishenev thanks! I've removed analysis flag, suppressed test and added comments to the issues
PR is ready for review @IgnatBeresnev / @vmishenev

vmishenev (approved):

IgnatBeresnev (approved):
👍 👍

PR #3358, opened 2023-11-20
Run analysis-kotlin-api tests with k2

IgnatBeresnev:
* Adds the ability to run analysis-kotlin-api tests with the K2 analysis


dokka-subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/configuration/TestDokkaConfigurationMapper.kt
```diff
@@ -109,6 +110,9 @@ private fun TestDokkaSourceSet.toDokkaSourceSet(relativeToDir: File): DokkaConfi
         override val analysisPlatform: Platform
             get() = analysisPlatform
 
+        override val displayName: String
+            get() = displayName
```
IgnatBeresnev, line 114: I would've added this ability as part of #3195, but after rebase #3236 started using `displayName` and the tests started failing, so I'm addressing it as a separate change for convenience

Issue #3141, opened 2023-08-24
[K2] Use only stable compiler API


IgnatBeresnev:
This issue is part of a larger issue: [Sustainability of dependencies (#3127)](https://github.com/Kotlin/dokka/issues/3127)

## Context

As described in #3127, the K1 compiler API that Dokka relies on is not stable, and internal compiler refactorings, new features and bugfixes sometimes (unwillingly and unknowingly) have a very negative impact on Dokka. We aren't the only ones with this problem, so one of the goals of the new K2 compiler is to introduce separate API-only artifacts that would be stable and not affected by internal compiler changes.

As of Kotlin 1.9.10, this stable compiler API is still in the works, so Dokka has to depend on [a bunch of artifacts](https://github.com/Kotlin/dokka/blob/83d6ad4c63d6674007d95e3744764580adf1b33a/subprojects/analysis-kotlin-symbols/build.gradle.kts#L20-L75) with questionable stability — some required for the needed API, some for runtime only.

## Goal

Dokka should, sooner or later, only rely on the stable compiler API, and all of the compiler runtime dependencies (== API implementations) should (ideally) be provided as a single or a number of artifacts with `runtimeOnly` scope.

This will require working with the team responsible for the compiler API to make sure that all of Dokka's use cases are covered. It will take time for that team to implement the API that Dokka needs, so this issue exists to keep track of progress.

IgnatBeresnev:
After a quick chat with @darthorimar, we agreed that the Dokka team will create a list of non-stable API that Dokka uses (in other words, the API we use from non-API / implementation artifacts). Then we can discuss it together and decide what to do with this API - either also extract / stabilize it, or come up with stable alternatives.

IgnatBeresnev:
Subtasks:

1. After implementing #3114, collect a list of non-stable API and discuss it with the compiler team.
2. After implementing #3137, we'll refine the list.

vmishenev:
Needed API for Dokka:
[KT-63339](https://youtrack.jetbrains.com/issue/KT-63339/Analysis-API-Provide-a-way-to-extract-KDoc-for-symbols)

vmishenev:
It is expected Analysis API's `sourcePsiSafe` should support Java sources as well. It depends on [KT-53669](https://youtrack.jetbrains.com/issue/KT-53669/Analysis-API-redesign-KtSymbolOrigin-to-distinguish-kotlin-java-source-library-declarations)
[...]

PR #3361, opened 2023-11-21
[K2] Fix sample resolving

vmishenev:


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolSampleAnalysisEnvironment.kt
```diff
@@ -78,13 +79,50 @@ private class SymbolSampleAnalysisEnvironment(
...
+        for (element in this) {
+            val result = transform(element)
+            if (result != null) {
+                return result
+            }
+        }
+        return null
+    }
+
+    // TODO: remove after KT-53669 and use [org.jetbrains.kotlin.analysis.api.symbols.sourcePsiSafe] from Analysis API
```
whyoleg, line 94: is there a Dokka issue to keep track of it?

whyoleg (approved):

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolSampleAnalysisEnvironment.kt
vmishenev: I have left [the comment](https://github.com/Kotlin/dokka/issues/3141#issuecomment-1820794232) in #3141

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolSampleAnalysisEnvironment.kt
```diff
@@ -78,13 +79,50 @@ private class SymbolSampleAnalysisEnvironment(
...
+            KtSymbolOrigin.JS_DYNAMIC -> null
+        }
+
+        return sourcePsi as? PSI
+    }
+
     private fun findPsiElement(sourceSet: DokkaSourceSet, fqLink: String): PsiElement? {
-        val ktSourceModule = samplesKotlinAnalysis.getModuleOrNull(sourceSet)
-            ?: projectKotlinAnalysis.getModule(sourceSet)
+        // fallback to default roots of the source set even if sample roots are assigned
```
IgnatBeresnev, line 117: good idea to add a comment! it explains the code below, but it doesn't explain the intention, which is the less obvious part :)

```suggestion
        // fallback to default roots of the source set even if sample roots are assigned,
        // because `@sample` tag can contain links to local project functions
```

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolSampleAnalysisEnvironment.kt
```diff
@@ -78,13 +79,50 @@ private class SymbolSampleAnalysisEnvironment(
...
+        val ktSourceModules =
+            listOfNotNull(samplesKotlinAnalysis.getModuleOrNull(sourceSet), projectKotlinAnalysis.getModule(sourceSet))
 
-        return analyze(ktSourceModule) {
-            resolveKDocTextLinkSymbol(fqLink)
-                ?.sourcePsiSafe()
+        return ktSourceModules.firstNotNullOfOrNull { ktSourceModule ->
```
IgnatBeresnev, line 121: Calling `listOfNotNull` to then do `list.firstNotNullOfOrNull` seems excessive in this case, and could be avoided like so

```kotlin
    private fun findPsiElement(sourceSet: DokkaSourceSet, fqLink: String): PsiElement? {
        // fallback to default roots of the source set even if sample roots are assigned,
        // because `@sample` tag can contain links to local project functions
        return samplesKotlinAnalysis.findPsiElement(sourceSet, fqLink)
            ?: projectKotlinAnalysis.findPsiElement(sourceSet, fqLink)
    }

    private fun KotlinAnalysis.findPsiElement(sourceSet: DokkaSourceSet, fqLink: String): PsiElement? {
        val ktSourceModule = this.getModuleOrNull(sourceSet) ?: return null
        return analyze(ktSourceModule) {
            resolveKDocTextLinkSymbol(fqLink)
                ?.kotlinAndJavaSourcePsiSafe()
        }
    }
```

IgnatBeresnev (approved):
Thanks for taking care of it! It thought it'd be more difficult for some reason

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolSampleAnalysisEnvironment.kt
vmishenev: Can you elaborate on what `local project functions` means? Why is it "local"?  I did not understand.

IgnatBeresnev: I understood the problem and the fix the following way:

If you have the code like

```kotlin
/src/kotlin/com/example/MyFile.kt

/**
 * @sample bar
 */
fun foo() {
    
}

fun bar() {}
```

where `@sample` references a function declared in the project sources itself, like literally the function below it, we need to use `projectKotlinAnalysis`. This is what I meant by a "local" function.

And if you store sample functions outside of the project sources, like in a separate directory somewhere, and you configure it using `DokkaSourceSet#samples`, then it needs to use `samplesKotlinAnalysis`. I'd consider this "external"

The problem was that it was always using `samplesKotlinAnalysis`, which pointed to external sample functions, but it needs to resolve "local" project functions as well

If I got the idea and the fix correctly, you can use whatever words you find more suitable, that was just a suggestion :) But I might've misunderstood it, in which case it definitely needs some sort of a comment :)

vmishenev: I have changed a little to avoid a "local" function.
```kt
// because `@sample` tag can contain links to functions from project sources
```

Issue #3041, opened 2023-06-09
Make Kotlin Playground for `@sample` rendering by default


IgnatBeresnev:
## Problem

Embedded [Kotlin Playground](https://play.kotlinlang.org/) scripts are used to render the code referenced by the `@sample` KDoc tag. It works wonderfully for Kotlin's own standard and popular libraries, such as stdlib and kotlinx.coroutines, because they are included as dependencies by default.

However, you cannot simply add external dependencies to Kotlin Playground scripts - to do that, you have to fork the playground repo, [add dependencies manually](https://github.com/JetBrains/kotlin-compiler-server#how-to-add-your-dependencies-to-kotlin-compiler-books), and then host the playground yourself.

This means that in most cases the user code cannot be run by the playground, because external types are used in the samples, and they cannot be resolved by default.

## Solution

Do not use Kotlin Playground to render `@sample` content by default. Instead, embed the code referenced in `@sample` as non-runnable code blocks. Support for Kotlin Playground runnable samples should be extracted into a separate plugin with its own settings. 

### Definition of Done

* [SamplesTransformer](https://github.com/Kotlin/dokka/blob/adfeed1b35b94ced80aba4e13dc926b2c389efb1/plugins/base/src/main/kotlin/transformers/pages/samples/SamplesTransformer.kt#L25) and relevant code should be extracted into a separate plugin.
* Users should be able to provide their own `kotlin-playground` JS script (now hardcoded)
* If the default script is used, it should be possible to specify a custom playground server URL (now hardcoded to `play.kotlinlang.org` in the JS script)
* This plugin should replace default non-runnable code blocks with interactive and runnable Kotlin Playground samples. 

___

Related: #1281

IgnatBeresnev:
As part of the overall improvement of handling samples, it would be nice to add a user-friendly log message if the `@sample` link is not resolved (API returns null), with the place where this link came from. 

That is, given the following code

```kotlin
package com.example

class MyClass {
    /*
     * @sample com.example.samples.sampleFunction
     */
    fun foo() {}
}
```

if `com.example.samples.sampleFunction` cannot be resolved, there should be a log message along the lines of "The sample link used in `com.example.MyClass.foo` could not be resolved. Please, make sure it points to ..."

aajtodd:
Any idea when this will be integrated? It would definitely be nice to disable this. All of the samples we plan to include are not usable in the playground due to the external dependencies factor but also due to authentication requirements.

[...]

PR #3364, opened 2023-11-22
Address review comments for Sample API

IgnatBeresnev:
Continuation of https://github.com/Kotlin/dokka/pull/3195#pullrequestreview-1740723020

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
@@ -37,7 +37,7 @@ internal inline fun DRI?.ifUnresolved(action: () -> Unit): DRI? = this ?: run {
  */
 internal fun KtAnalysisSession.resolveKDocTextLinkDRI(link: String, context: PsiElement? = null): DRI? {
     val kDocLink = createKDocLink(link, context)
-    return kDocLink?.let { resolveKDocLinkDRI(it) }
+    return kDocLink?.let { resolveKDocLinkToDRI(it) }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -37,7 +37,7 @@ internal inline fun DRI?.ifUnresolved(action: () -> Unit): DRI? = this ?: run {
  */
 internal fun KtAnalysisSession.resolveKDocTextLinkDRI(link: String, context: PsiElement? = null): DRI? {
```
vmishenev, line 38: `resolveKDocTextLinkToDRI`

vmishenev (approved):

IgnatBeresnev:
https://github.com/Kotlin/dokka/pull/3195#discussion_r1400365906:
> I missed in the first review the message does not have the location of sample link. It can be useful to find the link in the input code.

I'm not sure I can add it as the API function does not have a location per se, it can be called from anywhere with any fully qualified link.

I think this needs to be added to `DefaultSamplesTransformer` somewhere, which I didn't touch much. As I replied to Oleg, `DefaultSamplesTransformer` is going to be changed in #3041 anyway, so I added this comment into the issue, it'll be a better time to address it

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
IgnatBeresnev: I didn't rename this one because it was consistent with the function `resolveKDocTextLinkSymbol` below, but I guess I'll rename it to `resolveKDocTextLinkToSymbol` as well

Issue #3365, opened 2023-11-22
Core data model stabilization


IgnatBeresnev:
This issue serves as a parent issue for stabilizing the core data model in Dokka. It includes, but is not limited to:
* The `Documentable` model
* The `DocTag` and `TagWrapper` models
* `DokkaConfiguration` and `DokkaSourceSet` configuration classes
* Utility classes like `DokkaSourceSetID`, `DRI`, `KotlinModifier` and so on.

___

Some known problems that need to be addressed:
* All documentables are `data` classes, which makes maintaining backward compatibility difficult. They should likely be migrated to average classes, but then there will need to be a substitution to `.copy()`, which is widely used in transformers
* Due to documentables being data classes, their `equals` and `hashCode` compare all properties, which is undesirable in some cases. For instance, it means that the compiler descriptors and symbols are compared transitively, which can be a heavy operation and might backfire. See https://github.com/Kotlin/dokka/issues/2620 for more details
* There's barely any documentation.
* `sealed` classes should be re-considered. At the moment it means all classlikes have to be in the same file and custom types cannot be added in external plugins.
* `typealias SourceSetDependent<T> = Map<DokkaSourceSet, T>` - here, `DokkaSourceSet` is used as a key, which also leads to the `hashCode()` and `equals()` comparing all of its fields, including the heavy classpath. It needs to be reconsidered, maybe it should be a map, or maybe the key should be `DokkaSourceSetID`. See #3246 for more details
* Some information is exposed as documentable properties, while other as `extra`. For example, it's unclear why `IsVar`, `IsAlsoParameter`, `DefaultValue`, `Annotations` and others are extras, whereas they could be first-class properties like `DClass#modifier` and `DFunction#isExpectActual`.
* Interop with Java needs to be thought out, as at this moment it's difficult to represent some Java-specific things with the documentable model, so some information gets lost. Basically all things that differ between the languages: getters/setters, visibilities, etc. It should either have first-class support, or it should be separate from the core model, and at this moment it's neither.
* A lot of user-facing code depends on the implementations of `toString()`, for instance `DokkaSourceSetID#toString` is used directly in HTML files. `DRI` has a similar problem. 
* `Throws` tag wrapper has an inconvenient name, clashes with the annotation.
* `MARKDOWN_FILE` magic constant is used throughout the code without any explanation for why it was chosen, with some logic and tests depending on it
* (the list is to be updated)

___

Related: #2933, #2620, #3246

[...]

PR #3366, opened 2023-11-22
Expose MARKDOWN_FILE constant

IgnatBeresnev:
Part of #3112, since Dackka depends on it for creating a few custom tags.

The story with this constant is the following. There was a constant in the `org.jetbrains.markdown` library: `org.intellij.markdown.MarkdownElementTypes.MARKDOWN_FILE.name` whose value was `MARKDOWN_FILE`.

It was used as the `name` of `CustomDocTag`, and in a few other places.

Because Dokka used to expose PSI, kotlin-compiler and other internal libs as `api` dependencies, some downstream plugins (like Dackka) started using it as well, because they saw how Dokka used it and they just (allegedly) copied it.

When the analysis refactoring happened in #3034, all of these internal dependencies were no longer exposed to the user, so the code that was using `org.intellij.markdown.MarkdownElementTypes.MARKDOWN_FILE.name` couldn't be compiled anymore.

I looked at the use cases, but I couldn't figure out why it was needed in the first place. I mean, sure, it's used in `CustomDocTag` as the default `name`, so we should just rename it and call it `CUSTOM_DOC_TAG_DEFAULT_NAME`, right?

However, there are some dodgy checks like 

https://github.com/Kotlin/dokka/blob/5bb1c8a474e130221878ee2b618833fbed02ac65/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/pages/comments/DocTagToContentConverter.kt#L269

and

https://github.com/Kotlin/dokka/blob/5bb1c8a474e130221878ee2b618833fbed02ac65/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/JavaDocCommentParser.kt#L203

and it is mapped using this constant in 

https://github.com/Kotlin/dokka/blob/5bb1c8a474e130221878ee2b618833fbed02ac65/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/factories/DocTagsFromIElementFactory.kt#L61

So it makes me question whether `CustomDocTag` is used not only for custom tags, but also for something to do with markdown files? 

To not make any rushed decisions that will need to be revisited later on anyway, I decided to expose this constant as it is, with an honest KDoc. We'll need to figure out what it was actually used for, and come up with an alternative, most likely when we start stabilizing the core API and its data model. Created an issue to keep track of it: #3365


dokka-subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/internal/MarkdownApi.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * You can depend on it in your existing code, but please refrain from using it in the new code
+ * of your plugin unless absolutely necessary. If something does not work without using this constant,
+ * please report your use case to https://github.com/Kotlin/dokka/issues/3365, it will help a lot.
+ *
+ * This constant is not marked with [InternalDokkaApi] on purpose. Even though it is discouraged to use it,
+ * we understand that some existing code might depend on it, so once a replacement is provided,
+ * this constant should be deprecated with a message that will help users migrate to something stable.
+ */
+public const val MARKDOWN_ELEMENT_FILE_NAME: String = "MARKDOWN_FILE"
```
IgnatBeresnev, line 23: This is the only meaningful change in this PR. Everything else just changes `"MARKDOWN_FILE"` literals to use this constant, so that we at least know the places where it is used.

dokka-subprojects/analysis-java-psi/build.gradle.kts
```diff
@@ -8,6 +8,7 @@ plugins {
 
 dependencies {
     compileOnly(projects.dokkaSubprojects.dokkaCore)
+    compileOnly(projects.dokkaSubprojects.analysisKotlinApi)
```
whyoleg, line 11: Is it really a good idea to make java-psi and markdown analysis depend on kotlin-api?
For me it looks really strange...
May be it's better to move this declaration to `core`, in `internal` package? As we will anyway do something later with it?

vmishenev (commented):
To be honest, I don't understand why we can't figure out what it is actually used for in this PR. Instead of it, we expose the internal constant that plugins'll likely depend on. This constant's important as it's like a root name by default.  Anyway we will force Dakka and other users to migrate to the new constant again. 

We can try the new name (not `MARKDOWN_FILE`) here since Dakka should migrate to it anyway. We've already changed the package location. It's unnecessary to use the old name.
And we'll have a possibility to revisit it in https://github.com/Kotlin/dokka/issues/3365

> However, there are some dodgy checks like

I think the names of these checks stem from the name of constant. They are local.

> and it is mapped using this constant in

You can also see usages `MarkdownElementTypes.MARKDOWN_FILE` in the markdown artifact _(or ask it's team)_. This is just the name of the root element.

dokka-subprojects/analysis-kotlin-api/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/internal/MarkdownApi.kt
IgnatBeresnev: The last paragraph in the KDoc has the answer :)

>  * This constant is not marked with [InternalDokkaApi] on purpose. Even though it is discouraged to use it,
> * we understand that some existing code might depend on it, so once a replacement is provided,
> * this constant should be deprecated with a message that will help users migrate to something stable.

whyoleg:
Filled https://github.com/Kotlin/dokka/issues/3404 based on comment above

whyoleg (approved):

IgnatBeresnev:
I've rebased this PR onto master, and updated the KDoc to include the research results from above, along with a link to the issue: https://github.com/Kotlin/dokka/pull/3366/commits/d20a6dbd0e3317e8c5f9aa21e27e97e16bd15c29

I contemplated extracting this constant into `core`, naming it something like `ROOT_TAG_NAME` and describing the use case in more detail in KDocs, but in the end decided not to. 

With a normal-looking name and inside `core` it looks like proper stable API, and limits the use case to the name and the KDoc description. While we've done good research, I still feel like some key detail might missing due to the complexity of parsing, and for this reason it feels scary to improve this constant beyond necessary as part of this PR, as it might bring more unwanted questions / bugs / confusion / attention down the road. 

I feel like it's best to address this properly as part of #3404, and have this PR serve as an enabler for Dackka and be a tiny incremental improvement.

I've also moved it out of the `internal` package and removed the recommendation to not use it, because we now know for a fact that it is required in some cases, whereas before we weren't sure. So it felt weird to have it in the `internal` package if it's something that plugins can`t live without under certain rare conditions, but it doesn't seem to be used often enough to justify moving it to `core` - we'll be able to do it down the road with an easy `replaceWith` deprecation if it's needed.

PR #3374, opened 2023-11-24
Render member extensions correctly

### LINKED ISSUE #3187: No member extensions in the index
**Describe the bug**
The index page of `Duration.Companion` showing members:
image

The page of `Duration.Companion.days`:
image

As you can see, the member extension functions `days` are not displayed in the containing object members list.

**Expected behaviour**
Member extension functions are displayed in members list of the containing class/object.

whyoleg:
Fixes #3187 

**Context**
During building Classlike page, separate tabs for `members` and `members+extensions` are configured inside `divergentBlock`. Functions/Properties are divided in members and extensions by calling `Documentable.isExtension` (returns if function/property is extension or not based on existence of receiver).

**Problem**
Kotlin also has member extensions, which should be shown on `members` page, but `Documentable.isExtension` for them returns `true` because, you know, they are extensions.

To better illustrate the problem, here is a sample code:
```kotlin
interface Receiver {
    // 1. just ordinary member function
    // receiver = null
    fun memberFunction()

    // 2. member extension function
    // receiver = Int
    fun Int.intExtensionFunction()

    // 3. member extension function (yes it's valid Kotlin code)
    // receiver = Receiver
    fun Receiver.recieverExtensionFunction()

    companion object {
        // 4. extension function inside companion
        // receiver = Receiver
        fun Receiver.receiverCompanionExtensionFunction(): String = "bar"
    }
}

// 5. top-level extension function
// receiver = Receiver
fun Receiver.receiverTopLevelExtensionFunction(): String = "bar"
```

**Expected behaviour**
Functions 1, 2, 3 should be shown on `members` tab, and on `members+extensions` additionally shown 4 and 5

**Current behaviour**
Only function 1 is shown on `members` tab, and on `members+extensions` additionally shown 1, 2, 3, 4, 5.

**Code problem**
`DocumentableContentBuilder.divergentBlock` only accepts `List<Documentable>` (where `Documentable` could be function, property or type), and having only this we can't distinguish, is this function/property is `member extension in type Receiver` or it is `external extension with receiver of type Receiver`.
But we have this information upper in chain:
1. for Classlikes, member extensions are already present in functions/properties and extensions are coming from `CallableExtensions` extra
2. for packages, we can check `receiver != null` the same way as in `Documentable.isExtension`. We don't really render top-level declarations and top-level extensions differently in HTML or other formats, but we have tests for it :)

Another minor problem(inconvenience) that now processing types, functions and properties is done via single function (`DocumentableContentBuilder.divergentBlock`), and so it already has multiple checks, if we are processing different kinds of documentables and change behaviour accordingly

**Solution**
* Path extensions and members explicitly as separate lists and handle them correctly after it
* Split functions/properties and types logic, so that only related code for each kind of declarations is executed and code flow is easier to understand

**Questions**
* Do we really need `Documentable.isExtension` in core? (it was used only in `dokka-base`s `DefaultPageCreator.kt`)
* Is it ok to break `DefaultPageCreator` ABI? Now half of members there `protected open` and half `private`. In our codebase the only usage is in `kotlin-as-java` to override 1 function.
On GitHub there is also only one usage of it in [island-time](https://github.com/erikc5000/island-time/blob/fcb0e14d7b4e25e836e08ef6509296ed57031a45/tools/mkdocs-dokka-plugin/src/main/kotlin/MkdocsDocumentableToPageTranslator.kt#L69) - only `contentForModule` is overridden there (the only change is [headers of table](https://github.com/erikc5000/island-time/blob/fcb0e14d7b4e25e836e08ef6509296ed57031a45/tools/mkdocs-dokka-plugin/src/main/kotlin/MkdocsDocumentableToPageTranslator.kt#L92-L95) - looks like it is to fix rendering of packages in GHM format)

### DIFF
--- a/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableUtils.kt
+++ b/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableUtils.kt
@@ -24,4 +24,8 @@ public fun DTypeParameter.filter(filteredSet: Set<DokkaSourceSet>): DTypeParamet
         )
     }
 
+@Deprecated(
+    "Deprecated for removal",
+    ReplaceWith("this is Callable && this.receiver != null")
+)
 public fun Documentable.isExtension(): Boolean = this is Callable && receiver != null
--- a/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableUtils.kt
+++ b/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableUtils.kt
@@ -24,8 +24,4 @@ public fun DTypeParameter.filter(filteredSet: Set<DokkaSourceSet>): DTypeParamet
         )
     }
 
-@Deprecated(
-    "Deprecated for removal",
-    ReplaceWith("this is Callable && this.receiver != null")
-)
 public fun Documentable.isExtension(): Boolean = this is Callable && receiver != null
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
@@ -71,7 +71,8 @@ public open class DefaultPageCreator(
      * @see ActualTypealias
      */
     private fun List<Documentable>.filterOutActualTypeAlias(): List<Documentable> {
-        fun List<Documentable>.hasExpectClass(dri: DRI) = find { it is DClasslike && it.dri == dri && it.expectPresentInSet != null } != null
+        fun List<Documentable>.hasExpectClass(dri: DRI) =
+            find { it is DClasslike && it.dri == dri && it.expectPresentInSet != null } != null
         return this.filterNot { it is DTypeAlias && this.hasExpectClass(it.dri) }
     }
 
@@ -382,28 +383,42 @@ public open class DefaultPageCreator(
         extensions: List<Documentable> = emptyList()
     ): ContentGroup {
         val types = scopes.flatMap { it.classlikes } + scopes.filterIsInstance<DPackage>().flatMap { it.typealiases }
+        val (extensionProperties, extensionFunctions) = extensions.splitPropsAndFuns()
         return contentForScope(
-            @Suppress("UNCHECKED_CAST")
+            dri = @Suppress("UNCHECKED_CAST")
             (scopes as List<Documentable>).dri,
-            sourceSets,
-            types,
-            scopes.flatMap { it.functions },
-            scopes.flatMap { it.properties },
-            extensions
+            sourceSets = sourceSets,
+            types = types,
+            functions = scopes.flatMap { it.functions },
+            properties = scopes.flatMap { it.properties },
+            extensionFunctions = extensionFunctions,
+            extensionProperties = extensionProperties
         )
     }
 
     protected open fun contentForScope(
         s: WithScope,
         dri: DRI,
-        sourceSets: Set<DokkaSourceSet>,
-        extensions: List<Documentable> = emptyList()
+        sourceSets: Set<DokkaSourceSet>
     ): ContentGroup {
         val types = listOf(
             s.classlikes,
             (s as? DPackage)?.typealiases ?: emptyList()
         ).flatten()
-        return contentForScope(setOf(dri), sourceSets, types, s.functions, s.properties, extensions)
+
+        // can be used only for packages
+        val (extensionFunctions, functions) = s.functions.partition { it.receiver != null }
+        val (extensionProperties, properties) = s.properties.partition { it.receiver != null }
+
+        return contentForScope(
+            dri = setOf(dri),
+            sourceSets = sourceSets,
+            types = types,
+            functions = functions,
+            properties = properties,
+            extensionFunctions = extensionFunctions,
+            extensionProperties = extensionProperties
+        )
     }
 
     private fun contentForScope(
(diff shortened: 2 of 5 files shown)


vmishenev:
After the grooming:

I do not know if this function will be useful
```kt
fun <T> T.isTopLevelExtension() where T : Documentable, T : Callable = receiver != null && dri.classNames.isNullOrBlank()
```

whyoleg:
> I do not know if this function will be useful

For now, I deprecated function `isExtension` in `core` which was rather ambiguous in meaning and now isn't used anywhere in our code (and not sure, that we need to rely on it). If user will want this functionality, they will be able to implement it in one line

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -643,128 +671,180 @@ public open class DefaultPageCreator(
...
+        }
     }
 
-    protected open fun DocumentableContentBuilder.divergentBlock(
+    private class DivergentElementGroup(
+        val name: String?,
+        val kind: ContentKind,
+        val contentTypeOverride: BasicTabbedContentType?,
+        val elements: List<Documentable>
+    )
```
IgnatBeresnev, line 779: `contentTypeOverride` looks suspicious... I wonder if it can be improved in any way

I think I understand why it exists - to be able to set `EXTENSION_FUNCTION` tab type for some of the children inside the `FUNCTION` block. 

And it seems to be a very specific corner case, as I can't imagine a scenario in which someone would want to override the type of tabbed content inside `BasicTabbedContentType.TYPE` - it would probably lead to a bug or undesired behaviour. So having it in one place like before seems more solid, i.e if `contentType == FUNCTION && group.elements.all { it.isExtension() } EXTENSION_FUNCTION else contentType`.

So what looks suspicious to me is that we're trying to generalize and encourage a corner case that maybe shouldn't even exist or should be refactored. It's not a major problem given that the class is private, we can just document it and hopefully it'll do until the next refactoring/feature/bugfix, but it would be nice to reduce the complexity here a little, so that you don't have to think about what exactly `contentTypeOverride` overrides, how to use it and in which scenarios

Do you maybe have some thoughts on this, or maybe you've tried different approaches and something didn't work / you didn't like?

As a side note: I honestly don't even remember why the differentiation between `FUNCTION` and `EXTENSION_FUNCTION` was needed - I all for removing it altogether if it makes things easier and everything works the same way as before. I hope we're not going to have anything similar with other types, like `CLASS` and `CLASS_INNER` or something.

IgnatBeresnev (commented):
Well done with the tests! And it's nice to see that `typesBlock` was extracted from `divergentBlock`, I think it makes it much easier to understand what's going on in different tabs :+1:

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableUtils.kt
```diff
@@ -24,4 +24,8 @@ public fun DTypeParameter.filter(filteredSet: Set<DokkaSourceSet>): DTypeParamet
         )
     }
 
+@Deprecated(
```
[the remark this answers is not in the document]
whyoleg: Okay, I will revert changes here. My main concern is that the body of this function is rather straightforward code to write, but having such as an extension on `Documentable` feels strange and not obvious + we don't have a use-case for it and no-one asked for it. Not sure, that we need to make such functions our ABI.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -382,28 +383,42 @@ public open class DefaultPageCreator(
...
+            sourceSets = sourceSets,
+            types = types,
+            functions = scopes.flatMap { it.functions },
+            properties = scopes.flatMap { it.properties },
+            extensionFunctions = extensionFunctions,
+            extensionProperties = extensionProperties
         )
     }
 
     protected open fun contentForScope(
```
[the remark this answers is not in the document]
whyoleg: I will check what is possible here, thanks!

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -643,128 +671,180 @@ public open class DefaultPageCreator(
...
+            groups = groups
         )
     }
 
-    private fun DocumentableContentBuilder.propertiesBlock(
+    private fun DocumentableContentBuilder.declarationsBlock(
         name: String,
-        list: Collection<DProperty>
+        isFunctions: Boolean,
+        declarations: List<Documentable>,
```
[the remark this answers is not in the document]
whyoleg: > it may be better/easier (or not) to split extensions and regular callables inside this function

This is specifically the cause, why I've had to refactor `divergentBlock` and to path extensions externally, as we can't distinguish it here.

> A function with a very broad name declarationsBlock() with a param isFunctions: Boolean that does else -> Properties - it's asking for trouble

Yeah, I agree, though, it's hard to find a good name for a function which should work for member/extensions functions/properties and not other entities, while having possibility to customise there handling.
Let's check our variants:
1. pass `ContentKind` - this sounds even better than boolean flag for me, as in this case, in future, someone will be able to easily add one more condition to `when` and so one more time make this function a responsibility monster — which should have not, as handling of types (class likes) most likely will always be different comparing to properties and functions, which are rather similar in handling
2. Deducing the type from `declarations` on the other hand could be nice, but it can suffer from the same issue above.
3. Leave `isFunctions` boolean flag but rename function to something like `functionsOrPropertiesBlock` + for clarity introduce 2 functions near for just functions and properties

I'm more in favour of variant 3 to make behaviour more explicit, that it's applied only for functions and properties.

WDYT?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
whyoleg: So, I've tried some different ways on how to improve code here, but after several tries I decided that it's not worth it in the end. While the current logic is not ideal, I've added some comments to improve its understanding.

Just for future readers, the most promising idea (at least I thought), both to clean up the code and simplify understanding of the logic was to add `BasicTabbedContentType` attribute no to the tab itself but to every row - in this case we clearly mark different rows with different type and so in HTML renderer depending on tab we filter by specific declarations (this also allows to easily do any splits, like showing only properties, or only functions, or other combinations if needed).
But it causes a lot of changes in final HTML, like some tags becomes just empty groups and they are skipped or new attributes are added. Also, this is needed only for functions/properties and make no sense for types (may be it's possible to extract `Types` to separate function).
So in the end, it will cause a lot of refactoring which could give us not a lot of profit (and also, could be refactored in future one more time)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -380,73 +388,58 @@ public open class DefaultPageCreator(
...
-                "Properties", properties + extensionProps
-            )
-            functionsBlock("Functions", functions + extensionFuns)
+            propertiesBlock("Properties", properties, extensionProperties)
+            functionsBlock("Functions", functions, extensionFunctions)
         }
     }
 
-    private fun Iterable<DFunction>.sorted() =
-        sortedWith(compareBy({ it.name }, { it.parameters.size }, { it.dri.toString() }))
```
IgnatBeresnev, line 448: Hmm, did this sort get lost accidentally? I wonder if #1333 became relevant again, I don't see any tests in the PR that fixed it

IgnatBeresnev (approved):

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
whyoleg: Not accidentally but because of its redundancy. After https://github.com/Kotlin/dokka/pull/2846 sorting is now performed inside divergent group - specifically [here](https://github.com/Kotlin/dokka/blob/2bcd3da3fb2726fd56a316bb9d78ee4faa5b5b16/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt#L859C34-L859C34)

that's why no tests changed
but, I will recheck original issue #1333, just in case

vmishenev:
Approved from my side

vmishenev (approved):

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
whyoleg: Checked this case — even with shuffling of members, the order in documentation stays stable.

PR #3390, opened 2023-11-30
[K2] Run Gradle integration tests

### LINKED ISSUE #3239: [K2] Add  a check that source roots do not intersect
Different source sets should have disjoint source roots.
Otherwise, it causes
```
Modules are inconsistent
org.jetbrains.kotlin.utils.exceptions.KotlinIllegalStateExceptionWithAttachments: Modules are inconsistent
	at org.jetbrains.kotlin.analysis.low.level.api.fir.file.builder.LLFirFileBuilder$buildRawFirFileWithCaching$1.invoke(LLFirFileBuilder.kt:56)
```

Reproducer: add two source sets with the same source root.

**Note:** It can complicate the StdLib setting.

vmishenev:
Also, it fixes #3239 for integration tests

### DIFF
--- a/dokka-integration-tests/gradle/projects/it-basic-groovy/build.gradle
+++ b/dokka-integration-tests/gradle/projects/it-basic-groovy/build.gradle
@@ -13,6 +13,7 @@ dokkaHtml {
     outputDirectory = new File(buildDir, "/dokka/customHtml")
     failOnWarning = false
     dokkaSourceSets {
+        // create a new source set
         customSourceSet {
             sourceRoot(file("src/custom/java"))
             sourceRoot(file("src/custom/kotlin"))
--- a/dokka-integration-tests/gradle/projects/it-basic-groovy/build.gradle
+++ b/dokka-integration-tests/gradle/projects/it-basic-groovy/build.gradle
@@ -14,8 +14,8 @@ dokkaHtml {
     failOnWarning = false
     dokkaSourceSets {
         customSourceSet {
-            sourceRoot(file("src/main/java"))
-            sourceRoot(file("src/main/kotlin"))
+            sourceRoot(file("src/custom/java"))
+            sourceRoot(file("src/custom/kotlin"))
             displayName.set("custom")
             reportUndocumented.set(true)
         }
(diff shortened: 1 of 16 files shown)


dokka-integration-tests/gradle/projects/it-sequential-tasks-execution-stress/build.gradle.kts
```diff
@@ -27,9 +27,9 @@ fun createTask(name: String) {
         dokkaSourceSets {
             moduleName.set("Some example")
             register("kotlin-stdlib-common") {
-                sourceRoots.from("src/main/java")
-                sourceRoots.from("src/main/kotlin")
-                samples.from("src/main/kotlin")
+                sourceRoots.from("src/common/java")
+                sourceRoots.from("src/common/kotlin")
+                samples.from("src/common/kotlin")
```
IgnatBeresnev, line 32: Was it needed for something? It looks strange because this test applies `kotlin("jvm")`, so it's unclear why the source set should be named `common` (or is it not supposed to be the same as `commonMain`?)

dokka-integration-tests/gradle/projects/it-basic-groovy/build.gradle
```diff
@@ -14,8 +14,8 @@ dokkaHtml {
     failOnWarning = false
     dokkaSourceSets {
         customSourceSet {
-            sourceRoot(file("src/main/java"))
-            sourceRoot(file("src/main/kotlin"))
+            sourceRoot(file("src/custom/java"))
+            sourceRoot(file("src/custom/kotlin"))
```
IgnatBeresnev, line 18: Same question here about why this change was needed. `custom` is less confusing than `common`, but it's still unclear what will happen if I revert / change this :)

dokka-integration-tests/gradle/projects/it-sequential-tasks-execution-stress/build.gradle.kts
IgnatBeresnev: Oh, nevermind, the name of [the commit](https://github.com/Kotlin/dokka/pull/3390/commits/24b03240a80773ee3c24fb580f39814a57ecf425) that introduced it explains it

Adding a short comment here and in `it-basic-groovy` (in the similar place) might help a lot in the future, otherwise it's not very obvious that because the new source set is created, it cannot use the same source directory as the default source set, even though the default source set doesn't have any files in it

IgnatBeresnev (approved):

PR #3391, opened 2023-12-01
Support parsing javadoc `@param` for classes

### LINKED ISSUE #3199: `@param` on Java classes is ignored
## Background

The `@param` Javadoc tag can be used on classes to annotate type parameters. Example:

```java
/**
 * a class
 *
 * @param <Bar> a bar
 * @param <Baz> a baz
 */
public class Foo<Bar extends String, Baz extends String> implements java.util.Map<Bar, Baz> {
    ...
}
```

## Expected behaviour

The tag and value are not ignored and are both present in the documentable model and in the rendered documentation.

## Actual behaviour

Dokka seems to be ignoring this case altogether.

## Regression

Looks like it worked in Dokka 1.8.20, but it doesn't work in Dokka 1.9.0 anymore. Possibly, due to changes in #3034. 

<details>

<summary>Dokka 1.8.20</summary>

image

</details>

<details>

<summary>Dokka 1.9.0</summary>

image

</details>

## Notes

For type parameters, `@param <Bar>` seems to be the correct way to use it (surrounded with `<` and `>`), whereas `@param Bar` is highlighted in red by IDEA and it's ignored by the native javadoc format. Previously, both variants worked in Dokka 1.8.20, but it's only expected that the first one should work.
___

Related to: #3112

whyoleg:
Fixes #3199

### DIFF
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/JavaDocCommentParser.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/JavaDocCommentParser.kt
@@ -4,6 +4,7 @@
 
 package org.jetbrains.dokka.analysis.java.parsers
 
+import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiNamedElement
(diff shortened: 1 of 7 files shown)


dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/JavaDocCommentParser.kt
```diff
@@ -73,18 +74,38 @@ internal class JavaPsiDocCommentParser(
         analysedElement: PsiNamedElement
     ): TagWrapper? {
         val paramName = tag.dataElements.firstOrNull()?.text.orEmpty()
-
-        // can be a PsiClass if @param is referencing class generics, like here:
-        // https://github.com/biojava/biojava/blob/2417c230be36e4ba73c62bb3631b60f876265623/biojava-core/src/main/java/org/biojava/nbio/core/alignment/SimpleProfilePair.java#L43
-        // not supported at the moment
```
IgnatBeresnev, line 79: Shoot, I think it's one of my comments that I didn't turn into a TODO/issue, so I forgot to fix it before the merge :sweat_smile:

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavadocAnalysisTest.kt
```diff
@@ -0,0 +1,122 @@
...
+package org.jetbrains.dokka.analysis.test.jvm.java
+
+import org.jetbrains.dokka.analysis.test.api.javaTestProject
+import org.jetbrains.dokka.analysis.test.api.parse
+import org.jetbrains.dokka.model.SourceSetDependent
+import org.jetbrains.dokka.model.doc.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JavadocAnalysisTest {
```
IgnatBeresnev, line 14: Could you please add a single test for `@param Bar` on a class/function (not `<Bar>`, but `Bar`)? To verify that it either resolves `Bar` correctly, or it doesn't resolve it at all.

It's highlighted in red by IDEA, but Dokka 1.8.20 resolves it correctly in the resulting documentation. Whether it's right by specification and whether Dokka should support it is a different question, I'm not proposing to change the resolution logic now, I want the test to only "freeze" this behaviour so that we are notified in case it changes. And once/if it does change - we'll figure out what to do with it

IgnatBeresnev (approved):
:+1: :+1:

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavadocAnalysisTest.kt
```diff
@@ -0,0 +1,122 @@
...
+            function.documentation,
+            listOf(
+                Description(customDocTag("function")),
+                Param(customDocTag("parameter, long parameter description"), "bar"),
+            )
+        )
+    }
+
+    private fun customDocTag(text: String): CustomDocTag {
+        return CustomDocTag(listOf(P(listOf(Text(text)))), name = "MARKDOWN_FILE")
```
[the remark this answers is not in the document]
whyoleg: So, `MARKDOWN_FILE` is needed here just because it's already created in other code — here it's needed only to test that the content is the same.

But overall, I've tracked why this was introduced and will write some info into #3366

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/java/JavadocAnalysisTest.kt
whyoleg: Ok, I will add.
But, as far as I remember, during implementation, I found that both java and kotlin analysis for `param` tag doesn't validate if such parameter exists. So f.e., if you will have class like this (kotlin):
```kotlin
/**
 * @param non_existent documentation
 **/
fun something(param: String) {}
```
In the end, documentation for `non_existent` param will be present in generated HTML file. Same for Java.
Also, there will be no links for params - as they have nowhere to go - as parameters are already shown on the same page under this documentation

You can take a look here on what I mean: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html (parameters are underlined but have no links)

Issue #3392, opened 2023-12-01
[K2] Typealias to functional type with reciever leads to IndexOutOfBoundsException


atyrin:
The code
```Kotlin
typealias HttpExceptionCallback<T> = String.(T) -> String

fun <T> exception(
    callback: HttpExceptionCallback<T>,
){}
```
leads to a failure:
```
Caused by: java.lang.IndexOutOfBoundsException: fromIndex: 0, toIndex: -1
	at kotlin.collections.EmptyList.subList(Collections.kt:50)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider$funType$1.invoke(KotlinSignatureProvider.kt:486)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider$funType$1.invoke(KotlinSignatureProvider.kt:466)
	at org.jetbrains.dokka.base.translators.documentables.PageContentBuilder.contentFor(PageContentBuilder.kt:38)
	at org.jetbrains.dokka.base.translators.documentables.PageContentBuilder.contentFor$default(PageContentBuilder.kt:29)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider.funType(KotlinSignatureProvider.kt:466)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider.signatureForProjection(KotlinSignatureProvider.kt:417)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider.signatureForProjection$default(KotlinSignatureProvider.kt:405)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider$functionSignature$1$1$10.invoke(KotlinSignatureProvider.kt:331)
	at org.jetbrains.dokka.base.signatures.KotlinSignatureProvider$functionSignature$1$1$10.invoke(KotlinSignatureProvider.kt:326)
	at org.jetbrains.dokka.base.signatures.JvmSignatureUtils$parametersBlock$1$2.invoke(JvmSignatureUtils.kt:222)
	at org.jetbrains.dokka.base.signatures.JvmSignatureUtils$parametersBlock$1$2.invoke(JvmSignatureUtils.kt:221)
	at org.jetbrains.dokka.base.translators.documentables.PageContentBuilder.contentFor(PageContentBuilder.kt:50)
	at org.jetbrains.dokka.base.translators.documentables.PageContentBuilder$DocumentableContentBuilder.buildGroup(PageContentBuilder.kt:522)
[...]

PR #3396, opened 2023-12-04
[K2] Fix typealiased type that leads to functional type

vmishenev:


whyoleg (approved):

IgnatBeresnev (approved):
Could you please also do something with the unsafe list index access in `KotlinSignatureProvider#486`? It's from the exception in the reported [issue](https://github.com/Kotlin/dokka/issues/3392). I don't think it should fail the build even if something wasn't resolved properly by analysis.

It could generate nothing inside parentheses and emit a warning, it could write `Error class` in its place (if it should've been resolved, but there's nothing there), or maybe you'll come up with some other solution

vmishenev:
> Could you please also do something with the unsafe list index access in KotlinSignatureProvider#486? I

 I have added a simple check on it since we have no constraints on a number of arguments in a functional type.

PR #3401, opened 2023-12-07
Make integration tests debuggable

vmishenev:


whyoleg (approved):
LGTM, just in case, checked locally - works at my side

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -34,6 +34,7 @@ public abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest()
             .forwardOutput()
             .withJetBrainsCachedGradleVersion(buildVersions.gradleVersion)
             .withTestKitDir(File("build", "gradle-test-kit").absoluteFile)
+            .withDebug(!TestEnvironment.isDisabledDebug)
```
IgnatBeresnev, line 37: I'm of a strong opinion that debug mode should not be enabled by default. The first and simple reason is that it is not needed it in the overwhelming majority of cases (like 90%+ of the time), but the second and more complex reason - it might produce unwanted side-effects that can, for example, trigger false positives or make the code execution different to how it's executed in "production", let alone slow it down.

To expand on the second reason:

I don't know how the flag is implemented in Gradle and what it does under the hood, but if it starts the build in the same JVM instead of a separate process, we might experience false positives due to classpath conflicts in debug mode, and you will end up in a situation where you run the tests locally (debug **on** by default) and it fails, but you run the same test in TeamCity (debug **off** by default due to a special condition) and it doesn't fail, but the same test on GitHub Actions fails (debug **on** by default). Now go figure out what's wrong and try to find this flag :)

And if you run a JVM with `jdwp` (debug mode), it _can_ put restrictions on the types of optimizations the JVM can do (things like escape analysis/dead code elimination/inlining), and it _can_ have some overhead, because it needs to maintain debugging information and add various checks. And I've seen some "smart code" that works locally while its bytecode is being interpreted, but starts behaving differently once it's compiled and optimized heavily by C2 - trying to understand the reason was not fun.

Seeing that you disabled debugging for teamcity runs makes me think there might already be a problem :)

So I'd urge you to reconsider and change the `DISABLE_DEBUG` flag to `ENABLE_DEBUG` (so that it's enabled on demand and disabled by default), but either way it's not critical as the integration tests are going to be re-written anyway

IgnatBeresnev (approved):

PR #3405, opened 2023-12-11
fix package in KDoc for DokkaBuildProperties accessors

adam-enko:
Very minor fix for the KDoc of the util accessors


IgnatBeresnev (approved):
If only KDocs could check link resolution...

qwwdfsad:
>If only KDocs could check link resolution...

Nice idea for explicit API mode :)

PR #3409, opened 2023-12-12
Overcome fastutil issue

### LINKED ISSUE #3329: Dokka failing for email-like string in angle brackets
```kotlin
/**
 * <me@mail.com>
 */
class Kclass {
}
```

dokka generation fails with

```
Caused by: java.lang.NoSuchMethodError: it.unimi.dsi.fastutil.ints.IntArrayList.pop()Ljava/lang/Integer;
	at org.intellij.markdown.flavours.gfm.lexer._GFMLexer.advance(_GFMLexer.kt:456)
	at org.intellij.markdown.lexer.MarkdownLexer.advanceBase(MarkdownLexer.kt:60)
	at org.intellij.markdown.lexer.MarkdownLexer.calcNextType(MarkdownLexer.kt:55)
	at org.intellij.markdown.lexer.MarkdownLexer.locateToken(MarkdownLexer.kt:48)
	at org.intellij.markdown.lexer.MarkdownLexer.advance(MarkdownLexer.kt:38)
	at org.intellij.markdown.parser.sequentialparsers.LexerBasedTokensCache$Companion.cacheTokens(LexerBasedTokensCache.kt:51)
	at org.intellij.markdown.parser.sequentialparsers.LexerBasedTokensCache$Companion.access$cacheTokens(LexerBasedTokensCache.kt:25)
	at org.intellij.markdown.parser.sequentialparsers.LexerBasedTokensCache.<init>(LexerBasedTokensCache.kt:15)
	at org.intellij.markdown.parser.MarkdownParser.doParseInline(MarkdownParser.kt:80)
	at org.intellij.markdown.parser.MarkdownParser.parseInline(MarkdownParser.kt:38)
	at org.intellij.markdown.parser.MarkdownParser$InlineExpandingASTNodeBuilder.createLeafNodes(MarkdownParser.kt:110)
	at org.intellij.markdown.parser.TopLevelBuilder.createASTNodeOnClosingEvent(TopLevelBuilder.kt:21)
	at org.intellij.markdown.parser.TreeBuilder.buildTree(TreeBuilder.kt:40)
	at org.intellij.markdown.parser.MarkdownParser.doParse(MarkdownParser.kt:74)
	at org.intellij.markdown.parser.MarkdownParser.parse(MarkdownParser.kt:26)
	at org.intellij.markdown.parser.MarkdownParser.buildMarkdownTreeFromString(MarkdownParser.kt:21)
	at org.jetbrains.dokka.analysis.markdown.jb.MarkdownParser.parseStringToDocNode(MarkdownParser.kt:37)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.KdocMarkdownParserKt.parseFromKDocTag$parseStringToDocNode(KdocMarkdownParser.kt:27)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.KdocMarkdownParserKt.parseFromKDocTag(KdocMarkdownParser.kt:39)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.KdocMarkdownParserKt.parseFromKDocTag$default(KdocMarkdownParser.kt:17)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor.resolveDocumentation(DefaultDescriptorToDocumentableTranslator.kt:1056)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor.resolveDescriptorData(DefaultDescriptorToDocumentableTranslator.kt:926)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor.access$resolveDescriptorData(DefaultDescriptorToDocumentableTranslator.kt:152)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor$resolveClassDescriptionData$2.invokeSuspend(DefaultDescriptorToDocumentableTranslator.kt:961)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor$resolveClassDescriptionData$2.invoke(DefaultDescriptorToDocumentableTranslator.kt)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor$resolveClassDescriptionData$2.invoke(DefaultDescriptorToDocumentableTranslator.kt)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:78)
	at kotlinx.coroutines.CoroutineScopeKt.coroutineScope(CoroutineScope.kt:264)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor.resolveClassDescriptionData(DefaultDescriptorToDocumentableTranslator.kt:958)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor.classDescriptor(DefaultDescriptorToDocumentableTranslator.kt:413)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor.visitClassDescriptor(DefaultDescriptorToDocumentableTranslator.kt:201)
	at org.jetbrains.dokka.analysis.kotlin.descriptors.compiler.translator.DokkaDescriptorVisitor$visitClasslikes$2$invokeSuspend$$inlined$parallelMap$1$1.invokeSuspend(parallelCollectionOperations.kt:28)
```

**Installation**
- Dokka version: 1.9.20

whyoleg:
fixes #3329 

I've added description of the issue directly in code (same for both K1 and K2 modules). Will duplicate/expand on it here:
Related KT issue: https://youtrack.jetbrains.com/issue/KT-47150

What is the issue: `kotlin-compiler` artifact includes unshaded (not-relocated) STRIPPED `fastutil` dependency, STRIPPED here means, that it doesn't provide full `fastutil` classpath, but only a part of it which is used and so when shadowJar task is executed it takes classes from `fastutil` from `kotlin-compiler` and adds it to shadow-jar then adds all other classes from `fastutil` coming from `markdown-jb`, but because `fastutil` from `kotlin-compiler` is STRIPPED, some classes (like `IntStack`) lacks some methods and so such classes are not replaced afterward by `shadowJar` task - it visits every class once

Current solution (presented in PR) is to exclude `fastutil` from `shadowJar` and add runtime dependency on `fastutil` which will be resolved like if using `runtimeOnly`.
This is not ideal solution, as `theoretically` we could have issues with third-party plugins, that could provide `fastutil` on their side, but I think, that it should be more-or-less fine for now because:
1. analysis-kotlin-* dependencies will be only used for execution of dokka in a separate classpath, so there should be no issues for final users except they will for some reason use some old version of `fastutil`
2. analysis-markdown-jb (which is used by base and all-modules plugins) depends on markdown library which also depends on `fastutil` as `api`  - and so it will be available in classpath anyway

Other ways to fix the issue:
1. May be a better solution will be to include `fastutil` in shadow JAR - in this case we will need to create additional shadowJar task, which will first build jar without `fastutil` from `kotlin-compiler` and then build another shadowJar task which will depend on it + will include full `fastutil` (same as used in markdown lib). There is no now possibility in `shadow` plugin to select, from which dependency to get classes, when they differs (because it's really should not happen in real world).
2. Another possible way to fix the issue, is to use `kotlin-compiler-embeddable` which shoud correctly embed and shade dependencies (such as `fastutil`), but in this case, we should explicitly add other dependencies from `kotlin-compiler` which are relocated in `kotlin-compiler-embeddable` - then it will be needed to investigate which dependencies are really used (from my experiment I found, that at least some of the intelliij dependencies will be needed). Though, it gets more complicated with K2 analysis as on current moment there is no `kotlin-compiler-embeddable` artifact in declared repositories for version 2.0.0-dev-8561, even [kotlin/dev](https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev/) doesn't contain it - most likely analysis API and `kotlin/dev` are published independently.
3. Another possibility, when https://youtrack.jetbrains.com/issue/KT-47150 will be fixed we need to do nothing - as `fastutil` will be correctly relocated in `kotlin-compiler`

So, in the end, what is implemented in PR is fine, but if there are some concerns we can do what described above in point1 - other ways are not available at the moment.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -29,6 +29,7 @@ kotlinx-html = "0.9.1"
 
 ## Markdown
 jetbrains-markdown = "0.5.2"
+fastutil = "8.5.12"
 
 ## JSON
 jackson = "2.12.7" # jackson 2.13.X does not support kotlin language version 1.4, check before updating
@@ -112,6 +113,7 @@ korlibs-template = { module = "com.soywiz.korlibs.korte:korte-jvm", version.ref 
 
 #### Markdown ####
 jetbrains-markdown = { module = "org.jetbrains:markdown", version.ref = "jetbrains-markdown" }
+fastutil = { module = "it.unimi.dsi:fastutil", version.ref = "fastutil" }
 
 #### Jackson ####
 jackson-kotlin = { module = "com.fasterxml.jackson.module:jackson-module-kotlin", version.ref = "jackson" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -29,6 +29,10 @@ kotlinx-html = "0.9.1"
 
 ## Markdown
 jetbrains-markdown = "0.5.2"
+# This version SHOULD be compatible with both used in `koltin-compiler` and in `jetbrains-markdown`
+# version used in `kotlin-compiler` is located in https://github.com/JetBrains/kotlin/blob/c9ad09f4dcaae7792cdf27fbdc8c718443d0ad51/gradle/versions.properties#L14
+# verison used in `jetbrains-markdown` is located in https://github.com/JetBrains/markdown/blob/0f10d90f8f75a6c261e6f5a4f23d29c6cf088a92/build.gradle.kts#L92
+# In most cases using the latest version should work (be careful during MAJOR or MINOR version changes)
 fastutil = "8.5.12"
 
 ## JSON
--- a/dokka-integration-tests/gradle/projects/it-basic/src/main/kotlin/RootPackageClass.kt
+++ b/dokka-integration-tests/gradle/projects/it-basic/src/main/kotlin/RootPackageClass.kt
@@ -2,6 +2,8 @@
 
 /**
  * A class that lives inside the root package
+ *
+ * <me@mail.com>
  */
 class RootPackageClass {
     val description = "I do live in the root package!"
--- a/dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/documentable/SpecialCharacterTest.kt
+++ b/dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/documentable/SpecialCharacterTest.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.analysis.test.documentable
+
+import org.jetbrains.dokka.analysis.test.api.kotlinJvmTestProject
+import org.jetbrains.dokka.analysis.test.api.parse
+import org.jetbrains.dokka.model.doc.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class SpecialCharacterTest {
+
+    @Test
+    fun `should be able to parse email`() {
+        val project = kotlinJvmTestProject {
+            ktFile("Klass.kt") {
+                +"""
+                    /**
+                     * <me@mail.com>
+                     */
+                    class Klass
+                """
+            }
+        }
+
+        val module = project.parse()
+
+        val pkg = module.packages.single()
+        val cls = pkg.classlikes.single()
+
+        assertEquals("Klass", cls.name)
+
+        val text = P(
+            listOf(
+                Text("<", params = mapOf("content-type" to "html")),
+                Text("me@mail.com"),
+                Text(">", params = mapOf("content-type" to "html"))
+            )
+        )
+        val description = Description(CustomDocTag(listOf(text), name = "MARKDOWN_FILE"))
+
+        assertEquals(
+            DocumentationNode(listOf(description)),
+            cls.documentation.values.single(),
+        )
+    }
+}
--- a/dokka-subprojects/analysis-kotlin-descriptors/build.gradle.kts
+++ b/dokka-subprojects/analysis-kotlin-descriptors/build.gradle.kts
@@ -27,3 +27,25 @@ tasks.shadowJar {
     // from the dependencies are loaded, and not just a single one.
     mergeServiceFiles()
 }
+
+/**
+ * hack for shadow jar and fastutil because of kotlin-compiler
+ *
+ * KT issue: https://youtrack.jetbrains.com/issue/KT-47150
+ *
+ * what is happening here?
+ *   fastutil is removed from shadow-jar completely,
+ *   instead we declare a maven RUNTIME dependency on fastutil;
+ *   this dependency will be fetched by Gradle at build time (as any other dependency from maven-central)
+ *
+ * why do we need this?
+ *   because `kotlin-compiler` artifact includes unshaded (not-relocated) STRIPPED `fastutil` dependency,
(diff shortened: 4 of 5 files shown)


gradle/libs.versions.toml
```diff
@@ -29,6 +29,7 @@ kotlinx-html = "0.9.1"
 
 ## Markdown
 jetbrains-markdown = "0.5.2"
+fastutil = "8.5.12"
```
whyoleg, line 32: just in case, I haven't mentioned it, but this is the same version, which is used in `jetbrains-markdown` right now

IgnatBeresnev:
Thank you for such a detailed explanation, and especially for the in-code comments! Everything makes sense now, and I think this solution is indeed decent given the constraints

> analysis-markdown-jb (which is used by base and all-modules plugins)

oof, this doesn't look right; I must've forgotten about it when breaking analysis down. after #3366 it should be pretty easy to get rid of the dependency on `analysis-markdown-jb` in base and all-modules-page. I'll send a follow up PR, hopefully it'll make the chances of something blowing up smaller

> Another possible way to fix the issue, is to use kotlin-compiler-embeddable which shoud correctly embed and shade dependencies (such as fastutil)

I think I researched it a bit, and yeah, I definitely remember having problems with it. If I recall correctly, kotlin-compiler ships with its own relocated PSIs, but we actually need PSIs in Java analysis, especially when it comes to inheriting documentation, and so I think I couldn't pass PSIs from `analysis-kotlin` into `analysis-java` and vise versa, because they were different classes. Something like that

gradle/libs.versions.toml
IgnatBeresnev: Does this version need to be aligned with the version of `fastutil` from some other dependency, like the version used in `jetbrains-markdown` or in `kotlin-compiler`? 

If it needs to be aligned, I think there should definitely be a comment about it with a link, similar to how it's done for `intellij-platform` up above, so that we are not trigger-happy with updating it

IgnatBeresnev (approved):

gradle/libs.versions.toml
whyoleg: Overall, the version should be compatible with what is used in `kotlin-compiler` (AFAIK the same version is used in `IntelliJ` dependencies) and with `markdown-jb` library.
For now, `markdown-jb` uses newer version - so I would think we need to align with it.
I will write additional comment here

whyoleg:
> I'll send a follow up PR, hopefully it'll make the chances of something blowing up smaller

Note, that in `all-modules` it's used to parse `includes` files - so full blown `MarkdownParser` is initiated - not sure, that we will be able to easily remove there dependency on `analysis-markdown`

> kotlin-compiler ships with its own relocated PSIs, but we actually need PSIs in Java analysis

Hm, yes, that sounds legit. Then, I think, may be we need to somehow discuss with team developing `analysis-api` regarding this/similar things - so that at least with K2 stable API we will somehow be able to get rid of such `hacks`, it that's possible

IgnatBeresnev:
> Note, that in all-modules it's used to parse includes files - so full blown MarkdownParser is initiated - not sure, that we will be able to easily remove there dependency on analysis-markdown

That's why the `internal` package in `analysis-kotlin-api` exists :) Could just add a signature for the markdown parser

vmishenev (approved):

PR #3410, opened 2023-12-12
refactor: improved navigation controls style

### LINKED ISSUE #3300: Verify CSS styles and the icon of the homepage link button
#3235 added a homepage link with an icon to the HTML format, as per #2948.

The current icon is a royalty-free placeholder, so we need to see if there's a preferred icon-pack within JetBrains / Kotlin, and change the icon if needed.

The styles also need to be verified, as they were mostly made to just work. There might be issues with scaling and/or the mobile layout, but we lack the expertise to be certain.

berezinant:
fixes #3300 

* replaced icon
* simplified css
* fixed outline for navigation controls for better keyboard navigation

Demo:
Monosnap HelloKotlin 2023-12-12 23-03-32

Also some html to test the result
[test-page.zip](https://github.com/Kotlin/dokka/files/13653673/test-page.zip)


dokka-subprojects/plugin-base/src/main/resources/dokka/templates/includes/header.ftl
```diff
@@ -22,10 +22,10 @@
     </div>
     <div class="navigation-controls">
         <#if homepageLink?has_content>
-            <div class="navigation-controls--btn navigation-controls--homepage" id="homepage-link" role="button"><a href="${homepageLink}"></a></div>
+            <a class="navigation-controls--button navigation-controls--button_homepage" id="homepage-link" href="${homepageLink}" ></a>
```
berezinant, line 25: @IgnatBeresnev I've just found out about header/footer html templates being used in user's projects and about corresponding problems caused by changes in html. So how bad is this change and should I revert it maybe for this time and butch all such changes to a special release?

berezinant:
Fixed style for custom home page icon, e.g. github
image

dokka-subprojects/plugin-base/src/main/resources/dokka/templates/includes/header.ftl
IgnatBeresnev: Good thought! The homepage button specifically hasn't been released yet, so it's perfectly fine and expected to change the styles related to it, see https://github.com/Kotlin/dokka/pull/3235/files#diff-316e3e9b3d50ff73dcc848844e5eaac72b2d38b7f1e84dca8a383a8b2c5a162c

As for other "unnecessary" / refactoring changes, I think it makes sense to minimize them in this release (1.9.20). We can break it down into two PRs: the first one fixes the issue in the most compatible way, we'll merge it and then have the code freeze at the end of the week. The second one introduces some additional changes from this PR, we'll merge it after code freeze

berezinant: Sounds good, I've rolled back unnecessary changes to the template

IgnatBeresnev (approved):

zoobestik (approved):

Issue #3414, opened 2023-12-15
Accessibility violations


lauzadis:
**Is your feature request related to a problem? Please describe**
We were recently made aware that the documentation generated by Dokka violates various accessibility standards.

1. Ensure all interactive functionality is operable with the keyboard: [W3C's SC 2.1.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html)
    - Arrow icons to expand/collapse side menu sections
    - "X" button to clear the search field
2. Ensure custom controls provide proper textual name, role, and state information: [W3C's SC 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
    - X button to clear the search field
    - Arrow icons to expand/collapse side menu sections lack name, role, state information
3. Provide a mechanism for skipping past repetitive content [W3C's SC 2.4.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html)
    - Side menu items can not be skipped. Our package has over 300 modules and navigating with the Tab key means touching each of these items before reaching the main content.
4. Ensure keyboard focus is indicated visually [W3C's SC 2.4.7](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html)
    - "Types" and "Functions" filter buttons don't have an outline.
    - **I will submit a PR for this**
6. Ensure pages reflow without requiring two-dimensional scrolling without loss of content or functionality [W3C's SC 1.4.10](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) 
    - Links and code blocks are not properly wrapped, leading to two-dimensional scrollbar appearing on small screens. Reflow style can be added to make this wrap properly.
    - **I will submit a PR for this**

**Describe the solution you'd like**
It would be great if these violations could be addressed, it's especially useful for screen reader users.

**Describe alternatives you've considered**
None

**Additional context**
AWS SDK for Kotlin has fixed a variety of these accessibility issues by customizing Dokka's HTML / CSS files, but it would be good to get this upstreamed. I will be submitting a number of PRs which address these violations, you can take them if you like, or come up with your own solutions.

Here is a link to our docs where you can see the various violations listed: https://sdk.amazonaws.com/kotlin/api/1.0.17/index.html

**Are you willing to provide a PR?**
For some of them, yes. Some are more difficult to solve (such as skipping past repetitive content) and would require design discussion.

[...]

PR #3415, opened 2023-12-15
Accessibility: Wrap links and code blocks

lauzadis:
Wrap links and code blocks to prevent two-dimensional scrollbar from appearing on small screens.

Solves violation number 5 in this issue: https://github.com/Kotlin/dokka/issues/3414

### DIFF
--- a/dokka-subprojects/plugin-base/src/main/resources/dokka/styles/style.css
+++ b/dokka-subprojects/plugin-base/src/main/resources/dokka/styles/style.css
@@ -802,6 +802,7 @@ code:not(.block) {
     top: 8px;
     right: 8px;
     overflow-wrap: break-word;
+    word-break: break-word;
 }
 
 .sideMenuPart > .overview {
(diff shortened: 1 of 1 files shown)


dokka-subprojects/plugin-base/src/main/resources/dokka/styles/style.css
```diff
@@ -1023,6 +1024,7 @@ h1.cover {
     font-size: inherit;
     line-height: inherit;
     transition: color .1s, border-color .1s;
+    word-wrap: break-word;
```
berezinant, line 1027: nit: probably it'd be more consistent to rename this to `overflow-wrap` since it is a more common synonym

berezinant (commented):
Thanks, @lauzadis, I've tested it and this definitely makes things better!

However there are some other cases when two dimensional scroll still preserves
image

Would you prefer to fix such cases in this PR?

e.g. adding something like `word-break: break-word` could lead to such picture
image

/cc @IgnatBeresnev

berezinant (approved):

PR #3416, opened 2023-12-15
Accessibility: add outline when section-tab is focused

lauzadis:
Adds an outline to section-tabs when they are focused. The color matches the underline color when they are hovered over. 

Solves violation number 4 in this issue: https://github.com/Kotlin/dokka/issues/3414

### DIFF
--- a/dokka-subprojects/plugin-base/src/main/resources/dokka/styles/style.css
+++ b/dokka-subprojects/plugin-base/src/main/resources/dokka/styles/style.css
@@ -580,7 +580,7 @@ td:first-child {
     border-bottom: 2px solid var(--active-tab-border-color);
 }
 
-.section-tab:focus, .section-tab:active {
+.section-tab:focus-visible {
     outline: 2px solid var(--active-tab-border-color);
 }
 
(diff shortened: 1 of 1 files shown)


berezinant (commented):
Thanks @lauzadis, this improvement is useful 👍 

I've only noticed that there are two types of tabs so could you please add the fix for `platform-bookmark` as well?

image

dokka-subprojects/plugin-base/src/main/resources/dokka/styles/style.css
```diff
@@ -580,6 +580,10 @@ td:first-child {
     border-bottom: 2px solid var(--active-tab-border-color);
 }
 
+.section-tab:focus, .section-tab:active {
```
berezinant, line 583: `focus-visible` would be a slightly better option here because it only cause outline when navigation via keyboard and not when clicking which leads to a slightly clear UI

```suggestion
.section-tab:focus-visible {
```

lauzadis:
I took your suggestion about `focus-visible` selector and expanded the fix to cover `platform-bookmark`. Thanks!

berezinant (approved):

PR #3420, opened 2023-12-18
KT-64377 - Update Dokka build properties

adam-enko:
Updates a couple of integration test properties to use the DokkaBuildProperties class. 

#### Depends on 

* ~#3419~
* #3418

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
@@ -48,6 +48,20 @@ abstract class DokkaBuildProperties @Inject constructor(
     val kotlinLanguageLevel: Provider<KotlinVersion> =
         dokkaProperty("kotlinLanguageLevel", KotlinVersion::fromVersion)
 
+    /** Control [org.gradle.api.tasks.testing.Test.maxParallelForks] in integration tests. */
+    val integrationTestParallelism: Provider<Int> =
+        dokkaProperty("integration_test.parallelism", String::toInt)
+
+    /** Not currently used - should be dropped. */
+    val integrationTestExhaustive: Provider<Boolean> =
+        dokkaProperty("integration_test.exhaustive", String::toBoolean)
+            .orElse(false)
+
+    /** Control whether integration tests should use the `org.jetbrains.dokka.experimental.tryK2` flag. */
+    val integrationTestUseK2: Provider<Boolean> =
+        dokkaProperty("integration_test.useK2", String::toBoolean)
+            .orElse(false)
+
 
     private fun <T : Any> dokkaProperty(name: String, convert: (String) -> T) =
         providers.gradleProperty("org.jetbrains.dokka.$name").map(convert)
--- a/dokka-integration-tests/gradle/build.gradle.kts
+++ b/dokka-integration-tests/gradle/build.gradle.kts
@@ -2,7 +2,6 @@
  * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-
 plugins {
     id("dokkabuild.test-integration")
 }
@@ -22,10 +21,6 @@ dependencies {
 val aggregatingProject = gradle.includedBuild("dokka")
 
 tasks.integrationTest {
-    // pass the property to a test fork
-    project.findProperty("org.jetbrains.dokka.experimental.tryK2")
-        ?.let { systemProperty("org.jetbrains.dokka.experimental.tryK2", it) }
-
     dependsOn(aggregatingProject.task(":publishToMavenLocal"))
 
     environment("DOKKA_VERSION", project.version)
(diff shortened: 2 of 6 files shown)


dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -22,10 +22,6 @@ dependencies {
 val aggregatingProject = gradle.includedBuild("dokka")
 
 tasks.integrationTest {
-    // pass the property to a test fork
-    project.findProperty("org.jetbrains.dokka.experimental.tryK2")
-        ?.let { systemProperty("org.jetbrains.dokka.experimental.tryK2", it) }
-
```
adam-enko, line 28: I moved this configuration to the integration test convention plugin. The Maven/CLI integration tests might not use it, but to be it made sense to move as much config as possible into convention plugins to keep the project build scripts smaller.

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
```diff
@@ -48,11 +49,46 @@ abstract class DokkaBuildProperties @Inject constructor(
...
+                        it.toBoolean()
+                    }
+            )
+            .orElse(
+                providers.environmentVariable("DOKKA_INTEGRATION_TEST_IS_EXHAUSTIVE")
+                    .map {
+                        logger.warn("DOKKA_INTEGRATION_TEST_IS_EXHAUSTIVE is deprecated - use ORG_GRADLE_PROJECT_org.jetbrains.dokka.integration_test.exhaustive instead")
+                        it.toBoolean()
+                    }
+            )
```
IgnatBeresnev, line 79: JFYI this particular property (`dokka_integration_test_is_exhaustive)` can actually be dropped altogether. The versions of Kotlin/Gradle/etc that are run under it are not compatible (and haven't been for 2 years), so the tests fail under this flag. And I don't know if it makes sense to invest time into making it work now, so I'd just drop it and return to it when/if it's needed

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
```diff
@@ -48,11 +49,46 @@ abstract class DokkaBuildProperties @Inject constructor(
...
+    val integrationTestParallelism: Provider<Int> =
+        dokkaProperty("integration_test.parallelism", String::toInt)
+            // fallback & warn if the old property is used
+            .orElse(
+                providers.gradleProperty("dokka_integration_test_parallelism")
+                    .map {
+                        logger.warn("dokka_integration_test_parallelism is deprecated - use org.jetbrains.dokka.integration_test.parallelism instead")
+                        it.toInt()
+                    }
+            )
```
IgnatBeresnev, line 61: A nice thing to have, although I don't think it's used much outside of our CI (GitHub/TC) and especially by external people. 

So if the use of this property is corrected in CI (you now have access to change TC configurations), and you can notify Dokka's maintainers in Slack that this property has been changed, I think there's no need to keep the compatibility like that. This idea applies to other internal things too :)

IgnatBeresnev (approved):

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
```diff
@@ -48,11 +49,46 @@ abstract class DokkaBuildProperties @Inject constructor(
...
+            .orElse(
+                providers.environmentVariable("DOKKA_INTEGRATION_TEST_IS_EXHAUSTIVE")
+                    .map {
+                        logger.warn("DOKKA_INTEGRATION_TEST_IS_EXHAUSTIVE is deprecated - use ORG_GRADLE_PROJECT_org.jetbrains.dokka.integration_test.exhaustive instead")
+                        it.toBoolean()
+                    }
+            )
+
+    val tryK2: Provider<Boolean> =
```
whyoleg, line 81: BTW: while we are here, if this is used only in integration tests, may be it's better to rename it to `integration_test.useK2` instead of using the same flag which is used to run `dokka` with K2 - now it feels a bit confusing IMO

whyoleg (approved):

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
adam-enko: Can that be done in a later PR? :) I just want to focus on updating the Gradle config in this chain of PRs, and the `isExhaustive` variable is still used in code.

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
adam-enko: Can do! I was concerned about forgetting to update these properties in whatever miscellaneous places (which could also be on on developers' machines too) after this chain of PRs is completed, but it's not such a big problem.

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
adam-enko: Good idea. Just to be clear the full path will be `org.jetbrains.dokka.integration_test.useK2`

whyoleg: Note: TC config should be updated to use this new property - after merge to master only

adam-enko:
I've rebased this branch to remove #3419

PR #3421, opened 2023-12-19
Display Java's `Deprecated` message

### LINKED ISSUE #2668: Move text under @depracation tag to Deprecation block in Java sources
For the Java code like:

```Java
public class DeprecatedJClass {
    /**
     * Don't do anything
     * @deprecated
     * This method is no longer acceptable to compute time between versions.
     * <p> Use {@link DeprecatedJClass#getStringMethodNew()} instead.
     */
    @Deprecated(since = "18.0.2", forRemoval = true)
    public void getStringMethod(){}
    public void getStringMethodNew(){}
}
```

The HTML output look like:
image

It seems like it would be nice to move the text from the Depreciation paragraph to the Deracation table to have everything in one place.

IgnatBeresnev:
Fixes #2668

___

This PR is a follow-up from our pair-programming onboarding session with @berezinant. We've made the fix together, but Anton will re-implement it on his own


IgnatBeresnev:
Pushed the fix while I still have it

whyoleg (approved):

Issue #2925, opened 2023-03-16
Integration tests refactoring


IgnatBeresnev:
Dokka's [integration tests](https://github.com/Kotlin/dokka/tree/12e2a3c6b8bf1533148ddf29c77e73f5045754a6/integration-tests) run Dokka on real projects using various versions of Gradle, Kotlin and Android, and make sure there are no abnormalities, all links are present, everything is resolved and is able to be built, etc. This is extremely useful, and the same behaviour should be preserved.

Additionally, the integration tests are [used](https://github.com/Kotlin/dokka/blob/12e2a3c6b8bf1533148ddf29c77e73f5045754a6/.github/workflows/s3-snapshots.yml#L26) for generating and publishing live demos of documentation to S3, which is very useful for reviewing UI/UX changes. This feature should be preserved as well.

___

However, the way that the tests are set up now is not optimal, and it's causing problems in various places:

1. The tests cannot be cached by Gradle as they define incorrect inputs/outputs. [More information here](https://github.com/Kotlin/dokka/blob/3ea414251d32108dab8b14d41a672e9834824690/build-logic/src/main/kotlin/org/jetbrains/conventions/dokka-integration-test.gradle.kts#L23).
2. The tests are lumped together under the same task, so it's more difficult to have unique configuration for a project ([more information here](https://github.com/Kotlin/dokka/pull/2918#discussion_r1138607721)).
3. External projects, such as `kotlinx.coroutines`, are imported as git submodules. This is just annoying to maintain and update, and not obvious to someone encountering it for the first time. Maybe these projects could be added as dependencies, [more information here](https://github.com/Kotlin/dokka/pull/2884#discussion_r1138575188)
4. (to be continued)

The outlined problems should be researched and addressed if possible.

aSemy:
There was a discussion on Gradle Slack this week and https://github.com/gradle/exemplar was recommended for running integration tests. Although I find it pretty tough to understand how to use it, I think it would be worth investigating.

IgnatBeresnev:
Grooming notes

## Introduction

We have some integration tests already, but they've proven to be somewhat difficult to maintain, to extend, and they have some problems in their core that are getting in the way of other related things. These tests were a good start, but since their inception we've figured out additional use cases, figured out what works and what doesn't, and what's useful or would be nice to have in general.

[...]

PR #3427, opened 2023-12-20
KT-64377 use Java Test Suites for Gradle integration tests

adam-enko:
Restructure Gradle Integration tests to use Java Test Suites

- create a test suite for each project under test (a separate test suite helps with Gradle caching, task avoidance, and parallelising)
- update GitHub Action steps to run specific test-suite tasks
- disable caching integration test tasks due to dependency on Maven Local publishing (this will be fixed in #3433)

Contains:

* #3418 
* ~#3419~ 
* #3420 
* #3428


dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,41 +1,237 @@
...
  */
-
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+import org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED
+import org.gradle.api.tasks.testing.logging.TestLogEvent.SKIPPED
+import org.gradle.language.base.plugins.LifecycleBasePlugin.VERIFICATION_GROUP
+import org.jetbrains.kotlin.gradle.dsl.ExplicitApiMode.Disabled
 
 plugins {
-    id("dokkabuild.test-integration")
```
adam-enko, line 7: I've removed this convention for now because there's a split between the Maven/CLI/Gradle subprojects. 

I intend to refactor the Maven & CLI projects to use a similar format to this project. Once that's done it will be easier to extract common configuration into a new convention plugin.

For now I've copied the relevant configuration from the `dokkabuild.test-integration` convention into the Gradle subproject build config.

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,41 +1,237 @@
...
+        dokkaBuild.isCI.map { isCi -> if (isCi) "ALWAYS" else "ON_SUCCESS" }.get(),
+    )
+
+    testLogging {
+        exceptionFormat = FULL
+        events(SKIPPED, FAILED)
+        showExceptions = true
+        showCauses = true
+        showStackTraces = true
+    }
```
adam-enko, line 78: Copied from the integration-test convention plugin - see https://github.com/Kotlin/dokka/pull/3427/files#r1432756930 for reasoning

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,41 +1,237 @@
...
+
+    doNotTrackState("uses artifacts from Maven Local")
```
adam-enko, line 80: This will be fixed in an upcoming PR.

dokka-integration-tests/settings.gradle.kts
```diff
@@ -10,8 +10,8 @@ pluginManagement {
     includeBuild("../build-logic")
 
     repositories {
-        gradlePluginPortal()
         mavenCentral()
+        gradlePluginPortal()
```
adam-enko, line 14: Minor unrelated change. Maven Central is generally more stable than the Gradle Plugin portal, so prioritising Maven Central helps with build stability.

adam-enko:
I don't know why `Publish preview to web (S3) / kotlinx-serialization` is failing, the error seems unrelated.

```
    * Where:
    Script '/tmp/junit4521301651836114530/project/gradle/native-targets.gradle' line: 102

    * What went wrong:
    A problem occurred evaluating script.
    > No signature of method: static java.util.Arrays.compare() is applicable for argument types: ([Ljava.lang.String;, [Ljava.lang.String;, native_targets_1w0e7ne0z5gs2bcbvygztfbme$_run_closure1$_closure5$_closure8) values: [[1, 9, 0], [1, 9, 20], native_targets_1w0e7ne0z5gs2bcbvygztfbme$_run_closure1$_closure5$_closure8@64766701]
```


I guess this is the code that's failing https://github.com/Kotlin/kotlinx.serialization/blob/8baa04b4259274305eea6b02f5c5590b44ea126c/gradle/native-targets.gradle#L102


Is the test important?

build.gradle.kts
```diff
@@ -6,6 +6,10 @@ plugins {
     id("dokkabuild.base")
 }
 
+tasks.check {
```
whyoleg, line 9: AFAIK we already have this dependency below: `addDependencyOnSameTasksOfIncludedBuilds`

whyoleg: Also, in this root build script we have `integrationTest` aggregate task, which, as far as I understand, will now run only maven and cli tests and not Gradle. Are we planning to migrate `maven/cli` tests to test suites soon? If no, I think it will be better to somehow align this.

whyoleg: Note: before, when running `check` (`gradle check`) integration tests were not executed, but now they are. It's not convenient as integration tests are long :)

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,208 @@
...
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+import org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED
+import org.gradle.api.tasks.testing.logging.TestLogEvent.SKIPPED
+import org.gradle.language.base.plugins.LifecycleBasePlugin.VERIFICATION_GROUP
+import org.jetbrains.kotlin.gradle.dsl.ExplicitApiMode.Disabled
 
 plugins {
-    id("dokkabuild.test-integration")
+    id("dokkabuild.kotlin-jvm")
+    `test-suite-base`
```
whyoleg, line 12: What is the difference between `jvm-test-suite` and `test-suite-base` plugins and why we use `base` one?

dokka-integration-tests/utilities/build.gradle.kts
```diff
@@ -16,6 +16,8 @@ dependencies {
     implementation(libs.kotlinx.coroutines.core)
     implementation(libs.jsoup)
     implementation(libs.eclipse.jgit)
+
+    implementation(gradleTestKit())
```
whyoleg, line 20: why it's needed here? `utilities` is shared between other runners integration tests?

dokka-integration-tests/utilities/src/main/kotlin/org/jetbrains/dokka/it/gradleRunnerUtils.kt
```diff
@@ -1,4 +1,7 @@
-package org.jetbrains.dokka.dokkatoo.utils
+/*
+ * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package org.jetbrains.dokka.it
```
whyoleg, line 4: why it's placed in utilities and not in `gradle` subproject?

dokka-integration-tests/utilities/src/main/kotlin/org/jetbrains/dokka/it/systemVariableProviders.kt
```diff
@@ -1,4 +1,8 @@
-package org.jetbrains.dokka.dokkatoo.utils
+/*
+ * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.it
```
whyoleg, line 5: it looks like declarations from this file are not used now (at least IDEA shows no usages), is it for future or can be just removed?

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,208 @@
...
+    `java-test-fixtures`
 }
 
 dependencies {
-    implementation(projects.utilities)
+    api(projects.utilities)
```
whyoleg, line 17: just curious, why we need `api` here?

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,208 @@
...
+            }
+        }
+
+        // register a separate test suite for each 'template' project
+        registerTestProjectSuite(
+            "testTemplateProjectAndroid",
+            "it-android-0",
+            jvm = JavaLanguageVersion.of(11), // AGP requires JVM 11+
+        )
+        registerTestProjectSuite("testTemplateProjectBasic", "it-basic")
```
whyoleg, line 140: minor/question: as we now have a separate folder for tests with specific template, may be it will be a good idea to move also a template project there?
From one side it could be a little harder to find template project, from the other side it's convenient when test file and project will be near each other.

Or may be at least to align names of template projects and sourceSets to use one convention - up to you

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,208 @@
...
+            }
+        }
+
+        /**
+         * Create a new [JvmTestSuite] for a Gradle project.
+         *
+         * @param[projectPath] path to the Gradle project that will be tested by this suite, relative to [templateProjectsDir].
+         * The directory will be passed as a system property, `templateProjectDir`.
+         */
+        fun registerTestProjectSuite(
```
whyoleg, line 109: IMO, it would be better to move this function to the bottom of the file (or to `build-logic`) if possible to reduce context-switching while reading the file

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,208 @@
...
+    }
+
+    // TODO remove this in https://github.com/Kotlin/dokka/pull/3433
+    doNotTrackState("uses artifacts from Maven Local")
+}
+
+@Suppress("UnstableApiUsage")
+testing {
+    suites {
+        withType<JvmTestSuite>().configureEach {
```
whyoleg, line 87: minor: configuration here can be moved inside `registerTestProjectSuite` to easier understand all configuration. Not sure, that we will have any other suite which is not created by it

whyoleg (approved):
LGTM.

The only thing that Im afraid off, is that if/when we will enable `configuration-cache` for `dokka` project itself, running Gradle integration tests will use all my 64 GB of RAM on my Mac :)

adam-enko:
> The only thing that Im afraid off, is that if/when we will enable `configuration-cache` for `dokka` project itself, running Gradle integration tests will use all my 64 GB of RAM on my Mac :)

More incentive to reduce Dokka's memory usage 😆 

In practice the memory-usage should be limited. It has to run on TeamCity, which has 16GB!

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Yes, the organisation crossed my mind too!

I see two options

1. keep the template projects as they are now, in a specific project directory
    
    ```
    .
    └── dokka-integration-tests/gradle/
        ├── projects/
        │   ├── it-android-0
        │   ├── it-basic
        │   ├── it-basic-groovy
        │   └── ...
        └── src/
            ├── testTemplateProjectAndroid/
            ├── testTemplateProjectBasic/
            └── testTemplateProjectBasicGroovy/
    ```


3. move the template projects to be in the source
    
    ```
    .
    └── dokka-integration-tests/gradle/
        └── src/
            ├── testTemplateProjectAndroid/
            │   ├── kotlin/
            │   └── project/
            ├── testTemplateProjectBasic/
            │   ├── kotlin/
            │   └── project/
            ├── testTemplateProjectBasicGroovy/
            │   ├── kotlin/
            │   └── project/
            └── ...
    ```


I wasn't sure what would be better, both have pros/cons. I slightly prefer option 1 because it's easier to compare/contrast the template projects, but only slightly. I decided to minimize the changes in this PR.

dokka-integration-tests/utilities/src/main/kotlin/org/jetbrains/dokka/it/gradleRunnerUtils.kt
adam-enko: Good point, I'll move it

dokka-integration-tests/utilities/src/main/kotlin/org/jetbrains/dokka/it/systemVariableProviders.kt
adam-enko: thanks, you're right it can be moved back.

dokka-integration-tests/utilities/build.gradle.kts
adam-enko: it was required for gradleRunnerUtils.kt, but that's now moved to the gradle integration test project

build.gradle.kts
adam-enko: > AFAIK we already have this dependency below: `addDependencyOnSameTasksOfIncludedBuilds`


Unfortunately `addDependencyOnSameTasksOfIncludedBuilds` adds a dependency on `:dokka-integration-tests:check`, not `:dokka-integration-tests:gradle:check`.

TBH I found the logic about setting up task dependencies a bit confusing in this file, so there's probably a better way of using the functions, but I found being explicit easier to understand :). 

I get that it's a workaround for a Gradle flaw https://github.com/gradle/gradle/issues/22335 - I think using a task rule might be help https://stackoverflow.com/a/69179714/4161471


> Are we planning to migrate maven/cli tests to test suites soon? 

[Yes, for sure](https://github.com/Kotlin/dokka/pull/3427#discussion_r1432756930). I wanted to do it one subproject at a time, to minimize PR size, and then it will be easier to determine what functionality can be extracted into conventions & build-logic utils.

> Note: before, when running check (gradle check) integration tests were not executed, but now they are. It's not convenient as integration tests are long :)

Yes, that's intentional :) 'check' is the defacto standard lifecycle task for _all_ verification.

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: I'd prefer to keep JVM-specific config in the `withType<JvmTestSuite>().configureEach { }` block and test-project specific config in the `registerTestProjectSuite()` function. That's a clear split to me, but maybe that's not clear in the code and I can add a comment to explain?

Could you take another look, now that I've moved `registerTestProjectSuite()` to be a top-level function? I think it should help with readability.

dokka-integration-tests/gradle/build.gradle.kts
whyoleg: Yeah, let's not change that in this PR, but it could be improved later.
I don't know what is better (separate folder with templates or project in sources of test suite), but I think we can at least align names of test suites (sourceSets) and template projects.
F.e
* templateFolder=`basic` -> sourceSet=`basicTemplateTest` (or `testBasicTemplate` or `testTemplateBasic`)
* templateFolder=`multimodule` -> `sourceSet=`multimoduleTemplateTest` (or ...)
* templateFolder=`multimoduleSpecificCase` -> `sourceSet=`multimoduleSpecificCaseTemplateTest` (or ...)
comparing to current:
* templateFolder=`it-basic` -> sourceSet=`testTemplateProjectBasic`,
* templateFolder=`it-multimodule` -> `sourceSet=`testTemplateProjectMultimodule`
* templateFolder=`it-multimodule-specific-case` -> `sourceSet=`testTemplateProjectMultimoduleSpecificCase`
Though, not in this PR.

dokka-integration-tests/gradle/build.gradle.kts
whyoleg: now it looks better :)
Could you still add a comment there, regarding why we need this configuration (specifically `implementation(project)`), with link to Gradle docs or something (it's not super obvious behaviour, as I've written above)

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Yes, agreed to both! Names could be better, and not in this PR.

adam-enko: BTW I just had a discussion with @ALikhachev about how to improve the performance - specifically how to reduce the memory usage. One cause of high memory usage is that many of the tests use multiple different versions of Gradle, which means they each have a separate KotlinCompileDaemon.

Something that might help is to change limit what versions of Gradle run concurrently. One idea we had was to 1. move the tests into `src/main`, and 2. generate a test-suite for each Gradle/Kotlin/Android version. That way Gradle will only launch one task per Gradle version.

build.gradle.kts
whyoleg: Let's start from the end:)

> Yes, that's intentional :) 'check' is the defacto standard lifecycle task for all verification.

1. Yes, but in day-to-day workflow I would like to run just `test` + `apiCheck` (so old `check`) to test if my changes didn't broke anything. Sometimes it's fine to run only one module/test, but not all the time. At my side running `check` now will take 1-2 minute I think. If we will put running integration tests into `check` - it will become much more, so I will start to use `test` and `apiCheck` tasks separately. Or I will just run `check` for some subproject(s). External contributors also could just run `check` (as it's `defacto standard lifecycle task for all verification.`) and Im not sure, that all of them will want to wait for all tests or some of the tests could even fail, f.e. android, if there is no setup, and they will spend more time to fix the issue or will just ignore running test, or even finishing the feature. This is a little inconvenient.
3. as far as I understand, test suites created by Gradle test-suite plugin doesn't run when running `check` - and I think for a reason. `check` is day-to-day task for **fast** versification - integration tests are not fast.

So I think, that running `check` should not run integration tests - they should run by default on CI (in same/separate workflow), it should be enough for most cases. If needed, they can be easily run locally, all of them (rarely) or specific one (more frequently).

> I wanted to do it one subproject at a time, to minimize PR size

Agree, though, I've just mentioned inconsistency in behaviour of different integration tests after merging this PR - not a blocker, but just something to be aware of.

> TBH I found the logic about setting up task dependencies a bit confusing in this file, so there's probably a better way of using the functions, but I found being explicit easier to understand :).

yeah, still, it's very convenient to have such `aggregated` tasks in root project. If you have ideas on how to improve this, I will be glad to review separate PR :)
I've never used task rules before, so I don't know will it be possible/simpler/complex, so up to you.

dokka-integration-tests/gradle/build.gradle.kts
whyoleg: > Something that might help is to change limit what versions of Gradle run concurrently.

Up to you, will wait for the best solution, not sure what is the best approach here with all combination of so much different parameters.
But keep in mind, that some tests are run for different versions of Gradle+Kotlin combination, so looks like it will be a lot of combinations :)
Also, don't forget about IDEA (run/debug) experience if it's possible.

.github/workflows/tests-thorough.yml
```diff
@@ -30,11 +30,11 @@ jobs:
         # Running tests with the Gradle daemon on windows agents leads to some very strange
         # JVM crashes for some reason. Most likely a problem of Gradle/GitHub/Windows server
         run: >
-          ./gradlew test --stacktrace --no-daemon --no-parallel 
+          ./gradlew test --stacktrace --no-daemon --no-parallel --continue
```
IgnatBeresnev, line 33: Not entirely sure how adding this flag affects the build - could you please share why it was needed or what it improves?

build.gradle.kts
IgnatBeresnev: Sorry, but I strongly believe that `check` or `test` or any other similar basic task should not run anything as heavyweight / complicated as all of Dokka's "integration tests". 

I don't know if the reworked tests have the same problems as the current ones, but now you are unable to run integration tests locally if you
* didn't initialize git submodules (not done by default, fails with obscure errors like "unable to find a file")
* don't have Android SDK installed (the android test is executed regardless)
* have less than 32gb of ram
* can't leave some hungry process running for 1h (I had to close my IDEs and chrome to run tests when I had 16gb of ram, and it still didn't work half the time)

If you do manage to run these tests and they fail, it's not obvious at all to external contributors or members without context as to what to do about it and how to fix it (especially if it's a false positive or is expected), so for them it's virtually useless.

Integration tests were run by default before, and it was very annoying for the reasons outlined by Oleg. I ran pretty much every build with `test -x integrationTest`, but it might be impossible after this change because there's no taskgraph between tasks of included builds (all the `addDependencyOnSameTasksOfIncludedBuilds` calls), so they will be run whether you want it or not

___

As a side note (not a requirement or a comment for this PR specifically, there's no need to implement it). 

Ideally, I think the workflow should be the following: 
1. run basic checks locally (<= 1-2 min)
2. push a commit that triggers more complex CI checks (smoke, <= 30m), merge the PR
3. even more complex CI checks are triggered for every commit in master / nightly / manually (>= 30m)

 This idea was mentioned in https://github.com/Kotlin/dokka/issues/2925#issuecomment-1840852198:
> The current integration tests take 1h+ to run, which is painful in PRs and hogs CI agents. There should be a a differentiation between "full tests" that run everything, and their subset in the form of "smoke tests", that only test specific versions (maybe only the latest) or projects (maybe only 2 out of 10 user projects). Smoke tests are intended to be run for commits in PRs, and full tests - overnight or triggered manually.

dokka-integration-tests/gradle/src/testExternalProjectKotlinxCoroutines/kotlin/CoroutinesGradleIntegrationTest.kt
```diff
@@ -53,7 +53,7 @@ class CoroutinesGradleIntegrationTest : AbstractGradleIntegrationTest(), TestOut
         val result = createGradleRunner(
             buildVersions,
             ":dokkaHtmlMultiModule", "-i", "-s",
-            jvmArgs = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=500m")
+            jvmArgs = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=2g")
```
IgnatBeresnev, line 56: Is there a specific reason for increasing metaspace size? These tests also helped us find metaspace memory leaks a couple of times

.github/workflows/tests-thorough.yml
adam-enko: Without `--continue` Gradle will stop when the first task fails. With `--continue` Gradle will keep running tasks (if possible). Continuing even if one task fails is useful when running all tests because it helps to determine if your change has broken just one test vs many tests.

dokka-integration-tests/gradle/src/testExternalProjectKotlinxCoroutines/kotlin/CoroutinesGradleIntegrationTest.kt
adam-enko: I increased it because the test failed :) I didn't realise it was intentionally low - I'll set it back and add a comment to clarify why it should be kept low.

build.gradle.kts
adam-enko: Thanks for explaining, it definitely makes sense to make the UX of the Gradle tasks more comprehensible. Anyone should be able to run suitable tests with a single Gradle task.

I see there are some good options for this, proposed at the bottom, but first I want to clarify about the `check` task.

### Check task

Personally, I find the current behaviour (`check` doesn't run all tests) confusing because in all other projects it's a lifecycle task that runs all verifications.

That's because the `check` task is provided by the [Gradle Base Plugin](https://docs.gradle.org/current/userguide/base_plugin.html), which all other plugins use, and it's supposed to run all verifications.

image

Of course, Gradle is very flexible so Dokka doesn't have to follow this convention, but it would be unusual. It's important to align the conventions within Kotlin projects - so while "check runs all tests" might not meet your understanding at the moment, I hope it will make more sense in the future.

### Proposal

I see two options

1. Update the `test` task so that it will also run the `apiCheck` task.
    
    Update the README instructions:
    
    * `./gradlew check` - run all verifications (which might be slow):
      * `./gradlew test` - run unit tests (and also `apiCheck`)
      * `./gradlew apiCheck` - BCV check
      * `./gradlew integrationTest` - run all integration tests (Slow! Requires Prep!)

2. Add a custom `quickCheck` (or `checkFast` - whatever name you want) lifecycle task that will run unit tests + apiCheck (and any other future 'fast' checks).

    Update the README instructions:
    
    * `./gradlew check` - run all verification (which might be slow):
      * `./gradlew quickCheck` - runs quicker checks (`test` and `apiCheck`)
      * `./gradlew test` - run unit tests
      * `./gradlew apiCheck` - BCV check
      * `./gradlew integrationTest` - run all integration tests (Slow! Requires Prep!)

I have a slight preference towards 1, because currently there's not many verification tasks apart from `test` and `apiCheck` that would justify a new task.

Which do you prefer?

I can also split out these changes to the test-task dependencies into a separate PR to keep this PR focused.

IgnatBeresnev: > it's supposed to run all verifications.

Can you try to define "all" here? :) 

I think it's very important to draw the line somewhere, otherwise it's easy to fall into unproductive malicious compliance and make the `check` task perform **literally** all verifications: 
* Running tests on different platforms and not only on the user's one (i.e start docker containers for missing OSes like MacOS/Linux/Windows)
* Publishing verification checks (not done now)
* UI tests that might reside in a different repo (we'll have them soon, #3469)
* Run first with K1, then with K2 (even if not needed by default)
* Tests that run Dokka with the latest/arbitrary commit of `kotlin-compiler` / KGP / etc (also planned)
* Stress tests / Performance tests / Benchmarks

I would draw the line somewhere close to "Runs all embedded verifications that don't require special knowledge or setup, and that are reasonably expected to be run locally by developers". By that definition, Dokka's "integration tests" (which are really completely separate e2e tests) are not it for several reasons

___

About the proposals - I would first try to understand why we need the `check` task to run Dokka's "integration tests" specifically. It's taken for granted, but I don't think it should be :) 

Who or what is going to be using it **if** `check` also runs Dokka's "integration tests"? How is it going to be useful?

.github/workflows/tests-thorough.yml
IgnatBeresnev: It's indeed useful :+1:  For some reason I thought that it was the default behaviour, and that adding this flag had something to do with the rest of the changes, but perhaps it's just an unrelated minor improvement

build.gradle.kts
whyoleg: Yeah, Im totally agree with Ignat.
Still I will add some additional points:
1. AGP doesn't make `check` dependsOn instrumented tests (which are executed on emulator/real device) - they are slow and environment dependent
2. even Gradle by default doesn't do this for test suites - this means, that it's not strictly true, that we need to run all kind of tests by default when we run `check`
3. If we will make `test` dependOn `apiCheck` or any other check task (f.e. if we will add `detekt`) - if we will run tests from IDEA via gutter (just one test or test class), we will execute `apiCheck` - not sure it's expected. May be it's possible to workaround it, but not sure that by doing this we will simplify both build setup and user expectations.

So I believe `integrationTests` for Dokka should be executed in separate task. Additionally, we could created lifecycle task `checkAll`/`checkLong` if we really do want to execute it.

Another possibility: make global `check` run all tests except integration tests. make `dokka-integration-tests` `check` run all tests there, even integration.
It can cause confusion, in this case we will have some split between `check` tasks. But IMO, it will be still better then running all `integration tests` by calling global `check`

dokka-integration-tests/gradle/build.gradle.kts
ALikhachev: > the tests use multiple different versions of Gradle, which means they each have a separate KotlinCompileDaemon.

Not exactly. Each Kotlin version will have its own separate Kotlin daemon, but e.g. in combinations `Gradle 7.0 – Kotlin 1.7.0` and `Gradle 7.5 – Kotlin 1.7.0` there will be single Kotlin daemon and two Gradle daemons

ALikhachev (changes requested):
In general, LGTM 👍

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,235 @@
...
+    systemProperty("org.jetbrains.dokka.experimental.tryK2", useK2)
+    // allow inspecting projects in temporary dirs after a test fails
+    systemProperty(
+        "junit.jupiter.tempdir.cleanup.mode.default",
+        dokkaBuild.isCI.map { isCi -> if (isCi) "ALWAYS" else "ON_SUCCESS" }.get(),
+    )
 
     environment("DOKKA_VERSION", project.version)
+    environment("isExhaustive", dokkaBuild.integrationTestExhaustive)
+    environment("ANDROID_HOME", dokkaBuild.androidSdkDir.get().invariantSeparatorsPath)
```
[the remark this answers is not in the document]
adam-enko: Good spot, thanks. `environment()` still isn't Provider API compatible https://github.com/gradle/gradle/issues/11534

I've added a workaround.

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,43 +1,235 @@
...
+    api(gradleTestKit())
 }
 
 kotlin {
     // this project only contains test utils and isn't published, so it doesn't matter about explicit API
-    explicitApi = ExplicitApiMode.Disabled
+    explicitApi = Disabled
+
+    compilerOptions {
+        allWarningsAsErrors = false
```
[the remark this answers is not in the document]
adam-enko: Good point. I think this was left by mistake - I've removed it.

build.gradle.kts
adam-enko: I've removed the check task dependency.


> > it's supposed to run all verifications.
> 
> Can you try to define "all" here? :)

See https://docs.gradle.org/current/userguide/base_plugin.html

> Plugins and build authors should attach their verification tasks, such as ones that run tests, to this lifecycle task using `check.dependsOn(task)`.



>     * Running tests on different platforms and not only on the user's one (i.e start docker containers for missing OSes like MacOS/Linux/Windows)
>     * Publishing verification checks (not done now)

There are no Gradle tasks for this, so `check` can't depend on them.

>     * UI tests that might reside in a different repo (we'll have them soon, [Infrastructure to run UI tests #3469](https://github.com/Kotlin/dokka/issues/3469))

`check` can't depend on tasks in other repos.

>     * Run first with K1, then with K2 (even if not needed by default)

This is a good idea. It would help prevent the problem we saw recently where the K2 tests weren't running on TeamCity.

>     * Tests that run Dokka with the latest/arbitrary commit of `kotlin-compiler` / KGP / etc (also planned)

If this is achieved with a Gradle task then yes, `check` should depend on that task.

>     * Stress tests / Performance tests / Benchmarks

This depends:

* If the task has checks, then yes, `check` should depend on it

  For example, the existing `SequentialTasksExecutionStressTest`

* If the tasks are measuring performance and reporting on it, but not verifying it, then that's not a verification task so `check` shouldn't depend on it.

   For example: if JMH tests were added


> I would draw the line somewhere close to "Runs all embedded verifications that don't require special knowledge or setup, and that are reasonably expected to be run locally by developers". By that definition, Dokka's "integration tests" (which are really completely separate e2e tests) are not it for several reasons

I agree, there's a good need for such lifecycle task, and the meaning you explain is clear.

It's confusing to overload an existing task with a different meaning. It would make more sense to create a new lifecycle task that meets this definition.

It would also help prevent issues on CI where verifications have been skipped. E.g. K2 tests not being run, and up until recently apiCheck wasn't run on TeamCity.

Using `check` to run all test tasks aligns how `check` works both within JetBrains, but also within the Gradle ecosystem. Many other plugins use `check` to run all verifications.

IgnatBeresnev (approved):

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -1,52 +1,246 @@
...
 
     environment("DOKKA_VERSION", project.version)
 
-    inputs.dir(templateProjectsDir)
+    // environment() isn't Provider API compatible yet https://github.com/gradle/gradle/issues/11534
+    dokkaBuild.integrationTestExhaustive.orNull?.let { exhaustive ->
+        environment("isExhaustive", exhaustive)
+    }
+    dokkaBuild.androidSdkDir.orNull?.let { androidSdkDir ->
+        environment("ANDROID_HOME", androidSdkDir)
```
ALikhachev, line 77: Previously it was set with transformation using `invariantSeparatorsPath`, does it work correctly without it?

ALikhachev (approved):

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Good catch. Yes, it will work the same, but it should use an invariant path to help with remote Build Cache.

PR #3432, opened 2023-12-20
test: automation script for local Dokka testing

berezinant:
.testDokka.sh makes Dokka development and testing a bit faster

It compiles current Dokka version from the source code,
publishes it to the local Maven repository,
then runs the Dokka documentation generation against the specified project
and finally runs a webserver to open the generated documentation in the browser.

Demo of usage: https://monosnap.com/file/m1xuNF7IVlHXtYUZWovOBduYY3GsER

PS this is a prototype based on the most frequent things I've to do while testing Dokka, it could be improved or done in some other way, let's discuss


testDokka.sh
```diff
@@ -0,0 +1,83 @@
...
+echo "Dokka path: $DOKKA_REPO_PATH"
+echo "Port: $PORT"
+
+# 1. Publish to local Maven repository
+cd $DOKKA_REPO_PATH
+./gradlew publishToMavenLocal -Pversion=$NEW_VERSION
+
+# 2. Update Dokka version in test project
+cd $TEST_PROJECT_PATH
+sed -i "" "s/id(\"org.jetbrains.dokka\") version \"\(.*\)\"/id(\"org.jetbrains.dokka\") version \"$NEW_VERSION\"/"  build.gradle.kts
```
zoobestik, line 52: Looks like it's possible shorter:

```suggestion
sed -i "" "s/\(id(\"org\.jetbrains\.dokka\") version\) \".*\"/\1 \"$NEW_VERSION\"/"  build.gradle.kts
```

testDokka.sh.md
```diff
@@ -0,0 +1,42 @@
...
+### Specify port
+
+```bash
+./testDokka.sh  -p 8001
+```
+
+### All together
+
+```bash
+./testDokka.sh -d './examples/gradle/dokka-gradle-example' -v "1.9.20-SNAPSHOT" -p 8001
```
zoobestik, line 39: I think, it would be nice use for default not real version (for ex. `1.9.10-my-fix-SNAPSHOT` like [CONTRIBUTING.md](https://github.com/Kotlin/dokka/blob/master/CONTRIBUTING.md))

zoobestik: And may be do link in contribution.md.
/cc @IgnatBeresnev

testDokka.sh
```diff
@@ -0,0 +1,83 @@
...
+echo "Open http://localhost:$PORT in browser"
+
+
+# 5.2 Start Python server to view results
+cd "./build/dokka/html"
+
+echo 'Start Python server in directory'
+echo "$TEST_PROJECT_PATH/build/dokka/html"
+
+python3 -m http.server $PORT
```
zoobestik, line 81: Add this requirement to .md?

testDokka.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
```
zoobestik, line 2: Exit when any command fails?
 
 ```suggestion
set -e
```

File: testDokka.sh [hunk 5]
```diff
@@ -0,0 +1,83 @@
...
+    esac
+done
+
+echo "Test locally Dokka version $NEW_VERSION"
+echo "Test project path: $TEST_PROJECT_PATH"
+echo "Dokka path: $DOKKA_REPO_PATH"
+echo "Port: $PORT"
+
+# 1. Publish to local Maven repository
+cd $DOKKA_REPO_PATH
```
zoobestik, line 47: Will it work correctly without quotes?

testDokka.sh
```diff
@@ -0,0 +1,83 @@
...
+./gradlew publishToMavenLocal -Pversion=$NEW_VERSION
```
zoobestik, line 48: Is it required`./gradlew clean build ...`?

berezinant: Usually no, I haven't encountered any problems while not doing `clean` here

testDokka.sh
berezinant: This is not convenient because it stops webserver after the script is finished

testDokka.sh
berezinant: Yep, works well with and without quotes but I added it just in case

zoobestik (approved):

zoobestik (approved):

examples/gradle/dokka-customFormat-example/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ buildscript {
 }
 
 repositories {
+    mavenLocal()
```
whyoleg, line 21: minor/major: not sure that it's a good idea to add `mavenLocal` to examples which are used also by external users which want to look how `dokka` is configured.
The main issue is that users often just `copy-paste` everything what is in file and then they can leave `mavenLocal` there and later accidentally have issues because of it.

I think, it will be better to add `mavenLocal` in this case via script if possible

testDokka.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
```
whyoleg, line 1: minor: may be it's better to move script and it's description into `scripts` dir to not pollute root folder? if that's will be fine (I understand, that it could become inconvenient because of paths)

whyoleg (approved):
LGTM.
@IgnatBeresnev WDYT regarding `mavenLocal` in examples?

examples/gradle/dokka-customFormat-example/build.gradle.kts
berezinant: Yep, thanks, I agree that adding `mavenLocal()` via script is more flexible

IgnatBeresnev (approved):
Oleg pretty much already left the two comments that I had :))

PR #3438, opened 2023-12-30
Fix logic to declare incompatibility with configuration cache

3flex:
The existing logic does not work correctly (at least on Gradle 8.5) because the name of the first parameter of `notCompatibleWithConfigurationCache` method is `null`, not `reason`, so `containsNotCompatibleWithConfigurationCache` incorrectly evaluates to `false`.

This PR removes unnecessary reflection and corrects the logic.


whyoleg (approved):
Hey! Thank you for the PR! Somehow we missed it earlier.
LGTM!
I've additionally checked, that with this fix everything works as expected on both artificial projects on different Gradle version with both enabled/disabled configuration cache and real ones (like detekt) - everything works fine.

PR #3440, opened 2024-01-08
[K2] Display `override` keyword for members implemented by delegation

vmishenev:


vmishenev:
Closed in favour of #3566

Issue #3207, opened 2023-10-13
[K2] Support `[this]` link for extensions


vmishenev:
IDE and K1 support `[this]` links for documentation of extensions, but only for them. (not member)
We already have such links  in user projects, e.g. in https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.descriptors/get-contextual-descriptor.html. 
In Dokka K1 `[this]` has an incorrect link that leads to a page of containing package. For example, `fun pkcg.SomeClass.f()` a `[this]` link references to the `pkcg` page.
In IDE `[this]` points to a receiver parameter, but Dokka does not support direct links to parameters (old Dokka does, [example](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter.html#kotlin.collections$filter(kotlin.Array((kotlin.collections.filter.T)),%20kotlin.Function1((kotlin.collections.filter.T,%20kotlin.Boolean)))/predicate)) , so Dokka K2 should just open a page of the receiver type. 
It depends on :heavy_check_mark: https://youtrack.jetbrains.com/issue/KT-62407

Also, see https://github.com/Kotlin/dokka/issues/3197

PR #3445, opened 2024-01-11
[K2] Support `[this]` link for extensions

vmishenev:
#3207

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DRIFactory.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DRIFactory.kt
@@ -106,8 +106,8 @@ private fun KtAnalysisSession.getDRIFromReceiverType(type: KtType): DRI {
         is KtDefinitelyNotNullType -> getDRIFromReceiverType(type.original)
         is KtTypeErrorType -> DRI(packageName = "", classNames = "$ERROR_CLASS_NAME $type")
         is KtClassErrorType -> DRI(packageName = "", classNames = "$ERROR_CLASS_NAME $type")
+        is KtDynamicType -> DRI(packageName = "", classNames = "$ERROR_CLASS_NAME $type") // prohibited by a compiler, but it's a possible input
 
-        is KtDynamicType -> throw IllegalStateException("Dynamic receiver is prohibited while creating DRI $type")
         is KtCapturedType -> throw IllegalStateException("Unexpected non-denotable type while creating DRI $type")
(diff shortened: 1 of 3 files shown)


whyoleg (approved):

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DRIFactory.kt
```diff
@@ -93,6 +93,28 @@ internal fun KtAnalysisSession.getDRIFromValueParameter(symbol: KtValueParameter
...
+private fun KtAnalysisSession.getDRIFromReceiverType(type: KtType): DRI {
+    return when(type) {
+        is KtNonErrorClassType -> getDRIFromNonErrorClassType(type)
+        is KtTypeParameterType -> getDRIFromTypeParameter(type.symbol)
+        is KtDefinitelyNotNullType -> getDRIFromReceiverType(type.original)
+        is KtTypeErrorType -> DRI(packageName = "", classNames = "$ERROR_CLASS_NAME $type")
+        is KtClassErrorType -> DRI(packageName = "", classNames = "$ERROR_CLASS_NAME $type")
+
+        is KtDynamicType -> throw IllegalStateException("Dynamic receiver is prohibited while creating DRI $type")
+        is KtCapturedType -> throw IllegalStateException("Unexpected non-denotable type while creating DRI $type")
```
IgnatBeresnev, line 111: When I read these two lines, "unexpected" (from the second line) doesn't cause many questions, it likely means this type just hasn't been supported for whatever reason, but I can implement it if I need to fix a bug or something, so it's all good

"prohibited", however, immediately causes the question of "why". It reads as a strong requirement that it must not be implemented, and if I do implement it (in case there's a bug associated with it), something very bad will happen. and that if this error is even thrown in a user project, it's an indicator of a major problem.

is that the case? if it is, could you please add a code comment or change the exception message to explain why it's prohibited?

dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
```diff
@@ -121,6 +123,105 @@ class LinkTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    @OnlySymbols("#3207")
```
IgnatBeresnev, line 127: Will these tests not pass with K1? Is the behaviour still different in K1 and K2? I thought we were going to make it work the same way as in K1, so that the common tests pass

vmishenev: #3207: In Dokka K1 `[this]` has an incorrect link that leads to a page of containing package.  For example, `fun pkcg.SomeClass.f()` a `[this]` link references to the `pkcg` page.

PR #3446, opened 2024-01-12
Configure JetBrains Gradle Enterprise & Build Cache

adam-enko:
Use JetBrains Gradle Enterprise for builds.

This should improve the build speeds both for local development and on CI.

### Summary

* Introduces a new `build-settings-logic` project for settings conventions.

  It's named so that it's next to the existing `build-logic`.

* Convention plugins for Build Cache and Gradle Enterprise copied from https://github.com/JetBrains/kotlin/
  * https://github.com/JetBrains/kotlin/blob/2675531624d42851af502a993bbefd65ee3e38ef/repo/gradle-settings-conventions/build-cache/src/main/kotlin/build-cache.settings.gradle.kts
  * https://github.com/JetBrains/kotlin/blob/19073b96a7ed53dbda61337465ca898c1482e090/repo/gradle-settings-conventions/gradle-enterprise/src/main/kotlin/gradle-enterprise.settings.gradle.kts

### TODO

- [x] ~Update the readme/contributing guide? What properties are necessary?~ I added some brief docs. Build Scans aren't as important so I've opted out by default. 
- [x] ~update TeamCity config to add necessary properties.~ Dokka has a Build Cache account. I've added the user/pass to TeamCity
- [ ] update GitHub Actions config to add environment variable `ORG_GRADLE_PROJECT_org.jetbrains.dokka.build.scan.url=https://ge.jetbrains.com/` @IgnatBeresnev are you able to do this? I don't have access [to the config](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,7 @@ Notable builds:
 ### Gradle Build Scans
 
 [Gradle Build Scans](https://scans.gradle.com/) can provide insights into a Dokka Build. 
-JetBrains runs a Gradle Build Scan server that can be used to automatically upload reports.
+JetBrains runs a Gradle Build Scan server https://ge.jetbrains.com/ that can be used to automatically upload reports.
 
 To automatically opt in add the following to `$GRADLE_USER_HOME/gradle.properties`. 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,8 @@ Notable builds:
 ### Gradle Build Scans
 
 [Gradle Build Scans](https://scans.gradle.com/) can provide insights into a Dokka Build. 
-JetBrains runs a Gradle Build Scan server https://ge.jetbrains.com/ that can be used to automatically upload reports.
+JetBrains runs a Gradle Build Scan server https://ge.jetbrains.com/scans?search.rootProjectNames=dokka
+that can be used to automatically upload reports.
 
 To automatically opt in add the following to `$GRADLE_USER_HOME/gradle.properties`. 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,6 +154,21 @@ Notable builds:
   runs Dokka's integration tests, which are designed to test compatibility with different Kotlin versions, with different
   multiplatform targets and with various user scenarios.
 
+### Gradle Build Scans
+
+[Gradle Build Scans](https://scans.gradle.com/) can provide insights into a Dokka Build. 
+JetBrains runs a Gradle Build Scan server that can be used to automatically upload reports.
+
+To automatically opt in add the following to `$GRADLE_USER_HOME/gradle.properties`. 
+
+```properties
+org.jetbrains.dokka.build.scan.url=https\://ge.jetbrains.com/
+# optionally provide a username that will be attached to each report
+org.jetbrains.dokka.build.scan.username=Hannah Clarke
+```
+
+A Build Scan may contain identifiable information. See the Terms of Use https://gradle.com/legal/terms-of-use/.
+
 ## Contacting maintainers
 
 * If something cannot be done, not convenient, or does not work &mdash; submit an [issue](#submitting-issues).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,7 @@ Notable builds:
 ### Gradle Build Scans
 
 [Gradle Build Scans](https://scans.gradle.com/) can provide insights into a Dokka Build. 
-JetBrains runs a Gradle Build Scan server https://ge.jetbrains.com/scans?search.rootProjectNames=dokka
+JetBrains runs a [Gradle Develocity server](https://ge.jetbrains.com/scans?search.rootProjectNames=dokka).
 that can be used to automatically upload reports.
 
 To automatically opt in add the following to `$GRADLE_USER_HOME/gradle.properties`. 
--- a/build-settings-logic/settings.gradle.kts
+++ b/build-settings-logic/settings.gradle.kts
@@ -1,0 +1,29 @@
+import org.gradle.api.initialization.resolve.RepositoriesMode.PREFER_SETTINGS
+
+/*
+ * Copyright 2014-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+rootProject.name = "build-settings-logic"
+
+pluginManagement {
+    repositories {
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
+
+@Suppress("UnstableApiUsage")
+dependencyResolutionManagement {
+    repositoriesMode.set(PREFER_SETTINGS)
+    repositories {
+        mavenCentral()
+        gradlePluginPortal()
+    }
+
+    versionCatalogs {
+        create("libs") {
+            from(files("../gradle/libs.versions.toml"))
+        }
+    }
+}
--- a/build-settings-logic/src/main/kotlin/DokkaBuildSettingsProperties.kt
+++ b/build-settings-logic/src/main/kotlin/DokkaBuildSettingsProperties.kt
@@ -22,7 +22,7 @@ abstract class DokkaBuildSettingsProperties @Inject constructor(
 
 
     //region Gradle Build Scan
-    val buildScanServer: Provider<String> =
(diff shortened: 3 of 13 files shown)


dokka-integration-tests/settings.gradle.kts
```diff
@@ -8,17 +8,19 @@ rootProject.name = "dokka-integration-tests"
 
 pluginManagement {
     includeBuild("../build-logic")
+    includeBuild("../build-settings-logic")
 
     repositories {
-        gradlePluginPortal()
         mavenCentral()
+        gradlePluginPortal()
```
adam-enko, line 15: Unrelated change: prefer Maven Central over Gradle Plugin Portal.

Maven Central is more stable than Gradle Plugin Portal (which I think still connects to JCenter, which is unstable).

settings.gradle.kts
```diff
@@ -4,28 +4,30 @@
...
-
         maven("https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide")
         maven("https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies")
 
         maven("https://cache-redirector.jetbrains.com/intellij-repository/releases")
         maven("https://cache-redirector.jetbrains.com/intellij-third-party-dependencies")
 
+        mavenCentral()
+        google()
+
```
adam-enko, line 29: Unrelated change: put the JB cached repositories first, otherwise Gradle will find the dependency in Maven Central / Google and then not search the cached repositories.

settings.gradle.kts
```diff
@@ -94,16 +97,6 @@ include(
...
 
-val isCiBuild = System.getenv("GITHUB_ACTIONS") != null || System.getenv("TEAMCITY_VERSION") != null
-
-gradleEnterprise {
-    buildScan {
-        termsOfServiceUrl = "https://gradle.com/terms-of-service"
-        termsOfServiceAgree = "yes"
-        publishAlwaysIf(isCiBuild)
-    }
-}
```
adam-enko, line 105: Moved to the `dokkasettings.gradle-enterprise` convention plugin.

build-settings-logic/settings.gradle.kts
```diff
@@ -0,0 +1,29 @@
+import org.gradle.api.initialization.resolve.RepositoriesMode.PREFER_SETTINGS
```
whyoleg, line 1: imports should be below copyright (IDEA auto-import for some reason do it otherwise in `build-scripts`)

CONTRIBUTING.md
```diff
@@ -154,6 +154,22 @@ Notable builds:
   runs Dokka's integration tests, which are designed to test compatibility with different Kotlin versions, with different
   multiplatform targets and with various user scenarios.
 
+### Gradle Build Scans
+
+[Gradle Build Scans](https://scans.gradle.com/) can provide insights into a Dokka Build. 
+JetBrains runs a Gradle Build Scan server https://ge.jetbrains.com/scans?search.rootProjectNames=dokka
+that can be used to automatically upload reports.
+
+To automatically opt in add the following to `$GRADLE_USER_HOME/gradle.properties`.
```
whyoleg, line 163: not sure, that suggesting to put those properties in `global` directory is good idea for both internal and external contributors :)
I would expect that it will be possible to put it into `local.properties` file in the project

build-settings-logic/src/main/kotlin/DokkaBuildSettingsProperties.kt
```diff
@@ -0,0 +1,73 @@
...
+            .orElse(BUILD_SCAN_USERNAME_DEFAULT)
+            .map(String::trim)
+    //endregion
+
+
+    //region Gradle Build Cache
+    val buildCacheLocalEnabled: Provider<Boolean> =
+        dokkaProperty("build.cache.local.enabled", String::toBoolean)
+            .orElse(!buildingOnCi)
+            .orElse(true)
```
whyoleg, line 42: not sure, that this line will be ever executed, as `orElse` above will set value, isn't it?
(the same for `buildCachePushEnabled`)

whyoleg:
> update GitHub Actions config to add environment variable ORG_GRADLE_PROJECT_org.jetbrains.dokka.build.scan.url=https://ge.jetbrains.com/ @whyoleg are you able to do this?

Nope. @IgnatBeresnev this if for you I believe.

Also, just a question, how will it affect local development or only CI? AFAIU local build cache will work as before, remote build cache will not be available by default (and will be used only on CI)?

whyoleg (approved):
LGTM

CONTRIBUTING.md
adam-enko: What's not a good idea about it? :) The property is prefixed with `org.jetbrains.dokka` so it shouldn't collide with other properties. It's a little inconvenient, but unfortunately Gradle declined support for local.properties https://github.com/gradle/gradle/issues/12283, so Dokka would have to add a custom implementation. I can do it, but I think it's more complication than its worth.

adam-enko:
> Also, just a question, how will it affect local development or only CI? AFAIU local build cache will work as before, remote build cache will not be available by default (and will be used only on CI)?

Remote Build Cache will always be enabled, but in read-only mode. TeamCity has the Build Cache credentials, and so 'push' is enabled. For local development this will help by being able to re-use the build results from CI.

(Build Scan is opt-in, and is disabled by default.)

CONTRIBUTING.md
whyoleg: The only thing that I don't like, is that `project-specific` properties are located in `global` place - so you can very easily forgot about them :)
But, it's a minor, not sure, that someone external will need to put them into `global` place.
At least for me, I would just put it into `gradle.properties` and add to separate `Change list` in IDEA, to not commit it accidentally.

whyoleg:
> Remote Build Cache will always be enabled, but in read-only mode. 

But `build.cache.url` has no default, so no remote build cache will be enabled by default. So this means, that we will need additional configuration locally, right?

adam-enko:
> > Remote Build Cache will always be enabled, but in read-only mode.
> 
> But `build.cache.url` has no default, so no remote build cache will be enabled by default. So this means, that we will need additional configuration locally, right?

`org.jetbrains.dokka.build.cache.url` is set in the project root `gradle.properties` https://github.com/Kotlin/dokka/pull/3446/files#diff-3d103fc7c312a3e136f88e81cef592424b8af2464c468116545c4d22d6edcf19

whyoleg:
> org.jetbrains.dokka.build.cache.url is set in the project root gradle.properties

🤦, somehow missed it, sorry...
No more questions, good to go from my side

CONTRIBUTING.md
adam-enko: Yes, normally I would agree, but I think Build Scan is different and should be explicitly opted in, because it uploads identifiable information.

`kotlin.git` has a more complicated set up, where it logs a message to console and provides guidance on how to opt in. It then generates a local.properties file for Build Scan and other `kotlin.git` specific properties. I could recreate the same functionality, but I think that's a bit too complicated :)

adam-enko:
> 🤦, somehow missed it, sorry... No more questions, good to go from my side

No problem, thanks!

CONTRIBUTING.md
whyoleg: yeah, better to keep it simple

build-settings-logic/src/main/kotlin/dokkasettings.gradle-enterprise.settings.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+gradleEnterprise {
+    buildScan {
+        if (buildScanServer != null) {
+            server = buildScanServer
+            publishAlways()
+
+            capture {
+                isTaskInputFiles = true
+                isBuildLogging = true
+                isBuildLogging = true
```
Tapchicoma, line 47: Nit: duplicated line

Tapchicoma (approved):

build-settings-logic/src/main/kotlin/dokkasettings.gradle-enterprise.settings.gradle.kts
adam-enko, line 47:
```suggestion
                isBuildLogging = true
```

Issue #3250, opened 2023-10-25
[K2] Order of packages is unstable


vmishenev:
It can affect order of packages in a navigation menu.

Presumably it happens due to unstable order of source files. The order (traversing) of files depends on a system (linux/windows) https://github.com/Kotlin/dokka/actions/runs/6630172896/job/18012742052?pr=3227. 

See the test `two classes from different packages`, but more tests of package order are required.

IgnatBeresnev:
The order of functions also seems to be unstable compared to K1. This test reproduces it: 

https://github.com/Kotlin/dokka/blob/4e47ed400ae51734a798c26b63baf58b50b7c0a4/dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/mixed/SampleMixedJvmAnalysisTest.kt#L75-L81

IgnatBeresnev:
It's unclear what order we want to have in Dokka, by alphabet or as in code. First need to decide the product part.

IgnatBeresnev:
Need to research which user-visible pages have undetermined / unstable order, and ask @d-ambatenne to help. If it's only internal and related to our plugin API, we can decide it internally.

vmishenev:
It also affects a KDoc link of overload. See tests in #3447

vmishenev:
I have created [KT-65110](https://youtrack.jetbrains.com/issue/KT-65110/Analysis-API-In-Standalone-mode-the-order-of-symbols-is-unstable) to have a stable order internally.

IgnatBeresnev:
Andrei has a local patch to fix it for testing, so it's not blocking #3550 per se. The priority can be lowered

whyoleg:
Action point:
* try to fix it on AA side ourselves

vmishenev:
Just for the record, the unit test `two classes from different packages `  passes on only Windows, but not on Linux/Mac. 
https://github.com/Kotlin/dokka/actions/runs/10633139023
![image](https://github.com/user-attachments/assets/cc5eb6ad-a8f4-4577-879d-138c8601f603)

PR #3447, opened 2024-01-13
[K2] Fix order of KDoc link candidates

### LINKED ISSUE #3375: [K2] `List` reference leads to the stdlib collections package
```Kotlin
/**
 * ref to [List] or [Set]
 */
fun x(){}
```

`List` link leads to https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/ instead of https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/ (in K1).
For `Set` the link is correct.

**Installation**
- Dokka version: 1.9.20

Parent: #3328

vmishenev:
Fixes #3375

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
@@ -9,7 +9,7 @@ import org.jetbrains.dokka.analysis.kotlin.symbols.translators.getDRIFromSymbol
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.utilities.DokkaLogger
 import org.jetbrains.kotlin.analysis.api.KtAnalysisSession
-import org.jetbrains.kotlin.analysis.api.symbols.KtSymbol
+import org.jetbrains.kotlin.analysis.api.symbols.*
 import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.kdoc.psi.api.KDoc
 import org.jetbrains.kotlin.kdoc.psi.impl.KDocLink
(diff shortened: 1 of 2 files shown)


whyoleg (approved):

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -78,5 +78,20 @@ internal fun KtAnalysisSession.resolveKDocLinkToDRI(link: KDocLink): DRI? {
...
+ */
+private var linkCandidatesComparator: Comparator<KtSymbol> = compareBy{
+    when(it) {
+        is KtClassifierSymbol -> 1
+        is KtPackageSymbol -> 2
+        is KtFunctionSymbol -> 3
+        is KtVariableSymbol-> 4
+        else -> 5
+    }
+}
```
IgnatBeresnev, line 97: Could you please cover all cases with tests? I see a test for the initial bug where class < function, which is good, and it would be very helpful to make sure other cases don't break either. Especially once KT-64190 is fixed - it will help us verify that it was implemented the way we expected

And a side question: can there be two symbols of the same type? For example, two `KtClassifierSymbol`? Asking in case there's a non-obvious corner case related to source sets, expect/actual, `internal` visibility, java interop or something else? If there is, does it make sense to add a second comparison with `thenBy { it.name }` or something similar, to make it predictable?

vmishenev: > can there be two symbols of the same type? 

It is possible for `KtFunctionSymbol` when there are overloads. In this case, the order is not specified at all and it does not affect a result URL. It is not easy to make the order like in K1 on Dokka's side and consistent with IDEA. See added `tests for overloads`

For the other types of symbols, it is difficult to come up with a case when we have the same name in the same scope.

vmishenev: I think a stable order of overloads can be addressed in #3250

Luqman2302:
gh pr checkout 3447

PR #3465, opened 2024-01-23
[K2] Add a check that source roots do not intersect

vmishenev:
The pre-generation check throws `DokkaException` [here](https://github.com/Kotlin/dokka/blob/8e5c63d035ef44a269b8c43430f43f5c8eebfb63/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/generation/SingleModuleGeneration.kt#L105). Probably it could be  a logging error messages instead of an exception. Need to discuss.

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/SourceRootIndependentChecker.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/SourceRootIndependentChecker.kt
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.analysis.kotlin.symbols.plugin
+
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.validity.PreGenerationChecker
+import org.jetbrains.dokka.validity.PreGenerationCheckerOutput
+import java.io.File
+
+/**
+ * It checks that different source sets should have disjoint source roots and samples.
+ *
+ *
+ * K2's analysis API does not support having two different [org.jetbrains.kotlin.analysis.project.structure.KtSourceModule] with the same file system directory or intersecting files, it throws the error "Modules are inconsistent".
+ *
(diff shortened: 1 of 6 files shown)


whyoleg (commented):
LGTM.

I think, that it should be still an error, because if not, we will anyway fail during execution with K2, so error log message doesn't make sense to me, or do you have a reason why it could be better?

But, may be we need to add checker in K1 which will show warning message if something like this happens, to prepare users, that after switching to K2 current setup will not work? Or warning in KGP (from some version) will be enough?

IgnatBeresnev (approved):
:+1: :+1: 

I also believe it makes sense to throw an exception in this case, so that it fails with a user-understandable error instead of a very technical compiler error

And Oleg's idea also seems very useful: to log a warning in K1 that in K2 it will be an error, and to ask users who see this log warning in K1 to report their use case. Something like

> Source sets X and Y have common source roots: Z. Every Kotlin source file should belong to only one source set (module). This will become an error when Dokka migrates to K2-based code analysis. If you cannot fix this and you think you have a valid use case, please report it: https://github.com/Kotlin/dokka/issues/3239

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/SourceRootIndependentChecker.kt
```diff
@@ -0,0 +1,57 @@
...
+                // check source roots
+                val sourceRoot1 = sourceSets[i].sourceRoots.normalize()
+                val sourceRoot2 = sourceSets[j].sourceRoots.normalize()
+                val intersection = intersect(sourceRoot1, sourceRoot2)
+                if (intersection.isNotEmpty()) {
+                    messages += "Source sets '${sourceSets[i].displayName}' and '${sourceSets[j].displayName}' have the common source roots: ${intersection.joinToString()}. Every Kotlin source file should belong to only one source set (module)."
+                }
+            }
+        }
+        return PreGenerationCheckerOutput(messages.isEmpty(), messages)
```
IgnatBeresnev, line 39: Took me a while to figure out if / how / why Dokka would throw an exception in K2, but not in K1 (because the code looks the same), so leaving a message for Oleg in case he also doesn't see it right away: the first parameter of `PreGenerationCheckerOutput` is basically a `isSuccess` boolean, so Dokka fails if it's `false`.

IgnatBeresnev (approved):

PR #3476, opened 2024-01-29
Fix fastutil shadowing

whyoleg:
Fixes: https://github.com/Kotlin/dokka/issues/3329#issuecomment-1906072161

I've updated the description of the fix in files, where the workaround is applied.

Also there are two changes in CLI integration tests:
1. set `isTransitive = false` when building `shadowJar` which is used as `pluginsClasspath` in CLI - this way we can better control/understand which dependencies are provided to CLI. This will help to find issues with missing dependencies in future.
2. support for running tests with K2 with property - this was done to correctly test, that the the current fix works.

If needed, I can split CLI integration tests changes to separate PR.

### DIFF
--- a/dokka-integration-tests/cli/build.gradle.kts
+++ b/dokka-integration-tests/cli/build.gradle.kts
@@ -15,14 +15,18 @@ dependencies {
     implementation(projects.utilities)
 }
 
-/* Create a fat base plugin jar for cli tests */
-val basePluginShadow: Configuration by configurations.creating {
+// Configuration for plugins/dependencies required to run CLI with base plugin
+val cliPluginsClasspath: Configuration by configurations.creating {
     attributes {
         attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage::class.java, "java-runtime"))
     }
+
+    // we don't fetch transitive dependencies here to be able to control external dependencies explicitly
+    isTransitive = false
 }
 
-val cliConfiguration: Configuration by configurations.creating {
+// Configuration for CLI jar
+val cliClasspath: Configuration by configurations.creating {
     attributes {
         attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage::class.java, Usage.JAVA_RUNTIME))
         attribute(Bundling.BUNDLING_ATTRIBUTE, project.objects.named(Bundling::class.java, Bundling.SHADOWED))
@@ -32,22 +36,33 @@ val cliConfiguration: Configuration by configurations.creating {
 }
 
 dependencies {
-    cliConfiguration("org.jetbrains.dokka:runner-cli")
+    cliClasspath("org.jetbrains.dokka:runner-cli")
 
-    basePluginShadow("org.jetbrains.dokka:plugin-base")
+    cliPluginsClasspath("org.jetbrains.dokka:plugin-base")
+    // required dependencies of `plugin-base`
+    cliPluginsClasspath(libs.freemarker)
+    cliPluginsClasspath(libs.kotlinx.html)
 
-    // TODO [beresnev] analysis switcher
-    basePluginShadow("org.jetbrains.dokka:analysis-kotlin-descriptors") {
+    val tryK2 = project.providers
+        .gradleProperty("org.jetbrains.dokka.experimental.tryK2")
+        .map(String::toBoolean)
+        .orNull ?: false
+
+    val analysisDependency = when {
+        tryK2 -> "org.jetbrains.dokka:analysis-kotlin-symbols"
+        else -> "org.jetbrains.dokka:analysis-kotlin-descriptors"
+    }
+
+    cliPluginsClasspath(analysisDependency) {
         attributes {
             attribute(Bundling.BUNDLING_ATTRIBUTE, project.objects.named(Bundling::class.java, Bundling.SHADOWED))
         }
     }
 }
 
-val basePluginShadowJar by tasks.register("basePluginShadowJar", ShadowJar::class) {
-    configurations = listOf(basePluginShadow)
-    archiveFileName.set("fat-base-plugin-${project.version}.jar")
-    archiveClassifier.set("")
+val cliPluginsShadowJar by tasks.registering(ShadowJar::class) {
+    archiveFileName.set("cli-plugins-${project.version}.jar")
+    configurations = listOf(cliPluginsClasspath)
 
     // service files are merged to make sure all Dokka plugins
     // from the dependencies are loaded, and not just a single one.
@@ -55,10 +70,10 @@ val basePluginShadowJar by tasks.register("basePluginShadowJar", ShadowJar::clas
 }
 
 tasks.integrationTest {
-    dependsOn(cliConfiguration)
-    dependsOn(basePluginShadowJar)
+    dependsOn(cliClasspath)
+    dependsOn(cliPluginsShadowJar)
 
     inputs.dir(file("projects"))
-    environment("CLI_JAR_PATH", cliConfiguration.singleFile)
-    environment("BASE_PLUGIN_JAR_PATH", basePluginShadowJar.archiveFile.get())
+    environment("CLI_JAR_PATH", cliClasspath.singleFile)
+    environment("BASE_PLUGIN_JAR_PATH", cliPluginsShadowJar.get().archiveFile.get())
 }
--- a/dokka-integration-tests/cli/build.gradle.kts
+++ b/dokka-integration-tests/cli/build.gradle.kts
@@ -15,21 +15,21 @@ dependencies {
     implementation(projects.utilities)
 }
 
-// Configuration for plugins/dependencies required to run CLI with base plugin
 val cliPluginsClasspath: Configuration by configurations.creating {
+    description = "plugins/dependencies required to run CLI with base plugin"
     attributes {
-        attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage::class.java, "java-runtime"))
+        attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage.JAVA_RUNTIME))
     }
 
     // we don't fetch transitive dependencies here to be able to control external dependencies explicitly
     isTransitive = false
 }
 
-// Configuration for CLI jar
 val cliClasspath: Configuration by configurations.creating {
+    description = "dependency on CLI JAR"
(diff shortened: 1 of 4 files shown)


dokka-integration-tests/cli/build.gradle.kts
```diff
@@ -15,14 +15,18 @@ dependencies {
...
         attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage::class.java, "java-runtime"))
     }
+
+    // we don't fetch transitive dependencies here to be able to control external dependencies explicitly
+    isTransitive = false
 }
 
-val cliConfiguration: Configuration by configurations.creating {
+// Configuration for CLI jar
+val cliClasspath: Configuration by configurations.creating {
```
adam-enko, line 29:
```suggestion
val cliClasspath: Configuration by configurations.creating {
    description = "dependencies for CLI JAR"
```

Suggestion: use description to document the usage, so it's visible in more places.

dokka-integration-tests/cli/build.gradle.kts
```diff
@@ -15,14 +15,18 @@ dependencies {
     implementation(projects.utilities)
 }
 
-/* Create a fat base plugin jar for cli tests */
-val basePluginShadow: Configuration by configurations.creating {
+// Configuration for plugins/dependencies required to run CLI with base plugin
+val cliPluginsClasspath: Configuration by configurations.creating {
```
adam-enko, line 19:
```suggestion
val cliPluginsClasspath: Configuration by configurations.creating {
    description = "plugins/dependencies required to run CLI with base plugin"
```

Suggestion: use description to document the usage, so it's visible in more places.

dokka-subprojects/analysis-kotlin-descriptors/build.gradle.kts
```diff
@@ -47,5 +47,34 @@ tasks.shadowJar {
  *   and so such classes are not replaced afterward by `shadowJar` task - it visits every class once
  *
  */
-dependencies.shadow(libs.fastutil)
-tasks.shadowJar { exclude("it/unimi/dsi/fastutil/**") }
+
+// configuration for dependencies which we need to replace with original ones because `kotlin-compiler` minimizes them
+val shadowOverride: Configuration by configurations.creating {
+    attributes {
+        attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage::class.java, "java-runtime"))
```
adam-enko, line 54:
```suggestion
        attribute(Usage.USAGE_ATTRIBUTE, project.objects.named("java-runtime"))
```

use kotlin-dsl overload for `ObjectFactor.named()`

dokka-subprojects/analysis-kotlin-descriptors/build.gradle.kts
```diff
@@ -47,5 +47,34 @@ tasks.shadowJar {
  *   and so such classes are not replaced afterward by `shadowJar` task - it visits every class once
  *
  */
-dependencies.shadow(libs.fastutil)
-tasks.shadowJar { exclude("it/unimi/dsi/fastutil/**") }
+
+// configuration for dependencies which we need to replace with original ones because `kotlin-compiler` minimizes them
+val shadowOverride: Configuration by configurations.creating {
```
adam-enko, line 52:
```suggestion
val shadowOverride: Configuration by configurations.creating {
    description = "dependencies which we need to replace with original ones because `kotlin-compiler` minimizes them"
```

dokka-subprojects/analysis-kotlin-symbols/build.gradle.kts
```diff
@@ -110,6 +109,34 @@ tasks.shadowJar {
...
+    // and include `fastutil` from the result
+    configurations = listOf(project.configurations.runtimeClasspath.get())
+    exclude("it/unimi/dsi/fastutil/**")
+}
+
+tasks.shadowJar {
+    // override configurations to remove dependencies handled in `shadowJarDependencies`
+    configurations = emptyList()
+    from(shadowOverride, shadowDependenciesJar)
+}
```
adam-enko, line 142: It looks like this code is duplicated - should, or could, it be extracted into a convention plugin?

whyoleg: Yeah, the code is the same, but it's intentional at my side. While for now the issue is the same for both modules (with fastutil only); symbols and descriptors dependencies are rather different and even use different versions of Kotlin Compiler under the hood. So if something will change in future (in our build or in dependencies) it will be easier to fix.
This is rather a workaround for now than a long term solution, and we hope it will be fully removed for `symbols` module when Standalone Kotlin Analysis API will become stable

dokka-integration-tests/cli/build.gradle.kts
whyoleg: Thanks!

dokka-subprojects/analysis-kotlin-descriptors/build.gradle.kts
whyoleg: Thanks!

whyoleg: Updated manually to fix some other places

dokka-subprojects/analysis-kotlin-symbols/build.gradle.kts
adam-enko: Ah right, yes it makes sense that not everything will be duplicated. But the `shadowDependenciesJar` task and `shadowOverride` will be the same in each subproject? Could those definitions be moved to the existing `dokkabuild.publish-shadow.gradle.kts` plugin, and then each subproject would have independent workarounds, and add their own dependencies and exclusions?


E.g. add this to the convention plugin:

```kotlin
// dokkabuild.publish-shadow.gradle.kts

// ...

val shadowOverride: Configuration by configurations.creating {
  // ...
}

val shadowDependenciesJar by tasks.registering(ShadowJar::class) {
  //...
}

tasks.shadowJar {
    // override configurations to remove dependencies handled in `shadowJarDependencies`
    configurations = emptyList()
    from(shadowOverride, shadowDependenciesJar)
}
```

And then each subproject could set up things differently:

```kotlin
dependencies {
    shadowOverride(libs.fastutil)
}

tasks.shadowDependenciesJar {
    exclude("it/unimi/dsi/fastutil/**")
}
```

adam-enko: Up to you - I just wanted to clarify about what I think could be deduplicated, but perhaps it's not worth it :)

dokka-integration-tests/cli/build.gradle.kts
```diff
@@ -15,50 +15,65 @@ dependencies {
...
+        tryK2 -> "org.jetbrains.dokka:analysis-kotlin-symbols"
+        else -> "org.jetbrains.dokka:analysis-kotlin-descriptors"
+    }
 
-    // TODO [beresnev] analysis switcher
-    basePluginShadow("org.jetbrains.dokka:analysis-kotlin-descriptors") {
+    cliPluginsClasspath(analysisDependency) {
         attributes {
-            attribute(Bundling.BUNDLING_ATTRIBUTE, project.objects.named(Bundling::class.java, Bundling.SHADOWED))
+            attribute(Bundling.BUNDLING_ATTRIBUTE, project.objects.named(Bundling.SHADOWED))
```
adam-enko, line 58: 👍

adam-enko (approved):

dokka-subprojects/analysis-kotlin-symbols/build.gradle.kts
whyoleg: > But the shadowDependenciesJar task and shadowOverride will be the same in each subproject?

Who knows? :)
On current moment current workaround works for `fastutil` and we have no idea now, how other things will work.
So here Im more in favour of copy-paste workaround instead of generalising workaround based on one use-case.
Will wait for other reviews on this.

dokka-integration-tests/cli/projects/it-cli/src/main/kotlin/RootPackageClass.kt
```diff
@@ -0,0 +1,10 @@
+@file:Suppress("unused")
+
+/**
+ * A class that lives inside the root package
+ *
+ * <me@mail.com>
```
IgnatBeresnev, line 6: Some context sharing:

I'd add a comment that this input was added very intentionally and it's supposed to check for some corner cases, because someone might think it's just some random strings, and delete it / not adapt it when things are changing around here

```suggestion
 * <me@mail.com> - checks markdown parsing of emails and non-trivial cases like #3329, should compile
```

Something similar is done in https://github.com/Kotlin/dokka/blob/8e5c63d035ef44a269b8c43430f43f5c8eebfb63/dokka-integration-tests/gradle/projects/it-js-ir-0/src/main/kotlin/RootPackageClass.kt#L16-L26 after the test data became invalid and I had to find a subtitution (https://github.com/Kotlin/dokka/pull/2287#discussion_r846667111)

IgnatBeresnev (approved):
It's actually not as scary as I thought it would be :+1:

dokka-subprojects/analysis-kotlin-symbols/build.gradle.kts
IgnatBeresnev: I agree with Oleg here :+1:  I think it would be mega useful to extract this logic into a convention plugin if we had more than one use case for it that we expected to have long-term

This particular hack, however, ideally should only stay in K1 (if at all), and the K2 analysis shouldn't have this problem to begin with. We're waiting for K2's Analysis API to set up proper artifact publishing, and hopefully after [KT-61404](https://youtrack.jetbrains.com/issue/KT-61404/Analysis-API-implement-proper-library-publishing-structure) is implemented, we'll be able to remove this workaround from the `analysis-kotlin-symbols` build scripts 

So I wouldn't generalize this logic now, and wait until we stumble upon this problem again (which might be never, but might be next month)

dokka-integration-tests/cli/projects/it-cli/src/main/kotlin/RootPackageClass.kt
whyoleg: Yeah, sounds good, I thought about something like this, but was worrying that it will be to adhoc :)

vmishenev (approved):

PR #3477, opened 2024-01-29
Update task-selector workaround to use task rules

adam-enko:
Update the workaround for https://github.com/gradle/gradle/issues/22335 to use [task rules](https://docs.gradle.org/8.5/userguide/more_about_tasks.html#sec:task_rules).

[Discussed here.](https://github.com/Kotlin/dokka/pull/3427#discussion_r1463547746)

### Goals

* It should be more clear how to add/link tasks.
* Tasks should be easier to use, understand, run.
* tasks should follow conventions (specifically: `check` runs all verifications, `test` runs JVM verifications, `integrationTest` runs all integration tests)

### Details

* Create task rules for propagating tasks:
  * `includedBuildTasks_<TASK>` → runs <TASK> in included builds
  * `subprojectTasks_<TASK>` → runs <TASK> in subprojects (this shouldn't be invoked manually, only in by the `includedBuildTasks_` rule, otherwise it creates cyclical task dependencies).

* Update tasks:
  - check runs all validations
  - test runs apiCheck
  - introduce 'publishAllPublicationsToRemoteRepositories' lifecycle task

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild.base.gradle.kts
+++ b/build-logic/src/main/kotlin/dokkabuild.base.gradle.kts
@@ -28,3 +28,72 @@ val integrationTestPreparation by tasks.registering {
         "lifecycle task for preparing the project for integration tests (for example, publishing to the test Maven repo)"
     group = VERIFICATION_GROUP
 }
+
+
+val subprojectTasksPrefix = "subprojectTasks_"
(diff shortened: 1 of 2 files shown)


adam-enko:
Closing: this change isn't necessary.

PR #3478, opened 2024-01-30
Improve task cacheability

adam-enko:
- manually set [`cacheIf { true }`](https://docs.gradle.org/8.5/userguide/build_cache.html#using_the_runtime_api) on Sync/Exec tasks
  - by default Sync tasks are not cached because they are usually 'light' and just move files. But there are knock-on effects on later tasks if they are not cached.
  - Exec tasks aren't usually cached because they don't have well-defined outputs, but in our case they do
- use [path normalization](https://docs.gradle.org/8.5/userguide/build_cache_concepts.html#normalization) on Task file inputs
- add property names to task inputs (improves debugging/reporting)


adam-enko:
Split into separate PRs

* https://github.com/Kotlin/dokka/pull/3479
* https://github.com/Kotlin/dokka/pull/3480

PR #3479, opened 2024-01-30
improve NPM Gradle task caching

adam-enko:
- bump Gradle Node Plugin version
- fix npmInstall caching
- amend Gradle Node Plugin settings to allow npm (not Gradle) to manage `node_modules/` caching
- add path-sensitivity to task inputs, so cache can be re-used across machines
- add property names to task inputs (easier debugging/reporting)

This should improve build performance, and avoid an issue where npmInstall could not be up-to-date.

https://ge.jetbrains.com/s/rumw2jkk7scia/timeline?details=xhxfjcjtknpfe

build scan screenshot showing npmInstall warning: Not cacheable Overlapping outputs: Gradle does not know how file 'package-lock.json' was created (output property

### DIFF
--- a/dokka-subprojects/plugin-base-frontend/build.gradle.kts
+++ b/dokka-subprojects/plugin-base-frontend/build.gradle.kts
@@ -6,7 +6,6 @@ import com.github.gradle.node.npm.task.NpmTask
 import org.gradle.api.tasks.PathSensitivity.RELATIVE
 import org.jetbrains.kotlin.util.parseSpaceSeparatedArgs
 
-@Suppress("DSL_SCOPE_VIOLATION") // fixed in Gradle 8.1 https://github.com/gradle/gradle/pull/23639
 plugins {
     id("dokkabuild.setup-html-frontend-files")
     alias(libs.plugins.gradleNode)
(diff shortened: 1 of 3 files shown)


whyoleg (approved):

dokka-subprojects/plugin-base-frontend/build.gradle.kts
```diff
@@ -21,18 +22,36 @@ node {
 
 val distributionDirectory = layout.projectDirectory.dir("dist")
 
+tasks.npmInstall {
+    // enable caching - workaround for https://github.com/node-gradle/gradle-node-plugin/issues/81
+    outputs.file(layout.projectDirectory.file("node_modules/.package-lock.json"))
+        .withPropertyName("nodeModulesPackageLock")
+    outputs.cacheIf { true }
+}
```
adam-enko, line 30: This doesn't work as I expected, and additionally causes a problem because the NpmInstall task [already has a `nodeModulesPackageLock` output property](https://github.com/node-gradle/gradle-node-plugin/blob/c07b5ea719b757eabb4e3076741a2d069cc50a2d/src/main/kotlin/com/github/gradle/node/npm/task/NpmInstallTask.kt#L110) which clashes with this task (although sometimes it is null, so I think it only clashes sometimes?).


Ideally I'd want Gradle to skip the npmInstall task if npmRunBuild is UP-TO-DATE or FROM-CACHE, but I think that's something that needs to change in the Gradle Node Plugin.

Instead, I propose removing `node_modules/` from the clean task. That should improve performance locally.

ilgonmic (approved):

PR #3481, opened 2024-01-31
Gradle TestKit: reduce daemon timeouts

adam-enko:
Limit Gradle Workers:

- decreases the memory pressure 
- prevents daemons from previously failed build lingering on TeamCity agents, which then sabotage future builds by preventing directories from being deleted.

The aim is improved CI test consistency, which might increase overall test execution duration. I expect that the increased duration will be mitigated by other measures (aiming for enabling configuration-cache to allow for parallel test execution, and build-cache compliance allowing for avoiding re-running tests where possible).

### DIFF
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
@@ -13,6 +13,7 @@ import org.jetbrains.dokka.it.AbstractIntegrationTest
 import java.io.File
 import java.net.URI
 import kotlin.test.BeforeTest
+import kotlin.time.Duration.Companion.seconds
 
 abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
 
@@ -47,6 +48,12 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
                         "-P${TestEnvironment.TRY_K2}=true"
                     else
                         null,
+
+                    "-Porg.gradle.workers.max=1",
+                    // Decrease Gradle daemon idle timeout to prevent old agents lingering on CI.
+                    // A lower timeout means slower tests, which is preferred over OOMs and locked processes.
+                    "-Dorg.gradle.daemon.idletimeout=" + 10.seconds.inWholeMilliseconds, // default is 3 hours!
+                    "-Pkotlin.daemon.options.autoshutdownIdleSeconds=10",
 
                     * arguments
                 )
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
@@ -49,7 +49,7 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
                     else
                         null,
 
-                    "-Porg.gradle.workers.max=1",
+                    "-Dorg.gradle.workers.max=1",
                     // Decrease Gradle daemon idle timeout to prevent old agents lingering on CI.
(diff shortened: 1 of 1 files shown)


dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -48,6 +49,12 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
                     else
                         null,
 
+                    "-Porg.gradle.workers.max=1",
```
whyoleg, line 52: one question regarding all added properties: how will it affect local dev experience?
Locally most of the time I want to run just one integration test and so I have enough resources for it. Have you checked, will it slowdown (or maybe improve) local tests?

adam-enko: Good point, I'm not sure. It probably differs per machine. Could try enabling Build Scans and checking this branch vs master? I'll do the same.

whyoleg: Run all tests? one? which one? :)

adam-enko: This change should only affect the Gradle integration tests, so it should only be necessary to run those :)

whyoleg: I mean't in scope of integration tests (I don't want to run tests and leave a laptop for an hour :) )

Before each run I've deleted `build` folder in `integration-tests/gradle` to make tests cleaner. Additionally for serialisation I've done 2 tests (with and without cleaning)

I was expecting difference - but there is no - that's why I rechecked how tests are executed and found, that in both runs all workers were used :)
So I rechecked and found, that to limit workers, we need to set `-Dorg.gradle.workers.max=1` not `-Porg.gradle.workers.max=1` :)

MultiModule0IntegrationTest (no big changes in execution time ±30 seconds per test):
* master - https://ge.jetbrains.com/s/5mo4adfbqzrc4
* this branch - https://ge.jetbrains.com/s/kb6herwvfw3sq
* this branch + fix workers - https://ge.jetbrains.com/s/uda3yogtmdmka

SerializationGradleIntegrationTest (no big changes):
* master
  1. first run, after clean: https://ge.jetbrains.com/s/yhiadfmh7cs7m (2.5 min)
  2. no clean: https://ge.jetbrains.com/s/oupmjixkmikje (30 seconds)
* this branch
  1. first run, after clean: https://ge.jetbrains.com/s/a7z7buad4vrx6 (2.5 min)
  2. no clean: https://ge.jetbrains.com/s/d5e6vvdctpftk (30 seconds)
* this branch + fix workers
  1. first run, after clean: https://ge.jetbrains.com/s/b6zxbrmlqaau2 (4 min)
  2. no clean: https://ge.jetbrains.com/s/qb33fkt6dr3nu (50 seconds)

For me it's fine to reduce workers amount, but please check, that by doing so we will not increase integration tests time significantly

adam-enko, line 52:
```suggestion
                    "-Dorg.gradle.workers.max=1",
```

thanks for spotting this @whyoleg!

dkrasnoff (approved):

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
adam-enko: The local increase is significant, so I've removed `-Dorg.gradle.workers.max=1` from this PR and added it to the TeamCity K1 and K2 integration tests - hopefully that's sufficient for now.

whyoleg (approved):

PR #3483, opened 2024-02-02
Replace git submodules for test projects with custom GitCheckoutTask

adam-enko:
This improves the experience when running Dokka integration tests. 

* Automatically checking out the external projects used for integration tests means that anyone can checkout and run all tests without needed to setup and configure git submodules.
* Because the commitId is defined in code it's easier to see what version the external projects are used.

I have made the minimal number of changes in this PR to make it small and focused. I plan to replace the git-diff files with a less fragile alternative in a later PR.

### TODO

- [ ] Verify that the documentation is up to date.
    
   I thought there were some instructions on how to checkout and setup the Git submodules, but I couldn't find any. Do any docs need to be updated?

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
@@ -58,7 +58,7 @@ abstract class GitCheckoutTask @Inject constructor(
                 fs.sync {
                     from(temporaryDir)
                     into(destination)
-                    // exclude the .git dir to prevent the root Dokka repo getting confused with the nested subproject
+                    // exclude the .git dir to prevent the root git repo getting confused with a nested git repo
                     exclude(".git/")
                 }
 
--- a/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package dokkabuild.tasks
+
+import org.eclipse.jgit.api.Git
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.file.FileSystemOperations
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+import javax.inject.Inject
+
+/**
+ * Clones a remote Git repository into [GitCheckoutTask.destination].
+ */
+@CacheableTask
+abstract class GitCheckoutTask @Inject constructor(
+    private val fs: FileSystemOperations
+) : DefaultTask() {
+
+    @get:OutputDirectory
+    abstract val destination: DirectoryProperty
+
+    /** Public HTTP URI of the Git repo. */
+    @get:Input
+    abstract val uri: Property<String>
+
+    /** Specific git hash to checkout. */
+    @get:Input
+    abstract val commitId: Property<String>
+
+    init {
+        group = "git checkout"
+    }
+
+    @TaskAction
+    fun action() {
+        val uri = uri.get()
+        val commitId = commitId.get()
+
+        fs.delete { delete(temporaryDir) }
+
+        Git.cloneRepository()
+            .setURI(uri)
+            .setDirectory(temporaryDir)
+            .call().use { git ->
+                git.pull().call()
+
+                git.checkout()
+                    .setName(commitId)
+                    .call()
+
+
+                fs.sync {
+                    from(temporaryDir)
+                    into(destination)
+                    // exclude the .git dir to prevent the root Dokka repo getting confused with the nested subproject
+                    exclude(".git/")
+                }
+
+                logger.lifecycle("initialized git repo $uri in ${destination.asFile.orNull}")
+            }
+    }
+}
--- a/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
@@ -4,6 +4,12 @@
 package dokkabuild.tasks
 
 import org.eclipse.jgit.api.Git
+import org.eclipse.jgit.api.ResetCommand.ResetType.HARD
+import org.eclipse.jgit.lib.NullProgressMonitor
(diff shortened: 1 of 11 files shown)


build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        fs.delete { delete(temporaryDir) }
+
+        Git.cloneRepository()
+            .setURI(uri)
+            .setDirectory(temporaryDir)
+            .call().use { git ->
+                git.pull().call()
+
+                git.checkout()
```
whyoleg, line 53: is it intentional that we download full repository and then checkout specific commit instead of just downloading specific commit (head=1 or something like this)?

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
```diff
@@ -0,0 +1,69 @@
...
+                    .setName(commitId)
+                    .call()
+
+                fs.sync {
```
whyoleg, line 57: minor: may be we can move `fs.sync` outside of `use { git -> ... }` block to better understand steps here?

whyoleg (approved):

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
adam-enko: That would definitely be an improvement. If you can figure out how to do that with JGit I would be grateful and impressed 😄

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
adam-enko: sure, makes sense to me

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
```diff
@@ -0,0 +1,69 @@
...
+                    from(temporaryDir)
+                    into(destination)
+                    // exclude the .git dir:
+                    // - prevent the root git repo getting confused with a nested git repo
+                    // - improves Gradle caching
+                    exclude(".git/")
+                }
+
+                logger.lifecycle("initialized git repo $uri in ${destination.asFile.orNull}")
+            }
```
adam-enko, line 67:
```suggestion
        Git.cloneRepository()
            .setURI(uri)
            .setDirectory(temporaryDir)
            .call().use { git ->
                git.pull().call()

                git.checkout()
                    .setName(commitId)
                    .call()
            }

            fs.sync {
                from(temporaryDir)
                into(destination)
                // exclude the .git dir:
                // - prevent the root git repo getting confused with a nested git repo
                // - improves Gradle caching
                exclude(".git/")
            }

            logger.lifecycle("initialized git repo $uri in ${destination.asFile.orNull}")
```

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
adam-enko: Maybe `reset --hard $commitId` is all that's needed?

```kotlin
Git.cloneRepository()
    .setURI(uri)
    .setDirectory(temporaryDir)
    .call().use { git ->
        git.reset()
            .setMode(HARD)
            .setRef(commitId)
            .call()
    }
```

whyoleg: maybe? :) 
```kotlin
        Git.cloneRepository()
            .setNoCheckout(true)
            .setURI(uri)
            .setDirectory(temporaryDir)
            .call().use { git ->
                git.checkout().setName(commitId).call()
            }
```

There is also `setDepth()` available on `cloneRepository` in 6.x, but not in 5.x version of jgit
Still, it should be fine any way, as it should still do this fast :)

whyoleg (approved):

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
```diff
@@ -0,0 +1,116 @@
...
+                .setName(commitId)
+                .call()
+
+            // git reset --hard
+            git.reset()
+                .setProgressMonitor(gitOperationsPrinter)
+                .setMode(HARD)
+                .call()
+
+            // git clean -fdx
```
Tapchicoma, line 108: Probably I should add the same for Kotlin build regression benchmarks :+1:

Tapchicoma (approved):

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
```diff
@@ -0,0 +1,116 @@
...
+            Git.cloneRepository()
+                .setProgressMonitor(gitOperationsPrinter)
+                .setNoCheckout(true)
+                .setURI(uri)
+                .setDirectory(localRepoDir)
+                .call()
+        }
+
+        repo.use { git ->
+            git.checkout()
```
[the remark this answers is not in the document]
adam-enko: Sure thing, I've added some docs. Please let me know if they make sense and help - I think it will also answer [your question above](https://github.com/Kotlin/dokka/pull/3483#discussion_r1478077283)

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
adam-enko: which bit? :) My changes followed your example

https://github.com/JetBrains/kotlin/blob/f262247b2bcd7cc0e9d64d274f2fd412031187ab/libraries/tools/gradle/regression-benchmark-templates/src/main/kotlin/org/jetbrains/kotlin/gradle/benchmark/BenchmarkTemplate.kt#L133-L145

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
adam-enko: I discussed with @Tapchicoma and he showed me a similar process in `kotlin.git` - see https://github.com/JetBrains/kotlin/blob/f262247b2bcd7cc0e9d64d274f2fd412031187ab/libraries/tools/gradle/regression-benchmark-templates/src/main/kotlin/org/jetbrains/kotlin/gradle/benchmark/BenchmarkTemplate.kt#L133-L145

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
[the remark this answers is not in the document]
whyoleg: thanks! now everything is clear!

whyoleg (approved):

PR #3485, opened 2024-02-03
Build script tweaks

3flex:
Pretty low value stuff, but it tidies things up a little here and there.

Easiest to review commit by commit.


adam-enko:
Hi @3flex, thanks for the contribution. 

I've had a quick scan through and I've applied some of these changes in my own PRs. It makes sense to split them into a separate PR, so thanks for creating this.

Updating the scripts to use the new Kotlin Assignment feature is a more notable change however. Dokka needs to support a wide range of Gradle versions, and using Kotlin Assignment requires Gradle 8.2+. We can use Kotlin Assignment in our own build scripts, but not in the tested projects, and probably not in the docs (but I'll check further).

dokka-runners/dokkatoo/buildSrc/build.gradle.kts
```diff
@@ -14,6 +14,6 @@ dependencies {
 
 java {
   toolchain {
-    languageVersion.set(JavaLanguageVersion.of(11))
+    languageVersion = JavaLanguageVersion.of(11)
```
adam-enko, line 17: This version of Dokkatoo isn't connected to anything at the moment, and probably won't be worked on for a while. While editing it won't hurt, it also isn't necessary, so can you remove all of the changes in the `dokka-runners/dokkatoo` dir please?

build-logic/src/main/kotlin/dokkabuild.java.gradle.kts
```diff
@@ -26,9 +26,9 @@ tasks.withType<Test>().configureEach {
         (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
```
adam-enko, line 26:
```suggestion
        (Runtime.getRuntime().availableProcessors() / 2).coerceAtLeast(1)
```

Some more tidying, since we're here.

build-logic/src/main/kotlin/dokkabuild.base.gradle.kts
```diff
@@ -23,7 +23,7 @@ tasks.withType<AbstractArchiveTask>().configureEach {
     isReproducibleFileOrder = true
 }
 
-val integrationTestPreparation by tasks.registering {
+tasks.register("integrationTestPreparation") {
```
adam-enko, line 26: Can you explain more about this change? Personally I prefer the delegated accessor version over the string-based version.

adam-enko (dismissed):

build-logic/src/main/kotlin/dokkabuild.base.gradle.kts
3flex: It's just shorter and simpler. If you prefer the other way I'll revert.

whyoleg:
Hey @3flex !
Do you still willing to work on this PR?
If so, could you please:
* rebase on latest master (some changes could be applied already there by @adam-enko)
* revert changes of `.set` to `=` in integration tests projects - tests for those projects are executed with different Gradle versions where property assignment via `=` doesn't exist
* revert changes in `dokkatoo` - as stated above - this project is not connected for now

whyoleg (approved):
Thank you!
Sorry for the long review, somehow missed the notification :)

PR #3486, opened 2024-02-05
Remove no-op `parallelMapNotNull { it }`

adam-enko:
I presume that `parallelMapNotNull { it }` does nothing.

Alternatively: is it supposed to do something, but it doesn't?


IgnatBeresnev (approved):
👍 I think it was parsing sources within `parallelMapNotNull` before, but during the huge analysis refactoring the parsing itself was extracted up, with this left behind

whyoleg (approved):

PR #3489, opened 2024-02-06
Update integration-tests README

adam-enko:
Update instructions for integration tests

* move README up a level, because it's relevant for Maven and CLI too
* Git submodules were removed in https://github.com/Kotlin/dokka/pull/3483, so manual management is no longer required
* update instructions


dokka-integration-tests/gradle/README.md
```diff
@@ -8,21 +8,6 @@ locally (`./gradlew publishToMavenLocal`).
 
 To debug Gradle tests, the environment variable `ENABLE_DEBUG=true` should be declared.
 
-### To update git submodules 
-
-Integration tests have fixed git revision number, with the diff patch applied from the corresponding file (e.g. [`coroutines.diff`](projects/coroutines/coroutines.diff)).
```
IgnatBeresnev, line 13: While having outdated documentation could be worse than having none at all, it'd be nice to describe the current process of updating the versions of external IT projects here, similarly to how it's done in this block. 

Some context: the process of updating the tested coroutines/serialization projects isn't straightforward, you have to know not only how these tests work, but also where to look for these hardcoded values and diffs. While the members of the Dokka team now how to do all that, there are cases when people outside of the Dokka team need to update these (as Dokka and its tests are sometimes used for reproducing bugs in other projects), and in such cases it's convenient to point people to an instruction of some sort, like this one

dokka-integration-tests/README.md
```diff
@@ -0,0 +1,55 @@
...
+    val checkoutKotlinxSerialization by tasks.registering(GitCheckoutTask::class) {
+        uri = "https://github.com/Kotlin/kotlinx.serialization.git"
+        commitId = "<new-commit-sha-here>"
+        destination = templateProjectsDir.dir("serialization/kotlinx-serialization")
+    }
+    ```
+   Note that the Gradle task will clone the repo into a `build/tmp` directory
+   before copying it to a subdirectory inside `projects/`
+
+3. Manually write the diff (or apply the existing one and tweak) to have the project buildable against locally published Dokka of version `for-integration-tests-SNAPSHOT`
```
whyoleg, line 42: AFAIR there is no `for-integration-tests` version anymore and we just use `project.version` (currently 2.0.0 for main branch)
Looks like we can cleanup it a bit, while we are here

whyoleg (approved):

IgnatBeresnev (approved):
Awesome, even better than before - thank you!

PR #3492, opened 2024-02-09
Refactor Integration Test project setup steps

adam-enko:
### Summary

- Gradle projects:
  - deduplicate project copying code into `AbstractGradleIntegrationTest`
  - move custom repos into `template.settings.gradle.kts` & remove redundant `template.root.gradle.kts`
  - pass `template.settings.gradle.kts` as a Gradle task input, so the project-copying code can use it
  - added exclusive filter for MavenLocal in `template.settings.gradle.kts`, so Gradle only searches MavenLocal for Dokka
- refactor `copyAndApplyGitDiff()` to try and make it a bit more legible

### Verification

To test locally I recommend enabling configuration cache, as this will make the tasks run in parallel

```
./gradlew integrationTest --configuration-cache --configuration-cache-problems=warn
```


dokka-integration-tests/gradle/src/testExternalProjectKotlinxCoroutines/kotlin/CoroutinesGradleIntegrationTest.kt
```diff
@@ -38,12 +38,12 @@ class CoroutinesGradleIntegrationTest : AbstractGradleIntegrationTest(), TestOut
     override val projectOutputLocation: File by lazy { File(projectDir, "build/dokka/htmlMultiModule") }
 
     @BeforeTest
-    fun prepareProjectFiles() {
-        val templateProjectDir = File("projects", "coroutines/kotlinx-coroutines")
-        templateProjectDir.listFiles().orEmpty()
-            .forEach { topLevelFile -> topLevelFile.copyRecursively(File(projectDir, topLevelFile.name)) }
-
-        copyAndApplyGitDiff(File("projects", "coroutines/coroutines.diff"))
+    override fun beforeEachTest() {
```
whyoleg, line 41: Is there a reason why serialisation and coroutines setup differs?
In serialisation we don't override this method, but introduce another one?

whyoleg (approved):

whyoleg:
I haven't run changes locally, but looks legit :)

dokka-integration-tests/gradle/src/testExternalProjectKotlinxCoroutines/kotlin/CoroutinesGradleIntegrationTest.kt
adam-enko: Good spot! I've made both setup tasks work the same

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -101,6 +102,13 @@ testing {
 
             targets.configureEach {
                 testTask.configure {
+
+                    inputs.file(templateSettingsGradleKts)
+                    systemProperty(
```
dkrasnoff, line 107: If I understand correctly, we set a system property, which is used in `AbstractGradleIntegrationTest`. However, it is not clear to me where we set the second `templateProjectDir` system property.

dkrasnoff (approved):

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Good point, it's not clear.

I've moved `inputs.file(templateSettingsGradleKts)` to the `registerTestProjectSuite()` util function, so it's next to the `templateProjectDir` input.

This also triggered me to add property names and path sensitivity.

Thanks!

PR #3493, opened 2024-02-09
update developer docs build config

adam-enko:
Update MkDocs tasks to improve compatibility with Configuration Cache.

- use `com.pswidersk.python-plugin` (although it's less capable than `ru.vyarus.mkdocs` and requires more manual setup, `ru.vyarus.mkdocs` is [not Configuration Cache compatible](https://github.com/xvik/gradle-mkdocs-plugin/issues/23))
- Specify requirements.txt (helps with reproducible builds, Dependabot)
- Specify the output directory `site_dir: build/mkdocs` explicitly. Move config and site files so they're not in a subdirectory, and so the generated `site_dir` doesn't have to use confusing relative paths)
- Added dark/light mode toggle (an unrelated change, but one that helped my eyes during development)
- Define tasks for installing & building mkdocs.
- Update README with the new tasks.
- If possible, log link to site using IntelliJ built-in server (Because the 'serve' command doesn't work with `com.pswidersk.python-plugin` due to a Gradle issue that doesn't kill spawned processes when the Gradle build ends.)

The output of the build is still in the same location, so the only modification to the GitHub Actions is to the task names.

## TODO

- [x] Re-implement multiversion support & redirecting index.html


docs-developer/README.md
```diff
@@ -23,18 +23,22 @@ Alternatively, you can use Docker:
...
 
-This will build the docs and start a web server under [localhost:8000/Kotlin/dokka](http://localhost:8000/Kotlin/dokka/).
+This will build the docs and start a web server under
+[localhost:8000/Kotlin/dokka](http://localhost:8000/Kotlin/dokka/).
 
-### Livereload server
+### Live-reload server
 
-Alternatively, you can run a livereload server that automatically rebuilds documentation on every change:
+If you are using IntelliJ then a link to the docs will be logged in the console:
```
IgnatBeresnev, line 31: If the link is opened using IntelliJ's built in web server, it was surprising for me to find that it is only available in IntelliJ Ultimate, but not in community, so I think it makes sense to clarify it here if that's the case. Also might wanna link to https://www.jetbrains.com/help/idea/editing-html-files.html#ws_html_preview_output

I found out after recommending it to one of Dokka's users, and they spent quite some time looking for how to do it only to realize they don't have that feature :[

docs-developer/build.gradle.kts
```diff
@@ -1,16 +1,98 @@
...
+import org.gradle.api.tasks.PathSensitivity.RELATIVE
+import org.jetbrains.kotlin.util.parseSpaceSeparatedArgs
 
 plugins {
-    id("ru.vyarus.mkdocs") version "2.4.0"
+    id("com.pswidersk.python-plugin") version "2.4.0"
 }
 
-if (project.version.toString().endsWith("-SNAPSHOT")) {
-    // Do not generate the root index.html file with the redirect
```
IgnatBeresnev, line 10: Unfortunately, I can't test it myself now and don't wanna delay the review, so I assume you've understood what this hack does and the new plugin does everything in the same way as the old one: 
* For non-snapahot versions, it should build and commit a directory with the version name and create a root index.html file that redirects to it, so that `kotlin.github.io/dokka` always redirects to the latest docs
* For snapshot versions, it should just build and commit a directory with the version name, no redirects (useful for when users use dev/rc/preview versions and for us to review upcoming changes)

docs-developer/requirements.txt
```diff
@@ -0,0 +1,28 @@
...
+click==8.1.7
+colorama==0.4.6
+ghp-import==2.1.0
+idna==3.6
+Jinja2==3.1.3
+Markdown==3.5.2
+MarkupSafe==2.1.5
+mergedeep==1.3.4
+mkdocs==1.5.3
+mkdocs-material==9.5.10
```
IgnatBeresnev, line 13: If we want to update the version of mkdocs, I assume we need to bump `mkdocs-*` versions, but when it comes to the rest of dependencies here - how do we figure out which ones need updating as well, and to which versions? If it was copy-pasted from mkdocs documentation/examples or you somehow generated it, I think it's worth mentioning it in the README while you still remember :)

IgnatBeresnev (approved):

docs-developer/build.gradle.kts
adam-enko: Ahh okay, I missed that. I saw that the redirecting index.html was a ru.vyarus.mkdocs specific issue, and I assumed that Dokka config was disabling this feature.

It sounds like a pretty simple algorithm to set up in Gradle logic, so I'll set it up. It's a shame that ru.vyarus.mkdocs is incompatible with Config Cache though, because less custom code is better!

docs-developer/requirements.txt
adam-enko: This file was generated by running `pip freeze` (via the handy task `./gradlew :docs-developer:pipPrintRequirements`).

I've forgotten how Python works so I'm not sure how to document the correct procedure for updating Python dependencies :)

docs-developer/build.gradle.kts
adam-enko: Implemented, and tested by manually changing the version in `gradle.properties`, running `./gradlew buildMkDocsSite`, and checking in `docs-developer/build/mkdocs/`

#### 2.0.0 

image


#### 2.0.0-SNAPSHOT

image

docs-developer/README.md
adam-enko: I thought the built-in server was available in IJ Community, at least the code is there https://github.com/JetBrains/intellij-community/blob/8685843f6ad2a45ef772bf0cdf34590036a1d9e8/platform/built-in-server/src/org/jetbrains/builtInWebServer/BuiltInServerOptions.java

adam-enko: I've just installed IJ Community and the built-in server link works.

IgnatBeresnev: 0_o I definitely remember having problems with it in community, and the linked documentation page says "Ultimate" at the top

That's good to know, thanks for checking!

IgnatBeresnev (approved):

docs-developer/build.gradle.kts
```diff
@@ -1,16 +1,179 @@
...
-    // to a snapshot version, otherwise GitHub Pages-based documentation
-    // will always lead to the non-yet-released documentation.
-    // For more details, see https://github.com/Kotlin/dokka/issues/2869.
-    // For configuration details, see https://xvik.github.io/gradle-mkdocs-plugin/3.0.0/examples/#simple-multi-version.
-    mkdocs.publish.rootRedirect = false
+val dokkaVersion = provider { project.version.toString() }
+val isDokkaSnapshotVersion = dokkaVersion.map { it.endsWith("-SNAPSHOT") }
```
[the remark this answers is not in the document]
adam-enko: I believe Dokka uses SemVer, so using `KotlinToolingVersion` should work, but I'm wary about using an external class for a simple check, but maybe I'm missing something. What benefits are there in using `KotlinToolingVersion`?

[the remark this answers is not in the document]
dkrasnoff: If we have the same approach of naming versions for all Kotlin-related products, then having a single point of truth would be better. In this case, we will catch the problems in all related products once we change the naming approach.
Even if you are distant from the Kotlin project, I hope that using a single util to check versions in different places is a better solution than using some different ones in different places.

dkrasnoff (approved):

docs-developer/build.gradle.kts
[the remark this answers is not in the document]
dkrasnoff: Anyway I'm okay with the current changes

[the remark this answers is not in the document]
adam-enko: Oh, by "dokka's version aligns with kotlin's version", did you mean that Kotlin and Dokka should have the exact same version?

I _think_ it was the original idea that Dokka would be lock-step with Kotlin, so each Kotlin release would have a Dokka release with the exact same version. But I think that's changed.

wdyt @IgnatBeresnev?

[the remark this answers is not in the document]
IgnatBeresnev: > I think it was the original idea that Dokka would be lock-step with Kotlin, so each Kotlin release would have a Dokka release with the exact same version. But I think that's changed.

It's true for now with the classic Gradle plugin, but we might need to separate the DGP and the Engine versions once the new DGP is out, yeah.  

I don't mind the proposed idea with `KotlinToolingVersion`, it's easy enough to revert / fix if it gets in the way, so it's your call :)

adam-enko:
The developer docs for 2.0.0-SNAPSHOT have been published correctly https://kotlin.github.io/dokka/2.0.0-SNAPSHOT/developer_guide/introduction/

Issue #1030, opened 2020-06-29
Extension rendering tests


Kordyjan:
The goal of this task is to create tests that check if all required extensions of type are properly rendered for different kinds of types.

AbhishekPSingh07:
Hii I would like to give it a go I have experience in kotlin language but haven't had any chances to work with unit test and i am new to this codebase and still going through the documentation

AbhishekPSingh07:
Could you guide me on how to approach this issue

IgnatBeresnev:
@AbhishekPSingh07 hi, thanks for showing interest!

To write unit tests, you don't need anything specific. Just fork the project, clone it locally, make sure you have Java 11 installed, and open it in IntelliJ IDEA. Everything should get imported and just work automatically.

Navigate to [SignatureTest](https://github.com/Kotlin/dokka/blob/0e00edc6fcd406fcf38673ef6a2f8f59e8374de2/plugins/base/src/test/kotlin/signatures/SignatureTest.kt) in `dokka/plugins/base/src/test/kotlin/signatures` - this is where the tests need to be written.

Your tests should look roughly like the test named [`fun with keywords, params and generic bound`](https://github.com/Kotlin/dokka/blob/0e00edc6fcd406fcf38673ef6a2f8f59e8374de2/plugins/base/src/test/kotlin/signatures/SignatureTest.kt#L227-L228), you can use it as a starting point and as inspiration. Have a look at other tests in that file too, they should be just as good.

Dokka has some special test DSL, so some of the things you'll see are project-specific, but if you use the test I mentioned above as an example, you should be fine :) The idea is that you write the source code you want to test, as if you were writing it in an IDE, and then the test framework generates Dokka HTML pages for that source code, and you can verify HTML itself, make sure everything has been rendered correctly.

Some gotchas you might encounter:
* Make sure you write asserts inside `renderingStage = { _, _ ->`  in `testInline` 
* Make sure you change the name of the HTML file in `renderedContent("root/example/simple-fun.html")`, it needs to be the name of your function/class/property from the source code.
* If you use some incorrect, unexpected or excessive assert tags inside `match`, the error might not be obvious - it'll complain about collection size mismatch or something like that. Or that a certain tag was expected, but not found. This basically means you either missed or added unnecessary tags.
* Make sure `ignoreSpanWithTokenStyle = true` is used within `match`. It ignores very simple or stylistic tags that don't make much of a difference in this case, but it simplifies assert tags by a lot.

[...]

PR #3497, opened 2024-02-16
Issue #1030 Extension rendering tests

YohannKovacs:
Solving issue #1030 
Added rendering tests for extension methods with
- nullables
- variable args
- default args
- generics
- infix function
- lambdas as params
- property extensions

Let me know if there is anything to improve/add.


IgnatBeresnev (approved):
Well done with the test cases! I don't think I could've done it better myself :)

Thank you for the contribution, much appreciated!

PR #3516, opened 2024-03-05
Bump Kotlin versions to 2.0.0-RC1

IgnatBeresnev:
* Bumps KGP to 2.0.0-RC1
* Bumps kotlin-compiler to 2.0.0-RC1 (only K1)

### DIFF
--- a/dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
+++ b/dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
@@ -2,6 +2,6 @@
 # Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
 #
 
-dokka_it_kotlin_version=2.0.0-Beta5
+dokka_it_kotlin_version=2.0.0-RC1
 dokka_it_android_gradle_plugin_version=4.2.2
 android.useAndroidX=true
--- a/dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
+++ b/dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
@@ -2,6 +2,6 @@
 # Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
 #
 
-dokka_it_kotlin_version=1.9.22
+dokka_it_kotlin_version=2.0.0-Beta4
 dokka_it_android_gradle_plugin_version=4.2.2
 android.useAndroidX=true
--- a/dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
+++ b/dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
@@ -2,6 +2,6 @@
 # Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
 #
 
-dokka_it_kotlin_version=2.0.0-Beta4
+dokka_it_kotlin_version=2.0.0-Beta5
 dokka_it_android_gradle_plugin_version=4.2.2
 android.useAndroidX=true
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
@@ -41,6 +41,24 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
         vararg arguments: String,
         jvmArgs: List<String> = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=1G")
     ): GradleRunner {
+
+        // TODO quick hack to add `android { namespace }` on AGP 7+ (it's mandatory in 8+).
+        //      This hack could be made prettier, or only test AGP 7+
+        val androidMajorVersion = buildVersions.androidGradlePluginVersion
+            ?.substringBefore(".")
+            ?.toIntOrNull() ?: 0
+        if (androidMajorVersion >= 7) {
+            projectDir.resolve("build.gradle.kts").appendText(
+                """
+                |
+                |android {
+                |    namespace = "org.jetbrains.dokka.it.android"
+                |}
+                |
+            """.trimMargin()
+            )
+        }
+
         return GradleRunner.create()
             .withProjectDir(projectDir)
             .forwardOutput()
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
@@ -14,7 +14,7 @@ open class AllSupportedTestedVersionsArgumentsProvider : TestedVersionsArguments
 
 object TestedVersions {
 
-    val LATEST = BuildVersions("8.6", "2.0.0-Beta4")
+    val LATEST = BuildVersions("8.6", "2.0.0-Beta5")
 
     /**
      * All supported Gradle/Kotlin versions, including [LATEST]
@@ -41,7 +41,7 @@ object TestedVersions {
     val ANDROID =
         BuildVersions.permutations(
             gradleVersions = listOf("8.4"),
-            kotlinVersions = listOf("2.0.0-Beta4"),
+            kotlinVersions = listOf("2.0.0-Beta5"),
             androidGradlePluginVersions = listOf("8.3.0")
         ) + BuildVersions.permutations(
             gradleVersions = listOf("7.4.2", *ifExhaustive("7.0")),
@@ -66,7 +66,7 @@ object TestedVersions {
         "1.9.0" to "18.2.0-pre.597",
         "1.9.10" to "18.2.0-pre.597",
         "1.9.23" to "18.2.0-pre.682",
-        "2.0.0-Beta4" to "18.2.0-pre.713",
+        "2.0.0-Beta5" to "18.2.0-pre.726",
     )
 }
 
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
@@ -14,7 +14,7 @@ open class AllSupportedTestedVersionsArgumentsProvider : TestedVersionsArguments
 
 object TestedVersions {
 
-    val LATEST = BuildVersions("7.6.2", "1.9.22")
+    val LATEST = BuildVersions("8.5", "2.0.0-Beta4")
 
     /**
      * All supported Gradle/Kotlin versions, including [LATEST]
@@ -24,7 +24,7 @@ object TestedVersions {
     val ALL_SUPPORTED =
         BuildVersions.permutations(
             gradleVersions = listOf("7.6.2"),
-            kotlinVersions = listOf("1.9.10", "1.8.20", "1.7.20", "1.6.21", "1.5.31"),
+            kotlinVersions = listOf("1.9.23", "1.8.20", "1.7.20", "1.6.21", "1.5.31"),
         ) + BuildVersions.permutations(
             gradleVersions = listOf(*ifExhaustive("7.0", "6.1.1")),
             kotlinVersions = listOf(*ifExhaustive( "1.8.0", "1.7.0", "1.6.0", "1.5.0"))
@@ -40,6 +40,10 @@ object TestedVersions {
      */
     val ANDROID =
         BuildVersions.permutations(
+            gradleVersions = listOf("8.4"),
+            kotlinVersions = listOf("2.0.0-Beta4"),
+            androidGradlePluginVersions = listOf("8.3")
+        ) + BuildVersions.permutations(
             gradleVersions = listOf("7.4.2", *ifExhaustive("7.0")),
             kotlinVersions = listOf("1.7.20", "1.6.21", "1.5.31", "1.4.32"),
             androidGradlePluginVersions = listOf("7.2.0")
@@ -47,7 +51,7 @@ object TestedVersions {
             gradleVersions = listOf("6.9", *ifExhaustive("6.1.1", "5.6.4")),
             kotlinVersions = listOf("1.8.0", "1.7.0", "1.6.0", "1.5.0", "1.4.0"),
             androidGradlePluginVersions = listOf("4.0.0", *ifExhaustive("3.6.3"))
-        ) + LATEST
+        )
 
     // https://mvnrepository.com/artifact/org.jetbrains.kotlin-wrappers/kotlin-react
     val KT_REACT_WRAPPER_MAPPING = mapOf(
@@ -63,6 +67,7 @@ object TestedVersions {
         "1.9.10" to "18.2.0-pre.597",
         "1.9.20" to "18.2.0-pre.635",
         "1.9.22" to "18.2.0-pre.682",
+        "2.0.0-Beta4" to "18.2.0-pre.713",
     )
 }
 
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
@@ -14,7 +14,7 @@ open class AllSupportedTestedVersionsArgumentsProvider : TestedVersionsArguments
 
 object TestedVersions {
 
-    val LATEST = BuildVersions("8.6", "2.0.0-Beta5")
+    val LATEST = BuildVersions("8.7", "2.0.0-RC1")
 
     /**
      * All supported Gradle/Kotlin versions, including [LATEST]
@@ -41,7 +41,7 @@ object TestedVersions {
     val ANDROID =
         BuildVersions.permutations(
             gradleVersions = listOf("8.4"),
-            kotlinVersions = listOf("2.0.0-Beta5"),
+            kotlinVersions = listOf("2.0.0-RC1"),
             androidGradlePluginVersions = listOf("8.3.0")
         ) + BuildVersions.permutations(
             gradleVersions = listOf("7.4.2", *ifExhaustive("7.0")),
@@ -66,7 +66,7 @@ object TestedVersions {
         "1.9.0" to "18.2.0-pre.597",
         "1.9.10" to "18.2.0-pre.597",
         "1.9.23" to "18.2.0-pre.682",
-        "2.0.0-Beta5" to "18.2.0-pre.726",
+        "2.0.0-RC1" to "18.2.0-pre.726",
     )
 }
 
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
@@ -14,7 +14,7 @@ open class AllSupportedTestedVersionsArgumentsProvider : TestedVersionsArguments
 
 object TestedVersions {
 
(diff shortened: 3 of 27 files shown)


gradle/libs.versions.toml
```diff
@@ -1,27 +1,24 @@
 [versions]
 
-gradlePlugin-kotlin = "1.9.22"
+gradlePlugin-kotlin = "2.0.0-Beta4"
 # See: https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin
-gradlePlugin-android = "4.2.2"
-gradlePlugin-android-dokkatoo = "8.0.2"
+gradlePlugin-android = "7.1.3"
```
IgnatBeresnev, line 5: > gradlePlugin-android = "4.2.2"
> gradlePlugin-android-dokkatoo = "8.0.2"

These are unused at the moment, so I've removed them (and the dependencies that depend on them) to reduce noise. Can be added once it's needed

> gradlePlugin-android = "7.1.3"
KGP 2.0.0 now requires AGP 7.1.3+, so I had to bump it as well

gradle/libs.versions.toml
```diff
@@ -1,27 +1,24 @@
...
 gradlePlugin-dokka = "1.9.20"
-gradlePlugin-gradle-customUserData = "1.12"
-gradlePlugin-gradle-enterprise = "3.14.1"
```
IgnatBeresnev, line 9: I think it makes sense to have the most important / frequently updated dependencies on the top, and it's unlikely that Gradle enterprise will be looked for / updated all that often. Just moved it closer to the end to reduce noise

gradle/libs.versions.toml
```diff
@@ -1,27 +1,24 @@
...
-kotlin-compiler = "1.9.22"
+kotlin-compiler = "2.0.0-Beta4"
 kotlin-compiler-k2 = "2.0.0-dev-8561"
 
 # MUST match the version of the intellij platform used in the kotlin compiler,
 # otherwise this will lead to different versions of psi API and implementations
 # on the classpath and will fail with hard to debug problems in runtime.
 # See: https://github.com/JetBrains/kotlin/blob/e6633d3d9214402fcf3585ae8c24213a4761cc8b/gradle/versions.properties#L1
-intellij-platform = "213.7172.25"
+intellij-platform = "213.7172.53"
```
IgnatBeresnev, line 21: IntelliJ platform was bumped in 2.0.0-Beta4

https://github.com/JetBrains/kotlin/blob/9554232c35d628b5948c23c1b96dde89bcab074a/gradle/versions.properties#L1

dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/AndroidAutoConfigurationTest.kt
```diff
@@ -14,22 +14,27 @@ import kotlin.test.*
...
+            project.plugins.apply("org.jetbrains.kotlin.android")
+            project.plugins.apply("org.jetbrains.dokka")
+            project.extensions.configure<LibraryExtension> {
+                compileSdkVersion(28)
+            }
         }
     }
 
     @Test
+    @Ignore("Current version of AGP requiers Java 11. TODO: if expected, disable this test for Java 8")
```
IgnatBeresnev, line 30: AGP 7.1.3 (which is a requirement for KGP 2.0.0) requires Java 11 to run. I've asked the build tools team if it's intentional, and if it is - we can just mute these tests when running under Java 8, I guess. Will come back to it

dokka-runners/runner-gradle-plugin-classic/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/KotlinNativeDistributionAccessor.kt
```diff
@@ -27,12 +27,23 @@ internal class KotlinNativeDistributionAccessor(
   project: Project
 ) {
   private val konanDistribution = KonanDistribution(
-    @Suppress("INVISIBLE_MEMBER")
-    project.konanHome
+      // see this comment for the explanation of what's happening:
```
whyoleg, line 30: minor: may be it will be better to create a top level function like `private fun konanHome(project: Project): String` with both reflection and `alternativeKonanHome` logic (possibly divided there in 2 local functions)?
To keep it the logic of konan home computation in a more clear way?

dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/AndroidAutoConfigurationTest.kt
whyoleg: any news here?

dokka-runners/runner-gradle-plugin-classic/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/KotlinNativeDistributionAccessor.kt
IgnatBeresnev: The idea was to minimize changes for review and git blame later on: before, `project.konanHome` was called, and now the same property is being called in the same place, but through reflection; `alternativeKonanHome` didn't exist, so it's added as an additional thing. If this wasn't a bugfix and needed to be maintained in the future - I'd certainly extract the logic, yeah

If it's an eyesore - let me know, I don't mind refactoring it a little like you proposed

dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
```diff
@@ -2,6 +2,6 @@
 # Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
 #
 
-dokka_it_kotlin_version=1.9.22
+dokka_it_kotlin_version=2.0.0-Beta4
```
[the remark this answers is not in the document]
IgnatBeresnev: Good question, I have no idea :) We can try to remove it in a separate PR, see if the tests fail

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
```diff
@@ -14,7 +14,7 @@ open class AllSupportedTestedVersionsArgumentsProvider : TestedVersionsArguments
 
 object TestedVersions {
 
-    val LATEST = BuildVersions("7.6.2", "1.9.22")
+    val LATEST = BuildVersions("8.5", "2.0.0-Beta4")
```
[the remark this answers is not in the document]
IgnatBeresnev: > latest Gradle version is 8.6, should we use it here?

Damn, time flies and Gradle ships fast. Updated it to 8.6

I know there's 2.0.0-Beta5 now, but I won't update to it due to #3533, let's do it after this PR is merged, it has enough hacks as it is 😅 

> should we also test on versions between 7.6 and 8.6, or you think just latest minor will be enough?

Also a good question. Testing on the latest minor versions has worked well so far, we haven't had any problems with it, but Gradle 8 seems to have more features in minor releases, so I don't know, it might be needed. I'd be ideal to test on all intermediate versions, of course, but then the execution time would skyrocket

Perhaps, more versions could be tested if the `isExhaustive` flag is enabled, but last time I checked it didn't work well, some other versions need to get adjusted

dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/AndroidAutoConfigurationTest.kt
IgnatBeresnev: Yep! AGP 7.1.3+ is compiled with Java 8, but requires Java 11 to actually run. So we are compatible, but need to skip these tests with Java < 11. Will use the same solution as https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/documentable/ObviousFunctionsTest.kt#L124-L136

IgnatBeresnev:
The integration tests are failing with

```kotlin
Execution failed for task ':commonizeNativeDistribution'.
    > Could not resolve all files for configuration ':kotlinNativeBundleConfiguration'.
        > Could not find org.jetbrains.kotlin:kotlin-native-prebuilt:2.0.0-Beta4.
```

At first I thought that it was an incompatible change in Kotlin / KGP (https://youtrack.jetbrains.com/issue/KT-64181), but after manually checking it, I realized it could be specific to our integration tests, and indeed this configuration might be causing the failure:

https://github.com/Kotlin/dokka/blob/d9f1efe6d2bad585eba5de0b85d1ac2d9bd4c6ab/dokka-integration-tests/gradle/projects/template.settings.gradle.kts#L102-L130

@adam-enko could you please help here? It's a recent change from #3492, so hopefully it's still fresh in memory

dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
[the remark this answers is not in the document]
adam-enko: I believe the version is set as a default in IT projects, so if someone wants to open them in IJ and edit them, it's easier.

dokka-runners/runner-gradle-plugin-classic/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/KotlinNativeDistributionAccessor.kt
whyoleg: Thanks for the reasoning, it's fine for me like it's now, no need to change

dokka-integration-tests/gradle/projects/it-android-0/gradle.properties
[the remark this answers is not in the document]
whyoleg: But the version is used mostly in `template.settings.gradle.kts` (with small exceptions), so to open the project in IDEA, you will still need to copy it inside project folder, or copy it's content
So, I would propose to remove this (in some future PR), if possible, to reduce amount of places where we need to update Kotlin version (or any other version), so that at some point we will not have some situation, that this version is not updated or not overridden in integration tests

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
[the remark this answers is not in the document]
whyoleg: > Damn, time flies and Gradle ships fast. Updated it to 8.6

And soon there will be 8.7 - several RCs already :)

> but Gradle 8 seems to have more features in minor releases

Yeah, that's why I thought about it. 8.x versions are implementing a lot of new features regarding CC and IP, so it would be nice to have, but not strictly required.
May be after IT will be CC friendly we could check if adding some of them is feasible or not.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -39,6 +39,24 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
         vararg arguments: String,
         jvmArgs: List<String> = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=1G")
     ): GradleRunner {
+
+        // TODO quick hack to add `android { namespace }` on AGP 7+ (it's mandatory in 8+).
```
whyoleg, line 43: can't we just set `namespace` in project?
I see that the property is available from `4.2.0` - https://developer.android.com/reference/tools/gradle-api/8.3/com/android/build/api/dsl/CommonExtension#namespace()

adam-enko: Good idea. I did try setting it in all projects, but it failed. I didn't look too much into it, I just added this quick fix to get all the tests working locally.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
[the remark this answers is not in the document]
whyoleg: too soon: https://github.com/gradle/gradle/releases/tag/v8.7.0

adam-enko:
> @adam-enko could you please help here? It's a recent change from #3492, so hopefully it's still fresh in memory

Sure thing, it's a simple fix.

Explanation of the situation:

* KGP 1.x automatically added a custom Ivy repo for downloading K/N artifacts. 
* `template.settings.gradle.kts` set `repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)`, which meant that this custom Ivy K/N repo was specified manually.
* In KGP 2.x the K/N artifacts are uploaded to Maven Central, so no custom repo is necessary.
* BUT the workaround I added used an `exclusiveContent {}` filter (to help Gradle performance, so it wouldn't unnecessarily try to search the Ivy repo for non-K/N content)
* So when KGP 2.0 tried to download K/N artifacts, Gradle would _only_ look in the custom Ivy repo, which doesn't contain K/N 2.0 content.

I mad a commit to just remove the `exclusiveContent {}` filter, so Gradle will look for K/N artifacts in either the Ivy repo _or_ Maven Central.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestedVersions.kt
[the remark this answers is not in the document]
whyoleg: FYI, on current moment we blocked by: https://youtrack.jetbrains.com/issue/KT-66961/Early-access-to-gradle.rootProject-leads-to-an-exception
Fixed in 2.0.0-RC1 (checked that it works with 2.0.0-RC1-86 from dev repository)
To prepare for this, I've updated Kotlin version to `2.0.0-Beta5` and so fixed #3533 (https://github.com/Kotlin/dokka/pull/3516/commits/fe5f5b3590f9f8ccbb3a15286fe95308be502408 - should be rather safe, as we only use `BindingContext.get` afterwards)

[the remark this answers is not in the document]
whyoleg: I've updated latest Gradle version for IT to 8.7 and Kotlin to 2.0.0-RC1

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
whyoleg: The problem is, that for some reason we execute integration tests for Android with AGP 4.0.0 (and it even works) - I believe we could simplify this in following PR by testing with AGP 7+

dokka-subprojects/analysis-kotlin-descriptors-ide/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/ide/DokkaResolutionFacade.kt
```diff
@@ -87,6 +87,10 @@ internal class DokkaResolutionFacade(
                     }
                 }
 
+                override fun getProject(): Project? {
+                    throw UnsupportedOperationException()
```
IgnatBeresnev, line 91: Theoretically, we do have some project as a property of DokkaResolutionFacade - I wonder if it should be returned instead? Or if the return type is nullable, maybe `null` could be returned as well

Throwing an exception generally feels kinda scary in analysis, god knows what corner case might cause `getProject()` to be called now or in future versions (assuming kotlin-compiler has access to this object for something)

whyoleg: 1. on current moment only `get` from `BindingContext` is called on returned object and it's not passed anywhere - so chances of calling this method are minimal. So to fail here we will need to explicitly call `getProject` and so most likely it will fail on first test run.
2. Other methods override there do the same - `throw UnsupportedOperationException()` - and I would say, it's expected behaviour, as they should not be called.

I could change it to `null`, but I don't see a real reason why we should handle this method differently comparing to other ones.

whyoleg (approved):
LGTM

dokka-subprojects/analysis-kotlin-descriptors-ide/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/ide/DokkaResolutionFacade.kt
IgnatBeresnev: > it's not passed anywhere

oh, I thought it was created to then be passed to the compiler or something. If it's only used by us, it's indeed not a problem 👍

adam-enko:
How important is the failing SequentialTasksExecutionStressTest?

https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Dokka_DokkaIntegrationTestsK2_virtual_Batch_1_1/4567095?buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildDeploymentsSection=false&expandBuildProblemsSection=true

whyoleg:
> How important is the failing SequentialTasksExecutionStressTest?

Theoretically it could mean, that there is some kind of memory leak. But also, this could just mean, that more memory is needed.
After some short investigation (with profiler/memory graphs), I haven't found anything suspicious. So may be some more in-depth checks should be performed.

IgnatBeresnev:
I think it's certainly worth investigating why the test is failing so consistently and for K2 only:

image

It's relatively stable with K1 and in the `master` branch. 

We've had instances where updating the Kotlin version led to memory leaks, but we've also had situations where the new Kotlin version required just a bit more memory than the previous one, in which case the allocated memory should be increased. Just have to rule memory leaks out as the result of the investigation, because if there's a problem somewhere in Kotlin - it's worth reporting it upstream as it's likely to affect other Kotlin users too

We'll try to take a look at it as part of this PR, so I guess we can halt the review

whyoleg:
I've done some additional checks on how much memory is used, here are some graphs:

<details>
<summary>Test with K2 analysis built with Kotlin 2.0.0-RC1: current PR, this is what is failing on CI, works fine locally</summary>
image
</details>

<details>
<summary>Test with K2 analysis built with Kotlin 1.9.22: current master</summary>
image
</details>

<details>
<summary>Test with K1 analysis built with Kotlin 2.0.0-RC1: current PR, doesn't fail on CI</summary>
image
</details>

<details>
<summary>Test with K1 analysis built with Kotlin 1.9.22: current master</summary>
image
</details>

Note: integration test was run locally with the same Gradle (8.7) and KGP (2.0.0-RC1) versions.

Overall, all the graphs look fine, and most likely there is no memory leaks.
Still, K2 analysis uses slightly more metaspace memory (±60 MB), that's why most likely at some point this test starts to fail, as we do reinitialize ClassLoader for Dokka execution in Gradle plugin for every task execution. While locally it works fine (no crash), on CI there could be less resources - and so at some point ClassLoader for Dokka execution can not be loaded, and so fails with OOM.

I would propose to increase `-XX:MaxMetaspaceSize` in this specific test from 400 to 500 and see what's going on.
If it works, then most likely there is no leaks.

adam-enko:
> Still, it's possible, as profiling Dokka execution in Gradle from multiple tasks is not that informative, as there is a lot of Gradle machinery involved under the hood...

I agree.

Just writing down what's on my mind, I don't have any action planned.: 

This performance test also makes the CI tests slower, it shouldn't need to be run every time (e.g. scheduling it to be once a day should be fine), and it makes the CI tests much more sensitive to limited resources (I suspect without the performance test we could probably just run all of the tests on one machine, without limiting parallelism or needing to batch the tests.

I think it'd make sense to create a separate subproject for profiling tests.

PR #3522, opened 2024-03-08
consistently apply Gradle Build Cache config in all includedBuild projects

adam-enko:
Currently the plugin-includedBuild projects `build-logic` and `build-settings-logic` do not have Build Cache enabled. This hurts build performance, and Dokka gets dinged with a build scan warning.

> [The build cache configuration of the root build differs from the build cache configuration of the early evaluated ':build-logic', ':build-settings-logic' included builds.](https://ge.jetbrains.com/s/hy6qye4r7yi4e/performance/build-cache#warning-different_configurations)

This PR shares the build cache config in all projects, improving build performance.

### Convention plugin workaround.

In `build-settings-logic` there's a chicken-and-egg situation: a project can't apply its own convention plugin to itself.

Instead, I've applied a hack to apply the build cache config as a script plugin. ([Space also applies the same quick-fix.](https://jetbrains.team/p/crl/repositories/space/files/bdbcb4955b269e8d578b00c2d0b28efce7dfee2f/gradle-settings-conventions/settings.gradle.kts?tab=source&line=22&lines-count=1))

### DIFF
--- a/build-settings-logic/src/main/kotlin/dokkasettings.build-cache.settings.gradle.kts
+++ b/build-settings-logic/src/main/kotlin/dokkasettings.build-cache.settings.gradle.kts
@@ -1,41 +1,62 @@
 /*
  * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
-import DokkaBuildSettingsProperties.Companion.dokkaBuildSettingsProperties
 
 /**
  * Gradle Build Cache conventions.
  *
- * See [DokkaBuildSettingsProperties] for properties.
+ * Because Dokka uses Composite Builds, Build Cache must be configured consistently on
+ * - the root settings.gradle.kts,
+ * - and the settings.gradle.kts of any projects added with `pluginManagement { includedBuild("...") }`
  *
- * Because Dokka uses Composite Builds, Build Cache should only be configured on the root `settings.gradle.kts`.
  * See https://docs.gradle.org/8.4/userguide/build_cache.html#sec:build_cache_composite.
  *
  * Based on https://github.com/JetBrains/kotlin/blob/2675531624d42851af502a993bbefd65ee3e38ef/repo/gradle-settings-conventions/build-cache/src/main/kotlin/build-cache.settings.gradle.kts
  */
 
-val buildSettingsProps = dokkaBuildSettingsProperties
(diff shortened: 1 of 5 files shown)


build-settings-logic/src/main/kotlin/dokkasettings.build-cache.settings.gradle.kts
```diff
@@ -1,41 +1,62 @@
 /*
  * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
-import DokkaBuildSettingsProperties.Companion.dokkaBuildSettingsProperties
 
 /**
  * Gradle Build Cache conventions.
  *
- * See [DokkaBuildSettingsProperties] for properties.
+ * Because Dokka uses Composite Builds, Build Cache must be configured consistently on
```
whyoleg, line 8: minor: may be it make sense to add small note, that this file should not depend on anything (other plugins/classes from `build-settings-logic`) and so be fully independent?

gradle.properties
```diff
@@ -12,7 +12,6 @@ org.jetbrains.dokka.kotlinLanguageLevel=1.4
 
 # Build Scans are disabled by default - see CONTRIBUTING.md to opt in
 org.jetbrains.dokka.build.scan.url=https\://ge.jetbrains.com/
```
whyoleg, line 14: while we are here and we removed `cache.url`, what is the reason to have `scan.url` here and not inside settings plugin (like with `cache.url`)?

whyoleg (approved):
LGTM, even if it looks strange :)

May be some day Gradle will provide better out-of-the-box integration for included builds, which are inside of the project to avoid all those beautiful hacks...

gradle.properties
adam-enko: Good point, I'll hardcode the build scan url too.

The reason the scan url is configurable is because originally I intended that users could opt-out by not defining the url, but that didn't work out so well (I can't remember why) so I added an explicit boolean opt-in/out, but didn't update the scan url.

build-settings-logic/src/main/kotlin/dokkasettings.build-cache.settings.gradle.kts
adam-enko: Good idea, I'll add a warning (but I guarantee it won't be read 😄)

dkrasnoff (approved):

PR #3523, opened 2024-03-08
Fix checking for annotations when rendering constructor keyword

### LINKED ISSUE #3426: Excessive `constructor` keyword rendered in class signatures
The `constructor` keyword is in some cases rendered excessively as it's public and doesn't have any parameters.

image

whyoleg:
Fixes #3426

There are 2 cases when annotation is not rendered and so no `constructor` keyword should be rendered:
1. when annotation is not annotated with `MustBeDocumented`
2. when annotation is ignored (and so handled in some other way, like `Deprecated`)

Note: PR has small API/ABI changes (no breaking changes), that's because ignored annotations are ignored later in pipeline, only during rendering. We can't ignore annotations in `JvmSignatureUtils`, as we need all annotations in other places to make a special handling of those ignored annotations (like `Deprecated`).
Also, we can introduce something like `Annotation.shouldBeRendered` instead of `Annotation.isIgnored` inside `JvmSignatureUtils`, but to use it under the hood in `annotationsInline`/`annotationsBlock` we will need to break API/ABI (or introduce new declaration + deprecate old one) as on current moment `ignored` annotations there provided explicitly...
So I've decided not to go this road.
If we think that it's fine to break API/ABI in `JvmSignatureUtils` and its inheritors (can be seen in API dump changes), then I can update PR.

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
@@ -1366,6 +1366,63 @@ class SignatureTest : BaseAbstractTest() {
         }
     }
 
+    @OnlyDescriptors("#3354")
+    @Test
+    fun `should not render constructor for Any from Wasm sources`() = testRender(
+        """
+            |/src/main/kotlin/Any.kt
+            |package kotlin
+            |annotation class WasmPrimitiveConstructor
+            |open class Any @WasmPrimitiveConstructor constructor()
+        """.trimMargin(),
+    ) {
+        renderedContent("root/kotlin/-any/index.html").firstSignature().matchIgnoringSpans(
+            "open class", A("Any")
+        )
+    }
+
+    @Test
+    fun `should not render parameterless constructor with annotation without mustBeDocumented annotation`() = testRender(
+        """
+            |/src/main/kotlin/SomeClass.kt
+            |package example
+            |annotation class SomeAnnotation
+            |class SomeClass @SomeAnnotation constructor()
+        """.trimMargin(),
+    ) {
+        renderedContent("root/example/-some-class/index.html").firstSignature().matchIgnoringSpans(
+            "class", A("SomeClass")
+        )
+    }
+
+    @Test
+    fun `should not render parameterless constructor with ignored annotation`() = testRender(
+        """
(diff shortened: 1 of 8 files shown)


dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
```diff
@@ -1366,6 +1366,63 @@ class SignatureTest : BaseAbstractTest() {
         }
     }
 
+    @OnlyDescriptors("#3354")
+    @Test
+    fun `should not render constructor for Any from Wasm sources`() = testRender(
```
vmishenev, line 1371: _nitpick:_ I would prefer to have a more general name for this test, e.g. `should not render constructor without a renderable annotation`

vmishenev (approved):

dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
whyoleg: There is such test below, though, named a little more explicit.
This specific case is what was happening in original issue. Mostly to check, that it works the same for `kotlin` package and `builtin` class, so Im not sure, what will be the best name here.

vmishenev: Among other tests this name made me think about a specific logic for Any (we have such a one) or wasm source set.

whyoleg: yeah, I see why there is a confusion... do you have any other idea on how to name this test, so that the intention will be clear?
Does something like this will be better may be `should not render constructor without a renderable annotation for kotlin.Any`? I mean, while there is no special casing for `kotlin.Any`, I would still expect to have a test that checks specifically reported problem here, as kotlin builtins are specific things and so it's good to have tests for them.

vmishenev: I think the name below can be reused as the test has the same logic. 
`should not render parameterless constructor with annotation without mustBeDocumented annotation - for kotlin.Any `

IgnatBeresnev (approved):

PR #3524, opened 2024-03-08
Update test-k2 convention to use test suites

adam-enko:
### Summary

- use [JVM Test Suites](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html) to create create separate targets for descriptor and symbols tests
- update Configurations to use newer Gradle guidelines for declarable/resolvable configurations

### Usage

* `./gradlew test` runs descriptor-compatible tests.
  (Additionally, `test` tasks will trigger `testDescriptors` tasks.)
* `./gradlew testDescriptors` runs descriptor-compatible tests (this is an alias of `test`, because it's easier to read).
* `./gradlew testSymbols` runs symbols-compatible tests.

### Benefits

* Using JVM Test Suites unifies common test configuration for each of the test/descriptorsTest/symbolsTest tasks, so custom tasks aren't required
* Avoiding custom tasks will help with Configuration Cache compliance.
* Tagged tests are run in separate tasks, so the same tests aren't run multiple times with different tasks.
* Remove deprecated behaviour - see [Build Scan warning](https://ge.jetbrains.com/s/yganljdgr3cja/deprecations?expanded=WyIzIl0#0) "Relying on the convention for Test.classpath in custom Test tasks has been deprecated."


adam-enko:
Hi @vmishenev 👋, I've added you as a reviewer since you created this convention plugin.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
```diff
@@ -9,39 +12,82 @@
...
+            val symbolsTags = listOf("onlySymbols")
+
+            // Configure the regular 'test' target
+            val testTarget = targets.named("test") {
+                testTask.configure {
+                    description = "Runs tests (excluding descriptor and symbols tags: ${descriptorTags + symbolsTags})"
+                    useJUnitPlatform {
+                        excludeTags.addAll(descriptorTags + symbolsTags)
+                    }
+                    classpath += descriptorsTestImplementationResolver.incoming.files
```
dkrasnoff, line 58: It is not clear to me why we have the same configuration as for `descriptorsTest` ?

adam-enko: Good question. Basically, Dokka needs to analyse source code.

* For K1 source code, it uses [a descriptor-based implementation](https://github.com/Kotlin/dokka/tree/v1.9.20/dokka-subprojects/analysis-kotlin-descriptors).
* And for K2, [symbols-based implementation](https://github.com/Kotlin/dokka/tree/v1.9.20/dokka-subprojects/analysis-kotlin-symbols).

The Dokka subprojects don't use these implementations directly, they need to be provided at runtime. So, to run the tests, Dokka needs one implementation or the other. Dokka is still being converted to K2, so at the moment K1 is the default. Hence, the K1 descriptor-based analyser is used for 'regular' unit tests.

Correct me if I'm wrong please Dokka team!

dkrasnoff: That makes sense, thanks!

dkrasnoff (approved):

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
whyoleg: So, overall the idea is to run ALL tests with both K1 and K2 analysis. + there are some tests where behaviour differs (because f.e K1 or K2 analysis has some bugs), so for this cases we have annotations like `OnlyDescriptors`/`OnlySymbols`

And AFAIU, approach in this PR is not what we want. If im not wrong now we have 3 test tasks:
* `test` will run ALL tests excluding marked with annotation using K1
* `descriptorsTest` will run only marked for descriptors tests with K1
* `symbolsTest` will run only marked for symbols tests with K2

So I believe it's not what we want.
I don't know how flexible are test-suites but finally we want to have two separate tasks to run analysis for K1 and K2, f.e:
* `symbolsTest` will run ALL tests excluding `OnlySymbols` with K1 runtime
* `descriptorsTest` will run ALL tests excluding `OnlyDescriptors` with K2 runtime
* `test` will just run both `symbolsTest` and `descriptorsTest`, but don't run ALL tests one more time

Does it help?

whyoleg (changes requested):

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
vmishenev: > symbolsTest will run ALL tests excluding OnlySymbols with K1 runtime
descriptorsTest will run ALL tests excluding OnlyDescriptors with K2 runtime

it seems @whyoleg  mixed up `symbolsTest` and `descriptorsTest`
The correct order is
- `descriptorsTest` will run ALL tests excluding `OnlySymbols` with K1 runtime
- `symbolsTest` will run ALL tests excluding `OnlyDescriptors` with K2 runtime

whyoleg: Oops, yes, sorry, will fix the comment :(

adam-enko: Thanks! I see that this PR is a change in behaviour. 

I'm still confused about the best way to proceed though, because when I look on master, `descriptorsTest` and `symbolsTest` are exactly the same

* https://github.com/Kotlin/dokka/blob/bc58001c58a8a827a31f44820634794dd452acd6/build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts#L28-L35
* https://github.com/Kotlin/dokka/blob/bc58001c58a8a827a31f44820634794dd452acd6/build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts#L37-L42

What I wanted to avoid is having different tasks run the same test twice, with exactly the same dependencies.

I see that `descriptorsTest` isn't added to `check` or `test`, so it won't always run. Maybe that's the only intended purpose of this task? Do you @whyoleg, @vmishenev, run `./gradlew descriptorsTest`, just to test non-symbols tests in some subprojects?

Anyway, this PR is supposed to be a drop-in replacement so I'll update it to match what's on master, so I'm just asking because it'd help if I could add some comments to make the intention clearer :)

whyoleg: So, on master we have:
* `descriptorsTest` - runs tests with K1
* `symbolsTest` - runs test with K2
* `test` - configured the same as `descriptorsTest`, so runs tests with K1 + runs `symbolsTest` - so when running `test` you will be able to run tests both with K1 and K2

> I see that descriptorsTest isn't added to check or test, so it won't always run

But, you can run it from IDEA gutter - so f.e I can run tests for some class/method with K1 only, or with K2 only
image

Ideally, I think, it will be fine that `test` will not run any tests by itself, but just run dependent tasks (for K1 and K2), but I'm not sure, that it's possible to make compatible with both CLI and IDE, so current approach (from master) should be a good compromise

Does it help to understand the idea better? May be Vadim has something additional to say here

adam-enko: Thanks for explaining! So the idea is to provide a more descriptive name, even if `test` and `descriptorsTest` are both the same, under the hood?

> Ideally, I think, it will be fine that `test` will not run any tests by itself, but just run dependent tasks (for K1 and K2), but I'm not sure, that it's possible to make compatible with both CLI and IDE, so current approach (from master) should be a good compromise

Yeah, that makes sense. It's what I tried to do in the original PR, but JUnit and/or Gradle doesn't like having a test tasks that don't do anything (which is very important, because if a bug means no tests run then you don't want the test task to succeed!)

Anyway, I've reverted the behaviour to master, so it should work the same. 

And here's a summary of the discussed options:

| Task           | uses        | master | initial PR | test-runs-nothing proposal | test-runs-no-tags proposal |
|:---------------|:------------|:-------|:-----------|:---------------------------|:---------------------------|
| test           | descriptors | !K2    | !K1 && !K2 | _nothing_                  | !K1 && !K2                 |
| descriptorTest | descriptors | !K2    | K1         | !K2                        | !K2                        |
| symbolTest     | symbols     | !K1    | K2         | !K1                        | !K1                        |

build-logic/src/main/kotlin/dokkabuild/utils/gradle.kt
```diff
@@ -0,0 +1,89 @@
...
+ * Declarable Configurations should be extended by [resolvable] and [consumable] Configurations.
+ * They must not have attributes.
+ *
+ * ```
+ * isCanBeResolved = false
+ * isCanBeConsumed = false
+ * isCanBeDeclared = true
+ * ```
+ */
+internal fun Configuration.declarable(
```
whyoleg, line 27: I believe we can already use newer APIs introduced in Gradle 8.4 for this, no?
It's for our build-logic only, so it's fine.
Also, KGP is already using them even under the hood [here](https://github.com/JetBrains/kotlin/blob/0df3a37303f95e86608aa5249115e11ea754c34f/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/utils/configurations.kt#L79), so this should be fine overall.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
```diff
@@ -9,39 +12,81 @@
...
+                    description = "Runs tests using descriptors-analysis (K1) (excluding tags: $onlySymbolsTags)"
+                    useJUnitPlatform {
+                        excludeTags.addAll(onlySymbolsTags)
+                    }
+                    classpath += descriptorsTestImplementationResolver.incoming.files
+                }
+            }
+
+            // Create a new target for _only_ running descriptor (K1) tests.
+            targets.register("descriptorsTest") {
```
whyoleg, line 63: BTW, after some thoughts, may be we need to rename tasks to `testSymbols` and `testDescriptors` - it will be more aligned with integration tests naming (like `testTemplateProject***`) and also tasks will be near each other in IDEA?

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
whyoleg: may be we can then add comment the reason why `test` task is configured that way? So it's just an alias to descriptor tests, to avoid future confusions?

whyoleg (approved):

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
adam-enko: I'd be happy to rename them! Your suggestions make sense to me.

I wanted to keep the names the same, so that this PR didn't disrupt anyone's habits.

build-logic/src/main/kotlin/dokkabuild/utils/gradle.kt
adam-enko: We could do, but I'd recommend waiting for now. I suspect that these new role-based Configurations will be changed in the future, because they're a bit more rough around the edges.

For example, there's no nice Kotlin delegated DSL for creating them, so you have to duplicate the names.


```kotlin
val fooConfig by configurations.resolvable("fooConfig") { 
}
```

And aside from being mentioned in the release notes, there's no documentation for them. 

They probably work fine, but then, so does the existing set up :)

But if you insist, I can update this PR to use the role-based Configs.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
adam-enko: Just let me know what you prefer, otherwise I'll leave them as-is in this PR. We can always change them later.

whyoleg: For me it's fine to change it now. @vmishenev WDYT?

build-logic/src/main/kotlin/dokkabuild/utils/gradle.kt
whyoleg: Ok, no need for now then.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
whyoleg: aside from comment, it would be good to actualise PR description

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
vmishenev: I think so, `testSymbols` and `testDescriptors` are better

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
vmishenev: I have a question. Will the `test` task run the symbols tests as well?


Our CI (TeamCity/GH Actions) is set to run only the `test` task, e.g. https://github.com/Kotlin/dokka/blob/ed3ab96896d0ae5b0a5733bb70dd9bf88fb5146a/.github/workflows/tests-smoke.yml#L35

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
adam-enko: Done!

vmishenev (approved):

Issue #2879, opened 2023-02-17
Incorrect link for inherited java method from collection


vmishenev:
- In stdlib AbstractMutableCollection: the functions like `spliterator, stream, toArray, toString` point anywhere except that `AbstractMutableCollection` member pages: some point to `AbstractMutableSet`, some - to `LinkedHashSet`.

- Another reproducer ([the unit test](https://github.com/Kotlin/dokka/blob/d31fc588dcba45bd0cbbc413419fdf4a3911cb39/plugins/base/src/test/kotlin/model/InheritorsTest.kt#L450)):
```kt
            
            import java.util.AbstractCollection

            abstract class MyCollection<E> : AbstractCollection<E>()
```

vmishenev:
I have opened [KT-56838](https://youtrack.jetbrains.com/issue/KT-56838/JavaMethodDescriptor.overriden.overriden-has-an-incorrect-value-for-inherited-methods-of-Java-collections)

IgnatBeresnev:
Might be related: #2464

IgnatBeresnev:
The problem is on our side in the place where internal links (anchor) are resolved.

Typealises have their own pages now, so the internal link resolution might not be needed anymore, and this issue might be fixed by removing this resolution.

The anchor link resolution is somewhat confusing, so it might take longer to fix.

vmishenev:
Related issue: #3054

vmishenev:
#3054 can fix it, but in this case it would cause #3542.

IgnatBeresnev:
Try to filter out such J2K declarations. If there's no API to do that (esp. in Analysis API), create an upstream issue for it, and change the status of this issue to blocked. Descriptors (K1) - optional

vmishenev:
This is fixed by #3694 as well

Issue #3054, opened 2023-06-29
External inherited members lead to itself


vmishenev:
**Describe the bug**
For example, 
```kt
public class MyString: String
```
All inherited members  from external `String`  have anchor links that lead to itself

**Expected behaviour**
Inherited members should lead to external documentation (in this example - [kotlinlang.org](https://kotlinlang.org/))

**Screenshots**
![image](https://github.com/Kotlin/dokka/assets/9636937/bb23011b-f697-4416-8c20-0861aeb08c1c)

**Dokka configuration**
Dokka 1.8.20

vmishenev:
Just example (https://github.com/Kotlin/dokka/issues/2879#issuecomment-2182863488):

```kt
class MyStringA : String
class MyStringB : String
```

MyStringA.length (or another declaration) leads to MyStringB.length

PR #3529, opened 2024-03-12
Fix incorrect links for inherited java methods from a collection

vmishenev:
#2879


dokka-subprojects/plugin-base/src/test/kotlin/linkableContent/LinkableContentTest.kt
```diff
@@ -418,4 +420,41 @@ class LinkableContentTest : BaseAbstractTest() {
...
+            renderingStage = { rootPage, ctx ->
+                val location = DokkaLocationProvider(rootPage, ctx, ".html")
+                val classA = rootPage.dfs { it is ClasslikePageNode && it.name == "A" }
+                val classB = rootPage.dfs { it is ClasslikePageNode && it.name == "B" }
+                val classC = rootPage.dfs { it is ClasslikePageNode && it.name == "C" }
+                val sourceSet = (classA as ClasslikePageNode).content.sourceSets
+                val dri = org.jetbrains.dokka.links.DRI("kotlin.collections", "Collection", Callable(name="isEmpty", receiver=null, params=emptyList()))
+                assertEquals("../-b/index.html#-719293276%2FFunctions%2F-1617659094", location.resolve(dri, sourceSet, classA))
+                assertEquals("index.html#-719293276%2FFunctions%2F-1617659094", location.resolve(dri, sourceSet, classB))
+                assertEquals("../-b/index.html#-719293276%2FFunctions%2F-1617659094", location.resolve(dri, sourceSet, classC))
```
vmishenev, line 456: This is a reproducer.
If we remove an anchor logic in  `DokkaLocationProvider`, https://github.com/Kotlin/dokka/issues/3054 will be fixed as well. We have not discussed it at grooming, but it can be a big change for the user.

@IgnatBeresnev @whyoleg WDYT? 
```kt
            open interface C : List<C> {
            /**
            * [Collection.isEmpty]
            **/
            val p = 0
            }
            interface A : C()
            interface B : C()
            
```

IgnatBeresnev: `p` in the example is a bit confusing, it looks like it's somehow related to the bug, but it seems like it isn't?

Given the example

```kotlin
interface C : List<C>

interface A : C
interface B : C
```

I can open the page for `B` and see inherited functions there, like `isEmpty`. If I click on the anchor link icon, it gives me a link to `B#isEmpty`, the same place where I generated it: 

```
http://localhost:63342/dokka-debug-kts/build/dokka/html/dokka-debug-kts/me.beresnev.dokka.debug/-b/index.html#-1000881820%2FFunctions%2F769193423
```

image

What's going to happen if we implement the change? Will the anchor link lead to `List#isEmpty` on kotlinlang.org? If yes, what's the expected behaviour for overridden functions?

```kotlin
interface C : List<C>

interface A : C
interface B : C {

    /**
     * Overriding documentation
     */
    override fun isEmpty(): Boolean = true
}
```

vmishenev: > p in the example is a bit confusing, it looks like it's somehow related to the bug, but it seems like it isn't?

The reproducer is in the unit test.
```kt
interface C : List<C>
interface A : C
interface B : C
```
The `p` is not related to the bug at all, but the link `[Collection.isEmpty]` is.

> I can open the page for B and see inherited functions there, like isEmpty. If I click on the anchor link icon, it gives me a link to B#isEmpty, the same place where I generated it:

Sorry, I got it where you are confused. I mean only the `resolving anchors by DRI` logic (see `DokkaLocationProvider .resolveDrI`), but not generating anchor links for an icon (see `DokkaBaseLocationProvider.anchorForDCI`).  
 If you open the page for A or C and click on the `isEmpty ` keyword, it opens `-b/index.html#-1000881820%2FFunctions%2F769193423`.
 
 > Will the anchor link lead to List#isEmpty on kotlinlang.org?
 
 No, the generating anchor link icon will left the same.
 
> If yes, what's the expected behaviour for overridden functions?

In this case, Dokka generates a dedicated page for the overridden functions.

IgnatBeresnev: Oh, I think I understand now! After the fix, all anchor icon links will work the same way as before, but if `C` inherits `isEmpty`, clicking on `C#isEmpty` will lead to kotlinlang instead of leading to itself or to A/B, right? If so, it looks logical to me 👍

vmishenev:
After a discussion at our meeting, opening external documentation for such members can be a UI issue, e.g. there is no way to come back to the original documentation. 
I will open an issue later.

Also, I have discovered https://github.com/Kotlin/dokka/issues/3542 It should be discussed as well before merging this PR.

PR #3530, opened 2024-03-13
set jvmToolchain 11 for build-settings-logic

adam-enko:
`build-settings-logic/build.gradle.kts` is missing a JVM toolchain, which means in some cases builds are not reproducible. This may be causing Build Cache misses.

Also, update the syntax to use the new, shorter util function.


whyoleg (approved):
LGTM

dkrasnoff (approved):

PR #3531, opened 2024-03-13
Improve Build Caching of Maven plugin tasks

adam-enko:
Update tasks for building the Dokka Maven plugin, so that Build Cache works across machines.

(Alternative title: TIL that the compileKotlin task output includes caches)

### Summary

- Update MvnExec task:
  - separate classes/resources inputs
  - filter and annotate classes input:
     1. only include `.class` files (the compileKotlin file includes caches)
     2. annotate with `@Classpath` so Gradle will be able to normalize the files

- Update `generatePom` task to manually register the version inputs, to ensure Gradle will re-run the task if they change.
- Update `helpMojoResources` to exclude `maven-plugin-help.properties`, because it contains a property with an absolute file path, meaning that Build Cache cannot be shared between CI and local machines.


whyoleg (approved):
LGTM

dkrasnoff (approved):

PR #3548, opened 2024-03-25
Remove `<a href="TODO">` in enum constant summary

MattiasBuelens:
The rendered signature already contains the correct link, but was accidentally wrapped in another `<a>` with a placeholder `href` link. This PR removes that dummy `<a>` wrapper.


whyoleg (approved):

Issue #3535, opened 2024-03-19
Update stdlib to the latest version of Dokka


IgnatBeresnev:
The version of Dokka used to build stdlib's docs should be updated to the latest `-dev` build.

1. Bump the version
2. Implement the sample re-writer interface introduced in #3108
3. Enable the `All Types` page introduced in #2887

Discuss with Ilya which branch it should target and how to proceed after the PR is sent.

whyoleg:
Issues found during update:

1 (warning): Intersected sources `Every Kotlin source file should belong to only one source set (module).`
In Dokka K2 it will be an error.
* Source sets 'Common' and 'JS' have the common source roots: 
  * core/builtins/native (and nested files)
  * core/builtins/src (and nested files)
* Source sets 'Native' and 'Wasm-JS' have the common source roots:
  * libraries/stdlib/native-wasm/src
* Source sets 'Native' and 'Wasm-WASI' have the common source roots:
  * libraries/stdlib/native-wasm/src
* Source sets 'Wasm-JS' and 'Wasm-WASI' have the common source roots:
  * libraries/stdlib/native-wasm/src
  * libraries/stdlib/wasm/src
  * libraries/stdlib/wasm/builtins
  * libraries/stdlib/wasm/internal
  * libraries/stdlib/wasm/stubs

2 (not regression): Unable to resolve a @sample link: "samples.collections.Collections.Aggregates.minBy" (and `maxBy`) - same behaviour for previous versions and on currently deployed to kotlinlang.org - [example](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/min-by.html)

3 (regression): After updating fronted dependencies (https://github.com/Kotlin/dokka/pull/3413) samples are not runnable anymore. If to replace `scripts/main.js` in built documentation with latest Dokka (2.0.0-dev-323) with `main.js` from older Dokka (1.9.10) - samples are runnable again.

Changes are available in [dokka-2.0.0](https://github.com/whyoleg/kotlin/tree/dokka-2.0.0) branch, [diff](https://github.com/JetBrains/kotlin/compare/master...whyoleg:kotlin:dokka-2.0.0)

whyoleg:
Dokka update for Kotlin stdlib was merged in: https://github.com/JetBrains/kotlin/commit/1c774643c9f71aa855772217a47ff81e46d17e53

PR #3413, opened 2023-12-15
chore: updated dependencies (#3372)

berezinant:
fixes #3372

PR #3509, opened 2024-03-01
Add a test project for UI screenshot tests

IgnatBeresnev:
Fixes #3468

Still WIP

PR #3553, opened 2024-03-29
samples issue reproducer

whyoleg:
related: https://github.com/Kotlin/dokka/issues/3535#issuecomment-2025757474

## Problem
During update of Dokka to latest version (2.0.0-dev-323) in kotlin stdlib I found, that samples are not runnable anymore (playground is not loading)
Example:
image

Not reproducible with Dokka 1.9.10 and 1.9.20.

After some investigation I found that `scripts/main.js` is different when building with 1.9.20 and 2.0.0-dev-323.
After some additional check of older 2.0.0-dev-* versions I found, that the issue started to appear after https://github.com/Kotlin/dokka/pull/3413
If to revert `package-lock.json` changes samples are runnable again:
image

## Reproducing

Reproducer project is in PR. It's not an integration test, as it's not possible to render produced HTML there, and as the issue in `js` script there is no way to test it.

I've left some comments on how to reproduce the issue n `examples/gradle/dokka-runnable-samples-example/README.md`, will inline here just to understand the idea:

Note: better to change the version every time there are some changes happened,
as Gradle caches something badly and so `main.js` could not change in final HTML distribution.

To reproduce (from the root of the project):

```
rm -rf dokka-subprojects/plugin-base-frontend/node_modules
./scripts/testDokka.sh -d examples/gradle/dokka-runnable-samples-example -v 2.0.0-samples-issue-SNAPSHOT
```

Everything works fine if to revert commit which changes `package-lock.json` (155efa42):

```
git revert 155efa425bad65f6171b9c39aa6c25ce3cf85a83 --no-edit --no-commit
rm -rf dokka-subprojects/plugin-base-frontend/node_modules
./scripts/testDokka.sh -d examples/gradle/dokka-runnable-samples-example -v 2.0.0-samples-issue-fixed-package-json-SNAPSHOT
```

@berezinant Could you take a look?


IgnatBeresnev:
JFYI: I had a similar problem that I could reproduce even with Dokka 1.9.20:

image

After @atyrin said that it worked for him without any problems, I tried to reproduce it again with a VPN on, and it worked! I have DNS-level ad blocking on my local network, which I figured was blocking some resources that were needed to load the scripts. After turning the VPN on, it started using other DNS servers (without adblock), so the scripts and thus the samples got resolved correctly. Do you, by any chance, have Pi-hole or something similar running locally?

> as the issue in js script there is no way to test it.

The test project added in #3509 will be used for UI screenshot tests (I believe I added some `@sample` usages to it), which should catch such kinds of problems (it should spin a web browser, wait for the page to load, including js scripts, then make a screenshot and compare it to the previous version's screenshot).

whyoleg:
New information:
* Google Chrome - works
* Safari - doesn't work

Enabling/Disabling VPN/adblock doesn't affect results for both Chrome (works) and Safari (doesn't work)

whyoleg:
The issue should be fixed in scope of https://github.com/Kotlin/dokka/pull/3596.
Testing of such issues will be done using `ui-showcase` project and https://github.com/Kotlin/dokka/issues/3579.
This PR is no longer needed.

PR #3560, opened 2024-04-09
[K2] Fix DRI for override symbols

vmishenev:
Due to KT-63741 https://github.com/JetBrains/kotlin/commit/7056ad5325324bb8162db4cef2b5bcf02b3c5483

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
@@ -851,7 +851,7 @@ internal class DokkaSymbolVisitor(
          *      - delegating members
          */
         val isDeclaration = callableSymbol.origin == KtSymbolOrigin.SOURCE || callableSymbol.origin == KtSymbolOrigin.LIBRARY
-        if (!isDeclaration) {
+        if (isDeclaration) {
             return DRIWithOverridden(getDRIFromSymbol(callableSymbol), wasOverriddenBy)
 
         } else { // fake, synthetic, delegating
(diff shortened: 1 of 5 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -858,7 +873,7 @@ internal class DokkaSymbolVisitor(
 
     private fun KtAnalysisSession.isObvious(functionSymbol: KtFunctionSymbol, inheritedFrom: DRI?): Boolean {
         return functionSymbol.origin == KtSymbolOrigin.SOURCE_MEMBER_GENERATED && !hasGeneratedKDocDocumentation(functionSymbol) ||
-                !functionSymbol.isOverride && inheritedFrom?.isObvious() == true
```
vmishenev, line 861: Due to KT-63741 https://github.com/JetBrains/kotlin/commit/7056ad5325324bb8162db4cef2b5bcf02b3c5483, some fake overrides (e.g. from `Enum`) have become `isOverride=true`. This conjunct was just in case.

dokka-subprojects/plugin-base/src/test/kotlin/model/InheritorsTest.kt
```diff
@@ -512,4 +512,58 @@ class InheritorsTest : AbstractModelTest("/src/main/kotlin/inheritors/Test.kt",
...
+                dri.packageName equals "inheritors"
+            }
+        }
+    }
+    @Test
+    fun `DRI of inherited members should lead to super member`() {
+        inlineModelTest(
+            """
+                |interface A { fun x() = 0 }
+                |interface B : A { }
```
whyoleg, line 538: should we add additional test with `interface` which overrides `fun`?

AFAIU there is a matrix of variants here:
* interface vs class (vs open/abstract class, not sure it makes a difference)
* fake-override vs override vs no-override

May be it will be good to add all those cases?

whyoleg (approved):

dokka-subprojects/plugin-base/src/test/kotlin/model/InheritorsTest.kt
vmishenev: There is no difference between interface and class (in general - `classlikes`) for our logic here.
I would prefer to not add these tests  (interface vs class) since we have a lot of tests where interfaces and classes are interchangeable with the same result.

PR #3561, opened 2024-04-10
Do not require signing artifacts when publishing locally

IgnatBeresnev:
Currently, the requirement for signing is set as

```kotlin
setRequired(provider { !project.version.toString().endsWith("-SNAPSHOT") })
```

which works fine when it comes to day-to-day stuff, but I found it to be insufficient when preparing the 1.9.20 release.

In release branches, the `version` property doesn't have any suffixes, example here: https://github.com/Kotlin/dokka/blob/1.9.20/gradle.properties#L7

It is needed because there are cases when this version is also used for publishing of other things that are not related to artifacts, such as publishing [developer docs](https://github.com/Kotlin/dokka/blob/5d4dc7148adf6abc9b579eb2948cca6d5fb2c4e8/docs-developer/build.gradle.kts#L13) (updating the documentation in the release branch should lead to the docs being deployed under the release version).

If you remove the `-SNAPSHOT` suffix, however, and try to run integration tests, the build will fail due to missing keys as the integration tests publish Dokka's artifacts locally, which requires signing (as it's not a `-SNAPSHOT`). Given that commits are sometimes cherry-picked into the release branch, the integration tests must be runnable. 

Thus the proposed solution was born - do not require signing if the artifacts are being published locally (maven local, maven dev) as there's no real need for it. This makes integration tests work in release branches.

___

I don't remember if `project.version` is used someplace else aside from the `docs-developer` module, but if it isn't - the alternative would be to introduce a separate version property for publishing developer docs, but
1. it would require us to not forget to update it. Given that it's always 1 to 1 with the main Dokka version, I think the same version property should be used
2. it's sometimes helpful that the version inside release branches doesn't have the `-SNAPSHOT` suffix, it helps to differentiate between branches: `master` always has the `-SNAPSHOT` suffix, release branches never do.

So I'd vote to keep the current workflow with removing the `-SNAPSHOT` suffix in release branches, but this requires some changes when it comes to signing to make the integration tests runnable

I'm not insisting on the proposed approach per se, this was the best solution we found when fixing the 1.9.20 release branch - maybe there are other options / solutions that solve this problem?

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts
+++ b/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts
@@ -75,7 +75,15 @@ signing {
         System.getenv("DOKKA_SIGN_KEY_PASSPHRASE")?.takeIf(String::isNotBlank),
     )
     sign(publishing.publications)
-    setRequired(provider { !project.version.toString().endsWith("-SNAPSHOT") })
+    setRequired(provider {
+        // no signing should be required for locally published artifacts,
+        // as they are used for manual testing and running integration tests only
+        val isPublishingLocally = gradle.taskGraph.allTasks.any {
+            it.name in listOf("publishToMavenLocal", "publishToDevMavenRepo")
+        }
+
+        !isPublishingLocally
+    })
 }
 
 // This is a hack for a Gradle 8 problem, see https://github.com/gradle/gradle/issues/26091
--- a/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts
+++ b/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts
@@ -75,15 +75,9 @@ signing {
         System.getenv("DOKKA_SIGN_KEY_PASSPHRASE")?.takeIf(String::isNotBlank),
     )
     sign(publishing.publications)
-    setRequired(provider {
-        // no signing should be required for locally published artifacts,
-        // as they are used for manual testing and running integration tests only
-        val isPublishingLocally = gradle.taskGraph.allTasks.any {
-            it.name in listOf("publishToMavenLocal", "publishToDevMavenRepo")
-        }
-
-        !isPublishingLocally
-    })
+    // no signing should be required for locally published artifacts,
+    // as they are used for manual testing and running integration tests only
+    setRequired(provider { gradle.taskGraph.allTasks.any { it is PublishToMavenRepository } })
 }
 
 // This is a hack for a Gradle 8 problem, see https://github.com/gradle/gradle/issues/26091


adam-enko:
Something like this definitely makes sense.

I suspect checking the task graph is a little more fragile. Instead, what about only signing if the signing key is present?

```kotlin
signing {
    val signKeyId = System.getenv("DOKKA_SIGN_KEY_ID")?.takeIf(String::isNotBlank)
    val signKey = System.getenv("DOKKA_SIGN_KEY")?.takeIf(String::isNotBlank)
    val signKeyPassphrase = System.getenv("DOKKA_SIGN_KEY_PASSPHRASE")?.takeIf(String::isNotBlank)
    useInMemoryPgpKeys(signKeyId, signKey, signKeyPassphrase)
    sign(publishing.publications)
    setRequired({ signKeyId != null })
}

```

> I don't remember if project.version is used someplace else aside from the docs-developer module

The project version is used in a few places (of the top of my head it's passed into the Maven and Gradle plugins).

whyoleg:
> Instead, what about only signing if the signing key is present?

Yeah, this sounds more clear. The only theoretical problem is if during changes to build logic or CI, we accidentally broke something and so signing will be disabled and we most likely will know about it only during publish to Maven Central, as only it checks if the artefacts are signed. At the same time we sign for both publish to MC and to Space dev/test repositories.

Still, both variants are fine for me, may be a little more in favour of @adam-enko proposal, because it looks more clear and concise.

IgnatBeresnev:
> Instead, what about only signing if the signing key is present?

It's certainly an option, but I'd like to keep the check if at all possible: while it likely won't be possible to publish to Maven Central without signing everything, we didn't sign our `-dev` artifacts for a very long time without realizing it, until it was brought up by Google. Having the check just feels more robust when it comes to CI refactorings and such (if the property is renamed or a new configuration is created), especially if done by people outside of Dokka (such as the build infra team)

Maybe it would be easier to do the inverse - require signing if publishing to dev/test repositories, without checking the task graph?

The ideal solution would be something like this, but it looks to be impossible

```kotlin
publishing {
    repositories {
        maven {
            name = "mavenCentral"
            ...
            signing { 
                setRequired(true)
            }
        }
        maven {
            name = "spaceDev"
            ..
            signing { 
                setRequired(true)
            }
        }
        maven {
            name = "spaceTest"
            ..
            signing { 
                setRequired(true)
            }
        }
    }
}
```

Or maybe something like `task.publishToSpaceDev.dependsOn { requireNotNull(System.getenv("DOKKA_SIGN_KEY")) }`

Basically, any solution that

1) Doesn't require signing locally published artifacts
2) Makes sure that `DOKKA_SIGN_KEY` and others are set for a number of repositories (dev/test/central)

I tried all I could think of - no luck thus far :(

adam-enko:
All of the publish-to-remote-repo tasks are of type `PublishToMavenRepository`, while the local publishing tasks are of type `PublishToMavenLocal`.

So, a typesafe way of the check in the PR would be:

```kotlin
signing {
    setRequired(provider { gradle.taskGraph.allTasks.any { it is PublishToMavenRepository } })
}
```


If, in the future, we needed to opt in/out of signing on a repo-by-repo basis, then we could do it by keeping track of the repository names.

Here's a quick example:


```kotlin
enum class PublicationRepos(val isSigningRequired: Boolean = true) {
    MavenCentral,
    SpaceDev,
    SpaceTest(isSigningRequired = false),
}

tasks.withType<PublishToMavenRepository>().configureEach {
    val signingRequired = provider { signing.isRequired }
    doFirst {
        val isSigningRequiredForRepo = PublicationRepos.values()
            .firstOrNull { it.name == publication.name }
            ?.isSigningRequired
            ?: error("Unknown ${repository.name} - either fix the name, or add it to PublicationRepos")

        if (isSigningRequiredForRepo) {
            require(signingRequired.get()) { "Cannot publish to ${repository.name}. Signing is required, but is not enabled." }
        }
    }
}
```

whyoleg:
> So, the check could use the task type

I do like this, it feels better than strings :)

> If, in the future, we needed to opt in/out of signing on a repo-by-repo basis

I do think, that we don't need to require signing only for local publishings, so I believe we will not need it. But thanks for the possible solutions here!

IgnatBeresnev:
> All of the publish-to-remote-repo tasks are of type PublishToMavenRepository, while the local publishing tasks are of type PublishToMavenLocal.
>
> So, the check could use the task type:
>
> ```
> signing {
>     setRequired(provider { gradle.taskGraph.allTasks.any { it is PublishToMavenRepository } })
> }
> ```

Sweet, I didn't know that! It's definitely better than using task names :+1: 

I've checked it just in case - seems to work! Works for `publishToMavenLocal`, fails for publishing to Space Dev.

Pushed the fix as per your proposal

whyoleg (approved):

adam-enko (approved):

PR #3562, opened 2024-04-10
Allow publishing to a pre-defined staging repository

IgnatBeresnev:
Due to Gradle running publishing tasks in parallel, multiple staging repositories can be created within a couple of seconds with all artifact files scattered throughout them. Even though the publishing in done on the same build agent.

Interestingly enough, setting `-Dorg.gradle.parallel=false` didn't help, I believe it's due to us using included builds, but I'm not sure.

Setting `--max-workers=1` additionally did help, but significant time was spent to find and test this solution.

Given it's easy to forget to set these params (especially when adding new build steps in TC), I think it's much easier to just allow to publish to a predefined staging repo, which will be created before the publishing build is run (as a prerequisite).

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts
+++ b/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts
@@ -1,3 +1,5 @@
+import java.net.URI
+
 /*
  * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
@@ -13,7 +15,7 @@ publishing {
     repositories {
         maven {
             name = "mavenCentral"
-            url = uri("https://oss.sonatype.org/service/local/staging/deploy/maven2/")
+            url = mavenCentralRepositoryUri()
             credentials {
                 username = System.getenv("DOKKA_SONATYPE_USER")
                 password = System.getenv("DOKKA_SONATYPE_PASSWORD")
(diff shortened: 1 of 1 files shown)


whyoleg (approved):

adam-enko:
Makes sense to me. 

It'd also be worth replacing the config-cache disabler here

https://github.com/Kotlin/dokka/blob/7fc9244aae018cbfc0c136cf0a6ce98a70fec39f/build-logic/src/main/kotlin/dokkabuild.publish-base.gradle.kts#L92-L97

And instead using a build service to limit task parallelization: 

```kotlin
//region Maven Central can't handle parallel uploads, so limit parallel uploads with a service.
abstract class MavenPublishLimiter : BuildService<BuildServiceParameters.None>

val mavenPublishLimiter =
  gradle.sharedServices.registerIfAbsent("mavenPublishLimiter", MavenPublishLimiter::class) {
    maxParallelUsages = 1
  }

tasks.withType<PublishToMavenRepository>().configureEach {
  usesService(mavenPublishLimiter)
}
//endregion
```

vmishenev (approved):

adam-enko:
> Good idea! Additionally, AFAIU, this could be dropped at all when `DOKKA_MVN_CENTRAL_REPOSITORY_ID` is set, as there should be no issues then. Am I correct?

In theory yes, AFAIK Maven Central is the only problematic repo. But in practice I would lean towards keeping the build scripts simpler, even if it introduces a performance hit when publishing :). Is parallel publishing important for Dokka? Would it help significantly?

whyoleg:
>  But in practice I would lean towards keeping the build scripts simpler, even if it introduces a performance hit when publishing :). Is parallel publishing important for Dokka? Would it help significantly?

Parallel publishing is not that important, but, as I mentioned in the end, as we intend to publish to MC only from TC with `DOKKA_MVN_CENTRAL_REPOSITORY_ID` set, then we don't need to have additional setup in build scripts to `de-parallelize` publishing, as it will be just not needed at all, as it's just a hack for MC behaviour.
So by removing this, we will keep build scripts simpler :)

IgnatBeresnev:
@adam-enko @whyoleg thank you for the proposals!

Indeed, it seems like the only repo struggling with parallel uploads is Maven Central, so if we start using `DOKKA_MVN_CENTRAL_REPOSITORY_ID`, the parallelization limitations could be removed.

I'll merge this PR in to unblock me with adapting TeamCity build configurations, and will create a separate PR with the removal of config-cache disabler - there might be more questions there

PR #3567, opened 2024-04-15
[K2] Fix the test `kotlin_Enum should not have obvious members via ex…

vmishenev:
…ternal documentable provider` for jdk >=18

https://github.com/Kotlin/dokka/actions/runs/8703942539


whyoleg (approved):
I don't really like, that we just duplicate the test with minor changes (hard to compare what is different), but looks like we have no other choice...

vmishenev:
I did some refactoring to better understand the difference between the tests.

PR #3568, opened 2024-04-15
Compare DokkaSourceSetImpl by DokkaSourceSetID only

### LINKED ISSUE #3246: Compare `DokkaSourceSetImpl` by `DokkaSourceSetID` only
## Context

This issue is a small subtask of #2620. Whereas the parent issue has a wider scope and will take more time to address, this issue should be easy and quick to resolve, but at the same time it should yield noticeable performance and other improvements.

## Background

`DokkaSourceSet` is used as a map key in `SourceSetDependent` (typealias to `Map<DokkaSourceSet, T>`), which is used in many documentable properties: `sources`, `visibility`, `modifier`, `supertypes`, `documentation` and others. `DokkaSourceSet` also contains many collection-based properties, some of which (like `classpath` and `sourceRoots`) can have quite a number of elements.

The only implementation `DokkaSourceSetImpl` is a data class, so its `equals` and `hashCode` implementations are auto-generated and compare all properties, including the huge lists.

Because of this, `DokkaSourceSet#hashCode` is called frequently enough and takes enough time to notice it when profiling Dokka (based on the research done by our Google colleagues).

Making it a non-data class can lead to breaking backward compatibility and other issues, so it should be addressed in #2620, but there's a simple quick fix for now.

## Task

In theory, `DokkaSourceSet` should be identifiable by its `DokkaSourceSet#sourceSetID`, which should be unique. However, we don't have this behaviour specified anywhere and there are no checks that verify it.

This issue can address it:

1. Specify in KDocs for `DokkaSourceSet` that it **must** be uniquely identifiable by its `sourceSetID` property, which **must** be unique across the whole project (even in a multimodule one).
2. Write a `preGenerationCheck` check that verifies that all source set ids are unique (don't forget about `AllModulesPagePlugin`).
3. Override `hashCode` and `equals` in `DokkaSourceSetImpl` to use `sourceSetID` only.

whyoleg:
probably fixes #3246 

Note: PR is not ready to be merged, documentation is not added.

There are several questions which should be resolved first:
1. After implementation of `pre-generation` check I found out, that those checks are not run by default in TESTS - in this PR, I've enabled them to run by default (one test started to fail: `JavadocDocumentableJVMSourceSetFilterTest` because of `js` sourceSet present, which is not supported by javadoc plugin) - does anyone have anything against enabling them, as I do expect them to run?
2. It's not possible on current moment to create a `pre-generation` check for sourceSet IDs uniqueness during multi-module generation, as we don't have information about sourceSets there, only the result of Dokka execution - Do you think it's fine? I'm a little worried about this.
3. when comparing execution time of Dokka running with and without this commit - I haven't seen a big difference (for serialisation and coroutines) - better to recheck
4. We do also implement DokkaSourceSet in CLI and in tests separately - I don't think that we need to change something there, as now they are compared by identity and everything works find - WDYT?
5. Just out of curiosity I've replaced comparing by sourceSet IDs with comparing by identity - this caused failure of several tests (screenshot attached) - in those tests we do compare or `DokkaConfiguration` or `DModule` in `assertEquals`. So if we will change `equals` of `DokkaSourceSet` to use ID only, such tests will just silently stop working - we could fix those, but what about future tests and overall complexity of understanding this behaviour? WDYT?

image

### DIFF
--- a/dokka-runners/runner-cli/src/main/kotlin/org/jetbrains/dokka/SourceSetArgumentsParser.kt
+++ b/dokka-runners/runner-cli/src/main/kotlin/org/jetbrains/dokka/SourceSetArgumentsParser.kt
@@ -161,5 +161,13 @@ internal fun parseSourceSet(moduleName: String, args: Array<String>): DokkaConfi
         override val suppressedFiles = suppressedFiles.toMutableSet()
         override val documentedVisibilities: Set<DokkaConfiguration.Visibility> = documentedVisibilities.toSet()
             .ifEmpty { DokkaDefaults.documentedVisibilities }
+
+        override fun equals(other: Any?): Boolean {
+            return sourceSetID == (other as? DokkaConfiguration.DokkaSourceSet)?.sourceSetID
+        }
+
+        override fun hashCode(): Int {
+            return sourceSetID.hashCode()
+        }
     }
 }
--- a/dokka-runners/runner-gradle-plugin-classic/build.gradle.kts
+++ b/dokka-runners/runner-gradle-plugin-classic/build.gradle.kts
@@ -25,6 +25,7 @@ dependencies {
     testImplementation(libs.gradlePlugin.kotlin)
     testImplementation(libs.gradlePlugin.kotlin.klibCommonizerApi)
     testImplementation(libs.gradlePlugin.android)
+    testImplementation("org.jetbrains.dokka:dokka-test-api:$version")
 }
 
 @Suppress("UnstableApiUsage")
--- a/dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/DokkaConfigurationJsonTest.kt
+++ b/dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/DokkaConfigurationJsonTest.kt
@@ -12,6 +12,7 @@ import org.jetbrains.dokka.PluginConfigurationImpl
 import org.jetbrains.dokka.gradle.utils.create_
 import org.jetbrains.dokka.gradle.utils.externalDocumentationLink_
 import org.jetbrains.dokka.gradle.utils.withDependencies_
+import org.jetbrains.dokka.testApi.assertDokkaConfigurationEquals
 import org.jetbrains.dokka.toCompactJsonString
 import java.io.File
 import java.net.URI
@@ -66,7 +67,7 @@ class DokkaConfigurationJsonTest {
         val configurationJson = sourceConfiguration.toCompactJsonString()
         val parsedConfiguration = DokkaConfigurationImpl(configurationJson)
 
-        assertEquals(sourceConfiguration, parsedConfiguration)
+        assertDokkaConfigurationEquals(sourceConfiguration, parsedConfiguration)
         println(parsedConfiguration)
     }
 }
--- a/dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/DokkaConfigurationSerializableTest.kt
+++ b/dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/DokkaConfigurationSerializableTest.kt
@@ -11,6 +11,7 @@ import org.jetbrains.dokka.PluginConfigurationImpl
 import org.jetbrains.dokka.gradle.utils.create_
 import org.jetbrains.dokka.gradle.utils.externalDocumentationLink_
 import org.jetbrains.dokka.gradle.utils.withDependencies_
+import org.jetbrains.dokka.testApi.assertDokkaConfigurationEquals
 import org.junit.jupiter.api.io.TempDir
 import java.io.File
 import java.io.ObjectInputStream
@@ -73,6 +74,6 @@ class DokkaConfigurationSerializableTest {
             stream.readObject() as DokkaConfiguration
         }
 
-        assertEquals(sourceConfiguration, parsedConfiguration)
+        assertDokkaConfigurationEquals(sourceConfiguration, parsedConfiguration)
     }
 }
--- a/dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/tasks/DokkaCollectorTaskTest.kt
+++ b/dokka-runners/runner-gradle-plugin-classic/src/test/kotlin/org/jetbrains/dokka/gradle/tasks/DokkaCollectorTaskTest.kt
@@ -16,6 +16,7 @@ import org.jetbrains.dokka.gradle.utils.all_
 import org.jetbrains.dokka.gradle.utils.allprojects_
 import org.jetbrains.dokka.gradle.utils.configureEach_
 import org.jetbrains.dokka.gradle.utils.withDependencies_
+import org.jetbrains.dokka.testApi.assertDokkaConfigurationEquals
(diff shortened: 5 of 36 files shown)


IgnatBeresnev:
> After implementation of pre-generation check I found out, that those checks are not run by default

Note to other reviewers: this is about unit tests only. I got scared for a second when reading this sentence :cold_sweat: I think it's correct to enable it, yeah 

> It's not possible on current moment to create a pre-generation check for sourceSet IDs uniqueness during multi-module generation, as we don't have information about sourceSets there, only the result of Dokka execution - Do you think it's fine? I'm a little worried about this.

Why are you worried about this, what problems could it lead to? 

To me that's expected: source sets are only visible when you open a specific module, and each module is processed separately. The plugins are also written as if for a single module, so hashCode / equals should as expected

> when comparing execution time of Dokka running with and without this commit - I haven't seen a big difference (for serialisation and coroutines) - better to recheck

When don't use `DokkaSourceSet` in maps/sets much, so it's also expected that we wouldn't see a noticeable difference. Google's plugin does use it heavily from what I understand, and the folks have done some profiling and identified it as one of the reasons (the other one being the inclusion of sources, such as descriptors/psi, in hashCode/equals of most documentables)

> We do also implement DokkaSourceSet in CLI and in tests separately - I don't think that we need to change something there, as now they are compared by identity and everything works find - WDYT?

If the spec (i.e documentation) states that the implementations should compare it by DokkaSourceSetID, I think the behavior should be aligned. I doubt that comparing by identity here would lead to bugs, but I'd change it just for consistency and to cause less questions later on

> Just out of curiosity I've replaced comparing by sourceSet IDs with comparing by identity - this caused failure of several tests (screenshot attached) - in those tests we do compare or DokkaConfiguration or DModule in assertEquals. So if we will change equals of DokkaSourceSet to use ID only, such tests will just silently stop working - we could fix those, but what about future tests and overall complexity of understanding this behaviour? WDYT?

Good thought to check it! 

I wouldn't say they will "stop working" - they will be working correctly from the specification's (documentation's) point of view, but I see your concern - they might deserialize some fields of a source set incorrectly and it will pass the equals check

I would actually expect specialized serialization/deserialization tests to be checking every field one by one, not just compare the two topmost objects by `equals`, because it is fragile otherwise. These tests basically already rely on the implementation and it being a data class that compares everything by default, which is not guaranteed at all (equals doesn't have to and most of the time doesn't compare everything - for instance, collections/arrays are frequent to be omitted), and I think the tests should be fixed regardless of this change (now or later)

As to how to fix the tests - that's a good question. Do you have any thoughts here? 

I only have two ideas atm: 
1. add more asserts that check individual fields one by one - more code to write, but better for the tests in case they fail
2. take inspiration from `Objects.deepEquals` and add a `fun deepEquals(other: DokkaSourceSet)` to either `DokkaSourceSetImpl` or even `DokkaSourceSet` - this would compare literally everything, same as before, and we can adapt the tests to use it instead of the regular equals; the documentation for it would state that it should compare all underlying fields, but it will be relatively slow

whyoleg:
> Note to other reviewers: this is about unit tests only.

Oops, my bad, fixed the PR description.

> Why are you worried about this, what problems could it lead to?

I believe there should be no problems here for now. May be I'm a bit too worried about such change, and so not feeling safe without all possible checks.
But mostly it's for reference to the original issue #3246 where it mentions `(don't forget about AllModulesPagePlugin)` :)

> We don't use Documentable / DokkaSourceSet in maps/sets/etc much

I would disagree, as we do have `SourceSetDependent` (which is a map with DokkaSourceSet as a key) used 45 times just in `org/jetbrains/dokka/model/Documentable.kt` (4/5 shared properties per documentable entity). And we use pattern `documentable.property[sourceSet]` quite frequently in our code. So I would expect some difference, as there is no need to perform a lot of computations there. May be it's reproduced on bigger projects or slower hardware.
I'm fine with letting Google test the performance difference afterwards on their setup.

> (the other one being the inclusion of sources, such as descriptors/psi, in hashCode/equals of most documentables)

Just in case, for reference, the issue is here: https://github.com/Kotlin/dokka/issues/2620

>  I doubt that comparing by identity here would lead to bugs, but I'd change it just for consistency and to cause less questions later on

yeah, exactly my thoughts, just wanted to recheck

> As to how to fix the tests - that's a good question. Do you have any thoughts here?

As `DokkaSourceSet` is just an interface and I can't think of any case when it could be needed to compare it's content except in our tests, I would say that we could just introduce `assertSourceSetEquals` (and may be the same for other entities needed) in `core-test-api`/`core textFixtures` and use it in places where those tests failed.
I will add documentation for `DokkaSourceSet`, that it's hashCode/equals relies on `ID` only and this should be enough for understanding, that `equals` should not be used for structural comparing.

I will have last provocative question - what about overall making `DokkaSourceSet` `equals/hashCode` use just identity? AFAIU all over the code we don't really modify it anyway? As this is the `configuration` part, not the `documentable` part (which properties are frequently changed)?

Big thanks for such detailed review! I will be able to finish the PR based on this discussion.

whyoleg:
PR is ready for review:
* I've decided to put asserts in `dokka-test-api` module
* to be able to use `dokka-test-api` from Gradle plugin, I was needed to rename `Gradle project name` (doesn't affect published artefact) in the same way as we have done for `dokka-core`

IgnatBeresnev:
Nice idea about adding `DokkaConfiguration` asserts to `dokka-test-api`!

> what about overall making DokkaSourceSet equals/hashCode use just identity? AFAIU all over the code we don't really modify it anyway? As this is the configuration part, not the documentable part (which properties are frequently changed)?

If we were designing it from scratch, I don't think I would mind, because it's indeed only created during the configuration, so it shouldn't change, and I can't think of a reason for why someone might create their own implementation during the rest of the steps (documentable, content, renderer, etc)

Right now though, given that we already have `DokkaSourceSetID` which is being used to identify DokkaSourceSets, I think it would add to confusion if equals started using identity (or anything else, really). I guess `DokkaSourceSetID` could be renamed to something like `DokkaSourceSetScope` to avoid that confusion

I've also just now remembered that we have `dependentSourceSets: Set<DokkaSourceSetID> `

https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/configuration.kt#L178

and `DokkaSourceSetID` is used to check the dependencies between source sets, pretty much doing `DokkaSourceSet.equals(other)`

https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt#L43-L44

I _think_ this part could also be refactored to use identity, but that comes with a whole new can of worms. But without this refactoring, I think it would be confusing and possibly error prone to use both `DokkaSourceSetID` and identity for comparing source sets

IgnatBeresnev (approved):

vmishenev (approved):
I would like to know if this change breaks, e.g. the Google plugin.

> it's also expected that we wouldn't see a noticeable difference. 

Me too. And I am not even sure about the Google plugin. In my opinion, the description of the issue is too promising) 
I would use a flame graph to compare the performance.
But, anyway, it is nice to fix the comparison.

whyoleg:
> I would like to know if this change breaks, e.g. the Google plugin.

`dokka-devsite-plugin` builds successfully (with the patch from https://github.com/Kotlin/dokka/compare/sample-analysis-manual), all tests are green. No significant tests execution time difference.

PR #3581, opened 2024-04-23
Update Maven Integration Tests to use JVM Test Suites

adam-enko:
KT-64200

- Updates `dokka-integration-tests/maven` to use jvm-test-suites
- De-duplicates integration-test configuration into the `dokkabuild.test-integration.gradle.kts` convention plugin.

  - Note that the cli integration test suite was renamed from `integrationTest` to `cliIntegrationTest` to avoid clashing with the `integrationTest` lifecycle task.

- Implement a new utility for passing in system properties to test tasks, and correctly registering them as Gradle inputs.

  Using System Properties is preferred to environment variables, because they are easier to register as Gradle task inputs.


whyoleg (approved):

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -63,223 +60,119 @@ kotlin {
...
-    doFirst("set DOKKA_VERSION environment variable (workaround for https://github.com/gradle/gradle/issues/24267)") {
-        environment("DOKKA_VERSION", dokkaVersion.get())
-    }
-
-    // environment() isn't Provider API compatible yet https://github.com/gradle/gradle/issues/11534
-    fun environmentProvider(name: String, provider: Provider<out Any>) {
-        inputs.property(name, provider).optional(true)
-        provider.orNull?.let { environment(name, it) }
-    }
+    val enableDebug = providers.environmentVariable("ENABLE_DEBUG")
```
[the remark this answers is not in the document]
adam-enko: Good question, it wasn't specified in the Gradle config, but `ENABLE_DEBUG` is read in the Gradle tests 

https://github.com/Kotlin/dokka/blob/09437c33bae16dd6432129db7b5706577103d3ba/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestEnvironment.kt#L16

dokka-integration-tests/maven/build.gradle.kts
```diff
@@ -43,25 +50,64 @@ dependencies {
...
     dependsOn(tasks.installMavenBinary)
-    val mvn = mavenCliSetup.mvn
-    inputs.file(mvn)
+    systemProperty.inputFile("mavenBinaryFile", mavenCliSetup.mvn)
+        .withPathSensitivity(NAME_ONLY)
+}
 
-    val dokkaVersion = provider { project.version.toString() }
-    inputs.property("dokkaVersion", dokkaVersion)
+testing.suites.withType<JvmTestSuite>().configureEach {
```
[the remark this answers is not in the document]
whyoleg: yeah, I get it. I just don't think, that we will ever need to remove dependency on project itself in integration tests :)

**Off-top**:
For me it's always looks very unnatural that for test-suites to use code declared in `main` I need to declare direct dependency on it. For me it looks like something super strange, that inside one project you need to manually wire `main` and `integrationTest` (example) source sets via declaring dependency on "itself"...
But, if Gradle decided that way - then there are use-cases to do this

dokka-integration-tests/maven/src/main/kotlin/settingsXml.kt
```diff
@@ -12,7 +12,7 @@ import java.nio.file.Paths
 
 /** Create `settings.xml` file contents, with the custom dev Maven repos. */
 @Language("xml")
-fun createSettingsXml(): String {
+public fun createSettingsXml(): String {
```
ALikhachev, line 15: Probably it makes sense to disable explicit mode in the integration tests convention plugin

ALikhachev (approved):
LGTM

a side note: would be nice to move test projects from the project leveldirectory `projects` to the resources of the related test suit source set

build-logic/src/main/kotlin/dokkabuild/DevMavenPublishExtension.kt
```diff
@@ -42,14 +49,13 @@ abstract class DevMavenPublishExtension(
...
 
-        if (task is JavaForkOptions) {
-            task.doFirst("devMavenRepositories systemProperty") {
-                // workaround https://github.com/gradle/gradle/issues/24267
-                task.systemProperty(
-                    "devMavenRepositories",
-                    devMavenRepositories.joinToString(",") { it.canonicalFile.invariantSeparatorsPath }
-                )
-            }
+        if (task is Test) {
```
whyoleg, line 52: BTW, is there a case when we need it for non-`Test` task? May be just make `configureTask` accept `Test`?
I mean, it has a rather specific usage pattern.

build-logic/src/main/kotlin/dokkabuild/utils/SystemPropertyAdder.kt
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Add a System Property (in the format `-D$key=$value`).
+     *
+     * [value] will be treated as if it were annotated with [org.gradle.api.tasks.Internal]
+     * and will _not_ be registered as a Gradle [org.gradle.api.Task] input.
+     * (Which is beneficial in cases where [value] is optional, or is not reproducible,
+     * because such a property might disrupt task caching.)
+     *
+     * If you want to register the property as a Task input, use the
+     * [Test.systemProperty][dokkabuild.utils.systemProperty] above instead.
```
whyoleg, line 112: should be rephrased now

whyoleg (commented):
just some minor things after latest commits :)

build-logic/src/main/kotlin/dokkabuild/DevMavenPublishExtension.kt
adam-enko: Good idea, done.

build-logic/src/main/kotlin/dokkabuild/utils/SystemPropertyAdder.kt
adam-enko: Sorry, how do you think it could be rephrased?

whyoleg: I mean, it's now a function inside an object which is created by `[Test.systemProperty]`, so may be suggesting just using other property will be more appropriate here, like:
```suggestion
     * [inputProperty] above instead.
```

adam-enko: Good point, thanks for explaining. I've updated the kdoc.

whyoleg:
@adam-enko looks like after merging this PR one of the workflows started to fail: https://github.com/Kotlin/dokka/actions/runs/9579287595/job/26411415134
Could you take a look?

whyoleg:
> I wonder why the failure wasn't caught in the PR?

It doesn't run on PRs from `dokka` repository - only for external PRs. Plus `master` of course :)

PR #3583, opened 2024-04-24
Setup running a UI test project via Gradle integration tests infrastructure

whyoleg:
After some thoughts I've decided to put UI test project near all other Gradle integration tests because:
* it's easier this way to reuse all the infrastructure that is available for running Gradle and storing it's output
* it will be nice to run at least sanity checks, f.e that ui project builds, produces output and there is no `ERROR CLASS` and unresolved links there, so that if something is broken during development we will know proactively

Project is now located in `dokka-integration-tests/gradle/projects/ui`.
To run UI test and get HTML output (execute in the root of repository):
```bash
export DOKKA_TEST_OUTPUT_PATH="build/uiproject"
./gradlew :dokka-integration-tests:gradle:testUiShowcaseProject
```
> output will be located in `dokka-integration-tests/gradle/build/uiproject` in this specific example;
> it's also possible to provide absolute path to `DOKKA_TEST_OUTPUT_PATH`

The test will use `TestedVersions.LATEST` for execution.

@berezinant Please check, that this will work fine at your side. Feel free to ask additional questions


whyoleg:
@adam-enko could you please help. I've added additional integration test and looks like now it's executed in every batch on [TC](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Dokka_IntegrationTests/4591294?buildTab=log&logView=flowAware&pageFullLogBuildId=4591302&focusLine=1349&linesState=510.512.513.1209.1297.1321) :)

Note: K2 tests are failed because K2 analysis doesn't resolve `CPointer` - I will mute new test for K2

adam-enko:
Sure thing @whyoleg! I don't see anything obviously wrong at first glance...


Ah, it's in the docs https://www.jetbrains.com/help/teamcity/cloud/parallel-tests.html#Known+limitations

> A newly added test which is not yet known to TeamCity will run in each batch during the first run. Related YouTrack ticket: [TW-75913](https://youtrack.jetbrains.com/issue/TW-75913/Newly-added-tests-run-in-all-parallel-auto-generated-dependencies).

IgnatBeresnev:
> After some thoughts I've decided to put UI test project near all other Gradle integration tests

It should still be fine to have the actual UI test infra set up in `/dokka-integration-tests/ui`, since the test project can output the docs to any directory that is provided, right?

dokka-integration-tests/gradle/projects/ui/README.md
IgnatBeresnev, line 1: I'd rename the project from just `ui` to something like `ui-showcase` or `ui-demo`, just to make it more obvious what the contents are, since the other projects in the `projects` directory also somewhat produce and assert UI :)

IgnatBeresnev (approved):

dokka-integration-tests/gradle/projects/ui/README.md
whyoleg: Agree! Renamed to `ui-showcase` as it really shows most of the UI elements used there. Demo for me sounds not descriptive enough.

berezinant (approved):
I've tested the new task and it works well for me 👍 

As an addition, it also would be nice to have an opportunity generate output on UI showcase project agains specific Dokka's version, it would allow to compare versions quickly and simplify testing and reviewing of PRs

whyoleg:
> it also would be nice to have an opportunity generate output on UI showcase project agains specific Dokka's version

Created separate PR https://github.com/Kotlin/dokka/pull/3585 to not block this one and continue discussion there.

If it's fine, I will merge this one.

PR #3585, opened 2024-04-25
Allow overriding a Dokka version in integration tests

whyoleg:
An addition for https://github.com/Kotlin/dokka/pull/3583.

For UI showcase project, after PR is merged, it will be possible to run it like this:
```bash
./gradlew :dokka-integration-tests:gradle:testUiShowcaseProject -Porg.jetbrains.dokka.integration_test.dokkaVersionOverride=2.0.0-dev-329
```

It's also possible to add `org.jetbrains.dokka.integration_test.dokkaVersionOverride=2.0.0-dev-329` to `dokka-integration-tests/gradle.properties` (but, it's less recommended, as it could be unnoticed and affect tests behaviour)

Note: PR is based on master and not `ui-test-project` branch, so `testUiShowcaseProject` will be not available.

Note regarding implementation:
* it's easier to allow overriding version for all integration tests and not just for single `ui-showcase` project. Additionally it could be useful for reproducing bugs with older versions if needed via just property;
* it's possible to use `project.version` to override `DOKKA_VERSION` and then do the logic based on it - but it will cause rebuilding of all artefacts (because of `devMavenPublish`). It's possible to overcome this, but I didn't want to change a lot of code;
* additionally using `DOKKA_VERSION` for detecting some override could cause some unknown issues;
* so I've decided that separate property and separate env variable will be more safe, clear and `additive` change for this specific use case.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,6 +105,16 @@ However, if you need to run all integration tests locally, you can use the `inte
 If you need to run a specific test locally, you can run it from your IDE or by calling the corresponding Gradle
 task (for example, `:dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines`).
 
+It's possible to run integration tests with a custom Dokka version published to 
+[MavenCentral](https://central.sonatype.com),
+[dev](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev) or
+[test](https://maven.pkg.jetbrains.space/kotlin/p/dokka/test) 
+via `org.jetbrains.dokka.integration_test.dokkaVersionOverride` Gradle property:
+
+```bash
+./gradlew :dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines -Porg.jetbrains.dokka.integration_test.dokkaVersionOverride=2.0.0-dev-329
+```
+
 ## Infrastructure
 
 ### Java version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,7 +109,7 @@ It's possible to run integration tests with a custom Dokka version published to
 [MavenCentral](https://central.sonatype.com),
 [dev](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev),
 [test](https://maven.pkg.jetbrains.space/kotlin/p/dokka/test) or
-`mavenLocal` (in this case the version should contain `-local-` suffix, f.e `2.0.0-local-reproducing-bug`) 
+`mavenLocal` 
 via `org.jetbrains.dokka.integration_test.dokkaVersionOverride` Gradle property:
 
 ```bash
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,10 +105,11 @@ However, if you need to run all integration tests locally, you can use the `inte
 If you need to run a specific test locally, you can run it from your IDE or by calling the corresponding Gradle
 task (for example, `:dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines`).
 
-It's possible to run integration tests with a custom Dokka version published to 
+It's possible to run integration tests with a custom Dokka version published to
 [MavenCentral](https://central.sonatype.com),
-[dev](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev) or
-[test](https://maven.pkg.jetbrains.space/kotlin/p/dokka/test) 
+[dev](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev),
+[test](https://maven.pkg.jetbrains.space/kotlin/p/dokka/test) or
+`mavenLocal` (in this case the version should contain `-local-` suffix, f.e `2.0.0-local-reproducing-bug`) 
 via `org.jetbrains.dokka.integration_test.dokkaVersionOverride` Gradle property:
 
 ```bash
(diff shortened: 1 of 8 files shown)


IgnatBeresnev (approved):
This can indeed come in handy :+1:

CONTRIBUTING.md
```diff
@@ -105,6 +105,16 @@ However, if you need to run all integration tests locally, you can use the `inte
...
 task (for example, `:dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines`).
 
+It's possible to run integration tests with a custom Dokka version published to 
+[MavenCentral](https://central.sonatype.com),
+[dev](https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev) or
+[test](https://maven.pkg.jetbrains.space/kotlin/p/dokka/test) 
+via `org.jetbrains.dokka.integration_test.dokkaVersionOverride` Gradle property:
+
+```bash
+./gradlew :dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines -Porg.jetbrains.dokka.integration_test.dokkaVersionOverride=2.0.0-dev-329
```
IgnatBeresnev, line 115: It'd be nice to add this example to `dokka-integration-tests/gradle/projects/ui-showcase/README.md` as well

whyoleg: Done

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -94,13 +94,15 @@ tasks.withType<Test>().configureEach {
     }
 
     // environment() isn't Provider API compatible yet https://github.com/gradle/gradle/issues/11534
-    dokkaBuild.integrationTestExhaustive.orNull?.let { exhaustive ->
-        environment("isExhaustive", exhaustive)
-    }
-    dokkaBuild.androidSdkDir.orNull?.let { androidSdkDir ->
-        environment("ANDROID_HOME", androidSdkDir.invariantSeparatorsPath)
+    fun environmentProvider(name: String, provider: Provider<out Any>) {
+        inputs.property(name, provider)
```
whyoleg, line 98: @adam-enko FYI, without adding properties which are passed to `environment` to `inputs` tasks will always stay `up-to-date` when those properties are changed

berezinant (approved):
I've tested it, works well both with published and local versions of Dokka 👍

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -120,12 +123,39 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
...
+                        Supported versions are:
+                        - release versions like '2.0.0'
+                        - dev versions like `2.0.0-dev-329`
+                        - test versions like `2.0.0-test-49477c44dfc58e2702f4c145ff41190b39d117fb`
+                        - locally published (to mavenLocal) versions with `-local-` suffix like `2.0.0-local-reproducing-bug`
+                        """.trimIndent()
+                    )
+                }.also { repository ->
+                    println("Dokka version overridden with ${dokkaVersionOverride}. Using $repository for resolving Dokka")
+                }
```
adam-enko, line 155: What about simplifying the logic here?

So long as the correct version of Dokka gets picked up, does it matter which repo are defined? The tests won't be as strict, but that's what the integration tests are for.

```suggestion
                println("Dokka version overridden with ${dokkaVersionOverride}")
                // if `DOKKA_VERSION_OVERRIDE` environment variable is provided,
                //  we allow running tests on a custom Dokka version from specific repositories
                """
                maven("https://maven.pkg.jetbrains.space/kotlin/p/dokka/test")
                maven("https://maven.pkg.jetbrains.space/kotlin/p/dokka/dev")
                mavenCentral()
                mavenLocal()
                """.trimIndent()
```

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Huh, that's weird! I guess that's a consequence of the configuration cache.

Looks like it might be fixed in https://github.com/gradle/gradle/issues/11534 (finally!)

In the meantime we could convert them to set using CommandLineArgumentProvider instead - but in another PR.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -153,17 +184,17 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
 
         fun File.updateProjectLocalMavenDir() {
 
-            val dokkaDevMavenRepoMarker = "/* %{PROJECT_LOCAL_MAVEN_DIR}% */"
+            val dokkaMavenRepoMarker = "/* %{PROJECT_LOCAL_MAVEN_DIR}% */"
```
adam-enko, line 187: I think it's a little odd that the the variable name `PROJECT_LOCAL_MAVEN_DIR` doesn't match new current usage. I think should probably update (e.g. `DOKKA_IT_MAVEN_REPO`). WDYT?

adam-enko (changes requested):

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
whyoleg: Yeah, let's do it like this. Tried to be too careful :)

PR #3589, opened 2024-04-26
Provide a way to manually create SampleAnalysisEnvironment

### LINKED ISSUE #3573: Provide a way to manually create SampleAnalysisEnvironment
## Context

The API introduced for analyzing samples in #3195 does not fit Dackka (Google's plugin) well: Dokka works through processing batches of documentables in a pipeline, but Dackka has more of a service-like architecture, where it can invoke sample resolution from later stages.

## Problem

Invoking `SampleAnalysisEnvironmentCreator#use` for every resolution request in Dackka leads to stalling and other problems. Refactoring the architecture to batch sample resolution is also tricky.

## Solution

The following API was discussed with Google, and tested on Dackka, which seems to work and solve the problem stated above.

1. Make `SampleAnalysisEnvironment` closeable (`java.io.Closeable`).
2. Add `public fun create(): SampleAnalysisEnvironment` to `SampleAnalysisEnvironmentCreator`

The commit that has these changes: https://github.com/Kotlin/dokka/commit/2b6eb92892fa8ca794a1ea40a1117efd32106d05 ([sample-analysis-manual](https://github.com/Kotlin/dokka/compare/sample-analysis-manual) branch)

In order to merge these changes in, the following needs to be done:
1. Add a KDoc to `SampleAnalysisEnvironmentCreator#create`, stating that:
    * it's delicate API and should be avoided in favor of `#use`, because it can lead to memory leaks, concurrency issues or other unexpected problems, whereas `use` has a controlled environment.
    * the `SampleAnalysisEnvironment` that is created should be closed by the user manually, otherwise it may lead to leaks
2. Ideally, an opt-in annotation like `@DelicateDokkaApi` should be added, similar to [DelicateCoroutinesApi](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-delicate-coroutines-api/).
3. Add unit tests (there is one test already, but maybe more are needed)

## Other

The branch [sample-analysis-manual](https://github.com/Kotlin/dokka/compare/sample-analysis-manual) may be revived, although it's not a requirement - a new branch can be created.

whyoleg:
Fixes #3573

minor changes applied to original branch.
Note: `DelicateDokkaApi` description was taken from coroutines with minimal changes - it's description matches perfectly for our use cases.

Note: I haven't implemented mine suggestion from https://github.com/Kotlin/dokka/issues/3573#issuecomment-2074718683, as may be `Delicate` annotation is enough and we don't need to do additional tracking for niche use-case

### DIFF
--- a/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/DescriptorSampleAnalysisEnvironment.kt
+++ b/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/DescriptorSampleAnalysisEnvironment.kt
@@ -50,20 +50,7 @@ internal class DescriptorSampleAnalysisEnvironmentCreator(
         // avoid memory leaks through the compiler's ThreadLocals.
         // Might not be relevant if the project stops using coroutines.
         return runBlocking(Dispatchers.Default) {
-            @OptIn(DokkaPluginApiPreview::class)
-            SamplesKotlinAnalysis(
-                sourceSets = context.configuration.sourceSets,
-                context = context,
-                projectKotlinAnalysis = descriptorAnalysisPlugin.querySingle { kotlinAnalysis }
-            ).use { kotlinAnalysis ->
-                val sampleAnalysis = DescriptorSampleAnalysisEnvironment(
-                    kdocFinder = descriptorAnalysisPlugin.querySingle { kdocFinder },
-                    kotlinAnalysis = kotlinAnalysis,
-                    sampleRewriter = sampleRewriter,
-                    dokkaLogger = context.logger
-                )
-                block(sampleAnalysis)
-            }
+            create().use(block)
         }
     }
 
@@ -87,7 +74,7 @@ internal class DescriptorSampleAnalysisEnvironment(
     private val kotlinAnalysis: KotlinAnalysis,
     private val sampleRewriter: SampleRewriter?,
     private val dokkaLogger: DokkaLogger,
-) : SampleAnalysisEnvironment, Closeable {
+) : SampleAnalysisEnvironment {
 
     override fun resolveSample(
         sourceSet: DokkaConfiguration.DokkaSourceSet,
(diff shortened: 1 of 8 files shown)


dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/impl/DescriptorSampleAnalysisEnvironment.kt
```diff
@@ -49,21 +50,22 @@ internal class DescriptorSampleAnalysisEnvironmentCreator(
         // avoid memory leaks through the compiler's ThreadLocals.
         // Might not be relevant if the project stops using coroutines.
         return runBlocking(Dispatchers.Default) {
```
whyoleg, line 52: minor question: do we still need this here?
As we now provide `create()` (even if it's delicate), for me it looks a little strange, that `creator.use(block)` and `creator.create().use(block)` will behave differently, and most likely users of this API need to know about `internals` to work with it properly.
WDYT about this?

IgnatBeresnev: > minor question: do we still need this here?

I believe we do, otherwise it can lead to the mentioned issues :) This is the controlled environment that addresses the issues that we're aware of, and more hacks and fixes might be added here in the future

> for me it looks a little strange, that creator.use(block) and creator.create().use(block) will behave differently

yeah, the naming is unfortunate :( this is a lesson for the future - do not use names that can clash with the names from stdlib. otherwise, I'd say it's expected:

> most likely users of this API need to know about internals to work with it properly.

indeed, that is the whole point of `create()` being unsafe, delicate and NOT recommended, and `use()` being the safest and recommended option
* You're using `use()` and have encountered a bug/leak/etc? Report it, we'll have a look and try to resolve it (most likely by adding some hacks/workarounds to `use()`)
* You're using `create()` and have encountered a bug/leak/etc? Well, we gave no guarantees and warned you, you better research and figure out how to mitigate it on your own. Or start using `use()`, then it becomes our problem

___

If we remove `use()`, we'll have to add the `runBlocking` detail (and any other) to the KDoc of `create()`, but I see several problems with this approach:
* The users will have to read the docs to apply it - the chances of it are slim as it is
* If more issues arise, the KDoc will have to be updated with the new workarounds, and the users will have to revise it every release, which is also very unlikely
* If some issues are no longer relevant and the workarounds need to be removed - same story
* It kinda breaks the abstraction. This API can have multiple implementations with their own limitations and workarounds. Some might apply to K1, some to K2, some to both. Not to mention any potential 3rd party ones that we're not aware of

So to me it makes sense to have `use()`, which takes care of all this for you

___

We can soft-rename `use` to something else to not clash with `kotlin.io.use`, but let's do it as a separate issue/PR

IgnatBeresnev (approved):
:+1: :+1:

vmishenev (approved):

PR #3592, opened 2024-04-30
Remove log4j coming from intellij dependencies

### LINKED ISSUE #3508: analysis-kotlin-descriptors-1.9.10.jar is bundled with log4j:1.2.17.2
**Describe the bug**
Similar to #2488 

Seems that analysis-kotlin-descriptors-1.9.10.jar is bundled with log4j version 1.2.17.2. [DependencyCheck](https://github.com/jeremylong/DependencyCheck) is marking it `CRITICAL` via `.gradle/caches/modules-2/files-2.1/org.jetbrains.dokka/analysis-kotlin-descriptors/1.9.10/.../analysis-kotlin-descriptors-1.9.10.jar/META-INF/maven/log4j/log4j/pom.xml`

**Expected behaviour**
Upgrade to log4j version without security issues.

**To Reproduce**
Check the dependency tree

**Installation**

Operating system: Linux
Build tool: Gradle v8.5
Dokka version: 1.9.10 & 1.9.20

whyoleg:
Fixes #3508

Checked `shadowJar` contents before and after this change: the only difference is in the absence of `log4j` classes.
Note: we still include 2 classes from `log4j` - `Level` and `Priority` - they are coming from `kotlin-compiler` and don't trigger any dependency vulnerability checks.
We can drop those classes, and tests are working fine without them, but I've decided not to do this, as if they are used somehow in some situation, it will cause `No class found` exception.

Additionally, I've updated some comments and removed confusing dependencies which are not needed in `analysis-kotlin-symbols`, as they are coming from `analysis-java-psi` and `kotlin-compiler` anyway - no changes to those classes in final `shadowJar` have happened.

### DIFF
--- a/dokka-subprojects/analysis-java-psi/build.gradle.kts
+++ b/dokka-subprojects/analysis-java-psi/build.gradle.kts
@@ -10,14 +10,14 @@ dependencies {
     compileOnly(projects.dokkaSubprojects.dokkaCore)
     compileOnly(projects.dokkaSubprojects.analysisKotlinApi)
 
-    api(libs.intellij.java.psi.api)
+    // We exclude `log4j` as it's not used in our codebase,
+    // and we do override intellij logger with NOOP logger
+    // `log4j` dependency triggers errors by dependency vulnerability checkers
+    implementation(libs.intellij.java.psi.impl) {
+        exclude("org.jetbrains.intellij.deps", "log4j")
+    }
 
     implementation(projects.dokkaSubprojects.analysisMarkdownJb)
-
-    implementation(libs.intellij.java.psi.impl)
-    implementation(libs.intellij.platform.util.api)
-    implementation(libs.intellij.platform.util.rt)
-
     implementation(libs.kotlinx.coroutines.core)
     implementation(libs.jsoup)
 }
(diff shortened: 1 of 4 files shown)


vmishenev (approved):

dokka-subprojects/analysis-java-psi/build.gradle.kts
```diff
@@ -10,14 +10,14 @@ dependencies {
...
+    // and we do override intellij logger with NOOP logger
+    // `log4j` dependency triggers errors by dependency vulnerability checkers
+    implementation(libs.intellij.java.psi.impl) {
+        exclude("org.jetbrains.intellij.deps", "log4j")
+    }
 
     implementation(projects.dokkaSubprojects.analysisMarkdownJb)
-
-    implementation(libs.intellij.java.psi.impl)
-    implementation(libs.intellij.platform.util.api)
```
vmishenev, line 18: Can we remove them from toml as well?

whyoleg: yeah, dropped them

IgnatBeresnev (approved):

PR #3594, opened 2024-04-30
Update dead link to versioning plugin in its example

krocard:
The link was not updated when the plugins were moved from `plugins` to `dokka-subprojects`.


krocard:
This should only take a second to review :)

PR #3596, opened 2024-05-02
fix: made code samples runnable (#3535)

berezinant:
fixes the problem with runnable code snippets detected in #3535 and described in https://github.com/Kotlin/dokka/pull/3553


whyoleg:
Just curious: in future similar issues probably should be detected by ui-showcase screenshot tests, right? Is there a possibility there to check different browsers (Safari, Chrome, Firefox)? I believe in most cases just Chrome will be enough, so I'm just curious.

whyoleg (approved):
LGTM, works at my side

IgnatBeresnev (approved):

berezinant:
@whyoleg Cases like this are rather hard to prevent only by tests from inside Dokka because of how scripts are included on a page is not totally under our control
But yes, we'll be able to run tests in different browsers and we'll try to cover such cases if possible

PR #3605, opened 2024-05-13
Update Kotlin to 2.0.0-RC3

whyoleg:


dokka-runners/runner-gradle-plugin-classic/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinCompilationUtils.kt
```diff
@@ -6,14 +6,12 @@ package org.jetbrains.dokka.gradle.kotlin
...
 import org.jetbrains.dokka.gradle.kotlin
-import org.jetbrains.kotlin.gradle.dsl.KotlinCommonOptions
 import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
 import org.jetbrains.kotlin.gradle.dsl.KotlinSingleTargetExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinCommonCompilation
 
-internal typealias KotlinCompilation =
-        org.jetbrains.kotlin.gradle.plugin.KotlinCompilation<KotlinCommonOptions>
+internal typealias KotlinCompilation = org.jetbrains.kotlin.gradle.plugin.KotlinCompilation<*>
```
whyoleg, line 14: `KotlinCommonOptions` is deprecated

IgnatBeresnev (approved):

dokka-runners/runner-gradle-plugin-classic/src/main/kotlin/org/jetbrains/dokka/gradle/kotlin/kotlinCompilationUtils.kt
IgnatBeresnev: I'm surprised that it worked :) I thought we would need to use the replacements here, which could lead to compatibility problems with the older versions of KGP

vmishenev (approved):

PR #3617, opened 2024-05-20
Get information about expect/actual for properties in K2 via Analysis API and not via PSI

### LINKED ISSUE #3360: [K2] KMP: Platform tabs missed common for properties
Create a KMP project:
```Kotlin
// commonMain
expect class Dokka {
    val data: String?
    fun f()
}

// platformMain
actual class Dokka {
    actual val data: String? = null
    actual fun f() {}
}
```

K2 output:
image

**Installation**
- Dokka version: 1.9.20

Parent: #3328

whyoleg:
fixes #3360

After short investigation I found out that the issue is on our side - we had our own extension properties for `expect/actual` handling which were using just PSI information.
Accessors via Analysis API were added in https://youtrack.jetbrains.com/issue/KT-54846 (Fix in builds: 1.9.20-dev-7565).
There were no issues with functions, because `isExpect/isActual` were used there after update to Analysis API 1.9.20-* (because they were introduced as member properties and so they come before our extension properties during overload resolution). With properties the story is a bit different, because there is no `isExpect/isActual` for `KtPropertySymbol` which is used in our code, the exists only on `KtKotlinPropertySymbol`

So looks like upstream issue can be closed

### DIFF
--- a/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
@@ -740,8 +740,8 @@ private class DokkaDescriptorVisitor(
     ): DFunction {
         val dri = parent.copy(callable = Callable.from(descriptor))
         val isGetter = descriptor is PropertyGetterDescriptor
-        val isExpect = descriptor.isExpect
-        val isActual = descriptor.isActual
+        val isExpect = propertyDescriptor.isExpect
+        val isActual = propertyDescriptor.isActual
 
         suspend fun PropertyDescriptor.asParameter(parent: DRI) =
             DParameter(
--- a/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
@@ -740,8 +740,6 @@ private class DokkaDescriptorVisitor(
     ): DFunction {
         val dri = parent.copy(callable = Callable.from(descriptor))
         val isGetter = descriptor is PropertyGetterDescriptor
-        val isExpect = propertyDescriptor.isExpect
-        val isActual = propertyDescriptor.isActual
 
         suspend fun PropertyDescriptor.asParameter(parent: DRI) =
             DParameter(
@@ -798,7 +796,7 @@ private class DokkaDescriptorVisitor(
                 type = descriptor.returnType!!.toBound(),
                 generics = generics.await(),
                 modifier = descriptor.modifier().toSourceSetDependent(),
-                expectPresentInSet = sourceSet.takeIf { isExpect },
+                expectPresentInSet = null,
                 receiver = descriptor.extensionReceiverParameter?.let {
                     visitReceiverParameterDescriptor(
                         it,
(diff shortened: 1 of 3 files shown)


dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
```diff
@@ -740,8 +740,8 @@ private class DokkaDescriptorVisitor(
     ): DFunction {
         val dri = parent.copy(callable = Callable.from(descriptor))
         val isGetter = descriptor is PropertyGetterDescriptor
-        val isExpect = descriptor.isExpect
-        val isActual = descriptor.isActual
+        val isExpect = propertyDescriptor.isExpect
```
whyoleg, line 743: this is a minor fix for K1 compiler (just to make test easier).
`descriptor` of getter/setter in properties doesn't contain information about `expect/actual` (the same is true in K2, there is no such properties there).
So I've decided to make behaviour consistent. Not sure, that will affect anything rather than the model itself

dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
```diff
@@ -9,11 +9,102 @@ import org.jetbrains.dokka.model.withDescendants
...
+                displayName = "native"
+                dependentSourceSets = setOf(commonId)
+            }
+        }
+    }
+    private val commonSourceSetId =
+        multiplatformConfiguration.sourceSets.single { it.displayName == "common" }.sourceSetID
+
+    @Test
+    fun `should recognize expect property in class`() = testInline(
```
vmishenev, line 46: Can you emphasize that a declaration should be without  the `expect` keyword, please?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -481,26 +472,18 @@ internal class DokkaSymbolVisitor(
...
-                        index,
-                        symbol,
-                        dri
-                    )
-                }
+            val (isExpect, isActual) = when (propertySymbol) {
+                is KtKotlinPropertySymbol -> propertySymbol.isExpect to propertySymbol.isActual
+                is KtSyntheticJavaPropertySymbol -> false to false
+            }
+            val generics = propertySymbol.typeParameters.mapIndexed { index, symbol ->
```
vmishenev, line 479: Can you left the formatting unchanged for convenience, please?

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
vmishenev: > descriptor of getter/setter in properties doesn't contain information about expect/actual 

`isExpect/isActual `always returns `false` for accessors.
What is about to keep the same behavior in K2 to avoid the change in Documentable model?

vmishenev (approved):

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt
whyoleg: yeah, that's also fine I believe. Do you mind if I will change it for K1 also to just use `false` instead of `isExpect/isActual` properties to make it more clear?

dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
whyoleg: what do you mean? It's not possible to declare `expect val/fun` inside `expect class/interface/etc`

PR #3641, opened 2024-06-03
Fix the Wasm integration tests and enable it for K2

vmishenev:


dokka-integration-tests/gradle/projects/it-wasm-basic/src/wasmMain/kotlin/RootPackageClass.kt
```diff
@@ -16,9 +16,6 @@ fun test(list: MutableList<Int>) = "list"
 @JsNonModule
 external fun <T> sorted(a: Array<T>): Boolean
 
-//  this declaration can be used to check deserialization of dynamic type
-external interface TextLinkProps: AnchorHTMLAttributes<HTMLAnchorElement>
```
vmishenev, line 20: It was copied from[ Js integration test ](https://github.com/Kotlin/dokka/blob/8e5c63d035ef44a269b8c43430f43f5c8eebfb63/dokka-integration-tests/gradle/projects/it-js-ir-0/src/main/kotlin/RootPackageClass.kt)
but the import `import react.dom.html.AnchorHTMLAttributes` was missed

whyoleg: Couldn't we just add import here?
Or remove also other not used import `org.w3c.dom.HTMLAnchorElement`.

whyoleg (approved):

dokka-integration-tests/gradle/projects/it-wasm-basic/src/wasmMain/kotlin/RootPackageClass.kt
vmishenev: No, `AnchorHTMLAttributes` is from the `kotlin-wrappers:kotlin-react` that is not published for Wasm.

PR #3650, opened 2024-06-19
Use JetBrains cache-redirector

adam-enko:
Use JetBrains cache-redirector in Dokka build, Gradle integration tests, and Maven integration tests.

Update the Gradle Wrapper task to use the cache-redirector too, and add an explanation.

Part of [KT-64200](https://youtrack.jetbrains.com/issue/KT-64200)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,19 +12,20 @@ Bug reports, feature requests and questions are welcome. Submit issues [here](ht
 
 * Search for existing issues to avoid reporting duplicates.
 * When submitting a bug report:
-   * Test it against the most recently released version. It might have been fixed already.
-   * Include code that reproduces the problem. Provide a complete reproducer, yet minimize it as much as 
-     possible. A separate project that can be cloned is ideal.
-   * If the bug is in behavior, explain what behavior you expected and what the actual result is.
+    * Test it against the most recently released version. It might have been fixed already.
+    * Include code that reproduces the problem. Provide a complete reproducer, yet minimize it as much as
+      possible. A separate project that can be cloned is ideal.
+    * If the bug is in behavior, explain what behavior you expected and what the actual result is.
 * When submitting a feature request:
-   * Explain why you need the feature.
-   * Explaining the problem you face is more important than suggesting a solution.
-     Report your problem even if you have no proposed solution.
+    * Explain why you need the feature.
+    * Explaining the problem you face is more important than suggesting a solution.
+      Report your problem even if you have no proposed solution.
 
 ## Submitting PRs
 
-Dokka has extensive [Developer Guides](https://kotlin.github.io/dokka/1.9.20/developer_guide/introduction/) documentation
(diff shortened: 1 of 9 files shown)


build.gradle.kts
```diff
@@ -82,3 +90,26 @@ fun includedBuildTasks(taskName: String, filter: (IncludedBuild) -> Boolean = {
         .filter { it.name != "build-logic" }
         .filter(filter)
         .mapNotNull { it.task(":$taskName") }
+
+
+tasks.wrapper {
+    val gradleVersion = "8.7"
```
whyoleg, line 96: just to clarify: after PR merged to update the version of Gradle, we need:
1. set `gradleVersion` to the new value
2. go to `https://gradle.org/release-checksums/` and copy the checksum
3. set `distributionSha256Sum` to the new value
4. run `gradle wrapper`

correct?

Previously it was just `gradle wrapper --gradle-version=8.8`

whyoleg (approved):
LGTM, would be nice to leave a small comment on how to update Gradle wrapper now

adam-enko:
> LGTM, would be nice to leave a small comment on how to update Gradle wrapper now

Very good idea, I will do that.

build.gradle.kts
adam-enko: Good point, I will double check and document the steps. Normally I update `gradle.properties` but yes, there are multiple ways.

> Previously it was just gradle wrapper --gradle-version=8.8

And then re-run `gradle wrapper`, otherwise the wrapper scripts won't be updated 🙃

adam-enko: TIL how to update Gradle using `distributionSha256Sum`. It is not simple :(

Hopefully the instructions are clear - WDYT?

CONTRIBUTING.md
```diff
@@ -78,6 +78,40 @@ plugins {
 
 There is an automation script for this routine, see [testDokka.sh.md](scripts/testDokka.sh.md) for details.
 
+### Updating Gradle
```
whyoleg, line 81: Okay, that doesn't look nice :)

Also, I've followed your [link](https://docs.gradle.org/current/userguide/gradle_wrapper.html#configuring_checksum_verification) and there is a CLI property `--gradle-distribution-sha256-sum` which could simplify things.

So I've tried to use it, and it looks like everything is not that complicated.
We can safely remove `distributionSha256Sum` from `tasks.wrapper` configuration (leaving only rewriting of URL), and so we will have two options on how to update Gradle:
1. run `gradle wrapper --gradle-version=8.8 --gradle-distribution-sha256-sum=a4b4158601f8636cdeeab09bd76afb640030bb5b144aafe261a5e8af027dc612` (twice, or just `gradle wrapper` second time)
2. manually update both version and checksum in `gradle-wrapper.properties` and just `gradle wrapper` (to update scripts)

Both options work fine and looks like no complicated setup is needed at all.
In the end, looks like we need to just drop specifying `distributionSha256Sum` in `tasks.wrapper`.

adam-enko: That makes much more sense! Thanks. I'll update the instructions.

dkrasnoff (approved):

PR #3662, opened 2024-07-03
Correctly handle case when we have `internal expect` but `public actual` declarations

whyoleg:
fixes #3633

Issue #3386, opened 2023-11-29
[K2], KMP: Extension function miss platform tab for shared source set with one target


atyrin:
See the reproducer attached. Case is taken from okhttp

ModuleA: JVM + iosX64
```Kotlin
// commonMain
expect interface Call

// jvmMain
actual interface Call

//nativeMain
actual interface Call
```

ModuleB: JVM + iosX64 Depends on ModuleA
```Kotlin
// commonMain
expect fun Call.dokka()

// jvmMain
actual fun Call.dokka(){}

//nativeMain
actual fun Call.dokka(){}
```

For K2 the `dokka()` extension miss `native` tab
<img width="406" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/4073cbf4-ecc2-49ac-8d7e-a4483168c073">

**Installation**
- Dokka version: 1.9.20

Parent: https://github.com/Kotlin/dokka/issues/3328
[dokka-repros.zip](https://github.com/Kotlin/dokka/files/13503480/dokka-repros.zip)

vmishenev:
It is reproduced for `linuxX64` target on Ubuntu.

It is blocked by https://youtrack.jetbrains.com/issue/KT-64321/Analysis-API-Extension-functions-are-missed-for-shared-source-set-with-one-target

whyoleg:
First of all, the issue is not only about extension functions, but overall about resolving types from other modules **when there are intermediate source sets for one target**.
In such cases KGP will not create tasks to compile those shared source sets to `metadata` and so no compiler invocation will be done for such source sets. So for `module` there will be produced: `jvm.jar`, `iosx64.klib` and metadata for `commonMain` only.
If we add additional native target, like iosArm64 , then KGP will create task to compile `nativeMain` (and also `appleMain` and `iosMain`, but they are empty, so no task will be executed and no metadata produced) to metadata and so there will be following artefacts: `jvm.jar`, `iosx64.klib` and metadata for `commonMain` and `nativeMain`. In this case, all declarations will be resolved correctly.

Here are example for top-level properties/functions. Those with `Native` suffix are declared in `nativeMain` sourceSet, other declared in `commonMain` (as `expect`) and actualised in `jvmMain` and `nativeMain`:
<img width="1002" alt="image" src="https://github.com/Kotlin/dokka/assets/50462752/9cbe4e77-7370-402b-9502-c05bb6b02c22">
[...]

PR #3663, opened 2024-07-03
Use both macos targets in ui-showcase

whyoleg:
TL;DR;
This is needed because of the issue with rendering expect/actual. I would propose for now to just add `macosArm64` target and resolve underlying issue during implementation of https://github.com/Kotlin/dokka/issues/3386 for Dokka K2. We will anyway need to do something with such sourceSets.

In https://github.com/Kotlin/dokka/pull/3662 test started to fail on `ui-showcase` module because of `ERROR CLASS`.
Checking the page where the fail happened, I found that for some reason `ui-showcase` currently (on `master`) doesn't contain an actual declaration for `macos` sourceSet:
image
`asyncWithDelay` doesn't have `macos` tab, but `actual` declaration exists in sourceSet.

If we run the same test with the changes from https://github.com/Kotlin/dokka/pull/3662, `macos` tab is now appeared but shows `ERROR CLASS`:
image
All other declarations of `asyncWithDelay` for other sourceSets look the same as before.

Also, other declarations (which don't depend on `coroutines`) like `getCurrentDate` which also has `expect/actual` defined in the same - work fine on `master` and `macos` tab exist:
image

Most probably the issue is very similar to https://github.com/Kotlin/dokka/issues/3386 (even though problem described here reproducible on K1).

If we build `ui-showcase` with Dokka K2:
* on `master` - `linux` and `macos` are not shown at all for `asyncWithDelay`.
* with https://github.com/Kotlin/dokka/pull/3662 - all source sets are shown but both `linux` and `macos` will show `ERROR CLASS`
* if we add `macosArm64` target (as in this PR) - `macos` source set will be shown.

### DIFF
--- a/dokka-integration-tests/gradle/projects/ui-showcase/kmp/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/ui-showcase/kmp/build.gradle.kts
@@ -12,7 +12,11 @@ kotlin {
     jvm()
     linuxX64()
     macosX64()
+
+    // adding linuxArm64 and macosArm64 is a workaround for https://github.com/Kotlin/dokka/issues/3386
+    linuxArm64()
     macosArm64()
+
     js {
         nodejs()
     }
--- a/dokka-integration-tests/gradle/projects/ui-showcase/kmp/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/ui-showcase/kmp/build.gradle.kts
@@ -12,6 +12,7 @@ kotlin {
     jvm()
     linuxX64()
     macosX64()
+    macosArm64()
     js {
         nodejs()
     }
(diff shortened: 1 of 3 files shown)


vmishenev (commented):
> all source sets are shown but both linux and macos will show ERROR CLASS

Currently, does the `linux` tab have `<ERROR CLASS>.asyncWithDelay...` for K2, but K1 renders normally? 
Can we fix it with an extra target?

whyoleg:
> Currently, does the linux tab have <ERROR CLASS>.asyncWithDelay... for K2, but K1 renders normally?

Yeah, for K1 `linux` renders fine for some reason. May be because `coroutines` does have some `apple`/`macos` related stuff in publication which confuses Dokka.

> Can we fix it with an extra target as well?

Probably yes, I will recheck it one more time for both K1 and K2 and adapt PR

whyoleg:
Adding extra `linuxArm64` target not only solves `ERROR CLASS` for `asyncWithDelay` but also resolves issue with `CPointer` for K2 - so now, `ui-showcase` works for both K1 and K2.

vmishenev (approved):

PR #3687, opened 2024-07-09
use `macos-latest` for `ui-showcase` build on CI

whyoleg:
In https://github.com/Kotlin/dokka/pull/3662 test started to fail on ui-showcase module because of `ERROR CLASS`.
This was fixed in https://github.com/Kotlin/dokka/pull/3663 and everything was fine locally.
But on CI we were building `ui-showcase` on Linux - and so before #3662 there were no `macOS` specific declarations there.
So we need to build it on macOS.
Looks like we need to do the same on TC. @adam-enko could you help here? Is it possible to run only `UiShowcaseIntegrationTest` on TC on macOS? Other tests should work fine with linux runner.

### DIFF
--- a/.github/workflows/preview-publish-ga.yml
+++ b/.github/workflows/preview-publish-ga.yml
@@ -26,12 +26,12 @@ jobs:
       - name: Document coroutines
         run: ./gradlew :dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines --stacktrace "-Dorg.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=500m"
         env:
-          DOKKA_TEST_OUTPUT_PATH: /home/runner/work/dokka/coroutines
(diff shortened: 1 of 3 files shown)


adam-enko:
> Is it possible to run only `UiShowcaseIntegrationTest` on TC on macOS?

I think the best approach would be to skip tests on a case-by-case basis using JUnit.

```kotlin
import org.junit.jupiter.api.Assumptions.assumeTrue

@Test
fun foo() {
  assumeTrue(System.getProperty("os.name").lowercase().contains("mac"), "test requires macOS")
  // ...
}
```

adam-enko (approved):

vmishenev (approved):
I do not know how `ui-showcase` is used in ui testing. Is it ok for @berezinant to build the project only on Mac Os?

berezinant:
To build the project only on macOS will not be a problem for development
The only thing that could be affected is that we plan to run screenshot tests for ui-showcase project on CI and maybe there would be difficulties

PR #3690, opened 2024-07-11
Update to Kotlin 2.0.20-Beta2

whyoleg:


dokka-subprojects/analysis-kotlin-descriptors-ide/src/main/kotlin/org/jetbrains/kotlin/idea/klib/KlibCompatibilityInfo.kt
```diff
@@ -16,31 +16,24 @@ import java.io.IOException
...
 
-
 internal fun <T> KotlinLibrary.safeRead(defaultValue: T, action: KotlinLibrary.() -> T) = try {
     action()
 } catch (_: IOException) {
     defaultValue
 }
 internal val KotlinLibrary.compatibilityInfo: KlibCompatibilityInfo
     get() {
-        val hasPre14Manifest = safeRead(false) { has_pre_1_4_manifest }
```
whyoleg, line 31: `has_pre_1_4_manifest` was dropped

dokka-subprojects/analysis-kotlin-descriptors-ide/src/main/kotlin/org/jetbrains/kotlin/idea/klib/KlibCompatibilityInfo.kt
```diff
@@ -16,31 +16,24 @@ import java.io.IOException
  */
 internal sealed class KlibCompatibilityInfo(val isCompatible: Boolean) {
```
vmishenev, line 17: JIC this class can be copy-pasted from https://github.com/JetBrains/intellij-community/blob/master/plugins/kotlin/base/project-structure/src/org/jetbrains/kotlin/idea/base/projectStructure/moduleInfo/KlibCompatibilityInfo.kt

vmishenev (approved):

dokka-subprojects/analysis-kotlin-descriptors-ide/src/main/kotlin/org/jetbrains/kotlin/idea/klib/KlibCompatibilityInfo.kt
whyoleg: Thanks, I will add a reference to this file to the top of the file.
But overall, we don't need `AbstractKlibLibraryInfo`, otherwise, the content is the same

PR #3693, opened 2024-07-11
Support specifying custom folder name for older versions for versioning plugin

### LINKED ISSUE #3692: Make older versions dir name configurable
Dokka's versioning plugin puts previous versions into the directory named `older`, which is the default and not configurable at the moment. There was never enough reason to change it or make it configurable, but it now could be used for stdlib to make paths more logical, like this:
`https://kotlinlang.org/api/core/version/1.9/kotlin-stdlib/kotlin/-enum/`
Or even without any additional dedicated folder:
`https://kotlinlang.org/api/core/1.9/kotlin-stdlib/kotlin/-enum/`
Instead of current `older` folder:
`https://kotlinlang.org/api/core/older/1.9/kotlin-stdlib/kotlin/-enum/`

whyoleg:
Fixes #3692 

* Introduce a new parameter `olderVersionsDirName` in `VersioningConfiguration` to customize the directory name for older versions HTMLs
* Add integration test with different folder names

### DIFF
--- a/dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
+++ b/dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
@@ -13,7 +13,8 @@ public data class VersioningConfiguration(
     var olderVersions: List<File>? = defaultOlderVersions,
     var versionsOrdering: List<String>? = defaultVersionsOrdering,
     var version: String? = defaultVersion,
-    var renderVersionsNavigationOnAllPages: Boolean? = defaultRenderVersionsNavigationOnAllPages
+    var renderVersionsNavigationOnAllPages: Boolean? = defaultRenderVersionsNavigationOnAllPages,
+    var olderVersionsDirName: String = defaultOlderVersionsDirName
 ) : ConfigurableBlock {
     internal fun versionFromConfigurationOrModule(dokkaContext: DokkaContext): String =
         version ?: dokkaContext.configuration.moduleVersion ?: "1.0"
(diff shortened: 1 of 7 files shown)


dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
```diff
@@ -13,7 +13,8 @@ public data class VersioningConfiguration(
     var olderVersions: List<File>? = defaultOlderVersions,
     var versionsOrdering: List<String>? = defaultVersionsOrdering,
     var version: String? = defaultVersion,
-    var renderVersionsNavigationOnAllPages: Boolean? = defaultRenderVersionsNavigationOnAllPages
+    var renderVersionsNavigationOnAllPages: Boolean? = defaultRenderVersionsNavigationOnAllPages,
+    var olderVersionsDirName: String = defaultOlderVersionsDirName
```
vmishenev, line 17: Should `olderVersionsDirName` be nullable like other optional parameters?

vmishenev (approved):

dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
whyoleg: AFAIU it shouldn’t be `nullable` - it's already `optional` because it has a default (`"older"`). The same pattern is used in [DokkaBaseConfiguration.footerMessage](https://github.com/Kotlin/dokka/blob/3a0a602c2fd87666116c5c6b7a1a6616a7e62a35/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/DokkaBaseConfiguration.kt#L15)

With `null` value, IMO, it is a bit confusing on what should happen in absence of the value (especially when it is coming from `json` configuration): if it is absent - should we use `default` value or not create a folder or fail or anything else.

PR #3695, opened 2024-07-17
Merge Dokkatoo into Dokka Gradle Plugin 

adam-enko:
Merge Dokkatoo into Dokka repo, and rename it to dokka-gradle-plugin.

DGP-classic has been retained, and can be toggled on/off via a property. (The used name is a placeholder, and can be re-implemented).

The DGP-classic source code has been organised into `src/classicMain` and `src/classicTest` to help keep it separated, and to prevent mixing it.

Only Dokkatoo's HTML format has been copied. GFM, Javadoc, and Jekyll were excluded.

TODO:

- [x] Rename
  - [x] package `dev.adamko...` -> `org.jetbrains...`
  - [x] Dokkatoo -> Dokka
  - [x] `enableDokkatoo` -> something else...
- [x] kotlinx-coroutines test breaks. Need to update Coroutines. 
      ```
      e: /private/var/folders/_3/ps18c9f56bx2j4fyxt0z7sn40000gp/T/junit17292292638531040000/project/buildSrc/src/main/kotlin/Dokka.kt: (18, 20): 
      Class 'org.jetbrains.dokka.gradle.AbstractDokkaLeafTask' was compiled with an incompatible version of Kotlin.
      The binary version of its metadata is 1.8.0, expected version is 1.5.1.
      ```
      UPDATE: fixed by using Java Toolchains & release flag

^OSIP-355


build-logic/build.gradle.kts
```diff
@@ -13,7 +13,6 @@ kotlin {
 }
 
 dependencies {
-    implementation(libs.gradlePlugin.dokka)
```
adam-enko, line 16: I'm not sure why this dependency is here, but it's not used, and it broke Gradle dependency resolution.

dokka-runners/dokka-gradle-plugin/settings.gradle.kts
```diff
@@ -2,7 +2,7 @@
  * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-rootProject.name = "runner-gradle-plugin-classic"
+rootProject.name = "dokka-gradle-plugin"
```
adam-enko, line 5: I renamed the directory because Gradle recommends keeping the artifact ID and the project name aligned. It also makes the publishing configuration simpler.

dokka-runners/dokka-gradle-plugin/src/classicMain/kotlin/DokkaClassicPlugin.kt
```diff
@@ -13,7 +13,7 @@ import org.gradle.kotlin.dsl.withType
 import org.gradle.util.GradleVersion
 import org.jetbrains.dokka.DokkaDefaults
 
-open class DokkaPlugin : Plugin<Project> {
+open class DokkaClassicPlugin : Plugin<Project> {
```
adam-enko, line 16: Renamed for clarity.

This rename should not affect user's project as the newly added code contains a class with the same FQN.

https://github.com/Kotlin/dokka/blob/26eba3f6b11f4d6c2a5c5e6e610557c3952cfc58/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaPlugin.kt#L21

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
```diff
@@ -0,0 +1,41 @@
...
+                    project.provider { project.extra.properties[PropName]?.toString() ?: "" }
+                )
+                .orNull
+                ?.split(",")
+                .orEmpty()
+                .map { it.lowercase() }
+                .toSet()
+
+            val dokkatooEnabledQuietly = "dokkatoo!" in flags
+            val dokkatooEnabled = dokkatooEnabledQuietly || "dokkatoo" in flags
```
adam-enko, line 33: Here's the flag that will toggle between DGP-classic and DGP-nu.

i.e. A user can add `DokkaGradlePluginMode=dokkatoo` to switch to the new plugin mode.

Feel free to bike-shed about the property name!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaPlugin.kt
```diff
@@ -0,0 +1,47 @@
...
+                logger.lifecycle("[${target.displayName}] Dokka Gradle Plugin: Dokkatoo mode enabled")
+            }
+            with(target.pluginManager) {
+                apply(type = DokkaBasePlugin::class)
+
+                // auto-apply the custom format plugins
+                apply(type = DokkaHtmlPlugin::class)
+            }
+        } else {
+            logger.lifecycle("[${target.path}] Dokka Gradle Plugin: classic mode, please migrate, see \$linkToDGPMigrationDocs")
```
adam-enko, line 39: I have added a placeholder warning, as a reminder that we should add some sort of message in the future.

build-logic/build.gradle.kts
whyoleg: I do think it was there before, when we published Dokka generated documentation of Dokka in javadoc :)
I do think that we don't need it anymore even in version catalog, so feel free to remove it

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
```diff
@@ -2,34 +2,65 @@
...
-
-        freeCompilerArgs.addAll(
-            // need 1.4 support, otherwise there might be problems
-            // with Gradle 6.x (it's bundling Kotlin 1.4)
-            "-Xsuppress-version-warnings",
-            "-Xjsr305=strict",
-            "-Xskip-metadata-version-check",
-        )
+        // Must use Kotlin 1.4 to support Gradle 7
+        languageVersion = @Suppress("DEPRECATION") KotlinVersion.KOTLIN_1_4
```
whyoleg, line 19: why we don't set `apiVersion` here anymore? But still set it below in for `compileKotlin` task
minor: what is the reason to split this configuration in 2 parts and then link it? just curious

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
```diff
@@ -2,34 +2,65 @@
...
+    }
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    options.release = dokkaBuild.mainJavaVersion.map { it.asInt() }
+}
+
+tasks.withType<Test>().configureEach {
+    javaLauncher = javaToolchains.launcherFor {
+        languageVersion = dokkaBuild.mainJavaVersion
```
whyoleg, line 63: we do the same in `dokkabuild.java` but with `testJavaLauncherVersion` - is it correct?
I was thinking, that the main reason to use `testJavaLauncherVersion` is to be able to run tests on CI with different JDK versions, but now we will run them with just JDK 8.
I'm not saying this is required, just curious if it's what expected

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
```diff
@@ -2,34 +2,65 @@
...
+tasks.withType<KotlinJvmCompile>().configureEach {
+    compilerOptions {
+        jvmTarget = dokkaBuild.mainJavaVersion.map {
+            JvmTarget.fromTarget(it.formattedName())
+        }
+    }
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    options.release = dokkaBuild.mainJavaVersion.map { it.asInt() }
```
whyoleg, line 58: again, just curios: why do we need to set the version for `JavaCompile`? does it even run?

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
```diff
@@ -2,34 +2,65 @@
...
 }
 
 tasks.validatePlugins {
     enableStricterValidation = true
 }
+
+//region Java version target/compile config
+
+kotlin {
+    jvmToolchain(21)
```
whyoleg, line 40: just curious: can't we just leave using JDK 8 toolchain for this PR? And then if needed migrate FULL project to use approach with JDK 21 toolchain and JDK `release` flag?

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
```diff
@@ -2,34 +2,65 @@
...
+    compilerOptions {
+        freeCompilerArgs.add(
+            dokkaBuild.mainJavaVersion.map { "-Xjdk-release=${it.formattedName()}" }
+        )
+    }
+}
+
+tasks.withType<KotlinJvmCompile>().configureEach {
+    compilerOptions {
+        jvmTarget = dokkaBuild.mainJavaVersion.map {
```
whyoleg, line 51: Q: we set it here and in `kotlin { compilerOptions {} }` again because of some Gradle defaults? If yes, I think it would be nice to have it documented, if no, then let's move it in `kotlin { compilerOptions {} }` :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
whyoleg: Would be nice to have something like this `org.jetbrains.dokka.experimental.useNewDokkaGradlePluginMode=true` , naming is hard, but would be nice to have at least the same prefix as in `org.jetbrains.dokka.experimental.tryK2`. Not sure why we need `quietly` as there are only 2 states:
1. classic plugin - `warning` log will be emitted
2. new (dokkatoo) plugin - no `warning` log will be emitted :)

Of course, it's fine to change it later, just let's not forget about it

whyoleg (approved):
Looks fine from my side!

I've added some comments, but they are mostly `out of curiosity` about some Gradle usages.
I have no objections to merge it in current state and resolve anything else later

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
adam-enko: Yes, renaming to `org.jetbrains.dokka.experimental.useNewDokkaGradlePluginMode` makes sense. Although I very much prefer enums over boolean flags in public API, because they are more flexible, descriptive, and fine-grained.

What about something like `org.jetbrains.dokka.experimental.tryGradlePlugin=v2`?

whyoleg: I would argue that we need any more flexibility here, because we really need just a `flag`: use new or old plugin.
As a user, in this case, I would read `v2` as a separate `version` field which could change in future once more.
Let's not give hints, that we would have `v3` :)

Also, boolean flag is consistent with what we have now with `tryK2` and most flags in KGP also follow the same idea: if there is just switch to new shiny behaviour or some experimental thing - it's just a flag.

Though, `new` in naming looks bad, as it's hard to understand when it becomes `old`.

Overall, I'm fine with any variant which has `org.jetbrains.dokka.experimental` prefix.

adam-enko: I added the 'quietly' function because there's also a logged message for if DGP is in 'dokkatoo' mode, which instantly annoyed me! But I think the 'no warning' option is still relevant.

The plan is to have a phased rollout for Dokkatoo:

1. Release the merged DokkaClassic/Dokkatoo plugin.
   DokkaClassic is the default mode.
   There's no warning for DokkaClassic.
   Users can opt-in to Dokkatoo, with a logged message with a link to the migration docs and 'create new GitHub issue'.
3. Dokkatoo is stable, and is made the default.
   Deprecate DokkaClassic.
   Users can revert to DokkaClassic, with a logged warning (with a link to the migration docs).
4. Remove DokkaClassic, remove the opt-in flag.

In steps 1 and 2 it would be nice to be able to suppress the warnings.

whyoleg: okay, agree, that quietly option is needed, let's stick then with what is in this PR and just rename property to `org.jetbrains.dokka.experimental.gradlePluginMode`

adam-enko: > Though, `new` in naming looks bad, as it's hard to understand when it becomes `old`.

Naming things as 'NewFoo' is a pet peeve of mine :)

IgnatBeresnev (approved):

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
Tapchicoma: If `jvmToolchain` and `jvmTarget` have the same value - we don't need to configure both and only `jvmToolchain` should be enough

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
```diff
@@ -0,0 +1,349 @@
...
+annotation class ProjectDirectoryDsl
+
+@ProjectDirectoryDsl
+interface ProjectDirectoryScope {
+    val projectDir: Path
+
+//    val testMavenRepoRelativePath: String
+//        get() = projectDir
+//            .relativize(GradleProjectTest.testMavenRepoDir)
+//            .invariantSeparatorsPathString
```
Tapchicoma, line 210: Shouldn't it be used?

dokka-runners/dokka-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,217 @@
...
+    testFixturesCompileOnly("org.jetbrains.dokka:dokka-core:${project.version}")
+    testFixturesImplementation(platform(libs.kotlinxSerialization.bom))
+    testFixturesImplementation(libs.kotlinxSerialization.json)
+
+    testFixturesApi(platform(libs.kotest.bom))
+    testFixturesApi(libs.kotest.junit5Runner)
+    testFixturesApi(libs.kotest.assertionsCore)
+    testFixturesApi(libs.kotest.assertionsJson)
+    testFixturesApi(libs.kotest.datatest)
+
```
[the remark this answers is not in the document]
adam-enko: I understand the concern, but the Kotest compiler plugin is only necessary when using Kotest to launch tests in Kotlin Multiplatform projects. Here we only use Kotest for assertions, and JUnit is used to launch the tests (using the Kotest JUnit engine).

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.gradle.kotlin.dsl.extra
+
+internal data class ExperimentalFlags(
+    val dokkatooEnabled: Boolean,
+    val dokkatooEnabledQuietly: Boolean,
+) {
+
+    companion object {
+        @Suppress("ConstPropertyName")
+        private const val GradlePluginMode = "org.jetbrains.dokka.experimental.gradlePlugin"
```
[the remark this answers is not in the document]
adam-enko: Personal preference. I think PascalCase is prettier than SCREAMING_SNAKE_CASE :)

dokka-runners/runner-gradle-plugin-classic/build.gradle.kts
```diff
@@ -1,71 +0,0 @@
...
-).forEach {
-    it.configure { excludeGradleCommonDependencies() }
-}
-
-/**
- * These dependencies will be provided by Gradle, and we should prevent version conflict
- * Code taken from the Kotlin Gradle plugin:
- * https://github.com/JetBrains/kotlin/blob/70e15b281cb43379068facb82b8e4bcb897a3c4f/buildSrc/src/main/kotlin/GradleCommon.kt#L72
- */
-fun Configuration.excludeGradleCommonDependencies() {
```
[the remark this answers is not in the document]
adam-enko: The build config has been reworked so that the dependencies aren't added, so this exclusion does nothing.

If you run `./gradlew :dokka-gradle-plugin:publishDokkaPluginMarkerMavenPublicationToDevMavenRepo` and check `dokka-runners/dokka-gradle-plugin/build/dev-maven-repo/org/jetbrains/dokka/dokka-gradle-plugin/2.0.20-SNAPSHOT/dokka-gradle-plugin-2.0.20-SNAPSHOT.pom`, then there's no extra files.

We should make a test for this though... I've created a task.

https://youtrack.jetbrains.com/issue/KT-70503/Add-test-to-verify-Dokka-Plugin-does-not-expose-conflicting-dependencies

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
adam-enko: Done

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
```diff
@@ -0,0 +1,41 @@
...
+
+        fun Project.dokkaMode(): ExperimentalFlags {
+
+            val flags = providers
+                .gradleProperty(GradlePluginMode)
+                .forUseAtConfigurationTimeCompat()
+                .orElse(
+                    project.provider { project.extra.properties[GradlePluginMode]?.toString() ?: "" }
```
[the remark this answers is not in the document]
adam-enko: You mean applying the property in extraProperties? I added this option for the unit tests, which use ProjectBuilder, and setting `gradle.properties` isn't possible.

https://github.com/Kotlin/dokka/blob/6b8f97c17da013a629b4af715d9ccc085238c14d/dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt#L20-L19

https://github.com/Kotlin/dokka/blob/0e001f09364b33cdaff9525cc37cf8471d973bc1/dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleProjectUtils.kt#L10-L9

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaPlugin.kt
adam-enko: I created the link (although it doesn't go anywhere yet), and increased the log level.

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
adam-enko: I updated the config to set `apiVersion`

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
```diff
@@ -0,0 +1,349 @@
...
+// * Load a project from the [GradleProjectTest.dokkaSrcIntegrationTestProjectsDir]
+// */
+//fun gradleKtsProjectIntegrationTest(
+//  testProjectName: String,
+//  build: GradleProjectTest.() -> Unit,
+//): GradleProjectTest =
+//  GradleProjectTest(
+//    baseDir = GradleProjectTest.dokkaSrcIntegrationTestProjectsDir,
+//    testProjectName = testProjectName,
+//  ).apply(build)
```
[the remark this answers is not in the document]
adam-enko: This might be used again when the Dokkatoo integration tests are merged in, but I'll remove it for this PR.

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
adam-enko: See https://github.com/Kotlin/dokka/pull/3695#discussion_r1705325594

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
adam-enko: Thanks, removed. (You might see this config again though, consistent compilation is the last piece for KT-64200, and some Dokka subprojects have Java code.)

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
adam-enko: You're right, thanks for spotting it. Updated so that tests use `testJavaLauncherVersion`.

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
adam-enko: Yes, let's continue using JDK8 for compilation, and in another PR update the JDK compilation for the whole project along with the `-release` flag.

build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
```diff
@@ -2,34 +2,50 @@
...
+        languageVersion = dokkaBuild.mainJavaVersion
+    }
+}
+
+tasks.withType<Test>().configureEach {
+    javaLauncher = javaToolchains.launcherFor {
+        languageVersion = dokkaBuild.testJavaLauncherVersion
+    }
+}
+//endregion
```
whyoleg, line 51: BTW: this is already configured in `dokkabuild.java` (via `java.toolchain`, but it doesn't matter)

dokka-runners/dokka-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,211 @@
...
+    devPublication("org.jetbrains.dokka:plugin-kotlin-as-java:${project.version}")
+    devPublication("org.jetbrains.dokka:plugin-mathjax:${project.version}")
+    devPublication("org.jetbrains.dokka:plugin-templating:${project.version}")
+    devPublication("org.jetbrains.dokka:plugin-versioning:${project.version}")
+
+    devPublication(project)
+}
+
+gradlePlugin {
+    isAutomatedPublishing = true
```
whyoleg, line 101: minor: it's `true` by default

dokka-runners/dokka-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,211 @@
...
+
+    kotlin("plugin.serialization") version embeddedKotlinVersion
+
+    id("dokkabuild.dev-maven-publish")
+    alias(libs.plugins.kotlinx.binaryCompatibilityValidator)
+    `java-test-fixtures`
+    `jvm-test-suite`
+
+    id("dokkabuild.publish-base")
+    alias(libs.plugins.gradlePluginPublish)
```
whyoleg, line 21: probably this should be replaced with `id("dokkabuild.publish-gradle-plugin")` to configure `gradlePlugin` extension defaults and `javadoc` publication

dokka-runners/dokka-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+description = "Gradle plugin for using Dokka Engine"
+
+kotlin {
+    compilerOptions {
+        optIn.addAll(
+            "kotlin.RequiresOptIn",
+            "org.jetbrains.dokka.gradle.internal.DokkaInternalApi",
+            "kotlin.io.path.ExperimentalPathApi",
```
whyoleg, line 31: minor: how hard it would be not to depend on experimental stdlib API?

whyoleg (commented):
just some small minor remarks

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
adam-enko: I've updated the properties and the messages in a new PR

https://github.com/Kotlin/dokka/pull/3736

dokka-runners/dokka-gradle-plugin/build.gradle.kts
adam-enko: At the moment it's the experimental path functions are only used in test sources. It's a bit tricky/ugly to only opt-in for non-main sources.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/ExperimentalFlags.kt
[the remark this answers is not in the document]
adam-enko: I've renamed them in https://github.com/Kotlin/dokka/pull/3736

dokka-runners/runner-gradle-plugin-classic/build.gradle.kts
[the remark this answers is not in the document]
Tapchicoma: serialization brings stdlib as a transitive dependency, that is why we need this code block

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaPlugin.kt
```diff
@@ -0,0 +1,47 @@
...
+            target.pluginManager.apply(ClassicDokkaPlugin::class)
+        }
+    }
+
+    companion object {
```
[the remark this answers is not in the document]
Tapchicoma: hm, still I am not sure that this is the best approach.

> I think it's useful to keep properties and functions that don't rely on the object's state in a companion object.

This I would make it `private val` on the file level.

> It's also nice to create a companion object, in case others want to create a custom extension on the class.

Is `companion object` required for this? Consumers could just add `fun DokkaPlugin.myFunction()` without companion object.

Tapchicoma (approved):

dokka-runners/runner-gradle-plugin-classic/build.gradle.kts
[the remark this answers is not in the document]
adam-enko: I tried hard to try and discover what the issue is, but I wasn't able to discover what could go wrong. Whenever Dokka was added as a buildSrc dependency, the kotlin-stdlib and other dependencies were correctly constrained to the embedded Gradle version.

When this exclusions config was added in https://github.com/Kotlin/dokka/pull/2570, at the time DGP used `kotlin("jvm")` to build the project. This is probably what caused the issue, and the exclusions helped with that. (KGP is also built with `kotlin("jvm")`.) DGP no longer uses `kotlin("jvm")`, so I don't believe it's necessary. However, out of an abundance of caution, I've added the exclusions.

PR #3712, opened 2024-07-29
Update GitCheckoutTask, so it re-runs if the repo is dirty

adam-enko:
Currently, GitCheckoutTask might not re-run if the checked-out repository is dirty. This can cause invalid test results.

So, update GitCheckoutTask to re-run if the repo does not exist, is invalid, or contains any modifications.

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
@@ -48,7 +48,11 @@ abstract class GitCheckoutTask @Inject constructor(
     init {
         outputs.upToDateWhen { task ->
             require(task is GitCheckoutTask)
-            task.gitRepoIsUpToDate().get()
+            val result = task.gitRepoStatus().get()
+            if (!result.isUpToDate) {
+                task.logger.lifecycle("[${task.path}] git repo ${task.localRepoDir} is not up to date. ${result.message}")
+            }
+            result.isUpToDate
         }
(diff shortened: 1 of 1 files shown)


IgnatBeresnev (approved):

build-logic/src/main/kotlin/dokkabuild/tasks/GitCheckoutTask.kt
```diff
@@ -121,4 +126,61 @@ abstract class GitCheckoutTask @Inject constructor(
                 .call()
         }
     }
+
+    private fun gitRepoIsUpToDate(): Provider<Boolean> =
+        providers.of(GitRepoIsUpToDate::class) {
+            parameters.repoDir = localRepoDir
```
Tapchicoma, line 132: `expectedCommitId` is not configured. Is it expected?

adam-enko: No, that's unexpected, thanks!

I wonder why `parameters.expectedCommitId.get()` didn't throw an error?

adam-enko: I found the reason: the check always failed.

`RepositoryCache.FileKey.isGitRepository(repoDir)` always returns false, it should be `RepositoryCache.FileKey.isGitRepository(repoDir.resolve(".git")`. 

I fixed it, and added better logging.

Tapchicoma (approved):

PR #3470, opened 2024-01-24
build: extracted UI components to ui-kit in plugin-base-frontend

berezinant:
Fixes:
#3460
#3461
#3462

The UI component is just a couple of js and css files which are conntected to the markup by bem, if possible
The assets (ts and scss) are compiled by webpack, the results are stored in plugin-base as plain files js and css (both minified and non-minified for the sake of possible customization by users) files and committed to the repo.

PR #3714, opened 2024-07-29
fix: resolved problem with outdated node-gyp version

berezinant:
Follow up #3470


adam-enko (approved):
This fixes the problems on my machine, and it looks like on CI too. Thanks!

PR #3715, opened 2024-07-29
[K2] Fix missed annotations

### LINKED ISSUE #3688: [K2] Synthetic property misses annotations from getter/setter
Create a Java class and annotate the setter method.
```java
// JavaParent.java
import java.beans.BeanProperty;

public class JavaParent {
    private int context = 0;

    public int getContext() {
        return context;
    }

    @BeanProperty(required = true)
    public void setContext(int context) {
        this.context = context;
    }
}
```
Use it as a superclass in Kotlin.
```Kotlin
class A : JavaParent()
```
The synthetic property `context` will miss the annotation in Dokka.

K1:
image

---

K2
image

vmishenev:
Fixes #3688

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
@@ -15,6 +15,7 @@ import org.jetbrains.kotlin.analysis.api.symbols.KaPropertySymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbol
 import org.jetbrains.kotlin.analysis.api.symbols.KaSymbolOrigin
 import org.jetbrains.kotlin.analysis.api.types.*
+import org.jetbrains.kotlin.builtins.StandardNames
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget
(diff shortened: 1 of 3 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -57,10 +58,17 @@ internal class AnnotationTranslator {
...
     private fun KaSession.mustBeDocumented(annotation: KaAnnotation): Boolean {
-        if (annotation.isNoExistedInSource()) return false
+        /**
+         * Do not document the synthetic [parameterNameAnnotation] annotation since Dokka K1 ignores it too.
+         * The annotation can be added by the compiler during "desugaring" functional types.
+         * e.g., `(x: Int) -> Unit` becomes `Function1<@ParameterName("x") Int, Unit>`
+         * @see ParameterName
+         * @see getPresentableName
+        */
+        if(annotation.classId == parameterNameAnnotation && annotation.isNoExistedInKotlinSource()) return false
```
vmishenev, line 68: I have not discovered other synthetic annotations added by the compiler.  
The PR aims to keep the K1 behaviour that might be reconsidered later.

JIC I have opened https://youtrack.jetbrains.com/issue/KT-70325/Analysis-API-KaAnnotation-returns-only-Kotlin-PSI but it does not affect us now.

whyoleg (approved):

PR #3716, opened 2024-07-29
Enable build cache by default in Gradle integration tests

adam-enko:
Enabling build cache by default will improve performance when re-running the tests locally.

Update the test assertions to check that tasks either succeed or are loaded from cache (which does not make a difference in terms of integration tests).

Additionally, use `GRADLE_RO_DEP_CACHE` to re-use the host machine's Gradle dependencies cache. This will improve test performance locally and on CI.


IgnatBeresnev (approved):

dokka-integration-tests/gradle/src/testTemplateProjectAndroid/kotlin/Android0GradleIntegrationTest.kt
```diff
@@ -43,7 +42,7 @@ class Android0GradleIntegrationTest : AbstractGradleIntegrationTest() {
     @ArgumentsSource(AndroidTestedVersionsArgumentsProvider::class)
     fun execute(buildVersions: BuildVersions) {
         val result = createGradleRunner(buildVersions, "dokkaHtml", "-i", "-s").buildRelaxed()
-        assertEquals(TaskOutcome.SUCCESS, assertNotNull(result.task(":dokkaHtml")).outcome)
+        result.shouldHaveTask(":dokkaHtml").shouldHaveOutcome(SUCCESS, FROM_CACHE)
```
whyoleg, line 45: just curious: in which case it will be `FROM_CACHE`?

adam-enko: One way would is:

1. Run the tests locally (thus populating the TestKit cache in `dokka-integration-tests/gradle/build/gradle-test-kit`),
2. then you edited `dokka-integration-tests/gradle/build.gradle.kts`.
3. The tests would re-run, but because the test project hasn't changed, the tasks could be loaded from cache.

dkrasnoff (approved):

dkrasnoff:
Good job 👍

PR #3722, opened 2024-07-31
build: bump node version

berezinant:
This should fix a bunch of warnings when istalling node dependencies like
image


dokka-subprojects/plugin-base-frontend/.nvmrc
```diff
@@ -0,0 +1 @@
+18.20.4
```
adam-enko, line 1: I'd like it if the version wasn't duplicated both here and in the version catalog.

We could make a Gradle task that will automatically update this file to match the value in `libs.version.toml`. WDYT?

(I can do this in a new PR, to protect everyone from Gradle headaches.)

gradle/libs.versions.toml
```diff
@@ -43,7 +43,7 @@ apacheMaven-pluginTools = "3.5.2"
 kotlinx-cli = "0.3.6"
 
 ## NPM | Frontend
-node = "16.13.0"
+node = "18.20.4"
```
adam-enko, line 46: As someone who doesn't know much about JS development, can you give a brief explanation why v18 should be chosen? Isn't v20 the latest stable lts release?

dokka-subprojects/plugin-base-frontend/.nvmrc
berezinant: The automatic sync between `libs.version.toml` and `.nvmrc` would be handy 👍

But it's worth noticing that this .nvmrc`  file is not very important, it's only used to make local development slightly easier
So this automation will pay off only if it's really easy and quick

adam-enko: Yeah, it should just be a few lines to set up.

But if it's a dev-local thing, maybe `.nvmrc` should just be git-ignored?

berezinant: Cool then!

The `.nvmrc` helps to automatically change local node version on a developer's machine when one switches between projects, so it's convenient to have it in repo

gradle/libs.versions.toml
berezinant: Honestly there was no specific reasons, just a step-by-step update. I've tested it with v20 and it seems to work ok so I updated to v20

PR #3732, opened 2024-08-05
Update Maven Binary install task to be cachable

adam-enko:
Currently the `installMavenBinary` task does not normalize the task inputs. This results in the task re-running unnecessarily.

To improve build performance, use classpath-normalization on the Maven files before copying them to a local directory.

Additionally, update the Configurations to follow Gradle conventions (create a 'task bucket' Configuration for declaring dependencies, and a 'resolver' to fetch the dependencies.)


adam-enko:
I realised that making `installMavenBinary` cacheable isn't necessary. The actual Maven binary file is cached as a Gradle dependency, so `installMavenBinary` is just a simple copy-file task. Caching it is unnecessary, and would make the task slower.

Issue #3668, opened 2024-07-04
Rework navbar component according to new design


berezinant:
Rework navbar component according to new design https://www.figma.com/design/p5V2MV7znMxdVekniseQkA/DokkaUI-library?node-id=16-6079&t=CuTgexYC4mvRI2fa-0

[KTL-1564](https://youtrack.jetbrains.com/issue/KTL-1564/Dokka-rework-navbar-component-according-to-new-design)

PR #3734, opened 2024-08-06
feat: reworked navbar and ToC according to new design (#3668)

berezinant:
image

image


dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/htmlPreprocessors.kt
```diff
@@ -104,6 +104,12 @@ public class StylesInstaller(private val dokkaContext: DokkaContext) : PageTrans
 
 public object AssetsInstaller : PageTransformer {
     private val imagesPages = listOf(
+        // ui-kit assets
```
berezinant, line 107: For now there is no way to register all content from the assets directory so we have to register it file by file

berezinant:
Closing as duplicate of https://github.com/Kotlin/dokka/pull/3917

PR #3765, opened 2024-08-27
Remove Gradle wrappers from Gradle examples

adam-enko:
A fixed version of Gradle isn't super important: Dokka should work with most versions of Gradle, and will log a warning for unsupported versions.

Updating these wrappers is frequently forgotten about, so removing them reduces the maintenance requirements.


adam-enko:
Ah okay, these are needed. Please don't review yet, I'll re-evaluate...

adam-enko:
The DGPv2 examples no longer need wrappers https://github.com/Kotlin/dokka/pull/3809

PR #3769, opened 2024-08-28
Simplify configuration of DokkaPublication output directories

adam-enko:
- Rename `outputDir` to `outputDirectory` for consistency.
- Set the publication/module output dirs on a DokkaPublication, so that they are defined in a single place. This also makes it easier for users to configure the publication output, if desired.

Improves [KT-70880](https://youtrack.jetbrains.com/issue/KT-70880/DGP-v2-puts-output-in-the-html-subfolder-for-a-customized-output-directory)

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/dokka/DokkaPublication.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/dokka/DokkaPublication.kt
@@ -47,7 +47,8 @@ constructor(
 
     abstract val moduleVersion: Property<String>
 
-    abstract val outputDir: DirectoryProperty
+    /** Output directory for the finished Dokka publication. */
+    abstract val outputDirectory: DirectoryProperty
 
     abstract val cacheRoot: DirectoryProperty
 
@@ -63,4 +64,7 @@ constructor(
 
     // TODO probably not needed any more, since Dokka Generator now runs in an isolated JVM process
     abstract val finalizeCoroutines: Property<Boolean>
+
+    /** Output directory for the partial Dokka module. */
+    internal abstract val moduleOutputDirectory: DirectoryProperty
 }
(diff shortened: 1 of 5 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/dokka/DokkaPublication.kt
```diff
@@ -63,4 +64,7 @@ constructor(
 
     // TODO probably not needed any more, since Dokka Generator now runs in an isolated JVM process
```
whyoleg, line 65: Q: just in case, does this `TODO` tracked via some issue? the same for other `TODO` left after merging dokkatoo - would be nice to have them tracked somehow

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dokka/DokkaPublication.kt
```diff
@@ -47,7 +47,8 @@ constructor(
 
     abstract val moduleVersion: Property<String>
 
-    abstract val outputDir: DirectoryProperty
+    /** Output directory for the finished Dokka publication. */
+    abstract val outputDirectory: DirectoryProperty
 
     abstract val cacheRoot: DirectoryProperty
```
whyoleg, line 53: AFAIK `cacheRoot` is not used inside dokka and so probably we don't need to expose it anywhere at all and just use some temp dir

whyoleg (approved):
LGTM

though, IMO, configuring output directory on task level feels more natural for me for some reason

P.S. I've left only `unrelated` to PR comments as I spotted them just now

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dokka/DokkaPublication.kt
adam-enko: Agreed, the TODOs should be removed or added in YouTrack. There are a few others scattered around.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dokka/DokkaPublication.kt
adam-enko: I agree it makes sense to remove it, but that should be done in the DokkaConfiguration. DGP is just a wrapper around Dokka's core configuration, and possibly other Dokka plugins use the cache dir.

PR #3770, opened 2024-08-29
When DGP is used in a pre-compiled script plugin, correctly generate accessors and disable warning logs

adam-enko:
Gradle generates accessors in a temp project that doesn't have Gradle properties. This breaks the `enableV2` flag.

This workaround detects when Gradle is generating accessors, and if so, tries to discover the `gradle.properties`.

##### Fixes

- https://youtrack.jetbrains.com/issue/KT-70798/When-Dokka-Gradle-Plugin-v2-mode-enabled-in-a-buildSrc-convention-plugin-the-dokka-project-extension-DSL-isnt-available
- https://youtrack.jetbrains.com/issue/KT-71010/DGP-v2-wrong-logging-from-inside-convention-plugins

---

Because DGPv2 is heavily promoting pre-compiled script plugins I think it's important to get it right, to make a good first impression.


dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -193,43 +196,81 @@ internal abstract class PluginFeaturesService : BuildService<PluginFeaturesServi
...
+            v2PluginEnabled.set(getFlag(V2_PLUGIN_ENABLED_FLAG))
+            v2PluginNoWarn.set(getFlag(V2_PLUGIN_NO_WARN_FLAG_PRETTY).orElse(getFlag(V2_PLUGIN_NO_WARN_FLAG)))
+            v2PluginMigrationHelpersEnabled.set(getFlag(V2_PLUGIN_MIGRATION_HELPERS_FLAG))
+            k2AnalysisEnabled.set(getFlag(K2_ANALYSIS_ENABLED_FLAG))
+            k2AnalysisNoWarn.set(
+                getFlag(K2_ANALYSIS_NO_WARN_FLAG_PRETTY)
+                    .orElse(getFlag(K2_ANALYSIS_NO_WARN_FLAG))
+            )
+
+            try {
```
whyoleg, line 239: It would be nice to extract this into separate function with a high-level description on why we are doing this

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -264,3 +306,58 @@ internal abstract class PluginFeaturesService : BuildService<PluginFeaturesServi
...
+    return rootProjectPath
+        .substringBeforeLast("/")
+        .endsWith("build/tmp/generatePrecompiledScriptPluginAccessors")
+}
+
+
+/**
+ * Walk up the file tree until we discover a `gradle.properties` file.
+ */
+private fun Project.findGradlePropertiesFile(): File? {
```
whyoleg, line 346: Oh, this is very hacky... not sure what I think about it

The main question from me: is there any possibility of false-positives? looks like there are enough guards, but still want to have some thoughts on it

Tapchicoma:
> Note that there are no tests for this yet. They will be added in upcoming PRs, in particular https://github.com/Kotlin/dokka/pull/3781.

It is a little strange to add tests for changes in follow-up PR. I would prefer to have change and tests for it in a single PR.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
whyoleg: Thanks! sounds reasonable

> Release multiple versions of DGP, with separate suffixes
> Or even with (slightly) more descriptive suffixes, like 2.0.0-v2-migration and 2.0.0-v2.

let's try not to do this. I believe current approach should be enough for migration period

BTW, until this is merged, the workaround is to apply plugin like `plugins.apply(DokkaHtmlPlugin::class)` and configure extension via `extensions.configure<DokkaExtension>("dokka") {}` ([example](https://github.com/whyoleg/cryptography-kotlin/blob/f2640426484e334faa7f5db2e6ae1abc691fd367/build-logic/src/main/kotlin/ckbuild.documentation.gradle.kts)) - in this case no warning is emitted

adam-enko:
> It is a little strange to add tests for changes in follow-up PR. I would prefer to have change and tests for it in a single PR.

Fair point. I'll either try and get #3781 ready for review, or add a test.

adam-enko:
> It is a little strange to add tests for changes in follow-up PR. I would prefer to have change and tests for it in a single PR.

`BuildSrcKotlinDslAccessorsTest` added, please re-review.

whyoleg (approved):

Tapchicoma (approved):

PR #3779, opened 2024-09-04
Relocate dokka package

adam-enko:
KT-70811

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaGeneratorParametersSpec.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaGeneratorParametersSpec.kt
@@ -1,7 +1,7 @@
 /*
  * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
-package org.jetbrains.dokka.gradle.dokka.parameters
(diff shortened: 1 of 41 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaGeneratorParametersSpec.kt
```diff
@@ -27,7 +27,7 @@ constructor(
     objects: ObjectFactory,
     /**
      * Configurations for Dokka Generator Plugins. Must be provided from
-     * [org.jetbrains.dokka.gradle.dokka.DokkaPublication.pluginsConfiguration].
+     * [org.jetbrains.dokka.gradle.engine.DokkaPublication.pluginsConfiguration].
```
whyoleg, line 30: this KDoc link is unresolved (should be `formats` instead of `engine`) - the same in several other places

whyoleg (approved):
LGTM

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaGeneratorParametersSpec.kt
adam-enko: Thanks! I will fix them.

Tapchicoma (approved):

PR #3780, opened 2024-09-04
DGPv2: Tidy configurable options

adam-enko:
- `delayTemplateSubstitution` is no longer used, it is configured automatically internally.
- `DokkaPublication` is not used as a task input, and so it does not need the Gradle task annotations. It also doesn't need the workarounds like `cacheRootPath` and `outputDirPath`.


adam-enko:
No longer needed - I already tidied DokkaPublication in https://github.com/Kotlin/dokka/commit/37f14daa3e4b47003ff2c5a6b5289e3f61dc2699

PR #3781, opened 2024-09-05
Update and test example Gradle projects

adam-enko:
WIP

https://youtrack.jetbrains.com/issue/KT-71346/Update-Dokka-Gradle-example-projects-to-work-with-DGPv2


adam-enko:
Continued in https://github.com/Kotlin/dokka/pull/3809

PR #3782, opened 2024-09-05
Update .gitattributes

adam-enko:
- More comprehensive coverage of languages
- Exclude documentation from GitHub language stats


whyoleg (approved):
Thanks!
Never used `gitattributes` :)
Would be nice to include a link to specific template used or some source why those specific configurations are needed in our case

adam-enko:
> Thanks! Never used `gitattributes` :) Would be nice to include a link to specific template used or some source why those specific configurations are needed in our case

There are a couple of links at the top for general git attribute info, and a link about the GitHub specific linguist options. Do they provide enough info?

PR #3783, opened 2024-09-05
Update IntelliJ excluded directories

adam-enko:
Mark directories as excluded so that they don't appear in IntelliJ's global search. This is particularly annoying for the KotlinDslAccessorsTest files.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@
 
 plugins {
     id("dokkabuild.base")
+    idea
 }
 
 val publishedIncludedBuilds = listOf("runner-cli", "dokka-gradle-plugin", "runner-maven-plugin")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -105,3 +106,20 @@ tasks.wrapper {
...
+
+idea {
+    module {
+        // Mark directories as excluded so that they don't appear in IntelliJ's global search.
+        excludeDirs.addAll(
+            files(
+                ".idea",
+                ".husky",
+                ".kotlin",
+                "dokka-runners/dokka-gradle-plugin/.kotlin",
```
whyoleg, line 118: Q: shouldn't those be excluded by default? May be there is an issue on Idea YT (or Gradle issues)? Or should we create one?

whyoleg (approved):

build.gradle.kts
adam-enko: Yeah, it should be fixed in IntelliJ, but it hasn't been yet. I'm sure I saw a YouTrack issue for it, but I can't find it - it's impossible to search for.

PR #3794, opened 2024-09-10
Remove configuration of transitive Dokka Engine dependencies

adam-enko:
Remove unnecessary configuration of Dokka Engine transitive dependencies.

DGP should only depend on Dokka Engine, and not prescriptively control what dependencies to declare.

[KT-71382](https://youtrack.jetbrains.com/issue/KT-71382/Remove-version-overrides-for-transitive-Dokka-Engine-dependencies)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaFormatPlugin.kt
```diff
@@ -134,8 +134,9 @@ abstract class DokkaFormatPlugin(
                     // https://github.com/gradle/gradle/issues/27435)
                     dependenciesContainer.resolutionStrategy.eachDependency {
```
Tapchicoma, line 135: FYI: Using `.eachDependency` to specify version could lead to Gradle complain. I've fixed similar issue inside `kotlin.git`. Proposed solution is to use constraints with required (or some other level) version.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
```diff
@@ -131,12 +131,13 @@ constructor(
...
-        val freemarker: Property<String>
-        val kotlinxHtml: Property<String>
-        val kotlinxCoroutines: Property<String>
+        /**
+         * The default version of Dokka dependencies that are used at runtime during generation.
+         *
+         * This value defaults to the current Dokka Gradle Plugin version, but can be overridden
+         * if you want to use a newer or older version of Dokka at runtime.
+         */
+        val dokkaEngine: Property<String>
```
whyoleg, line 140: why not to move this on extension level and call just `dokkaEngineVersion`/`engineVersion` - similar to how KGP has `compilerVersion` when using `build-tools-api`?

Not sure we need `Versions` interface anymore

dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaFormatPlugin.kt
adam-enko: Constraints require a list of all Dokka dependencies, and I don't think we have this.

I think improving this is on Gradle to fix https://github.com/gradle/gradle/issues/15089

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
adam-enko: Sure, I'll move it to the top level.

The reason I wanted to keep it nested inside a `versions {}` block was to avoid having too many options in the top level of `dokka {}`. I think keeping the lower-level options nested helps indicate that they are more complicated and advanced. Although I agree that a `versions {}` block seems pretty pointless without other versions, and it's unlikely that we'll be adding other versions.

tbh I think the top level `dokka {}` DSL is pretty disorganised and could do with a review, but it's hard to re-organise it without becoming too disconnected from the underlying Dokka configuration.

Anyway, we can revisit this later if needed.

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaPluginFunctionalTest.kt
```diff
@@ -144,22 +144,24 @@ class DokkaPluginFunctionalTest : FunSpec({
...
 
-                            dokkaConfigurations.keys.shouldContainExactlyInAnyOrder(
-                                listOf(
-                                    "Configuration dokka${Format}GeneratorRuntimeResolver~internal",
-                                    "Configuration dokka${Format}ModuleOutputDirectoriesResolver~internal",
-                                    "Configuration dokka${Format}PluginIntransitiveResolver~internal",
-                                    "Configuration dokka${Format}PublicationPluginResolver~internal",
-                                )
-                            )
-
```
adam-enko, line 162: I have no idea why this test started failing, or more accurately, I have no idea how this test ever worked.

The changes in this PR should be unrelated, so they shouldn't have affected this test. But also, I don't see how this test ever worked... The Configurations for both HTML and Javadoc should always be present.

whyoleg: Can we merge it via separate PR?

adam-enko: https://github.com/Kotlin/dokka/pull/3801

dokka-runners/dokka-gradle-plugin/src/testFunctional/resources/KotlinDslAccessorsTest/root-project-v2-migration-helpers.txt
```diff
@@ -493,61 +493,6 @@ fun DependencyHandler.`dokkaHtmlRuntime`(
 fun DependencyHandler.`dokkaHtmlRuntime`(dependencyNotation: Any): Dependency? =
     add("dokkaHtmlRuntime", dependencyNotation)
 
-@Deprecated(message = "The dokkaJavadocPartialPlugin configuration has been deprecated for dependency declaration. Please use the 'dokkaJavadocPlugin' configuration instead.")
```
whyoleg, line 496: Why those files were changed? Those changes are not present also in https://github.com/Kotlin/dokka/pull/3801

PR #3800, opened 2024-09-16
HTML rendering memory allocations optimisation

whyoleg:
Overall the main idea of the changes is to reduce usage of `createHTML` which by default create `StringBuilder` of 32KB size and in most cases it needs much less, specifically in source-set-dependent content and during resource links embedding.

Those optimisations were found during investigation of stdlib docs build ([internal discussion](https://jetbrains.slack.com/archives/C02H0GSH474/p1726159162133729))

### DIFF
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
@@ -947,9 +947,10 @@ public open class HtmlRenderer(
         }
 
     // same as createHTML but creates StringBuilder of default size instead of using StringBuilder with 32kb size (default)
-    // it's useful for cases with "embedded html" f.e source-set-dependent content
+    // it's useful for cases with "embedded html" f.e source-set-dependent content.
+    // capacity of the StringBuilder was selected after experiments based on generating documentation for kotlin-stdlib
     private fun createSmallHTML(prettyPrint: Boolean = true, xhtmlCompatible: Boolean = false): TagConsumer<String> =
-        HTMLStreamBuilder(StringBuilder(), prettyPrint, xhtmlCompatible)
+        HTMLStreamBuilder(StringBuilder(256), prettyPrint, xhtmlCompatible)
             .onFinalizeMap { sb, _ -> sb.toString() }
             .delayed()
 
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
@@ -5,6 +5,9 @@
 package org.jetbrains.dokka.base.renderers.html
 
 import kotlinx.html.*
+import kotlinx.html.consumers.delayed
+import kotlinx.html.consumers.onFinalizeMap
+import kotlinx.html.stream.HTMLStreamBuilder
 import kotlinx.html.stream.createHTML
 import org.jetbrains.dokka.DokkaSourceSetID
 import org.jetbrains.dokka.Platform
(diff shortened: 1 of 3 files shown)


dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
```diff
@@ -315,12 +318,12 @@ public open class HtmlRenderer(
     ): List<Pair<DisplaySourceSet, String>> {
         var counter = 0
         return nodes.toList().map { (sourceSet, elements) ->
-            val htmlContent = createHTML(prettyPrint = false).prepareForTemplates().div {
+            val htmlContent = createSmallHTML(prettyPrint = false).prepareForTemplates().div {
                 elements.forEach {
                     buildContentNode(it, pageContext, sourceSet)
                 }
             }.stripDiv()
-            sourceSet to createHTML(prettyPrint = false).prepareForTemplates()
```
vmishenev, line 323: Have you considered an optimal capacity? 
In this example, It always takes more than the default `16`.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
```diff
@@ -344,14 +347,14 @@ public open class HtmlRenderer(
                 val distinct =
                     groupDivergentInstancesWithSourceSet(it.value, it.key, pageContext,
                         beforeTransformer = { instance, _, sourceSet ->
-                            createHTML(prettyPrint = false).prepareForTemplates().div {
+                            createSmallHTML(prettyPrint = false).prepareForTemplates().div {
                                 instance.before?.let { before ->
```
vmishenev, line 351: I think we can  extract  `instance.before?.let` from `createHTML` to avoid the redundant invocation of `createHTML`

vmishenev (approved):
Just out of curiosity: have you measured the overall effect of this pr on memory consumption?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
whyoleg: Yeah, I was thinking about it, but failed to think of a good default... Will it be 128 or 1024? It's hard to calculate which value will be best.

But I agree, that probably it would be nice to initialise it with bigger value. Do you have anything in mind?
I can try to check different values if needed

whyoleg:
> Just out of curiosity: have you measured the overall effect of this pr on memory consumption?

Yes! I've rechecked once more to get numbers. Tested with stdlib 1.5, newer versions will probably have even bigger differences.
* memory allocated during the full process of Dokka execution:
  * before: 46 GB
  * after: 27 GB
* max memory used during Dokka execution
  * before: 5.07 GB
  * after: 4.48 GB
* `byte[]` allocations (mostly coming from operations on `StringBuilder` from `kotlinx-html`, e.g `createHTML`)
  * before: 21 GB
  * after: 6.5 GB
* `text.lines()` allocations (removed in this PR) used 4.5 GB

I can do the same with some other project of your choice if you want to validate if it affects anything else

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
whyoleg: I will take a look, thanks!

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/HtmlRenderer.kt
vmishenev: > Do you have anything in mind?

I think you can find it out on stdlib.
E.g. for 128 vs 1024  - choose one with the minimal memory allocated in a profiler.

whyoleg: I've tried several values (default, 128, 256, 512, 1024), the difference in memory allocations is ±3% from best to worst. With 1024 it started to behave even worse, though the diff is minimal and not always the same.
I've decided to stick with 256.

PR #3802, opened 2024-09-16
Update Analysis API to 2.1.20-dev-4370

vmishenev:
The goal of PR is to update Analysis API to the latest version - `2.1.20-dev-4370`.

After some changes in AA, it is necessary to dispose of global resources (including threads of workers) at the end of the analysis to avoid memory leaks when rerunning the Dokka task (see https://youtrack.jetbrains.com/issue/KT-71862/Analysis-API-Standalone-Memory-leak-LowMemoryWatcherManager-does-not-release-resources ).

Also, Dokka encountered exhausted metaspace memory during one of the stress tests. On JVM 9 and later, the test passed successfully, but it failed on JVM 8, even with doubled or tripled memory. We resolved the issue using the flag `-XX:SoftRefLRUPolicyMSPerMB=10` (collect soft references more aggressively) from [KT-55831](https://youtrack.jetbrains.com/issue/KT-55831/) on JVM 8.


vmishenev:
This is blocked by https://youtrack.jetbrains.com/issue/KT-71862/Analysis-API-Standalone-Memory-leak-LowMemoryWatcherManager-does-not-release-resources

gradle/libs.versions.toml
```diff
@@ -14,7 +14,7 @@ ktor = "2.3.11"
 
 ## Analysis
 kotlin-compiler = "2.0.20"
-kotlin-compiler-k2 = "2.1.0-dev-5441"
+kotlin-compiler-k2 = "2.1.0-dev-7103"
```
whyoleg, line 17: @vmishenev have you checked, by the way, if the behavior on latest available version of AA is the same?

vmishenev: The same(

whyoleg:
@vmishenev could you rebase on latest master? I see test failure on CI which is fixed there

whyoleg (changes requested):
The main question at my side is, how it will affect end-users?
I mean, okay, we "patched" (via scary flag `SoftRefLRUPolicyMSPerMB`)  our test, but what will be on user-projects without it?

Also, I think we should check, that it will work fine with DGPv2, at least in IT before merging.
And also we should test combination DGPv2 + K2 with bigger external projects (kotlinx, ktor) before 2.0.0 release, so that we are sure that it works fine keeping in mind the difference in how Dokka is executed

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/JavaAnalysisPlugin.kt
```diff
@@ -71,6 +71,20 @@ public class JavaAnalysisPlugin : DokkaPlugin() {
         DocCommentFinder(logger, docCommentFactory)
     }
 
+    /**
+     * Disposes global resources which would persist after unloading Analysis API (Symbols analysis) and IJ platform classes.
```
whyoleg, line 75: nice comment BTW!

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/JavaAnalysisPlugin.kt
```diff
@@ -71,6 +71,20 @@ public class JavaAnalysisPlugin : DokkaPlugin() {
...
+     *
+     * **Important:** Once this function has been called, Analysis API *and* IntelliJ platform classes should not be used anymore. The classes
+     * should either be unloaded or the whole program should be shut down.
+     *
+     * Note: Disposing of resources, including threads, allows unloading Dokka's class loader.
+     */
+    @InternalDokkaApi
+    public fun disposeGlobalStandaloneApplicationServices() {
```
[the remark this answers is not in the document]
vmishenev: Only from `analysis-java-psi`,  you can call the methods of the IJ Platform. The K2 module has the truncated version of IJ from the compiler dependencies.
On other side, the calling of `disposeGlobalStandaloneApplicationServices` might affect the PSI analysis somehow.

dokka-integration-tests/gradle/src/testTemplateProjectTasksExecutionStress/kotlin/SequentialTasksExecutionStressTest.kt
```diff
@@ -25,7 +25,10 @@ class SequentialTasksExecutionStressTest : AbstractGradleIntegrationTest() {
             buildVersions,
             "runTasks",
             "-Ptask_number=$iterations",
-            jvmArgs = listOf("-Xmx1G", "-XX:MaxMetaspaceSize=500m"),
+            jvmArgs = listOf(
+                "-Xmx1G", "-XX:MaxMetaspaceSize=500m",
+                "-XX:SoftRefLRUPolicyMSPerMB=10" // to free up the metaspace on JVM 8, see https://youtrack.jetbrains.com/issue/KT-55831/
```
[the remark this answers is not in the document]
vmishenev: > Overall, as far as I understand, this means, that K2 analysis could cause OOM on user projects with JDK 8? 

Yes, but as you said it is unknown how many real user projects it affects.
This test is quite artificial with almost no code.
On real projects with the same number of modules, the behavior of JVM 8 might be different.

vmishenev:
> The main question at my side is, how it will affect end-users?

I do not know)  This test is quite artificial with almost no code.
On real projects with the same number of modules, the behavior of JVM 8 might be different.

>  I mean, okay, we "patched" (via scary flag SoftRefLRUPolicyMSPerMB) our test, but what will be on user-projects without it?

The another "popular" solution is increasing memory. It does not work for the stress test since its goal is to check the consumption of memory.


There are some observations:
1. It seems there is no memory leak in AA. It works well on JDK 9+. On Jvm 8, the flag `SoftRefLRUPolicyMSPerMB` is only about collection soft references more aggressively. 
2. We can do nothing with JVM 8 right now. We cold write some optimizations/heuristics in the code to avoid `SoftRefLRUPolicyMSPerMB`, but it requires research.
3. I do know what types of projects it affects.

> Also, I think we should check, that it will work fine with DGPv2, at least in IT before merging.

Does it really block merging? Haven't we checked DGPv2+K2 in IT at all yet?

> And also we should test combination DGPv2 + K2 with bigger external projects (kotlinx, ktor) before 2.0.0 release

Sure

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/JavaAnalysisPlugin.kt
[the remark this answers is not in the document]
vmishenev: Anyway, Marco will merge [the endpoint](https://jetbrains.team/p/kt/reviews/18225/timeline) into AA. I left `TODO`

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/JavaAnalysisPlugin.kt
vmishenev: Actually it is not mine) Marco's

whyoleg:
> Does it really block merging? Haven't we checked DGPv2+K2 in IT at all yet?

Okay, let's merge at test it manually and add DGPv2 + K2 testing separately.
The main reason for this to work is that we will have both DGPv2 and K2 as opt-in and I think that most of the users who will try those flags will be the same, so DGPv2 + K2 combination will be rather popular

> We can do nothing with JVM 8 right now

Yeah, let's see how it will go.

whyoleg (approved):

vmishenev:
JIC I have tested K2 with DGPv2 on a project (50 submodules) from the stress test (where a special VM flag is used).  It works well on JVM 8 without any additional options or flags.

Issue #3805, opened 2024-09-17
CI does not run all DGP tests


adam-enko:
CI does not run all tests defined in the dokka-gradle-plugin subproject.

When running `./gradlew test` or `./gradlew check`, this misses tests in DGP, e.g. `:dokka-gradle-plugin:testFunctional`.

Not running all tests leads to failing tests sneaking into master (e.g. https://github.com/Kotlin/dokka/pull/3801)

In similar situations before the Dokka team decided that the build config shouldn't follow Gradle conventions for running tests (The convention is that `check` should run all verifications). Assuming that Dokka should still not follow the conventions, then I'm not sure what the best way forward is.

whyoleg:
AFAIK on current moment we do have `check` depends on both [integration tests](https://github.com/Kotlin/dokka/blob/880ac6527e1208cc867d3ad1f3b78072aac61936/build-logic/src/main/kotlin/dokkabuild.test-integration.gradle.kts#L92-L98) and [functional tests](https://github.com/Kotlin/dokka/blob/d5515e998d19e56e0c5121836a25bc511dddd862/dokka-runners/dokka-gradle-plugin/build.gradle.kts#L204) in Gradle plugin.
I would argue that `test` should depend on `testFunctional`.

> In similar situations before the Dokka team decided that the build config shouldn't follow Gradle conventions for running tests (The convention is that check should run all verifications). Assuming that Dokka should still not follow the conventions, then I'm not sure what the best way forward is.

IMO current situation clearly show that we need to remove dependency of `check` on `integration tests` so that we can easily use just `check` on CI (both TC and GA) by default and continue having separate TC workflow for integration tests for both K1 and K2 analysis. This is inline with how maintainers are working with Dokka - no need to run integration tests in day-to-day workflow.
It's fine to have `:dokka-integration-tests:check` to run integration tests, though root `check/build` should not run them. I thin that here we are not violating any "Gradle conventions" as `dokka-integration-tests` is an included build for a reason and could have different lifecycle comparing to the whole other build.
WDYT?

adam-enko:
Sorry, I'm still not understanding the reasoning. If you want to do something unconventional, then go ahead, but I can't help. What's important is that the tests run on CI.

PR #3808, opened 2024-09-18
Run `check` task on CI

whyoleg:
fixes https://github.com/Kotlin/dokka/issues/3805

Introduce a Gradle property to skip integration tests, which could be set on CI/locally to be able to run `check` task without running all integration tests.

If we will enable running integration tests unconditionally we will increase GA tests time from 10-15 minutes to more than an hour ([example](https://github.com/Kotlin/dokka/actions/runs/10918956278)).
Note: all tests are red in the above example because most of them failed with OOM or "not enough space" when downloading K/N toolchain for different version of Kotlin.
So enabling integration tests unconditionally on GA is not an option also for reasons other than "slow tests".

Note: running `check` on CI exposed issue with Windows and DGPv2 function tests. This should be fixed before the merging of this PR.
Additionally after the PR is merged we could change TC setup for `Dokka Tests` to also use `check` with `integration_test.skip=true` flag. Or even drop it, as it doesn't cover other OS like we have on GA.

### DIFF
--- a/dokka-integration-tests/gradle.properties
+++ b/dokka-integration-tests/gradle.properties
@@ -10,3 +10,7 @@ version=2.0.20-SNAPSHOT
 org.jetbrains.dokka.javaToolchain.mainCompiler=8
 org.jetbrains.dokka.javaToolchain.testLauncher=8
 org.jetbrains.dokka.integration_test.parallelism=2
+
+# !!! ATTENTION: do not commit this !!!
+# Uncomment next line to skip running Dokka integration tests
+#org.jetbrains.dokka.integration_test.skip=true
--- a/.github/workflows/tests-smoke.yml
+++ b/.github/workflows/tests-smoke.yml
@@ -32,7 +32,8 @@ jobs:
           develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
       - name: Run tests
         run: >
-          ./gradlew check --stacktrace --continue 
+          ./gradlew check --stacktrace --continue
+          "-Porg.jetbrains.dokka.integration_test.skip=true"
           "-Porg.jetbrains.dokka.javaToolchain.testLauncher=${{ env.JAVA_TEST_VERSION }}"
       - name: Upload build reports
         if: failure()
--- a/.github/workflows/tests-smoke.yml
+++ b/.github/workflows/tests-smoke.yml
@@ -32,7 +32,7 @@ jobs:
           develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
       - name: Run tests
         run: >
-          ./gradlew test --stacktrace --continue 
+          ./gradlew check --stacktrace --continue 
           "-Porg.jetbrains.dokka.javaToolchain.testLauncher=${{ env.JAVA_TEST_VERSION }}"
       - name: Upload build reports
         if: failure()
--- a/.github/workflows/tests-thorough.yml
+++ b/.github/workflows/tests-thorough.yml
@@ -30,7 +30,7 @@ jobs:
           develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
       - name: Run tests
         run: >
-          ./gradlew test --stacktrace --continue
+          ./gradlew check --stacktrace --continue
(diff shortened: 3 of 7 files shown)


whyoleg:
Hm, running `check` causes tests with K1 runs twice, because we have both `test` and `testDescriptors` do run tests with K1 :(

And looks like there is no good way to tell both Gradle and IDEA that we want to have `test` to just run both `testDescriptors` and `testSymbols` but not run anything by itself...

Tapchicoma:
@whyoleg maybe you should structure your tests using Gradle test-suite plugin and then via Gradle properties just disable some suites

adam-enko (approved):
I think this approach is fine, although I have some suggestions:

Instead of disabling the entire task, what about adding JUnit tags to the integration test and excluding them via a filter? This would help with test reporting, as the tests would be reported as 'skipped' in places like Build Scan and TeamCity.

It's strange that `test` and `testDescriptors` are identical tests.

https://github.com/Kotlin/dokka/blob/77330518e098ab5f44ce6714d39f092b7ad09de5/build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts#L62-L63

Instead of another custom workaround, maybe just remove `testDescriptors`?

Alternatively, try creating a lifecycle task `testDescriptors` that depends on `test`, and add `idea.internal.test=true` to [trick IntelliJ into thinking it's a test task](https://github.com/JetBrains/intellij-community/blob/7ea3644a3e0a1532e53032ec66ae7563903dbef3/plugins/gradle/tooling-extension-impl/src/com/intellij/gradle/toolingExtension/impl/model/taskModel/GradleTaskModelBuilder.kt#L53).

whyoleg:
> This would help with test reporting, as the tests would be reported as 'skipped' in places like Build Scan and TeamCity.

just curious: this is needed to understand if the test is `skipped` or just not `run`?
I can add this tag, but adding it to every test in integration tests feels not right :)

> Instead of another custom workaround, maybe just remove testDescriptors?

I've decided to just skip `test` task running and make it `aggregate`. In the end, looks like even now running `test` from IDEA works bad: as no filters are applied to `testSymbols` and so it runs all tests even if only one test selected.

> Rather than dynamically enabling/disabling tasks, I think it'd be more clear to explicitly list the tasks to be executed in GitHub CI.

Let's run `check` on CI so that if some new tests (or other checks) are added (like with DGPv2 `testFunctional`), they are run by default. If we will make `test` dependsOn `testFunctional`, we will have the same issue as with current `test` task in other subprojects in IDEA I described above.
Also, we will not solve initial issue in this case.

adam-enko:
> > This would help with test reporting, as the tests would be reported as 'skipped' in places like Build Scan and TeamCity.
> 
> just curious: this is needed to understand if the test is `skipped` or just not `run`? I can add this tag, but adding it to every test in integration tests feels not right :)

Yeah, fair enough. I think adding a `@SlowTest` JUnit tag would be useful, but maybe later. I was more thinking-out-loud than actually suggesting.


I'm working on fixing the failing Windows tests in https://github.com/Kotlin/dokka/pull/3811

dokka-integration-tests/gradle.properties
```diff
@@ -10,3 +10,7 @@ version=2.0.20-SNAPSHOT
 org.jetbrains.dokka.javaToolchain.mainCompiler=8
 org.jetbrains.dokka.javaToolchain.testLauncher=8
 org.jetbrains.dokka.integration_test.parallelism=2
+
+# !!! ATTENTION: do not commit this !!!
+# Uncomment next line to skip running Dokka integration tests
+#org.jetbrains.dokka.integration_test.skip=true
```
whyoleg, line 16: FYI: @atyrin / @vmishenev after merging this PR you can uncomment this line locally to not run integration tests when running `build`/`check` tasks

whyoleg:
@adam-enko There is still one more test failure on Windows. Do you think may be it will be better to merge this PR and then fix the test, so that the fix will be properly tested on CI?

adam-enko:
> @adam-enko There is still one more test failure on Windows. Do you think may be it will be better to merge this PR and then fix the test, so that the fix will be properly tested on CI?

It's a strange failure, I'm really not sure what's wrong.

Let's suppress the test in this PR and fix it later.

PR #3811, opened 2024-09-19
Update DGP opt-in warnings and messages

adam-enko:
- Write separate warning log message (fix https://youtrack.jetbrains.com/issue/KT-70998/DGP-v2-warning-is-not-shown-in-IDEA-log-only-in-console)
- Update styling and formatting.
- Add link to https://kotl.in/slack
- Remove the ascii box, it did not render correctly on GitHub Windows CI.

#### DGP v1 warning

image

#### DGP v2 with helpers

image

#### DGP v2

image


dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -249,6 +265,18 @@ internal abstract class PluginFeaturesService : BuildService<PluginFeaturesServi
         private const val K2_ANALYSIS_NO_WARN_FLAG_PRETTY =
             "$K2_ANALYSIS_ENABLED_FLAG.noWarn"
 
+        @Suppress("ObjectPrivatePropertyName")
+        private val `To learn about migrating read the migration guide` = /* language=text */ """
```
whyoleg, line 269: minor: that's an interesting usage of backticks in functions :)
But, IMO, it's an abuse and it will be better to just inline documentation. I don't think we will change it afterwards anyway after release

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MigrationMessagesTest.kt
```diff
@@ -26,141 +28,203 @@ class MigrationMessagesTest : FunSpec({
...
-                """.trimMargin()
-            output shouldNotContain "Dokka Gradle Plugin V1"
-        }
+        private val expectedV1Message = /* language=text */ """
+            |Dokka Gradle plugin V1 is deprecated, and will be removed in Dokka version 2.1.0
+            |Please migrate to Dokka Gradle plugin V2. This will require updating your project.
+            |To learn about migrating read the migration guide https://kotl.in/dokka-gradle-migration
+            |
+            |To start migrating to Dokka Gradle plugin V2 add
+            |    ${PLUGIN_MODE_FLAG}=V2EnabledWithHelpers
```
whyoleg, line 191: minor: I think for tests it's better to inline the string, so that we really see what the warning is and check if it's for some reason changed

whyoleg (approved):
LGTM with small remarks

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MigrationMessagesTest.kt
adam-enko: I would agree, if `PLUGIN_MODE_FLAG` was the same as `PluginFeaturesService#PLUGIN_MODE_FLAG`, but the test redefines the flags. So if the flags change in `PluginFeaturesService`, then the test will fail.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
adam-enko: It was a hassle trying to keep the messages in-sync, and using a variable for the repeated parts helped minimise mistakes.

Extracting the repeated parts into variables also helps highlight the differences between the messages.

dkrasnoff (approved):

dkrasnoff:
lgtm 👍

PR #3812, opened 2024-09-19
Fix external documentation link default values

adam-enko:
Prevent the default value of `true` from overriding the specific values provided from the top level Dokka DSL.

[KT-70908](https://youtrack.jetbrains.com/issue/KT-70908/DGP-v2-passed-Android-package-lists-to-non-Android-projects)

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -218,7 +218,7 @@ constructor(
                 }
 
                 maybeCreate("jdk") {
-                    enabled.convention(this@dss.enableJdkDocumentationLink)
+                    enabled.set(this@dss.enableJdkDocumentationLink)
                     url(this@dss.jdkVersion.map { jdkVersion ->
                         when {
                             jdkVersion < 11 -> "https://docs.oracle.com/javase/${jdkVersion}/docs/api/"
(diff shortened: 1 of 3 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
```diff
@@ -218,7 +218,7 @@ constructor(
                 }
 
                 maybeCreate("jdk") {
-                    enabled.convention(this@dss.enableJdkDocumentationLink)
+                    enabled.set(this@dss.enableJdkDocumentationLink)
```
Tapchicoma, line 221: Hm, why does convention not work here?

whyoleg (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
Tapchicoma: Could it skip these known names like `if(name !in setOf("jdk", "kotlinStdlib", "androidSdk", "androidX"))`?

Usually using actual values instead of conventions in entities that could be configured by users are a code smell.

adam-enko: I agree that it looks unusual, but I think it's correct to use `set()` instead of `convention() here.

The root cause of the non-idiomatic behaviour is that the Dokka DSL is unusual. It has two ways of configuring whether the external links are active - either via `DokkaSourceSetSpec#enableAndroidDocumentationLink` or via `DokkaExternalDocumentationLinkSpec#getEnabled`. Why are there two? Which one should take priority? This should be tidied instead.

The actual DSL is less important in DGPv2: the values are determined automatically based on the buildscript environment (e.g. if AGP is applied), so hopefully users don't configure them manually.

In the future I would like to revisit the external documentation links to improve them. I'd like to add a _lot_ more of them, and to dynamically enable/disable them based on the classpath, so no manual configuration is required at all. And to make it easier to register new ones.

Tapchicoma: Please create an issue for this to not forget about it

Tapchicoma (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
adam-enko: https://youtrack.jetbrains.com/issue/KT-71894/Revisit-external-documentation-links-in-DGPv2

Issue #3382, opened 2023-11-28
KMP: References is not linked if they are in shared code and there is an intermediate level between them


atyrin:
Create a KMP library with KGP 1.9.20. Declare 2 ios targets. That is required for generating a few layers of shared native source sets

```Kotlin
// commonMain
class Dokka

// iosMain
class IosDokka

// iosX64Main

/**
 * Ths fun consume [Dokka] and [IosDokka]
 */
fun consumeDokka(d: Dokka, iosD: IosDokka) {}
```

References to `commonMain` will not have a links
<img width="450" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/de93c7da-ddb4-4767-9c0e-7ee2696c9335">

**Installation**
- Dokka version: 1.9.20

Related issue for K2: #3377. Probably, it is even a duplicate.

fAntel:
I have the same issue and there is one more case. Documentation from common isn't inherited in ios. If you have an interface or smth. in common source set and implementation in ios source set there would be no documentation for overridden methods and there even would be uncodument warning for them. For example:
```Kotlin
// commonMain
/** Common interface */
interface Dokka {
    /** Method to implement */
    fun foo()
}

// iosMain
class IosDokka private constructor() : Dokka {
    override fun foo() {}
}

// jvmMain
class JvmDokka private constructor() : Dokka {
    override fun foo() {}
}
```

```
Undocumented: com.example/IosDokka/ (ios)
Undocumented: com.example/IosDokka/foo/#/ (ios)
Undocumented: com.example/JvmDokka/ (jvm)
```

No link for common class and no documentation for overridden method in for iosMain
![image](https://github.com/Kotlin/dokka/assets/10121347/181cf297-8c6d-4e59-bfbd-e111fbba4b7b)

Works fine for jvmMain
![image](https://github.com/Kotlin/dokka/assets/10121347/25f640f1-5e2f-4674-b81a-15b3a6bdaaa7)

PR #3814, opened 2024-09-19
Fix DGP/KMP integration, so Dokka can 'see' code from shared source sets in target source sets

adam-enko:
There are 3 main fixes in this PR:

1. Update determining the Kotlin target for each source set. (See `KotlinAdapter#determineKotlinPlatform`). Filter out metadata targets (they're not relevant for target determination), and only fall back to `KotlinPlatform.Common` if necessary.
2. Correctly resolve Konan dependencies: Use `compilation.konanTarget.name` instead of `compilation.target.name`.
3. Fix fetching the classpath: For Android projects, fetch both `jar` and `android-classes-jar` dependencies. For KGP projects, use `compileDependencyFiles`.

The remaining changes are to support these fixes:

- Add functional test to verify that Dokka can correctly generate docs when 'leaf' source sets (e.g. linuxX64) reference 'branch' source sets (e.g. commonMain, nativeMain).
- Update DGPv2 AGP integration test data.
- Create a factory for creating DokkaSourceSetSpecs, which means DGP can consistently use `convention()` to set default values.
- Add/update KDoc.

Should fix KT-70857


dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -320,8 +376,15 @@ private class KotlinCompilationDetailsBuilder(
                     .artifactView {
                         // Android publishes many variants, which can cause Gradle to get confused,
                         // so specify that we need a JAR and resolve leniently
-                        if (compilation.target.platformType == KotlinPlatformType.androidJvm) {
-                            attributes { artifactType("jar") }
+                        if (compilation.target.platformType == androidJvm) {
+                            attributes { artifactType(artifactType) }
+
+                            // Setting lenient=true is not ideal, because it might hide problems.
```
whyoleg, line 382: just curious: is it possible to somehow `log` those problems in case of `lenient=true` or are they just silently ignored? just for troubleshooting?

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KmpCommonSourceSharedWithDependentsTest.kt
```diff
@@ -0,0 +1,149 @@
...
+import io.kotest.matchers.sequences.shouldBeEmpty
+import org.gradle.testkit.runner.TaskOutcome.*
+import org.jetbrains.dokka.gradle.internal.DokkaConstants.DOKKA_VERSION
+import org.jetbrains.dokka.gradle.utils.*
+import kotlin.io.path.isRegularFile
+import kotlin.io.path.useLines
+import kotlin.io.path.walk
+
+@Ignored("KMP: References is not linked if they are in shared code and there is an intermediate level between them https://github.com/Kotlin/dokka/issues/3382")
+class KmpCommonSourceSharedWithDependentsTest : FunSpec({
```
whyoleg, line 18: I believe this test should be placed into `dokka-integration-tests` so that we will be able to check it with different Kotlin versions.
In my understanding, `functional tests` in DGPv2 are more about plugin behavior while this test is more about overall Dokka behavior and so should be in IT

whyoleg (approved):
Nice!
Changes to defaults are looks logical for me, only the test placement worries me :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: It might be possible to log them (`configuration.incoming.artifactView { ... }.artifacts.failures`), but I'd be cautious about using them. It's very easy to accidentally cause Configuration Cache errors with ArtifactView ([example](https://github.com/gradle/gradle/issues/27594)), and I would expect a lot of failures (because this code takes a 'shotgun' approach and just tries to grab a lot of dependencies) so any actual errors could be difficult to differentiate from noise.

I agree logging the failures could be useful though, but I think it'd be best to be careful and check properly.

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/kotestFiles.kt
```diff
@@ -91,7 +91,7 @@ private fun describeFileDifferences(
                     /* revisedFileName = */ actualFile.relativeTo(actualDir).invariantSeparatorsPathString,
                     /* originalLines = */ expectedLines,
                     /* patch = */ patch,
-                    /* contextSize = */ 3,
+                    /* contextSize = */ 0,
```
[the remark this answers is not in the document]
adam-enko: Yeah, when there are lot of diffs then the test failure output is HUGE and hides the actual human-readable failure message.

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KmpCommonSourceSharedWithDependentsTest.kt
adam-enko: Good idea, done.

Could you check the expected data in `dokka-integration-tests/gradle/projects/it-kotlin-multiplatform/expectedData/html`? There are errors, but I'm not sure if any are DGP's responsibility, or are caused by https://github.com/Kotlin/dokka/issues/3382

antohaby (approved):

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KmpCommonSourceSharedWithDependentsTest.kt
whyoleg: I've checked `expectedData/html` and it looks reasonable and the error is most likely caused by #3382.
let's ignore test and re-check it in scope of https://github.com/Kotlin/dokka/issues/3930 (DGPv2 + K2)

adam-enko: Hmmmmm it looks like locally I get `Error type: Unresolved type for NativeMainCls`, but not on TeamCity?? Why??

https://ge.jetbrains.com/s/2cdhjngaftnhk/tests/task/:dokka-integration-tests:gradle:test/details/org.jetbrains.dokka.it.gradle.KotlinMultiplatformIT/generate%20dokka%20HTML(DokkaGradleProjectRunner)%5B1%5D?focused-exception-line=2-100&top-execution=4

whyoleg: The difference in OS (linux vs macOS) and so probably the difference is in dependencies passed inside KGP. E.g `iOS` targets are not compilable on linux and so probably the classpath (or some other configuration) is somehow different, which causes this behavior.

In this case, I think that for this test probably having just `linux` targets should be enough to reproduce the issue, correct?
We could also try to add `mingw` target (or `androidNative` targets), or add additional intermediate sourceSets somewhere, so that all targets will be cross-compiled on all OSs.

adam-enko: I could understand it if locally, on my Mac, I would get no errors and on Linux the macOS targets would cause `Unresolved type` errors, but I don't see why it's the other way around. I'd like to understand the problem more before trying to work around it, but I'm not sure how.

I'll check with my team.

PR #3820, opened 2024-09-23
Add DGPv2 example projects

adam-enko:
- Convert the existing DGPv1 examples to demonstrate DGPv2.
- Import [Dokkatoo examples](https://github.com/adamko-dev/dokkatoo/tree/v2.4.0/examples).

Part of KT-71346

To keep the PR smaller, tests to ensure these projects work as expected will be added in a separate PR https://github.com/Kotlin/dokka/pull/3820.


examples/gradle-v2/versioning-multimodule-example/README.md
```diff
@@ -0,0 +1,26 @@
...
+[GitHub Pages](https://kotlin.github.io/dokka/examples/dokka-versioning-multimodule-example/htmlMultiModule/index.html).
+
+![screenshot demonstration of output](demo.png)
+
+### Running
+
+Run `dokkaHtmlMultiModule` task to generate documentation for this example:
+
+```bash
+./gradlew dokkaHtmlMultiModule
```
atyrin, line 22: Outdated task suggested

adam-enko: fixed, thanks!

atyrin (approved):

examples/gradle-v2/custom-format-example/README.md
```diff
@@ -0,0 +1,17 @@
+## Dokka custom format example
```
whyoleg, line 1: It would be better to rename this example (folder and title) to something like `custom-styles` to not confuse with Dokka formats

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+            dependencies {
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.1")
+            }
+        }
+    }
+}
+
+dokka {
+    // Create a custom source set not known to the Kotlin Gradle Plugin
+    dokkaSourceSets.register("customSourceSet") {
```
whyoleg, line 27: what is the reason for this?
I don't think we need to show it in `KMP` example.
If you think it's needed, than it's better to create a separate example for custom source set creation (later)

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+    jvm()
+    linuxX64()
+    macosX64()
+    js(IR) {
+        browser()
+    }
+    sourceSets {
+        commonMain {
+            dependencies {
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.1")
```
whyoleg, line 19: let's update version to the latest :)

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+    kotlin("multiplatform") version "1.9.25"
+    id("org.jetbrains.dokka") version "2.0.20-SNAPSHOT"
+}
+
+group = "org.dokka.example"
+version = "1.0-SNAPSHOT"
+
+kotlin {
+    jvm()
+    linuxX64()
```
whyoleg, line 11: it's better to add multiple targets here if you are willing to test those, because Dokka could have issues with intermediate sourceSets if there is single target (https://github.com/Kotlin/dokka/pull/3663)

examples/gradle-v2/versioning-multimodule-example/docs/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+    moduleName.set("Dokka Versioning Example")
+}
+
+val currentVersion = "1.0"
+val previousVersionsDirectory: Directory = layout.projectDirectory.dir("previousDocVersions")
+
+dokka {
+    pluginsConfiguration {
+        // Main configuration for the versioning plugin:
+        versioning {
```
whyoleg, line 24: BTW, what will happen if user will configure `versioning` DSL, but will not add `versioning-plugin` dependency?
Can't we just apply `versioning-plugin` dependency if user configured `versioning` DSL?

whyoleg (approved):
Nice!

just some small questions/remarks

examples/gradle-v2/custom-format-example/README.md
adam-enko: YES this same confusing has been bugging me for a while

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
adam-enko: I have no idea, it's in the original DGPv1 example

https://github.com/Kotlin/dokka/blob/7ccc48eda17ecd82e2974002faf7e5da188ae2c5/examples/gradle/dokka-multiplatform-example/build.gradle.kts#L39-L44

I can just remove it? I'm happy either way.

whyoleg: Let's drop it

examples/gradle-v2/versioning-multimodule-example/docs/build.gradle.kts
adam-enko: If versioning is configured, but the plugin isn't provided, then nothing will happen. 

I think we could definitely improve the DSL to automatically add the plugin. I think it'd need to be an explicit `versioning { enabled = true }` option though.

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
adam-enko: Sure, I think adding most targets here shouldn't be a problem.

examples/gradle-v2/versioning-multimodule-example/docs/build.gradle.kts
whyoleg: I think, that as we provide DSL for it then it will be better to do apply it also via DSL. Could you create an issue for this case?

whyoleg: I believe it's also fine, if it will work in the same way as with javadoc.
so, f.e user will need to apply `dokka-versioning` Gradle plugin to get access to `pluginsConfiguration.versioning`.
This will also allow us to understand, how frequently `versioning` plugin is used

whyoleg: created an issue for this: https://youtrack.jetbrains.com/issue/KT-72713/DGP-v2-improve-UX-of-configuring-versioning-plugin

PR #3821, opened 2024-09-23
Test DGPv2 example projects

adam-enko:
Test the DGPv2 examples by comparing using golden testing.

The example projects are executed using Gradle TestKit. The generated Dokka output is compared against data committed into the directory `./dokka-integration-tests/gradle/src/testExampleProjects/expectedData/`

To re-emphasise, everything in `testExampleProjects/` is test data, and these are the remaining modified files: 

image

[KT-71346](https://youtrack.jetbrains.com/issue/KT-71346/Update-Dokka-Gradle-example-projects-to-work-with-DGPv2)

### Dependencies

- This PR requires https://github.com/Kotlin/dokka/pull/3820


dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko, line 1: This is the main test class to review. I'm adding a comment to highlight it because the GitHub UI is hiding the diff.

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -155,8 +175,6 @@ fun registerTestProjectSuite(
                     .inputFile("templateSettingsGradleKts", templateSettingsGradleKts)
                     .withPathSensitivity(NAME_ONLY)
 
-                devMavenPublish.configureTask(this)
```
adam-enko, line 158: Moved into a `configureEach {}` on line 206.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko, line 1: This is the main test class to review. I'm adding a comment to highlight it because the GitHub UI is hiding the diff.

whyoleg (changes requested):
For some time we had an idea of having such tests (which compares HTML), so that's nice. But I'm a bit afraid that if we will have some expected HTML changes, probably the process of updating the data will be cumbersome.
Would be nice to understand: how hard it will be to maintain those examples?

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,347 @@
...
+import org.junit.jupiter.params.provider.ArgumentsProvider
+import org.junit.jupiter.params.provider.ArgumentsSource
+import java.nio.file.Path
+import java.util.stream.Stream
+import kotlin.io.path.*
+import kotlin.streams.asStream
+
+class ExampleProjectsTest {
+
+    class TestCaseProvider : ArgumentsProvider {
```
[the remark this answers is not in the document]
adam-enko: afaik, no :(

https://github.com/gradle/gradle/issues/19897

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -177,3 +195,42 @@ val checkoutKotlinxSerialization by tasks.registering(GitCheckoutTask::class) {
...
+    val testExampleProjects by suites.registering(JvmTestSuite::class) {
+        targets.configureEach {
+            testTask.configure {
+
+                val exampleGradleProjectsDir = projectDir.resolve("../../examples/gradle-v2")
+                systemProperty
+                    .inputDirectory("exampleGradleProjectsDir", exampleGradleProjectsDir)
+                    .withPathSensitivity(RELATIVE)
+
+                val expectedDataDir = layout.projectDirectory.dir("src/testExampleProjects/expectedData")
```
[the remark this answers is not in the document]
adam-enko: I found the workflow of updating the files fairly easy to follow, although I'm sure it could be improved further.

Failures log a clickable link to the actual generated docs, so you just need to open the directory and drag the file into the expectedData dir.

image

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,347 @@
...
+                )
+
+        //first build should store the configuration cache
+        configCacheRunner.build {
+            output shouldContain "BUILD SUCCESSFUL"
+            output shouldContain "Configuration cache entry stored"
+            output shouldNotContain "problems were found storing the configuration cache"
+        }
+
+        val ccReport = testCase.project.loadConfigurationCacheReportData().shouldNotBeNull()
```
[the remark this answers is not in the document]
adam-enko: Thanks, fixed.

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -177,3 +195,42 @@ val checkoutKotlinxSerialization by tasks.registering(GitCheckoutTask::class) {
...
+                systemProperty
+                    .inputDirectory("expectedDataDir", expectedDataDir)
+                    .withPathSensitivity(RELATIVE)
+
+                systemProperty.inputProperty(
```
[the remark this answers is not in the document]
adam-enko: TeamCity OOMs when the tests are run in parallel. I added a comment to explain.

abdulowork:
It seems the GH UI is lagging pretty heavily for be due to the test fixture; rebasing the fixture to the first (or last) commit and making a stacked branch will probably reduce the lag

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleCCReportData.kt
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Extract and parse the JSON data from [ccReport].
+ */
+private fun parseCCReportData(ccReport: Path): ConfigurationCacheReportData {
+    val reportData = ccReport.readText()
+        .substringAfter("// begin-report-data", "could not find 'begin-report-data'")
+        .substringBefore("// end-report-data", "could not find 'end-report-data'")
+
+    return Json.decodeFromString<ConfigurationCacheReportData>(reportData)
```
abdulowork, line 71: Looking at this CC check I have a couple of thoughts:
* Could we just `ignoreUnknownKeys` and only read the [totalProblemCount](https://github.com/Kotlin/dokka/pull/3821/files#diff-e60a0378c359ab13552c8a4083fb4f95be9a4b66f36d7d4cbf4419e8707600f1R332)? Otherwise this will break if the CC report json format changes
* Do we really want to parse this json at all? What is the benefit compared to just failing the test?
* Does this CC check ever actually run for a build that fails due to a CC serialization error? I think the test is going to fail prematurely with a build failure or an assertion failure before the `totalProblemCount` assertion runs

dokka-integration-tests/gradle/build.gradle.kts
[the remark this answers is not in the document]
abdulowork: It would be nice if you could just remove the test data and rerun the test to regenerate it

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,344 @@
...
+            val ccReport = testCase.project.loadConfigurationCacheReportData().shouldNotBeNull()
+            ccReport.asClue {
+                it.totalProblemCount shouldBe 0
+            }
+        }
+
+        // second build should reuse the configuration cache
+        configCacheRunner.build {
+            output shouldContain "BUILD SUCCESSFUL"
+            output shouldContain "Configuration cache entry reused"
```
abdulowork, line 341: I think it might make sense to `clean` before rerunning this build; this might catch something like a build output file becoming a CC cache entry. When would this fail otherwise?

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,344 @@
...
+            .build {
+                output shouldContain "BUILD SUCCESSFUL"
+            }
+
+        // Re-run generate, and verify that the Dokka tasks are cached.
+        testCase.project.runner
+            .addArguments(
+                testCase.dokkaGenerateTask,
+                "--stacktrace",
+                "--build-cache",
```
abdulowork, line 255: I think this test passes even if we remove `--build-cache`. Shouldn't this test do `clean` before rebuilding to make sure build cache entries are being reused? I would also expect this test to check that tasks are fetch `FROM-CACHE`, rather than them being `UP-TO-DATE`

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,344 @@
...
+                withClue({
+                    """
+                    expectedDataDir: ${expectedDataDir.toUri()}
+                    actualOutputDir: ${dokkaOutputDir.toUri()}
+                    """.trimIndent()
+                }) {
+                    withClue("expect file trees are the same") {
+                        val expectedFileTree = expectedDataDir.toTreeString()
+                        val actualFileTree = dokkaOutputDir.toTreeString()
+                        expectedFileTree shouldBe actualFileTree
```
abdulowork, line 208: This test will be somewhat flaky if you don't filter dot files (or filter files per `.gitignore`). E.g. Poking around test fixtures with Finder on macOS will create a bunch of `.DS_Store` files which will fail this test when run locally

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    private fun testDokkaOutput(
+        testCase: TestCase,
+        format: String,
+    ) {
+        val expectedDataDir = testCase.expectedDataDir.resolve(format)
+        val dokkaOutputDir = testCase.dokkaOutputDir.resolve(format)
+
+        assert(expectedDataDir.exists() && expectedDataDir.isDirectory()) {
```
abdulowork, line 187: nit: I think `isDirectory` is sufficient

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,344 @@
...
+                }) {
+                    withClue("expect file trees are the same") {
+                        val expectedFileTree = expectedDataDir.toTreeString()
+                        val actualFileTree = dokkaOutputDir.toTreeString()
+                        expectedFileTree shouldBe actualFileTree
+                    }
+
+                    withClue("expect directories are the same") {
+                        dokkaOutputDir.shouldHaveSameStructureAs(expectedDataDir, skipEmptyDirs = true)
+                        dokkaOutputDir.shouldHaveSameStructureAndContentAs(expectedDataDir, skipEmptyDirs = true)
```
abdulowork, line 213: Do we need `shouldHaveSameStructureAs` and `shouldHaveSameStructureAndContentAs` simultaneously?

nb: `shouldHaveSameStructureAs` calls `shouldHaveSameStructureAndContentAs2` overload which doesn't actually compare contents

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleCCReportData.kt
adam-enko: > * Could we just `ignoreUnknownKeys` and only read the [totalProblemCount](https://github.com/Kotlin/dokka/pull/3821/files#diff-e60a0378c359ab13552c8a4083fb4f95be9a4b66f36d7d4cbf4419e8707600f1R332)? Otherwise this will break if the CC report json format changes

Yes, we could add `ignoreUnknownKeys`, but at the moment it's simple to parse everything. We could add it in the future if necessary.

The CC test here is a 'soft launch'. In the future I want to specifically read the JSON CC report data and test the registered inputs are as expected. 

image

> * Do we really want to parse this json at all? What is the benefit compared to just failing the test?

The difference with checking the CC report JSON is we can be more certain if CC fails. Just parsing the console output can lead to false positives and uncertain results. We can also get and report more specific information about failures, which might not be present in the console output.

> * Does this CC check ever actually run for a build that fails due to a CC serialization error? I think the test is going to fail prematurely with a build failure or an assertion failure before the `totalProblemCount` assertion runs

iirc there are some CC problems that are reported, but don't fail the build. But I can't remember them right now...

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -0,0 +1,344 @@
...
+import java.util.stream.Stream
+import kotlin.io.path.*
+import kotlin.streams.asStream
+
+class ExampleProjectsTest {
+
+    class TestCaseProvider : ArgumentsProvider {
+
+        override fun provideArguments(context: ExtensionContext): Stream<Arguments> {
+            val exampleProjectDirs = exampleGradleProjectsDir
```
abdulowork, line 36: It might make sense to check that `exampleProjectDirs` is not empty or these tests might silently become noop in the future

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
```diff
@@ -164,11 +162,32 @@ class GradleProjectTest(
...
 
+    /**
+     * Update all repositories in any `settings.gradle(.kts)` file.
+     *
+     * All `repositories {}` will be modified to add the dev maven repos, [mavenRepositories].
+     */
+    fun updateSettingsRepositories() {
+        projectDir.walk()
+            .filter { it.name == "settings.gradle.kts" || it.name == "settings.gradle" }
+            .forEach { p ->
```
abdulowork, line 173: Let's replace `p` and `ind` with something more descriptive

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
```diff
@@ -164,11 +162,32 @@ class GradleProjectTest(
...
+                val repoLine = p.useLines { it.firstOrNull { l -> l.trim() == "repositories {" } }
```
abdulowork, line 174: This implementation looks somewhat fragile because it relies on seeking `repositories {` substring. Also what is the use-case for calling this function, but then doing nothing if the repositories are not specified in the test project?

Can we simplify this implementation to:
* Find and assert `settings.gradle(.kts)` file exists
* Inject `pluginManagement` and `dependencyResolutionManagement` blocks with necessary repositories

abdulowork:
I like that this PR expands on E2E testing in Dokka; however, is it really feasible to test the generated html output in the long term? Imho, at the E2E level, html is more of an implementation detail. I imagine if someone is going to implement a major Dokka redesign in the future, they will have to regenerate these html files after every UI change.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
abdulowork: > this might catch something like a build output file becoming a CC cache entry

Actually, I guess rebuilding after execution phase without cleaning is more likely to catch an issue like that

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
adam-enko: > This implementation looks somewhat fragile because it relies on seeking `repositories {` substring. Also what is the use-case for calling this function, but then doing nothing if the repositories are not specified in the test project?

If the repositories aren't correctly updated, then I would expect that the test will fail. The version of DGP is not published to a remote repository, and is only available in a local directory.

> Can we simplify this implementation to:
> 
> * Find and assert `settings.gradle(.kts)` file exists

Some of the example projects (e.g. the composite build example) have multiple `settings.gradle`s, so they need to be discovered and updated dynamically. We could try to manually specify this per project, but since the tests should fail without a correct repository, doing an automatic find/replace seems fine to me.

> * Inject `pluginManagement` and `dependencyResolutionManagement` blocks with necessary repositories

Do you mean appending multiple `pluginManagement {}` and `dependencyResolutionManagement {}` blocks at the end of the file?

That could work for `dependencyResolutionManagement {}`, because Gradle will allow repeated blocks, but for `pluginManagement {}` Gradle will fail if there are multiple blocks, so `pluginManagement{ repositories {} }` has to be updated in-place.

Maybe an init plugin could help? But I'm not sure if that could work with the composite build example.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko: If the example project directory is empty then the test will fail.

```
* What went wrong:
Directory '/private/var/folders/_3/ps18c9f56bx2j4fyxt0z7sn40000gp/T/9203619320687405921/basic-gradle-example' does not contain a Gradle build.
```

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko: It's not a big difference, but I found the `shouldHaveSameStructureAs` assertion failure message is more useful.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko: Good catch, fixed, thanks!

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko: Trying to filter out some files seems like a lot of work, and for something that might only happen sometimes, and can be resolved by manually deleting the files.

adam-enko:
> I like that this PR expands on E2E testing in Dokka; however, is it really feasible to test the generated html output in the long term? Imho, at the E2E level, html is more of an implementation detail. I imagine if someone is going to implement a major Dokka redesign in the future, they will have to regenerate these html files after every UI change.

For some context, Dokka's output formats don't change very often. I've used a similar sort of direct file comparison testing in Dokkatoo for over a year, and I haven't had any instability.

We can work on making the reference files easier to update, but that's something I'd rather look at if/when it becomes a problem (https://xkcd.com/1205/).

dokka-integration-tests/gradle/build.gradle.kts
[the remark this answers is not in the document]
whyoleg: may be more safe alternative: add an explicit Gradle property to force regeneration of the output

Additionally this will be very useful, as it will be possible to force regeneration and see the diff in IDEA, and not in Gradle output (which is not very nice).
Without this, I think that the output will be not very useful, as even simple change in Dokka generation could slightly change output

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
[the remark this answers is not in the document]
whyoleg: I agree here, it would be better to generate sources, so that it will be possible to run single test.

I would not expect amount of examples to change frequently

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleCCReportData.kt
whyoleg: For me this check with JSON parsing looks not so useful, as the build should definitely fail, if CC is not supported for some reason with CC enabled.
I feel like it brings more harm (maintainability, possible format changes, etc) than profits (which are unknown to me).
E.g ` We can also get and report more specific information about failures, which might not be present in the console output.` It's always possible to run the test locally, anyway, to fix/investigate the issue we will need it. Of course it's not an option for TC, but I think that we could just then store CC reports somewhere - IMO, this will be easier to maintain and will have more profit (as we could read the whole report)

whyoleg (changes requested):
I don't think, that we should merge this PR without some way to update `expectedData`.
E.g https://github.com/Kotlin/dokka/pull/3821#discussion_r1808688799

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
abdulowork: I disagree; false positive test failures add a lot of DevX friction. I feel it shouldn't be hard to filter out at least the `.DS_Store` files

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleTestKitUtils.kt
abdulowork: My main argument for not doing this substring search/replace is that I find reasoning about such code cumbersome if the behavior doesn't match my expectation about what this code should be doing. If you want to keep the substring search/replace, I would advise to instead use an explicit marker:
```
repositories {
  <repositories_marker>
}
```
which will explicitly fail to evaluate if it wasn't processed

> pluginManagement{ repositories {} } has to be updated in-place.

Can we avoid specifying `pluginManagement` to avoid having to patch it later?

> work with the composite build example

Have you tried it out? I would expect init scripts to apply to all project hierarchies in the composite build

dokka-integration-tests/gradle/build.gradle.kts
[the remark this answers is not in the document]
abdulowork: In my experience the most common DevX for test data generation is the following: when there is no test data, you run the test which writes the test data and immediately fails (with an explicit "test data generation run" message)

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleCCReportData.kt
abdulowork: >  there are some CC problems that are reported, but don't fail the build

Are these going to be reflected in the `totalProblemCount`? In case they are, I still wonder if specifying `org.gradle.configuration-cache.problems=fail` and `org.gradle.configuration-cache.max-problems=0` will achieve  failing the build with CC errors

> report more specific information about failures

Would it be possible to zip the entire `build/reports/configuration-cache` directory into the GH/TC artifacts to be able to inspect the full Gradle report?

> test the registered inputs are as expected.

I want to highlight that this might bring a maintenance cost where some CC inputs are going to change due to changes like KGP version, which are unlikely to be of interest to Dokka.

abdulowork:
I still want to challenge the idea that we should test html output:
> Dokka's output formats don't change very often

but then is there a point in testing the html output that doesn't really change? In general I would expect E2E testing to be rather carried out using a UI testing framework like Selenium

adam-enko:
> I don't think, that we should merge this PR without some way to update `expectedData`. E.g [#3821 (comment)](https://github.com/Kotlin/dokka/pull/3821#discussion_r1808688799)

There is already a way to update the expected data 

https://github.com/Kotlin/dokka/pull/3821#discussion_r1780789277

adam-enko:
> I still want to challenge the idea that we should test html output:
> 
> > Dokka's output formats don't change very often
> 
> but then is there a point in testing the html output that doesn't really change? In general I would expect E2E testing to be rather carried out using a UI testing framework like Selenium

Yes, there is a point. We want to make sure that changes in DGP/AGP/Gradle/KGP don't affect the HTML outpoint.

adam-enko:
> What are the ALL steps needed so that the developer, who will want to merge this PR need to do?

1. Run tests.
2. See the error message.
3. Copy the files from the test directory into the expected-data directory.

It takes a few minutes to do.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko: > I disagree; false positive test failures add a lot of DevX friction.

Yes, false positive test failures add friction, but there haven't been any false positive failures.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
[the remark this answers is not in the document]
adam-enko: Generating sources would be too much work and hard to maintain.

dokka-integration-tests/gradle/build.gradle.kts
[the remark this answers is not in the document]
adam-enko: Introducing machinery to automatically updating missing test data introduces a risk of the tests being silently updated and resulting in the tests passing, even if the data is outdated.

If updating the test data becomes a burden then we can look into it in the future. For now, it's not necessary.

whyoleg (approved):
Let's see how it goes with maintaining of such tests...

Also, let's add information on how to investigate failures and update expected data in https://github.com/Kotlin/dokka/blob/master/dokka-integration-tests/README.md before merging

PR #3797, opened 2024-09-11
Update DGP plugin mode flags

adam-enko:
Use `org.jetbrains.dokka.experimental.gradle.pluginMode=[V1Enabled|V2EnabledWithHelpers|V2Enabled]` to control the behaviour.

Use `org.jetbrains.dokka.experimental.gradle.pluginMode.noWarn=[true|false]` to enable/disable the warning logs.

KT-71236

PR #3822, opened 2024-09-23
Fix DGPv2 migration helpers

adam-enko:
Fix minor issues the DGPv2 migration helpers.

(The migration helpers are intended to ease the transition from DGPv1 to DGPv2. When a project with Dokka enables v2 mode, then the built scripts should still be valid. To that end, DGPv2 must register 'dummy' DGPv1 tasks that appear to match the existing DGPv1 tasks.)

- Set the DGPv1 tasks to `"other"` instead of `""` (fix [KT-71678](https://youtrack.jetbrains.com/issue/KT-71678))
- Configure all DGPv1 tasks by using `AbstractDokkaTask`, not just `DokkaTask`s.
- Add a test to verify the tasks exist.


whyoleg (approved):

adam-enko:
Marked as a draft because https://github.com/Kotlin/dokka/pull/3797 should be reviewed and merged first, and then I'll need to fix merge conflicts.

PR #3823, opened 2024-09-23
Update KotlinGradlePluginVersion.kt

GeorgCantor:
**Single Split Operation:** The split operation is now limited to 3 to directly get major, minor, and patch versions.

**Mapping with `takeWhile`:** This ensures that any suffixes are ignored while still allowing for clean extraction of numeric values.

**Destructuring Declaration:** Using destructuring to directly assign `major`, `minor`, and `patch` makes the code cleaner.

**Improved Numeric Check:** The numeric check is now integrated into the validation step for clarity.


adam-enko:
Hi @GeorgCantor, thanks for the interest in Dokka!

We're currently re-implementing the Dokka Gradle plugin, and the 'classic' plugin will be deprecated. As such, we want to avoid making unnecessary changes to the classic plugin, so we can't accept the PR at present.

You provided an explanation of your changes, and I agree with the principle of improving the code where possible, but I'm not sure what your motivation is. Does the current implementation cause issues? If so, then please open an new issue to describe any problems and we can take a further look.

PR #3825, opened 2024-09-24
Unify and update build settings convention

adam-enko:
- Merge `dokkasettings.build-cache.settings.gradle.kts` into the Develocity convention (build cache must use the develocity credentials).
- Create a fat jar for the Develocity plugins.
- Commit the fat jar to the repo, so the build-settings-logic project to use its own convention plugin.


build-settings-logic/src/main/kotlin/dokkasettings.gradle-enterprise.settings.gradle.kts
```diff
@@ -66,3 +66,17 @@ develocity {
         }
     }
 }
+
+buildCache {
```
CristianGM, line 70: buildcache now goes inside develocity extension

PR #3830, opened 2024-09-26
DGPv2: Add suppressInheritedMembers/suppressObviousFunctions to help with migrating

adam-enko:
KT-71760

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -161,11 +161,11 @@ constructor(
             moduleOutputDirectory.convention(dokkaExtension.dokkaModuleDirectory.dir(formatName))
             suppressInheritedMembers.convention(
                 @Suppress("DEPRECATION")
-                dokkaExtension.suppressInheritedMembers.orElse(false)
+                dokkaExtension.suppressInheritedMembers
             )
             suppressObviousFunctions.convention(
                 @Suppress("DEPRECATION")
-                dokkaExtension.suppressObviousFunctions.orElse(true)
+                dokkaExtension.suppressObviousFunctions
             )
         }
     }
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -123,6 +123,11 @@ constructor(
             }
         )
 
+        @Suppress("DEPRECATION")
+        dokkaExtension.suppressInheritedMembers.convention(false)
+        @Suppress("DEPRECATION")
+        dokkaExtension.suppressObviousFunctions.convention(true)
+
         dokkaExtension.dokkaSourceSets.configureDefaults(
             sourceSetScopeConvention = dokkaExtension.sourceSetScopeDefault
         )
@@ -154,8 +159,14 @@ constructor(
             offlineMode.convention(false)
             outputDirectory.convention(dokkaExtension.dokkaPublicationDirectory.dir(formatName))
             moduleOutputDirectory.convention(dokkaExtension.dokkaModuleDirectory.dir(formatName))
-            suppressInheritedMembers.convention(false)
-            suppressObviousFunctions.convention(true)
+            suppressInheritedMembers.convention(
+                @Suppress("DEPRECATION")
+                dokkaExtension.suppressInheritedMembers.orElse(false)
+            )
+            suppressObviousFunctions.convention(
+                @Suppress("DEPRECATION")
+                dokkaExtension.suppressObviousFunctions.orElse(true)
+            )
         }
     }
 
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
@@ -13,8 +13,8 @@ import org.gradle.api.tasks.Nested
 import org.gradle.kotlin.dsl.newInstance
 import org.gradle.workers.WorkerExecutor
 import org.jetbrains.dokka.gradle.dependencies.BaseDependencyManager
-import org.jetbrains.dokka.gradle.formats.DokkaPublication
 import org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec
+import org.jetbrains.dokka.gradle.formats.DokkaPublication
 import org.jetbrains.dokka.gradle.internal.*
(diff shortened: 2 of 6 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
```diff
@@ -154,8 +159,14 @@ constructor(
             offlineMode.convention(false)
             outputDirectory.convention(dokkaExtension.dokkaPublicationDirectory.dir(formatName))
             moduleOutputDirectory.convention(dokkaExtension.dokkaModuleDirectory.dir(formatName))
-            suppressInheritedMembers.convention(false)
-            suppressObviousFunctions.convention(true)
+            suppressInheritedMembers.convention(
+                @Suppress("DEPRECATION")
+                dokkaExtension.suppressInheritedMembers.orElse(false)
```
whyoleg, line 164: why `orElse` is needed here?

whyoleg (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
```diff
@@ -176,4 +176,42 @@ constructor(
...
+     *
+     *   // Replace with:
+     *   dokkaPublications.configureEach {
+     *     suppressInheritedMembers.set(true)
+     *   }
+     * }
+     * ```
+     * @see DokkaPublication.suppressInheritedMembers
+     */
+    @Deprecated("Moved to DokkaPublication#suppressInheritedMembers")
```
abdulowork, line 196: Should these `Deprecation`s include `ReplaceWith("dokkaPublications.configureEach { suppressInheritedMembers.set(true) }")`?

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
```diff
@@ -115,5 +120,74 @@ class DokkaPluginTest : FunSpec({
                 }
             }
         }
+
+        context("DokkaPublications") {
```
abdulowork, line 124: These tests are missing `test` blocks and `./gradlew :dokka-gradle-plugin:test --tests 'org.jetbrains.dokka.gradle.DokkaPluginTest' --info --rerun` doesn't put these in the junit XML output. Is this correct?

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
```diff
@@ -115,5 +120,74 @@ class DokkaPluginTest : FunSpec({
...
+
+            context("dokka dsl provides convention values for suppressInheritedMembers and suppressObviousFunctions") {
```
abdulowork, line 126: nit: I think a more accurate description of this test is that "extension level suppression settings propagate to published suppressions by convention" or something like that

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
```diff
@@ -115,5 +120,74 @@ class DokkaPluginTest : FunSpec({
...
+            }
+
+            context("dokka dsl convention values for suppressInheritedMembers and suppressObviousFunctions can be overridden") {
+                val project = ProjectBuilder.builder().build()
+                    .enableV2Plugin()
+                    .also { it.plugins.apply("org.jetbrains.dokka") }
+
+                val dokka = project.extensions.getByType<DokkaExtension>()
+
+                checkAll(
```
abdulowork, line 162: nit: I feel this test would be more expressive if it just specified opposing suppression settings for the extension and the publication. Otherwise half of these test cases don't test that the value is actually overriden.

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
abdulowork, line 124: nit: maybe it makes to test that publications have a default without extension level suppressions since this is a [part of the behavior](https://github.com/Kotlin/dokka/pull/3830/files#diff-55ceb7738ade299cded228f9d44aa544238c6afaf5ccb2c5e59b2b41717325abR164)

abdulowork (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
adam-enko: If you can think of a way of making that work, then that would be lovely. But I don't think it's possible ([internal discussion](https://jetbrains.slack.com/archives/C05NTFCF9NV/p1727283728288829)).

```kotlin
// I want IJ to quick-replace it with this:
publications.configureEach { enabled.set(true) }

// But it gets replaced with this :(
publications.configureEach { enabled }.set(true) // ERROR Unresolved reference.
```

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
adam-enko: These properties are required, and all tests will fail if they are unset.

https://github.com/Kotlin/dokka/blob/e3dc412e9185564a49ecc639a7291b0596abc0a0/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt#L46-L47

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
adam-enko: Writing out all the different options is quite verbose and error prone, which is why I really like the property-based tests.

Only testing opposing values is a good idea! I've implemented that.

dokka-runners/dokka-gradle-plugin/src/test/kotlin/DokkaPluginTest.kt
adam-enko: Thanks! Fixed. `checkAll() {]` must be inside a `test() {}` block.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
adam-enko: it's not, thanks for spotting it. I've removed it.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
abdulowork: TIL `ReplaceWith` preserves the call expression

Issue #2515, opened 2022-05-31
Remove double negation in configuration option names


mahozad:
Created this issue as discussed in #2513.

It would be great if the methods in Dokka Gradle plugin were converted to properties and also avoided [double negative names](https://refactoring.com/catalog/removeDoubleNegative.html).

So, for instance, `noAndroidSdkLink.set(false)` can be renamed to something like `useAndroidSdkLink = true` or `isAndroidSdkLinkEnabled = true` or `enableAndroidSdkLink = true` or similar.

As an example, here is some of Android Gradle Plugin (AGP) properties which were renamed in its recent versions:

```kotlin
android {
  buildTypes {
    release {
      isTestCoverageEnabled = true
      isMinifyEnabled = false
    }
  }
}

```

**Are you willing to provide a PR?**
Maybe (depends on if the maintainers want it and if I can).

PR #3832, opened 2024-09-26
DGPv2: fix noJdkLink/noStdlibLink/noAndroidSdkLink

adam-enko:
The [last PR](https://github.com/Kotlin/dokka/pull/3829) had mistakes 🤦

- `enableKotlinStdLibDocumentationLink` is _not_ a direct replacement for `noStdlibLink`, because the value must be inverted.
- The kdoc/`@Deprecate` for `noJdkLink` and `noAndroidSdkLink` were the wrong way around.

I decided to make the deprecation level ERROR because these DGPv2 is an opt-in and migration is expected. Additionally, these properties are completely non-functional without migration. 

#2515
[KT-71758](https://youtrack.jetbrains.com/issue/KT-71758)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
```diff
@@ -381,19 +381,64 @@ constructor(
...
+     * // Before:
+     * noStdlibLink.set(true)
+     *
+     * // Update the name, and set the value to `false` instead of `true`:
+     * enableKotlinStdLibDocumentationLink.set(false)
+     * ```
+     *
+     * @see enableKotlinStdLibDocumentationLink
+     */
+    @Deprecated("Replaced with to enableKotlinStdLibDocumentationLink. The value must be inverted.", level = ERROR)
```
antohaby, line 399: why deprecation level error? to force migration to `enableKotlinStdLibDocumentationLink` API?
Or it is technically difficult to map inverted value of `noStdlibLink`  to `enableKotlinStdLibDocumentationLink` and back?

For migration purposes I'd try to stick as much as possible to warnings first.
If I understand correctly, Dokka v1 has that declaration `noStdlibLink: Property<Boolean>`. 

I was thinking about creating something like this:

```
val deprecatedInverseProperty = objects
    .property(Boolean::class.java)

val newUserFacingProperty = objects
    .property(Boolean::class.java)
    .convention(true)


fun <T: Any, D: Any> Property<T>.getValueProviderWithDeprecatedProperty(
    deprecatedProperty: Property<D>,
    conventionValue: T,
    mapFromDeprecatedToNew: (D) -> T,
    reportContradiction: (Property<T>, Property<D>) -> Nothing
): Provider<T> {
    if (!deprecatedProperty.isPresent) return this
    
    val mappedDeprecatedValue = deprecatedProperty.map { mapFromDeprecatedToNew(it) }
    if (!isPresent) return mappedDeprecatedValue
    
    val newValue = this.get()
    val deprecatedValue = mappedDeprecatedValue.get()
    
    return when {
        newValue == conventionValue && deprecatedValue == conventionValue -> this
        newValue == conventionValue && deprecatedValue != conventionValue -> mappedDeprecatedValue
        newValue != conventionValue && deprecatedValue == conventionValue -> this
        newValue != conventionValue && deprecatedValue == newValue -> this
        newValue != conventionValue && deprecatedValue != newValue -> reportContradiction(this, deprecatedProperty)
        else -> throw IllegalStateException()
    }
}
```

But I guess this is too heavy.

antohaby (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
whyoleg: Those properties are not very popular and are really needed only in small amount of use-cases.
So I think it's fine to deprecate them with ERROR right away

whyoleg (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
adam-enko: > why deprecation level error?

I decided to make the deprecation level `ERROR` because these DGPv2 is an opt-in and when users enable v2 some migration is expected.

> to force migration to `enableKotlinStdLibDocumentationLink` API?

Yes, when v2 is enabled these properties are completely non-functional without migration.

- Or it is technically difficult to map inverted value of `noStdlibLink` to `enableKotlinStdLibDocumentationLink` and back?

Yes.

I like the idea of a workaround, but I don't want to spend the rest of the week writing tests for it :)

I think a property transformer like that would be the responsibility of Gradle to add (maybe https://github.com/gradle/gradle/issues/27315?).

PR #3833, opened 2024-09-26
Forward Dokka Generator messages to Gradle logger

adam-enko:
Update Dokka Gradle tasks to forward DokkaGenerator logs to the appropriate levels.

- debug → debug
- info → info
- progress → info
- warn → warn
- error → error

[KT-71719](https://youtrack.jetbrains.com/issue/KT-71719/DGP-v2-Show-logging-from-Dokka-engine-in-task-output)


dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorLoggingTest.kt
```diff
@@ -0,0 +1,221 @@
...
+
+            project.runner
+                .addArguments(
+                    ":dokkaGenerateModuleHtml",
+                    "--rerun",
+                )
+                .build {
+                    output.invariantNewlines() shouldContain """
+                        > Task :dokkaGenerateModuleHtml
+                        e: [:dokkaGenerateModuleHtml] test error message
```
whyoleg, line 28: TBH, I don't think that duplicating task name will be useful, but I'm not against having it :)

whyoleg (approved):

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorLoggingTest.kt
adam-enko: I put the task name in because it can be confusing if there are multiple generations at the same time, or if running with `--quiet`. It's already a bit confusing when aggregating, because sometimes the same error is shown for both module and publication generation. But we can revisit the message if necessary.

Tapchicoma (approved):

PR #3835, opened 2024-09-30
Use JVM 21 to run Gradle build

adam-enko:


adam-enko:
I'd like to pick this up when I have time. It should be easier to implement, now that Gradle supports auto-provisioning for the daemon https://docs.gradle.org/9.0.0/userguide/gradle_daemon.html#sec:detect_provision

PR #3864, opened 2024-10-16
Fixes Dokka Accepting Directories for classpath/includes/sourcesRoots/suppressedFiles

### LINKED ISSUE #3863: [Bug] Dokka Does Not Accept Using Directories for the Includes Property
### Describe the Bug
When setting the "includes" for dokka, it will not accept directories.

### Expected Behaviour
Directories are accepted when doing

```kotlin
dokka {
        includes.from(file("dokka")) // here, dokka is a directory containing multiple markdown files
    }
}
```

### Actual Behaviour
An error is produced when the above code is used.

### Environment
- Operating system: Linux
- Build tool: Gradle 8.10.1
- Dokka version: 2.0.0-Beta

### Additional Information
The issue here is with the use of `FileCollection.getFiles()` on this line: https://github.com/Kotlin/dokka/blob/7027da812e8b3120a56830d9790a27f57fa7d27c/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaSourceSetBuilder.kt#L77

As-per the documentation:
> Returns the contents of this collection as a Set. The contents of a file collection may change over time.
> Note that this method returns File objects that represent locations on the file system. These File objects **do not necessarily refer to regular files**. Depending on the implementation of this file collection and how it has been configured, the returned set **may contain directories**, or missing files, or any other kind of file system element.

(emphasis my own)

Also, this same issue will apply to
- `classpath`
- `samples`
- `suppressedFiles`

The solution to this would be to instead use a FileTree:
```kotlin
// files
classpath = spec.classpath.asFileTree.files.toList(),
includes = spec.includes.asFileTree.files,
samples = spec.samples.asFileTree.files,
sourceRoots = spec.sourceRoots.asFileTree.files,
suppressedFiles = spec.suppressedFiles.asFileTree.files,
```

however, I do believe it would be beneficial to expose `suppressedFiles` as a `ConfigurableFileTree` rather than a `ConfigurableFileCollection`. This is because it allows filtering:
```kotlin
dokka {
        suppressedFiles.apply {
            from("src")
            include("**/internal/**")
            exclude("**/api/**")
        }
    }
}
```
however, this would be a breaking api change. Unsure if breaking api changes would be possible for the beta.
I am going to submit a PR that changes it to use a file tree, however.

solonovamax:
This PR fixes #3863

See issue for more info.

The test case here is kinda jank and ugly, but whatever

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaSourceSetBuilder.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaSourceSetBuilder.kt
@@ -73,11 +73,11 @@ internal object DokkaSourceSetBuilder {
             sourceSetID = build(spec.sourceSetId.get()),
 
             // files
-            classpath = spec.classpath.files.toList(),
-            includes = spec.includes.files,
-            samples = spec.samples.files,
-            sourceRoots = spec.sourceRoots.files,
-            suppressedFiles = spec.suppressedFiles.files,
+            classpath = spec.classpath.asFileTree.files.toList(),
+            includes = spec.includes.asFileTree.files,
+            samples = spec.samples.asFileTree.files,
+            sourceRoots = spec.sourceRoots.asFileTree.files,
+            suppressedFiles = spec.suppressedFiles.asFileTree.files,
         )
     }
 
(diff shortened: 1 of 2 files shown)


adam-enko:
Converted to a draft, while we discuss further in #3863

PR #3866, opened 2024-10-16
Unhide DGPv2 migration guide

adam-enko:
Now that DGPv2 has a beta release, we can display the migration guide.


AlejandraPedroza (approved):
I checked in staging and everything looks ok! 👍

PR #3870, opened 2024-10-17
Fix Configuring Custom Dokka Plugin

### LINKED ISSUE #3869: [Bug] Cannot Configure Custom Dokka Plugin
### Describe the Bug
It is currently impossible to configure custom dokka plugins.

### Expected Behaviour
It should be possible to, and documented how to, configure custom dokka plugins.

### Actual Behaviour
When attempting to configure custom dokka plugins with the (entirely undocumented) `DokkaPluginParametersBuilder` class and `pluginParameters` function, eg.
```kotlin
dokka {
    pluginsConfiguration {
        pluginParameters("ca.solostudios.dokkascript.plugin.DokkaScriptsPlugin") {
            files("scripts") {
                from(fileTree(dokkaScripts))
            }
        }
    }
}
```
then the following error is produced:
```
Cannot create a DokkaPluginParametersBuilder because this type is not known to this container. Known types are: DokkaHtmlPluginParameters, DokkaVersioningPluginParameters
```

If `DokkaPluginParametersBuilder` then has a factory registered using
```kotlin
dokka {
    pluginsConfiguration {
        registerFactory(DokkaPluginParametersBuilder::class.java) { name ->
            objects.newInstance<DokkaPluginParametersBuilder>(name, name)
        }
    }
}
```

then, gradle *will* properly load, however gradle *will not* allow the `dokkaGenerate` task to run, and instead produces the following error:
```
Some problems were found with the configuration of task ':dokkaGenerateModuleHtml' (type 'DokkaGenerateModuleTask').
  - Type 'org.jetbrains.dokka.gradle.engine.plugins.DokkaPluginParametersBuilder' method 'jsonEncode()' should not be annotated with: @Internal.
```

### Environment
- Operating system: Linux
- Build tool: Gradle 8.10.1
- Dokka version: 2.0.0-Beta

solonovamax:
Fixes #3869

- Register binding for `DokkaPluginParametersBuilder`
- Remove `@Internal` annotation on `jsonEncode()` for `DokkaPluginParametersBuilder` (see #3869)
- Annotate the getter for `DokkaPluginParametersBuilder.objects` instead of the field (the annotation only works for getters)
- Add missing functions for adding files, directories, and maps, which were implied but not present for `DokkaPluginParametersBuilder`
- remove overriden `DokkaPluginParametersBuilder.pluginFqn` property as it just caused issues
- Add documentation for configuring custom plugins in dokka-migration.md

### DIFF
--- a/docs/topics/dokka-migration.md
+++ b/docs/topics/dokka-migration.md
@@ -104,6 +104,11 @@ options according to your project setup:
           customAssets.set(listOf("logo.png"))
           footerMessage.set("(c) Your Company")
       }
+      pluginConfiguration<YourCustomPlugin, YourCustomPluginConfiguration> {
+          stringProperty.set("bar")
+          booleanProperty.set(true)
+          fileListProperty.set(files("foo.txt", "bar.txt"))
+      }
   }
   ```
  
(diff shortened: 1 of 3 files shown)


adam-enko:
Thank you for the approach, it helped me investigate the issue further.

However, in addition to the problems you found, I found more issues, so we're going to remove it from DGP (see https://github.com/Kotlin/dokka/issues/3869#issuecomment-2419073268).

Thank you again for your efforts!

PR #3871, opened 2024-10-17
Add example to demonstrate how to configure a custom Dokka plugin

adam-enko:
Add an example to demonstrate how to configure a custom Dokka plugin using DGPv2.

DGPv2 has much more strict requirements for configuring a Dokka plugin, to ensure that inputs/outputs for a Gradle task are properly configured. However, this does make the configuration harder to write. This is something we could look at improving in the future, but for now, here is an example to demo how it can be done (albeit in a verbose and unintuitive/Gradle way).

#3869

To verify, check the expected docs. `fun main()` is not visible:

image


examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Define custom parameters for the HideInternalApi Dokka Plugin.
+ *
+ * Using a custom class for defining parameters means Gradle can accurately check for task up-to-date checks.
+ *
+ * If you find you need to re-use the plugin parameters configuration class in multiple buildscripts,
+ * move the class into a shared location, like `buildSrc` (or another included-build for build conventions).
+ * See [Sharing Build Logic between Subprojects](https://docs.gradle.org/8.10/userguide/sharing_build_logic_between_subprojects.html).
+ */
+@OptIn(InternalDokkaGradlePluginApi::class)
```
whyoleg, line 38: probably we need to drop this annotation from this constructor, it’s not a good idea to advertise usage of `internal` API's in example

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -215,5 +219,25 @@ testing {
...
+            .listDirectoryEntries()
+            .map { it.name }
+            .filter {
+                it.endsWith("-example")
+                // note: avoid checks like `isDirectory()` to prevent Gradle registering the files as CC inputs.
+            }
+            .forEach { exampleProjectName ->
+                val prettyName = exampleProjectName.split("-").joinToString("") { it.uppercaseFirstChar() }
+
+                targets.register("test$prettyName") {
```
whyoleg, line 234: minor: it would be better to have names `testExampleXXX` instead of `testXXXExample` so that they are easier to find

whyoleg (approved):
LGTM

examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
adam-enko: Yes, at the moment it's a problem that custom Dokka plugins can't provide their own implementation without using the internal `DokkaPluginParametersBaseSpec`. 

However, just creating a subtype of `DokkaPluginParametersBaseSpec` isn't nice. Users have to manually convert the values to JSON, which is annoying and easy to get wrong (like not properly escaping the values).

I think it'd be worth re-implementing `DokkaPluginParametersBuilder` (removed in https://github.com/Kotlin/dokka/pull/3872), but make it better and tested. I want something similar to `CommandLineArgumentProvider`, where users can just create a subclass and define the inputs in the constructor ([Dokka build example](https://github.com/Kotlin/dokka/blob/15986f67452872ac13956d37f7cf090a5c6cfc58/build-logic/src/main/kotlin/dokkabuild/utils/SystemPropertyAdder.kt#L195-L205))

whyoleg: > I think it'd be worth re-implementing

Not yet for sure, let's see how much users will need this based on feedback. For now, I think we have enough tasks with higher priority.

dokka-integration-tests/gradle/build.gradle.kts
whyoleg: Ah, I missed that it's in a separate group - interesting :)
Thanks!

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -86,6 +97,9 @@ class ExampleProjectsTest {
     ) {
         val exampleProjectName = ExampleProject.of(project.projectDir)
 
+        val isEnabled: Boolean =
+            project.projectDir.name == System.getProperty("exampleProjectFilter")
```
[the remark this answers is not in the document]
abdulowork: Would it be better to use something like:
```
val isEnabled: Boolean = System.getProperty("exampleProjectFilter").let {
    if (it == null) return@let true
    return@let it == project.projectDir.name
}
```
?

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -215,5 +219,28 @@ testing {
                 systemProperty("junit.jupiter.execution.parallel.mode.default", "CONCURRENT")
             }
         }
+
+        // Register specific Gradle test runs for each example so the examples can be tested individually.
```
abdulowork, line 223: nit: we discussed this in the previous PR, but I still feel that it would be more conventional to have a generated `testGen` source directory (under git and with an idempotency check) than a per test project Gradle task

abdulowork: One of the benefits is that you could then run an arbitrary subset of such tests using the standard `Test` task's `--tests` filter

dokka-integration-tests/gradle/build.gradle.kts
```diff
@@ -215,5 +219,28 @@ testing {
...
+        exampleGradleProjectsDir.toPath()
+            .listDirectoryEntries()
+            .map { it.name }
+            .filter {
+                it.endsWith("-example")
+                // note: avoid checks like `isDirectory()` to prevent Gradle registering the files as CC inputs.
```
abdulowork, line 229: I think it makes sense to explicitly make these directories the CC fingerprint input. These tasks now have an issue where the task list with CC and without is inconsistent:
```bash
# Cache CC for basic project
./gradlew :dokka-integration-tests:gradle:testExampleBasicGradle
# Delete basic example
mv ../../examples/gradle-v2/basic-gradle-example ../../examples/gradle-v2/basic-gradle-example.bak
# The task now runs with CC
./gradlew :dokka-integration-tests:gradle:testExampleBasicGradle
# Now it fails with "Cannot locate tasks" due to reconfiguration
./gradlew :dokka-integration-tests:gradle:testExampleBasicGradle --no-configuration-cache
```

examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    abstract val annotatedWith: ListProperty<String>
+
+    override fun jsonEncode(): String {
+        // Convert annotatedWith to a JSON list.
+        val annotatedWithJson = annotatedWith.orNull.orEmpty()
+            .joinToString(separator = ", ", prefix = "[", postfix = "]") { "\"$it\"" }
+
+        return """
+            {
+              "annotatedWith": $annotatedWithJson
```
abdulowork, line 58: I am a bit surprised Dokka doesn't support the standard Java serialization for this configuration

abdulowork: nit: wdyt about suggesting:
```
buildscript {
    dependencies {
        classpath("org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3")
    }
}
```
and serializing using:
```
return Json.Default.encodeToString(
    JsonObject.serializer(),
    JsonObject(
        mapOf("annotatedWith" to JsonArray(annotatedWith.get().map { JsonPrimitive(it) }))
    )
)
```
?

examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    name: String
+) : DokkaPluginParametersBaseSpec(
+    name,
+    // The plugin ID of the custom HideInternalApiPlugin
+    "demo.dokka.plugin.HideInternalApiPlugin",
+) {
+
+    @get:Input
+    @get:Optional
+    abstract val annotatedWith: ListProperty<String>
```
abdulowork, line 49: nit: IIUC this sample primarily shows how you parametrize a `dokkaPlugin`, but I am not sure whether the `ListProperty` makes sense for the use case of suppressing some documentation; would you want to suppress documentation using multiple different annotations? I feel it might be clearer to use a `Property<String>` with a single FQN and then the [JSON serialization](https://github.com/Kotlin/dokka/pull/3871/files#diff-ef71188fefb2f9aa503ea30a7275fb046dcdbbe5388824639e507cc6fbd2a5eeR53) can also be simplified

abdulowork (changes requested):
Could you please go through my comments. I am primarily concerned whether we might be accidentally [skipping](https://github.com/Kotlin/dokka/pull/3871#discussion_r1846636420) the example IT in CI runs

examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
adam-enko: If I were doing this in my own project, I would use Kotlinx Serialization. But because this is a basic example, I want to show the most simple configuration possible, even if manually creating Json is ugly and invites mistakes.

In the future we want to improve the experience of configuring custom Dokka plugins (#3869, #3925), and provide a better wrapper.

examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
adam-enko: Good idea, I'll implement this.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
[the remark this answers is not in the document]
adam-enko: Hmm I see your point, and I think it's correct, but I see the tests are run in CI 🤔 

https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Dokka_IntegrationTests/4885102?buildTab=tests&name=ExampleProjects

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Yes, those would be good benefits. I'm just struggling to think about how to generate tests dynamically though, because I usually find code generation inconvenient and prone to mistakes.

dokka-integration-tests/gradle/build.gradle.kts
adam-enko: Thanks, I've used a ValueSource, and I think it's working as intended now.

examples/gradle-v2/custom-dokka-plugin-example/demo-library/build.gradle.kts
abdulowork: Ok. I wonder if it would be simpler to just provide a `File` as a transport between Gradle -> Dokka Plugin and then let the user choose an arbitrary serialization format

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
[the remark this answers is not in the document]
abdulowork: Judging by the log the per-project tasks get pulled in by the `integrationTest` task though I wonder why this happens and how easily this is broken

dokka-integration-tests/gradle/build.gradle.kts
abdulowork: If you decide to look further into the direction of generating test cases, I think the setup is usually something like:
* `src/MyTestsInfrastructure.kt`
```kotlin
abstract class TestsInfrastructure  {
  // hand craft some eloquent testing infrastructure with an API that easy to generate in 1 LOC
  fun runTest(testDataPath: String) { ... }
}
```
* `src/tests-gen/Tests.kt`
```kotlin
// This class is generated
class Tests : TestsInfrastructure {
  @Test
  fun test_basicExample() = runTest("path/to/basicExample")
}
```

abdulowork (approved):

whyoleg (approved):

abdulowork:
@adam-enko  lgtm

PR #3873, opened 2024-10-17
Fix appending URI Path

adam-enko:
`URI#resolve(newPath)` doesn't escape spaces

Not escaping the space breaks logging the HTML URL logging task when the IntelliJ project name has a space.


whyoleg (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/uriUtils.kt
```diff
@@ -8,5 +8,14 @@ import java.net.URI
...
-    return resolve(newPath).normalize()
+
+    return URI(
+        /*    scheme = */ scheme,
+        /*  userInfo = */ userInfo,
+        /*      host = */ host,
+        /*      port = */ port,
+        /*      path = */ newPath,
+        /*     query = */ query,
+        /*  fragment = */ fragment,
```
Tapchicoma, line 19: Nit: passed names imho are self-explanatory :)

Tapchicoma (approved):

PR #3877, opened 2024-10-21
Add deprecated property to help migrate from JSON Dokka plugin configuration

adam-enko:
Add a deprecated property to help migrate from JSON-based plugin configuration.

([internal discussion](https://jetbrains.slack.com/archives/C02H0GSH474/p1729161083325359?thread_ts=1729150701.318319&cid=C02H0GSH474)).


dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaExtension.kt
```diff
@@ -213,5 +214,86 @@ constructor(
...
+     *     pluginsConfiguration {
+     *         versioning {
+     *             version.set("1.2")
+     *             olderVersionsDir.set(projectDir.resolve("dokka-docs"))
+     *         }
+     *     }
+     * }
+     * ```
+     *
+     * #### Dokka Plugin authors
```
whyoleg, line 258:
```suggestion
     * #### External Dokka Plugin configuration
```

Probably it would be nice to put all of this into migration guide, and then just add a link to it. Or copy-paste. I don't have preferences, but I think it should be in migration guide.

Also, is needed not only for plugin authors, but initially mostly for plugin consumers.

adam-enko: Let's merge this PR as-is and we can update the docs separately?

https://youtrack.jetbrains.com/issue/KT-72447/Document-Dokka-plugin-configuration

I think some of this docs can go into the Dokka dev docs too.

whyoleg: ok
dev docs could be updated later (after release)

whyoleg (approved):

Tapchicoma (approved):

PR #3878, opened 2024-10-21
Fix logging link to HTML output composite builds

adam-enko:
Log using the root-most Gradle project, taking composite builds into account.

[KT-71254](https://youtrack.jetbrains.com/issue/KT-71254)


adam-enko:
I'd like to fix this at some point, but it's not high priority, and it's a very niche problem. I only encounter it because I open Dokka's test projects.

Issue #3855, opened 2024-10-11
Prepare for migration to KGP 2.1.0


whyoleg:
Kotlin 2.1.0 will [drop support for language version 1.4 and 1.5](https://youtrack.jetbrains.com/issue/KT-60521/Drop-language-versions-1.4-and-1.5). But we need need to use older api/language version so that DGP will be compatible with Gradle 7.6.
To be able to migrate DGP (and dependent modules) to it, we will need to use fixed compiler version for compiling which still supports old language versions (e.g 2.0.20). 
Similar changes were already implemented in [kotlinx-benchmark](https://github.com/Kotlin/kotlinx-benchmark/pull/261/files).

PR #3896, opened 2024-11-01
Use fixated Kotlin compiler 2.0.20 for compiling Dokka

whyoleg:
~~NOTE: depends on #3899~~

Fixes https://github.com/Kotlin/dokka/issues/3855

For now the version of the Kotlin compiler and Kotlin Gradle plugin are the same, but it will change when we migrate to KGP 2.1.0

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
@@ -59,13 +59,10 @@ abstract class DokkaBuildProperties @Inject constructor(
         dokkaProperty("javaToolchain.testLauncher", JavaLanguageVersion::of)
             .orElse(mainJavaVersion)
 
-    /**
-     * The Kotlin language level that Dokka artifacts are compiled to support.
-     *
-     * Updating the language level is a breaking change.
-     */
-    val kotlinLanguageLevel: Provider<KotlinVersion> =
-        dokkaProperty("kotlinLanguageLevel", KotlinVersion::fromVersion)
+    /** Indicates whether Kotlin compatibility with older Gradle versions should be enforced */
+    val enforceGradleKotlinCompatibility: Provider<Boolean> =
+        dokkaProperty("enforceGradleKotlinCompatibility", String::toBoolean)
+            .orElse(true)
 
     /** Allows skipping running of integration tests */
     val integrationTestSkip: Provider<Boolean> =
--- a/build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
@@ -62,7 +62,7 @@ abstract class DokkaBuildProperties @Inject constructor(
     /** Indicates whether Kotlin compatibility with older Gradle versions should be enforced */
     val enforceGradleKotlinCompatibility: Provider<Boolean> =
         dokkaProperty("enforceGradleKotlinCompatibility", String::toBoolean)
-            .orElse(true)
+            .orElse(false)
 
     /** Allows skipping running of integration tests */
     val integrationTestSkip: Provider<Boolean> =
--- a/build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
+++ b/build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package dokkabuild.utils
+
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.dokkaBuild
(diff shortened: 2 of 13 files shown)


gradle/libs.versions.toml
```diff
@@ -1,5 +1,8 @@
 [versions]
 
+# We use a different compiler version to compile modules and Gradle Plugin
+# to be able to use Kotlin Language/API version 1.4 to be compatible with Gradle 7.
+# The logic leaves in build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```
adam-enko, line 5:
```suggestion
# The logic lives in build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
+
+// We must use Kotlin language/api version 1.4 (DEPRECATED) to support Gradle 7,
+// and to do that we need to use old Kotlin version, which has support for such an old Kotlin version (e.g., 2.0.20)
+fun Project.configureGradleKotlinCompatibility() {
+    if (!dokkaBuild.enforceGradleKotlinCompatibility.get()) return
+
+    extensions.configure<KotlinJvmProjectExtension>("kotlin") {
+        @OptIn(ExperimentalKotlinGradlePluginApi::class)
+        compilerVersion.set("2.0.20")
```
adam-enko, line 20: IntelliJ shows a warning here, but the `ExperimentalBuildToolsApi` opt-in can't be imported 🤔 

image

gradle/libs.versions.toml
```diff
@@ -142,8 +145,6 @@ kotlinx-cli = { module = "org.jetbrains.kotlinx:kotlinx-cli-jvm", version.ref =
 #### Test dependencies  ####
 eclipse-jgit = { module = "org.eclipse.jgit:org.eclipse.jgit", version.ref = "eclipse-jgit" }
 
-kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test-junit5", version.ref = "gradlePlugin-kotlin" }
```
adam-enko, line 145: Why was `kotlin-test` removed from the version catalog?

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
whyoleg: Yeah, that's true, I saw something like this, but if I remember it will go away when we will update KGP to more recent version. And we will be able to do it after this PR is merged

gradle/libs.versions.toml
whyoleg: JIC: `kotlin(xxx)` is defined as:
```kotlin
fun DependencyHandler.kotlin(module: String, version: String? = null): Any =
    "org.jetbrains.kotlin:kotlin-$module${version?.let { ":$version" } ?: ""}"
```
So I believe that just dropping version.ref in the version catalog should work nicely.
Also, we will be able to leave a comment which explains the idea why we don't have versions there.

So yes, I like your idea, but let's do it in another PR. I will create PR then shortly after this is merged.

build-logic/src/main/kotlin/dokkabuild/DokkaBuildProperties.kt
```diff
@@ -59,13 +59,10 @@ abstract class DokkaBuildProperties @Inject constructor(
...
-     * The Kotlin language level that Dokka artifacts are compiled to support.
-     *
-     * Updating the language level is a breaking change.
-     */
-    val kotlinLanguageLevel: Provider<KotlinVersion> =
-        dokkaProperty("kotlinLanguageLevel", KotlinVersion::fromVersion)
+    /** Indicates whether Kotlin compatibility with older Gradle versions should be enforced */
+    val enforceGradleKotlinCompatibility: Provider<Boolean> =
+        dokkaProperty("enforceGradleKotlinCompatibility", String::toBoolean)
+            .orElse(true)
```
whyoleg, line 65: While looking at the PR diff, I found that it's a bit inconsistent that it has a `true` default value.
I mean, to enforce language version 1.4, we anyway need to have `kotlin.compiler.runViaBuildToolsApi=true` in gradle.properties, so it makes more sense to also have `org.jetbrains.dokka.enforceGradleKotlinCompatibility=true` near it


So, @adam-enko, quick question (yes or no), WDYT about flipping the default here? :)

adam-enko: What about changing the default to be the value of `runViaBuildToolsApi`?

adam-enko: Either way makes sense to me - feel free to change the default.

adam-enko (approved):

gradle/libs.versions.toml
whyoleg: Done in https://github.com/Kotlin/dokka/pull/4143

PR #3898, opened 2024-11-01
Test with KGP 2.1.0-RC

whyoleg:
compilation is failing with:
```
[KOTLIN] e: org.jetbrains.kotlin.backend.common.CompilationException: Back-end: Please report this problem https://kotl.in/issue
dokka/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt:26:1
Problem with `@Serializable
@DokkaInternalApi
data class DokkaModuleDescriptionKxs
...
Details: kotlinx.serialization compiler plugin internal error: unable to transform declaration, see cause
...
Caused by: java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.ir.expressions.impl.IrWhenImpl org.jetbrains.kotlin.ir.builders.ExpressionHelpersKt.irIfThen$default(org.jetbrains.kotlin.ir.builders.IrBuilderWithScope, org.jetbrains.kotlin.ir.types.IrType, org.jetbrains.kotlin.ir.expressions.IrExpression, org.jetbrains.kotlin.ir.expressions.IrExpression, org.jetbrains.kotlin.ir.expressions.IrStatementOrigin, int, java.lang.Object)'
...
```

Probably there is some issue with kotlinx.serialization compiler plugin usage when using `kotlin.compiler.runViaBuildToolsApi`.
KGP version = 2.1.0-RC-330
Kotlin compiler version used to compile code = 2.0.20
I've tried to set all different versions of kotlinx.serialization plugins, but none of them work.

Not sure where the issue comes from, but probably the easiest way will be to drop `kotlinx.serialization` usage from DGPv2 so that there will be less moving parts :)
It's possible to replace it:
1. or with just runtime `kotlinx.serialization` usage via `JsonObject`
2. or `kotlinx.serialization` can be fully dropped as it's used only for [single class](https://github.com/Kotlin/dokka/blob/e3dc412e9185564a49ecc639a7291b0596abc0a0/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt#L28) which could be easily replaced with just a txt file with 4 lines :)

Note: last commit adds a workaround: https://github.com/Kotlin/dokka/pull/3898/commits/d43e162bdbe4bfd05338b679ada91b28385604c1

with applied workaround, there are several failed DGPv2 functional tests

### DIFF
--- a/build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
+++ b/build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
@@ -3,6 +3,7 @@
  */
 
 import dokkabuild.utils.excludeGradleEmbeddedDependencies
+import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
 plugins {
@@ -12,10 +13,13 @@ plugins {
 }
 
 kotlin {
+    @OptIn(ExperimentalKotlinGradlePluginApi::class)
+    compilerVersion = libs.versions.gradlePlugin.kotlin.compiler
+    @Suppress("DEPRECATION", "DEPRECATION_ERROR")
     compilerOptions {
         // Must use Kotlin 1.4 to support Gradle 7
-        languageVersion = @Suppress("DEPRECATION") KotlinVersion.KOTLIN_1_4
-        apiVersion = @Suppress("DEPRECATION") KotlinVersion.KOTLIN_1_4
+        languageVersion = KotlinVersion.KOTLIN_1_4
+        apiVersion = KotlinVersion.KOTLIN_1_4
     }
 }
 
--- a/build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
+++ b/build-logic/src/main/kotlin/dokkabuild.gradle-plugin.gradle.kts
@@ -3,6 +3,7 @@
  */
 
 import dokkabuild.utils.excludeGradleEmbeddedDependencies
+import org.jetbrains.kotlin.buildtools.api.ExperimentalBuildToolsApi
 import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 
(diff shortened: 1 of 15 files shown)


adam-enko:
DGPv2 uses kotlinx.serialization for encoding the parameters for the HTML and Versioning Dokka plugins, but yes, I agree it could be removed. It'd help keep the buildscripts cleaner.

adam-enko:
> I've extracted removing compiler plugin into separate PR (#3899). Also, @adam-enko, it would be nice if you could take a look on the reason of test failures in this PR. E.g here is the [build scan](https://ge.jetbrains.com/s/i4sh5osvgysfi) from this [GA build](https://github.com/Kotlin/dokka/actions/runs/11628778963/job/32384531878?pr=3898). Tests failed for `dokka-gradle-plugin` only. Not urgent, but would be nice to know the reason

1. Some tests are failing because there's extraneous logging in build scripts

   https://ge.jetbrains.com/s/i4sh5osvgysfi/tests/task/:dokka-gradle-plugin:testFunctional/details/org.jetbrains.dokka.gradle.DokkaGeneratorLoggingTest/DokkaGenerator%20logging:%20at%20lifecycle%20log%20level%20expect%20only%20error%20and%20warn%20logs?top-execution=1

  The test uses a custom Dokka plugin to cause Dokka to log info/warn/error logs, but it's compiled with an incompatible Kotlin version (the embedded Kotlin version).

   https://github.com/Kotlin/dokka/blob/master/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorLoggingTest.kt#L152

2. Some tests are failing because AGP

   ```
   Caused by: org.gradle.api.InvalidUserCodeException: Kotlin Gradle Plugin <-> Android Gradle Plugin compatibility issue:	
   The applied Android Gradle Plugin version (7.1.3) is lower than the minimum supported 7.3.1.
   ```

   https://ge.jetbrains.com/s/i4sh5osvgysfi/tests/task/:dokka-gradle-plugin:test/details/org.jetbrains.dokka.gradle.AndroidAutoConfigurationTest/at%20least%20one%20dokka%20task%20created()?top-execution=1

PR #3899, opened 2024-11-04
Drop kotlinx.serialization compiler plugin as it's unnecessary

whyoleg:
Extracted from https://github.com/Kotlin/dokka/pull/3898, required to be able to update to KGP 2.1.0

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
@@ -3,7 +3,10 @@
  */
 package org.jetbrains.dokka.gradle.engine.parameters
 
-import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.put
 import org.gradle.kotlin.dsl.java
 import org.jetbrains.dokka.DokkaConfiguration
 import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
@@ -23,7 +26,6 @@ import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
  * @see org.jetbrains.dokka.gradle.engine.parameters.DokkaModuleDescriptionKxs
  * @see org.jetbrains.dokka.DokkaModuleDescriptionImpl
  */
-@Serializable
 @InternalDokkaGradlePluginApi
 data class DokkaModuleDescriptionKxs(
     /** @see DokkaConfiguration.DokkaModuleDescription.name */
@@ -34,4 +36,20 @@ data class DokkaModuleDescriptionKxs(
     val moduleOutputDirName: String = "module",
     /** name of the sibling directory that contains the module includes */
     val moduleIncludesDirName: String = "includes",
-)
+) {
+    internal companion object {
+        fun toJsonObject(module: DokkaModuleDescriptionKxs): JsonObject = buildJsonObject {
+            put("name", module.name)
+            put("modulePath", module.modulePath)
+            put("moduleOutputDirName", module.moduleOutputDirName)
+            put("moduleIncludesDirName", module.moduleIncludesDirName)
+        }
+
+        fun fromJsonObject(obj: JsonObject): DokkaModuleDescriptionKxs = DokkaModuleDescriptionKxs(
+            name = obj["name"]!!.jsonPrimitive.content,
+            modulePath = obj["modulePath"]!!.jsonPrimitive.content,
+            moduleOutputDirName = obj["moduleOutputDirName"]!!.jsonPrimitive.content,
+            moduleIncludesDirName = obj["moduleIncludesDirName"]!!.jsonPrimitive.content,
+        )
+    }
+}
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
@@ -33,11 +33,14 @@ data class DokkaModuleDescriptionKxs(
     /** @see org.jetbrains.dokka.gradle.DokkaExtension.modulePath */
     val modulePath: String,
     /** name of the sibling directory that contains the module output */
-    val moduleOutputDirName: String = "module",
+    val moduleOutputDirName: String = defaultModuleOutputDirName,
     /** name of the sibling directory that contains the module includes */
-    val moduleIncludesDirName: String = "includes",
+    val moduleIncludesDirName: String = defaultModuleIncludesDirName,
 ) {
     internal companion object {
+        private val defaultModuleOutputDirName = "module"
+        private val defaultModuleIncludesDirName = "includes"
+
         fun toJsonObject(module: DokkaModuleDescriptionKxs): JsonObject = buildJsonObject {
             put("name", module.name)
             put("modulePath", module.modulePath)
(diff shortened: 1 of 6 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
```diff
@@ -34,4 +36,20 @@ data class DokkaModuleDescriptionKxs(
...
+            put("moduleOutputDirName", module.moduleOutputDirName)
+            put("moduleIncludesDirName", module.moduleIncludesDirName)
+        }
+
+        fun fromJsonObject(obj: JsonObject): DokkaModuleDescriptionKxs = DokkaModuleDescriptionKxs(
+            name = obj["name"]!!.jsonPrimitive.content,
+            modulePath = obj["modulePath"]!!.jsonPrimitive.content,
+            moduleOutputDirName = obj["moduleOutputDirName"]!!.jsonPrimitive.content,
+            moduleIncludesDirName = obj["moduleIncludesDirName"]!!.jsonPrimitive.content,
+        )
```
adam-enko, line 53: This needs to be updated, because properties with default constructor arguments are optional.

What do you think about adding some tests? To ensure any code changes don't break things in the future.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
```diff
@@ -23,15 +26,37 @@ import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
...
+            put("moduleOutputDirName", module.moduleOutputDirName)
+            put("moduleIncludesDirName", module.moduleIncludesDirName)
+        }
+
+        fun fromJsonObject(obj: JsonObject): DokkaModuleDescriptionKxs = DokkaModuleDescriptionKxs(
+            name = obj.getString("name"),
+            modulePath = obj.getString("modulePath"),
+            moduleOutputDirName = obj.getString("moduleOutputDirName", defaultModuleOutputDirName),
+            moduleIncludesDirName = obj.getString("moduleIncludesDirName", defaultModuleIncludesDirName),
+        )
```
adam-enko, line 56: Instead of these methods, what do you think about exposing a hand written `KSerializer<DokkaModuleDescriptionKxs>`? It'd be more consistent with the current behaviour, reduce our API surface, and not tie us to a specific format.

whyoleg: I don't really think it's needed here as the implementation of custom serializer will be much more complex than the implementation of those two functions.
I don't see how it will reduce our API surface, as it's `internal` (overall the whole class should be `internal` I believe).
The same with specific format - we don't need it now, and as it's an implementation detail we can implement it later if needed.

adam-enko: Ah I see, I missed the companion object being `internal`, so the api is the same.

Well... I really like writing custom serialisers and it's about the same amount of code. Please can I push a change to use a custom KSerializer, to satisfy my kotlinx-serialization love? :)

adam-enko (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaModuleDescriptionKxs.kt
whyoleg: > Well... I really like writing custom serialisers and it's about the same amount of code. Please can I push a change to use a custom KSerializer, to satisfy my kotlinx-serialization love? :)

You are welcome, feel free to check if I implemented everything correctly :)
But, as I said previosly, it's really much more complex, so let's stick with simple variant as in current PR.
```kotlin

private object DokkaModuleDescriptionKxsSerializer : KSerializer<DokkaModuleDescriptionKxs> {
    override val descriptor: SerialDescriptor =
        buildClassSerialDescriptor("org.jetbrains.dokka.gradle.engine.parameters.DokkaModuleDescriptionKxs") {
            element("name", String.serializer().descriptor)
            element("modulePath", String.serializer().descriptor)
            element("moduleOutputDirName", String.serializer().descriptor, isOptional = true)
            element("moduleIncludesDirName", String.serializer().descriptor, isOptional = true)
        }

    override fun deserialize(
        decoder: Decoder
    ): DokkaModuleDescriptionKxs = decoder.decodeStructure(descriptor) {
        var name: String? = null
        var modulePath: String? = null
        var moduleOutputDirName: String? = null
        var moduleIncludesDirName: String? = null
        while (true) {
            when (val index = decodeElementIndex(descriptor)) {
                0 -> name = decodeStringElement(descriptor, index)
                1 -> modulePath = decodeStringElement(descriptor, index)
                2 -> moduleOutputDirName = decodeStringElement(descriptor, index)
                3 -> moduleIncludesDirName = decodeStringElement(descriptor, index)
                CompositeDecoder.DECODE_DONE -> break
                else -> error("Unexpected index: $index")
            }
        }
        requireNotNull(name) { "Missing 'name' property" }
        requireNotNull(modulePath) { "Missing 'modulePath' property" }
        DokkaModuleDescriptionKxs(
            name = name,
            modulePath = modulePath,
            moduleOutputDirName = moduleOutputDirName ?: defaultModuleOutputDirName,
            moduleIncludesDirName = moduleIncludesDirName ?: defaultModuleIncludesDirName,
        )
    }

    @OptIn(ExperimentalSerializationApi::class) // for shouldEncodeElementDefault
    override fun serialize(
        encoder: Encoder,
        value: DokkaModuleDescriptionKxs
    ): Unit = encoder.encodeStructure(descriptor) {
        encodeStringElement(descriptor, 0, value.name)
        encodeStringElement(descriptor, 1, value.modulePath)
        if (shouldEncodeElementDefault(descriptor, 2) && value.moduleOutputDirName != defaultModuleOutputDirName) {
            encodeStringElement(descriptor, 2, value.moduleOutputDirName)
        }
        if (shouldEncodeElementDefault(descriptor, 3) && value.moduleIncludesDirName != defaultModuleIncludesDirName) {
            encodeStringElement(descriptor, 3, value.moduleIncludesDirName)
        }
    }
}
```

adam-enko: Oh I meant more like this:

```kotlin


    internal object Serializer : KSerializer<DokkaModuleDescriptionKxs> {
        override val descriptor: SerialDescriptor =
            JsonObject.serializer().descriptor

        override fun deserialize(decoder: Decoder): DokkaModuleDescriptionKxs {
            val jsonObject = decoder.decodeSerializableValue(JsonObject.serializer())

            fun getString(key: String, defaultValue: String? = null): String {
                return jsonObject[key]?.jsonPrimitive?.content ?: defaultValue
                ?: error("Missing required property '$key'")
            }

            return DokkaModuleDescriptionKxs(
                name = getString("name"),
                modulePath = getString("modulePath"),
                moduleOutputDirName = getString("moduleOutputDirName", defaultModuleOutputDirName),
                moduleIncludesDirName = getString("moduleIncludesDirName", defaultModuleIncludesDirName),
            )
        }

        override fun serialize(encoder: Encoder, value: DokkaModuleDescriptionKxs) {
            val jsonObject = buildJsonObject {
                put("name", value.name)
                put("modulePath", value.modulePath)
                put("moduleOutputDirName", value.moduleOutputDirName)
                put("moduleIncludesDirName", value.moduleIncludesDirName)
            }
            encoder.encodeSerializableValue(JsonObject.serializer(), jsonObject)
        }
    }

```

adam-enko: Let's merge it as-is and we can come back to it if I'm still feeling perfectionistic later.

PR #3902, opened 2024-11-05
Allow analysisPlatform to be overridden

adam-enko:
WIP - I need to set up a full integration test because AGP is required to trigger the bug (it adds multiple Kotlin compilations with different platforms), and AGP requires running with Java 11.


adam-enko:
Fixed in #3814

PR #3904, opened 2024-11-05
DGPv2 Android integration tests

adam-enko:
Create integration tests for using DGPv2 with Android projects.

Additionally, create a JUnit extension for helping set up and run the tests.

Part of [KT-70855](https://youtrack.jetbrains.com/issue/KT-70855) - the HTML currently contains errors, but this will be resolved in later PRs.

#### Overview

- DokkaGradlePluginTestExtension is a JUnit extension for automatically configuring and re-running tests.
  The behaviour can be configured by providing custom GradleTestProjectInitializer (for setting up the test projects),
  or custom GradlePropertiesProvider (for providing custom Gradle properties).

- The extension will re-run the tests multiple times with different versions.
  `TestedVersions.testedVersions()` provides all possible, valid combinations of versions used to run the tests.

- Add JUnit `@Tag` annotations, to help organise and categorise the tests.

- Add two new test projects in dokka-integration-tests/gradle/projects: it-android and it-android-compose.
  Each project has an `expectedData` directory containing the expected generated Dokka content.

- Split GradlePropertiesBuilder into separate class, so it can be re-used in the integration tests


dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/GradleTestProjectInitializer.kt
```diff
@@ -0,0 +1,110 @@
...
+                    .sorted()
+                    .joinToString(separator = "\n", postfix = "\n")
+            )
+        }
+
+        private fun updateVersions(
+            projectDir: Path,
+            testedVersions: TestedVersions,
+        ) {
+            projectDir.walk().filter { it.isRegularFile() }.forEach { file ->
```
whyoleg, line 74: Curious, if this will not become a performance issue.
AFAIU we are here iterating through ALL files in the project, reading and writing it.
F.e in `ui-showcase` we have versioning with 5 versions, and so 5 Dokka’s HTML outputs :)

Of course for now it should not be an issue, but still, I think that we could at limit those modifications to a list of well known files (build.gradle.kts, settings.gradle.kts, libs.versions.toml)

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
```diff
@@ -0,0 +1,228 @@
...
+ * for each valid combination.
+ *
+ * To access the specific values used for a test invocation, add a [TestedVersions]
+ * parameter in the test function.
+ */
+data class TestedVersions(
+    /** Kotlin Gradle Plugin version. */
+    val kgp: SemVer,
+    /** Gradle version. */
+    val gradle: SemVer,
```
whyoleg, line 24: nitpick: for some reason, I feel like `gradle` version should go first :)

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * The test must be annotated with [TestsAndroidGradlePlugin] for this value to be set.
+     */
+    val agp: SemVer?,
+    /**
+     * The Compose version being tested.
+     *
+     * The test must be annotated with [TestsCompose] for this value to be set.
+     */
+    val composeGradlePlugin: SemVer?,
```
whyoleg, line 38: nitpick: looking on this, it feels like it’s better just call properies `kotlin`/`dokka`/`android`/`compose`

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
```diff
@@ -0,0 +1,228 @@
...
+     * then the combination is invalid.
+     */
+    private fun isValid(): Boolean {
+        if (agp != null) {
+            if (agp.major != gradle.major) {
+                return false
+            }
+
+            when (kgp) {
+                SemVer("1.9.0") -> {
```
whyoleg, line 68: from where those values are comming? Could you add a link to the source?

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
```diff
@@ -0,0 +1,228 @@
...
+            put("Dokka", dgp)
+            put("Kotlin", kgp)
+            put("Gradle", gradle)
+            if (agp != null) put("Android", agp)
+            if (composeGradlePlugin != null) put("Compose", composeGradlePlugin)
+        }
+
+    /**
+     * Join the major and minor versions of each version.
+     * Used by [allTestedVersions] to filter out duplicated patch versions.
```
whyoleg, line 90: it’s not clear for me, why this could be needed?
If we do have different `patch` versions, may be we do need to test them? E.g may be `2.1.21` introduces bug-fix which is required for Dokka

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
```diff
@@ -0,0 +1,228 @@
...
+            }
+        return 0
+    }
+
+    companion object {
+
+        /** All possible Kotlin Gradle Plugin versions that could be tested. */
+        private val allKgpVersions = listOf(
+            "1.9.0", // required for Compose 1.5.1
+            "1.9.23", // required for Compose 1.5.12
```
whyoleg, line 138: curious: why do you think we need to test with different compose versions for patch versions of kotlin pre-2.0?

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
```diff
@@ -0,0 +1,228 @@
...
+            "1.9.24", // required for Compose 1.5.14
+            "1.9.25",
+            "2.0.20",
```
whyoleg, line 141: if there is both 2.0.20 and 2.0.21, then proabably 2.0.0 and 2.0.10 should be here—otherwise, I think it should be fine to have just one version per major Kotlin version (except if we have some specific issues there)

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/testTags.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * If a test is annotated with [TestsAndroidGradlePlugin] then
+ * [DokkaGradlePluginTestExtension] will run the test multiple times,
+ * and provide a [DokkaGradleProjectRunner] with a valid [TestedVersions.agp].
+ * Otherwise, [TestedVersions.agp] will be `null`.
+ */
+@Tag("AndroidGradlePlugin")
+@Target(FUNCTION, CLASS)
+@MustBeDocumented
+annotation class TestsAndroidGradlePlugin
```
whyoleg, line 39: may be just `TestsAndroid` (as with compose)?

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Integration test for the `it-android-compose` project.
+ */
+@TestsDGPv2
+@TestsAndroidGradlePlugin
+@TestsCompose
+@WithGradleProperties(GradlePropertiesProvider.Android::class)
+class AndroidComposeIT {
+
+    @DokkaGradlePluginTest(sourceProjectName = "it-android-compose")
```
whyoleg, line 30: is it possible to have `sourceProjectName` per class and not per method? As this is how most of our IT are configured

whyoleg (approved):
Overall LGTM.

A lot of "magic" under the hood in junit setup (I don't have a lot of experince with writing custom extensions), but looks more or less understandable.
But I do really like how `*IT.kt` files look: nice and clear (and kotest also helps here)!
Thanks for improving our IT even more!

dokka-integration-tests/gradle/projects/it-android/settings.gradle.kts
```diff
@@ -0,0 +1,64 @@
+rootProject.name = "it-android-0"
+
+pluginManagement {
```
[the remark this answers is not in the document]
adam-enko: Yes, the config is duplicated. Since these tests are a fresh start, I wanted to try and avoid using the existing test utilities initially, to try and see that would encourage some alternatives. 

Copy and pasting the config isn't ideal, but it works for now. I would like to revisit it in the future when more tests are added, and it becomes a bit more clear about how the DGPv2 integration tests can be best implemented.

Is it okay to leave the copy-pasted code for now?

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: I will order them alphabetically

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: The problem is there's no single 'compose' version. It's very confusing :(

1. Compose compiler plugin hosted by Google (for Kotlin < 2.0.0)
2. Compose compiler plugin hosted by JB
3. Compose Gradle plugin for applying Compose compiler plugin (2)
4. Jetpack Compose (UI framework for Android) as a part of AGP
5. Compose Multiplatform (JB’s UI framework that is using Jetpack Compose on Android) as a Gradle plugin

All these are usually referred as just Compose.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: The main purpose is to prevent unnecessary testing. We don't need to test both KGP 1.9.23 *and* KGP 1.9.24.

However, we do need to consider patch numbers when generating test versions, because Compose has very strict requirements for specific Kotlin versions, down to the patch level. So Compose compiler plugin 1.5.14 is only compatible with KGP 1.9.24, and Compose compiler 1.5.13 is only compatible with KGP 1.9.23.

https://developer.android.com/jetpack/androidx/releases/compose-kotlin


So, in the list of potential KGP versions we have to have the specific versions required by Compose compiler. Otherwise it might not be able to generate any valid combination for Compose and KGP.

When Compose is tested, then it will restrict the KGP version so there's only one KGP version.

However, when Compose isn't being tested, there's no restriction. The test infra will generate a list of all versions, which will contain combinations for both KGP 1.9.23 *and* KGP 1.9.24.

But for Dokka we don't really care about testing KGP 1.9.23 *and* KGP 1.9.24. So we filter out the lower patch versions.

Does that make sense? It's a bit hard to explain. If it helps, try removing the filtering and see that there's a lot of extra testing.

> may be 2.1.21 introduces bug-fix which is required for Dokka

If at some point DGP requires a specific version of KGP, then let's think of a solution then. We could add a custom requirement to make sure DGP gets tested.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: The Compose compiler has very strict requirements for specific Kotlin versions, down to the patch level.

https://developer.android.com/jetpack/androidx/releases/compose-kotlin

Without the exact KGP version Compose will throw an error.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/GradleTestProjectInitializer.kt
adam-enko: Good idea, I'll add a filter.

dokka-integration-tests/gradle/projects/it-android-compose/settings.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+dependencyResolutionManagement {
+    repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
+    repositories {
+        /* %{DOKKA_IT_MAVEN_REPO}% */
+        mavenCentral()
+        google()
+        maven("https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/public/p/compose/dev") {
```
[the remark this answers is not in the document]
adam-enko: Sorry, I don't know. For whatever reason some Compose compiler versions aren't published to Maven Central.

After Kotlin 2.0 Compose is easier to use, so I would avoid thinking about it too much and just wait until we drop Kotlin 1.9 :)

dokka-integration-tests/gradle/projects/it-android/settings.gradle.kts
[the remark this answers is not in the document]
whyoleg: yeah, okay, let's keep it like this for now

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
whyoleg: Yeah, I understand, but can't we just test it for 1.9.25?
I mean, it sounds a bit strange that we test with 3 latest patch versions - one should be enough, no?

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
whyoleg: I think that it should be or per-class or per-method, not both, otherwise yes, there are confusions. If KGP uses per-class, then let's leave it like this to be consistent.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: I'm pretty sure Compose will throw an error if it's used with KGP 1.9.25.

whyoleg: Looks like everything works fine with 1.9.24, and in the end, only tests for 1.9.25 (from 1.9.x) are run :)
image

Here, I've added `1.7.1` and `1.8.1` versions to `allComposeGradlePluginVersions` just to check, how it works. And here I see, that `AndroidProjectIT` runs for each compose version, even though it don't uses compose at all.

minor: printing Dokka version there doesn't make sense, as we will always run with single Dokka version (be it current snapshot or override)

I feel like current solution may work now (for current two tests with those specific versions), but it really has like foot-guns and I'm unsure what I think about it.
I like the direction of this PR, really, but it looks too complex for me and so it's hard to understand if it works as intended, and more important, that it will work in the same way when we will add more versions :)

I would like to hear an opinion of someone from KBT on this (e.g second reviewer :)), may be I'm too nitpicking...

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @DokkaGradlePluginTest(sourceProjectName = "it-android-compose")
+    fun `Dokka tasks should be build cacheable`(
+        project: DokkaGradleProjectRunner,
+    ) {
+        project.runner
+            .addArguments(
+                ":dokkaGenerate",
+                "--build-cache",
```
antohaby, line 92: tests appears to be testing only up-to-date checks on re-run.
why build cache is needed here?

shouldn't we also tests for some "changes" in outputs or inputs?

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
```diff
@@ -0,0 +1,158 @@
...
+        fun clearCcReports() {
+            project.file(".gradle/configuration-cache").deleteRecursively()
+            project.file("build/reports/configuration-cache").deleteRecursively()
+        }
+
+        val configCacheRunner =
+            project.runner.addArguments(
+                "clean",
+                ":dokkaGenerate",
+                "--no-build-cache",
```
antohaby, line 125: why is it necessary to specify "no build cache" here? I though configuration cache can work without knowing anything about build cache as it is more related to execution phase.

dokka-integration-tests/gradle/src/test/kotlin/AndroidProjectIT.kt
```diff
@@ -0,0 +1,163 @@
...
+            )
+            .build {
+                withClue("expect dokkaGenerate tasks are loaded from cache") {
+                    shouldHaveTask(":dokkaGenerate").shouldHaveOutcome(UP_TO_DATE)
+                }
+            }
+    }
+
+    @DokkaGradlePluginTest(sourceProjectName = "it-android")
+    fun `expect Dokka is compatible with Gradle Configuration Cache`(
```
antohaby, line 120: also it looks like AndroidComposeIT and AndroidProjectIT have shared code. Is it ok for you to keep those duplicated? i.e. you don't see lots of maintenance with this code in the future.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/testTags.kt
adam-enko: Good idea, renamed.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: Agreed, I removed 2.0.20, it wasn't necessary.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: I've added a provider, `TestedVersionsSource`, that uses a Sequence builder to generate the versions.

This can be used to explicitly add a new combination of versions, if necessary. But for now, it just needs to provide the combinations of all tested versions.

Does this work better?

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersions.kt
adam-enko: > And here I see, that `AndroidProjectIT` runs for each compose version, even though it don't uses compose at all.

Thanks, this was a bug, and I've fixed it.


> minor: printing Dokka version there doesn't make sense, as we will always run with single Dokka version (be it current snapshot or override)

My thinking here was to prefer to be explicit. Because the Dokka version can be overridden, I thought it'd be useful to always log the Dokka version. That way, it's easy to see what version of Dokka is being tested.

I looked into adding a filter to only log the version if the Dokka version override was not present, but tbh it was quite complicated and I think more effort than it was worth.

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
adam-enko: Thanks, good point, I've updated the assertions to test the actual Dokka generation work tasks, not just the `dokkaGenerate` lifecycle task.

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
adam-enko: Hmmm I suppose disabling build cache not strictly necessary, but I can imagine that a buggy Gradle task might produce some uncached output that could disrupt CC. 

TBH I can't remember why it's there. Perhaps it could be removed, but I don't think it does any harm.

antohaby (approved):

PR #3907, opened 2024-11-05
Improve directory comparison assertion

adam-enko:
Significantly improve comparison between directories.

Currently the failure messages are too obscure, and hard to diagnose. Example: https://ge.jetbrains.com/s/vl2jgh3ivevyi/tests/task/:dokka-integration-tests:gradle:test/details/org.jetbrains.dokka.it.gradle.AndroidProjectIT/generate%20dokka%20HTML(DokkaGradleProjectRunner)%5B1%5D?top-execution=1


adam-enko:
Ah, I forgot about binary files. I'll fix that, please wait before reviewing.

dokka-runners/dokka-gradle-plugin/src/test/kotlin/utils/ShouldBeADirectoryWithSameContentAsTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        val actualDir = tempDir().apply {
+            resolve("file.bin").writeBytes(PNG_BYTES.reversedArray())
+        }
+
+        val failure = shouldFail { expectedDir.shouldBeADirectoryWithSameContentAs(actualDir) }
+
+        failure.shouldHaveMessage(
+            """
+            file.bin has 2 differences in content:
+                --- file.bin
```
whyoleg, line 105: TBH, it would be better to just skip binary files difference printing if possible and check by hash/size only, it's not really useful in my opinion and diff doesn't really helps here.

E.g if we will change the image in our frontend, not sure that diff will help somehow :)

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/kotestFiles.kt
```diff
@@ -3,106 +3,140 @@
...
 
-    data class Context(
-        val expectParentPath: String,
-        val actualParentPath: String,
-    )
+/**
+ * Read lines from a file, leniently.
+ * Handles text and binary data (though binary data might be hard to read).
+ *
+ * ([kotlin.io.path.readLines] blows up when it reads binary files.)
```
whyoleg, line 110: as I mentioned before, IMO, I think for binary files it will be better to just check size/hash and not content

whyoleg (approved):
Overall, LGTM, thought it would be nice to see also the example of how the failure will be `after` this changes just to compare

dokka-runners/dokka-gradle-plugin/src/test/kotlin/utils/ShouldBeADirectoryWithSameContentAsTest.kt
adam-enko: I agree, but as far as I can tell JVM doesn't have a reliable way of determining if a file is a binary file or not. If I think about our priorities, this should be good enough for now.

What do you think?

whyoleg: Agree, the only reason I've mentioned it, is because we have custom logic for this, e.g `readByteLines`.

adam-enko:
> it would be nice to see also the example of how the failure will be `after` this changes just to compare

Are the test assertions helpful as brief examples?

https://github.com/Kotlin/dokka/blob/9c7a32aa8139c2aa34e289a99a5ff53e97aaddb4/dokka-runners/dokka-gradle-plugin/src/test/kotlin/utils/ShouldBeADirectoryWithSameContentAsTest.kt

whyoleg:
> Are the test assertions helpful as brief examples?

Yes and no :)
In the description you say `Currently the failure messages are too obscure, and hard to diagnose` and show the example for big project test failure, while in tests we do have small artificial project, and so I don't fully understand how the diff will look like for big project with nested files/folders etc.

The only thing that I understand which will be better is that there will be `diff`s. That's nice!

If that's hard to do, no problem, no need to spend time on this.
But in future it will be much better if it will be possible to see the difference of such changes and if it really solves the problem it's trying to solve :).
While you probably saw `before` and `after` on your own it's obvious for you what are the `pros`, but for me it's just comparing an `apple` (example in description) with `potato` (example in tests)

adam-enko:
So, [currently](https://ge.jetbrains.com/s/vl2jgh3ivevyi/tests/task/:dokka-integration-tests:gradle:test/details/org.jetbrains.dokka.it.gradle.AndroidProjectIT/generate%20dokka%20HTML(DokkaGradleProjectRunner)%5B1%5D?top-execution=1) the assertion failure has a pretty file tree, but it's hard to actually see what `expected:<[Change at line 44] │   │   │   ├── index.html` actually means.

Does this mean a file is missing or added? Or does it differ in content?

image


In this PR the assertion failure will provide a comprehensive message for all failures. It will list the files that are unexpected, missing, or differing in content. And, if the content is different, then show what expected/actual content is.

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/files.kt
```diff
@@ -21,3 +24,18 @@ fun Path.listRelativePathsMatching(predicate: (Path) -> Boolean): List<String> {
...
 }
+
+/**
+ * Create a temporary directory.
+ *
+ * Kotest will attempt to delete the file after the current spec has completed.
+ *
+ * (@see [io.kotest.engine.spec.tempdir], but this returns a [Path], not a [java.io.File].)
+ */
+fun TestConfiguration.tempDir(prefix: String? = null): Path {
```
Tapchicoma, line 35: Why not just use `io.kotest.engine.spec.tempdir.toPath()`? :thinking:

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/kotestFiles.kt
```diff
@@ -3,106 +3,140 @@
...
+ */
+private fun Path.readByteLines(): List<String> {
```
Tapchicoma, line 112: How about using either `.useLines { }` instead?

Tapchicoma (approved):

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/kotestFiles.kt
adam-enko: This also fails with MalformedInputException.

dokka-runners/dokka-gradle-plugin/src/test/kotlin/utils/ShouldBeADirectoryWithSameContentAsTest.kt
adam-enko: I figured out an improvement: just catch IOExceptions when reading the content, and print more details otherwise. So now binary file comparison failures look like this:


```
file.bin has 1 differences in content:
    --- file.bin
    +++ file.bin
    @@ -1,4 +1,1 @@
    -Failed to read file content
    -java.nio.charset.MalformedInputException Input length = 1
    -file size: 100
    -checksum: c+g2IzgALWMi8irrW5xr/gB32XVU8WTtaQ8hkD8qXzE=
    +This file in the actual dir has valid UTF-8 content
```

wdyt?

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/files.kt
adam-enko: I find having `toPath()` everywhere is repetitive and distracting, so it's nicer to have a convenient shortcut.

dokka-runners/dokka-gradle-plugin/src/test/kotlin/utils/ShouldBeADirectoryWithSameContentAsTest.kt
whyoleg: yeah, looks better!
minor: probably it would be better to leave binary contents for both expect and actual values as it was before, as this makes more sense I think

adam-enko: Cool, thanks.

The example used in the test is a bit atypical. In real life both files are likely to be binaries, so the logged message will be different.

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/files.kt
Tapchicoma: You could still have shortcut, but under-the-hood just invoke kotest method and convert it to path.

It is more about not reinventing the wheel :slightly_smiling_face:

PR #3908, opened 2024-11-06
Change default Dokka Generator worker isolation to 'classloader'

adam-enko:
[KT-72859](https://youtrack.jetbrains.com/issue/KT-72859)


adam-enko:
Classloader isolation causes OOMs, we will increase the heap instead 

- https://youtrack.jetbrains.com/issue/KT-72859/Change-default-Worker-isolation-to-Classloader#focus=Comments-27-11030570.0-0

PR #3915, opened 2024-11-12
Add debug utility for manually overriding the Dokka configuration JSON

adam-enko:
Being able to make minor manual modifications to the Dokka configuration is exceptionally useful for local debugging.

This should not be used in production, and it is protected by an opt-in API.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
@@ -128,7 +128,7 @@ constructor(
     ) {
         val dokkaConfiguration =
             if (overrideJsonConfig.isPresent) {
-                logger.warn("w: [$path] Overriding DokkaConfiguration with manualJsonConfig")
+                logger.warn("w: [$path] Overriding DokkaConfiguration with overrideJsonConfig")
                 DokkaConfigurationImpl(overrideJsonConfig.get())
             } else {
                 createDokkaConfiguration(generationType, outputDirectory)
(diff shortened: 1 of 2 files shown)


whyoleg (approved):
I'm not against to having this, though I don't fully understand, how it could help during local debugging. Could you explain it a bit?

adam-enko:
> I'm not against to having this, though I don't fully understand, how it could help during local debugging. Could you explain it a bit?

For example, @antohaby and I looked at improving KMP handling. It was convenient to be able to manually modify the Dokka configuration rather than trying to modify the Gradle config that DGP uses to produce the Dokka config.

For example, we could manually add/remove files from the dependencies that Dokka Generator analyses, or change the order of the files. Changing the order of the files would be quite difficult to do by modifying the Gradle build config.

So, instead we added this util. So we could run `:dokkaGenerate`, take the `dokka-config.json`, and quickly modify it to experiment with different configuration. Once we found some configuration that worked, we could focus on making DGP produce the correct config.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
```diff
@@ -114,7 +126,13 @@ constructor(
         generationType: GeneratorMode,
         outputDirectory: File,
     ) {
-        val dokkaConfiguration = createDokkaConfiguration(generationType, outputDirectory)
+        val dokkaConfiguration =
+            if (overrideJsonConfig.isPresent) {
+                logger.warn("w: [$path] Overriding DokkaConfiguration with manualJsonConfig")
+                DokkaConfigurationImpl(overrideJsonConfig.get())
```
antohaby, line 132: what's `manualJsonConfig `? I don't see it anywhere. Perhaps you meant `overrideJsonConfig ` :)

antohaby (approved):

antohaby:
I'm afraid it can be abused in the future. But on the other hand why not? someone can also apply patches to the json files when necessary.

PR #3920, opened 2024-11-14
Remove the dependency of `dokkaGenerate` on `dokkaGenerateModule*`

whyoleg:
`dokkaGenerate` is a lifecycle task, and should only trigger tasks that produce Dokka Publications.

Currently `dokkaGenerate` also triggers Dokka Module tasks, but this is unnecessary for single-module projects, which do not need to generate 'partial' Dokka Modules, because Dokka can generate them directly into a Publication.

Avoiding trigging the Dokka Module tasks will help with performance.


adam-enko (approved):
Great find!

adam-enko:
@whyoleg I'll leave it up to you to decide if another reviewer is needed, and to push the 'merge' button.

Issue #3702, opened 2024-07-19
[K2]  KMP: Unresolved declarations from another module in Android + JVM project


atyrin:
1. Create a project with 2 KMP module (android + jvm + any other target)
2. In Module1 declare a type
e.g.
```Kotlin
object Firebase
```

3. In Module2 declare two functions: one in the Android source set, another in the JVM source set that uses `Firebase` object. Two of them because I assume that the selected target may change

With K1 the extension page will have all platforms
<img width="500" alt="image" src="https://github.com/user-attachments/assets/fc2e94fc-97f0-4877-91fc-4c3b20d04d0c">

K2:
<img width="500" alt="image" src="https://github.com/user-attachments/assets/c723b915-99a5-4edd-b442-bc29f007a94e">

[dokka-repros.zip](https://github.com/user-attachments/files/16314276/dokka-repros.zip)

vmishenev:
Reproducer: https://github.com/Kotlin/dokka/commit/1cb62bd9656a99f000b341567dc054a0d1d4b6e4
This is blocked by https://youtrack.jetbrains.com/issue/KT-70384/Analysis-API-Standalone-The-same-class-in-the-same-two-renamed-jars-is-unresolved

whyoleg:
fixed in upstream: published in 2.1.0-dev-6739

PR #3941, opened 2024-11-28
Add a test for #3702

vmishenev:


dokka-subprojects/plugin-base/src/test/kotlin/translators/KMPTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        return File(resource).absolutePath
+    }
+
+    val configuration = dokkaConfiguration {
+        sourceSets {
+            sourceSet {
+                sourceRoots = listOf("src/androidMain/kotlin")
+                analysisPlatform = "jvm"
+                name = "android-example"
+                classpath = listOf(getResourceAbsolutePath("jars/jvmAndroidLib-jvm-copy.jar"))
```
whyoleg, line 29: would be nice to left some comment on how those jars were generated.

whyoleg (approved):

dokka-subprojects/plugin-base/src/test/kotlin/translators/KMPTest.kt
vmishenev: I left a comment about  what jars contain. I think it is not a problem  to build such jars.

PR #3942, opened 2024-11-28
KT-71784 Fix classpath for KMP shared source sets

adam-enko:


antohaby (approved):

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
```diff
@@ -118,6 +122,28 @@ class ExampleProjectsTest {
...
+
+        private fun updateGradleProperties() {
+            when (exampleProjectName) {
+                ExampleProject.KotlinMultiplatform -> {
+                    project.gradleProperties {
+                        // kotlin.native.enableKlibsCrossCompilation must be set to `true`
+                        // otherwise Kotlin can't generate a Klib for Coroutines in macosMain
+                        // when generating on Linux machines, resulting in 'Error class: unknown class'
+                        // for CoroutineScope appearing in the generated docs.
+                        kotlin.native.enableKlibsCrossCompilation = true
```
whyoleg, line 138: is there a reason why we set it here and also in example `gradle.properties?

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-    kotlin("multiplatform") version "1.9.25"
+    kotlin("multiplatform") version "2.1.0"
```
whyoleg, line 2: WDYT about creating separate PR to update all other examples to use latest Kotlin version?

examples/gradle-v2/kotlin-multiplatform-example/gradle.properties
```diff
@@ -5,3 +5,9 @@ org.gradle.parallel=true
 
 org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
 org.jetbrains.dokka.experimental.gradle.pluginMode.noWarn=true
+
+# kotlin.native.enableKlibsCrossCompilation must be set to `true`
+# otherwise Kotlin can't generate a Klib for Coroutines in macosMain
+# when generating on Linux machines, resulting in 'Error class: unknown class'
+# for CoroutineScope appearing in the generated docs.
+kotlin.native.enableKlibsCrossCompilation=true
```
whyoleg, line 13:
```suggestion
# kotlin.native.enableKlibsCrossCompilation could be set to `true`
# to be able to generate documentation via Dokka for macosMain on Linux machines
# Otherwise, Dokka will produce 'Error class: unknown class'
# for CoroutineScope appearing in the generated docs because of absent klibs
kotlin.native.enableKlibsCrossCompilation=false
```

Would be nice not to introduce experimental flags into example projects if possible

whyoleg (approved):
NIce!

The only suggestion I have is to not enable klib cross compilation by default in example because it's experimental and could cause other unknown to us issues.

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
adam-enko: Yeah, I think that's a good idea.

dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
adam-enko: The util that updates the Gradle properties doesn't merge them with the current ones, for better reproducibility

The test-util _used_ to merge the new properties with the old ones, but I took it out because it caused flaky tests https://github.com/Kotlin/dokka/commit/afab82d6755c419547a338e9fa0d552f29c02087#diff-32f73820e3e4e7cd35545577c35d32df857b8936909734336e493f74bd7a696eR37-R40

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
whyoleg: created PR: https://github.com/Kotlin/dokka/pull/3949

Issue #3681, opened 2024-07-08
[K2] Unresolved declarations are resolved in KDoc


atyrin:
```Kotlin
/**
 * Resolve value: [value]
 */
fun useValue() = value
```

`value` in the function body is unresolved in IDE. 
K2: `value` in KDoc is resolved and refers to `kotlin-stdlib-2.0.0-sources.jar!/commonMain/kotlin/util/Lazy.kt:44` for IDE and to https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/ for Dokka

K1: `value` in kdoc is unresolved

**Installation**
- Dokka version: 2.0.0

atyrin:
Ticket for Analysis API: https://youtrack.jetbrains.com/issue/KT-69736/K2-IDE-False-positive-resolution-from-KDoc-for-value

vmishenev:
The general case:
```kt
fun getProperty() = 0
/**
 * [property] is resolved in K2, but - not in K1
 */
fun usage() = 0
```

Related issue: https://github.com/Kotlin/dokka/issues/3586

Issue #3631, opened 2024-05-27
Unresolved KDoc link to two extensions with the same name


vmishenev:
```kt
class C
public fun C.ensureActive() {}
/**
 * [C.ensureActive]
 */
class B
/**
 * [C.ensureActive] is unresolved in IDE K2
 */
public fun B.ensureActive() {}
```
![image](https://github.com/Kotlin/dokka/assets/9636937/eed6480e-d8b2-4bc4-a1ea-3b4e9c383f55)

At the same time if an extension is renamed, it is resolved
![image](https://github.com/Kotlin/dokka/assets/9636937/c0ec9c3a-09bd-4bc7-9be1-9bcf2ba71d8b)

vmishenev:
One more case:
```
import kotlin.coroutines.*
class C
fun C.resume() = 0

/**
 * [Continuation.resume] // is unresolved
 */
fun usage() = 0
```

but (replaced star import )
```
import kotlin.coroutines.resume
class C
fun C.resume() = 0

/**
 * [Continuation.resume] // is resolved
 */
fun usage() = 0

vmishenev:
It is blocked by https://youtrack.jetbrains.com/issue/KT-69047/Analysis-API-Unresolved-KDoc-reference-to-extensions-with-the-same-name

PR #3947, opened 2024-12-04
Add new tests for KDoc links

vmishenev:
Checks #3681 and #3631

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
@@ -1099,6 +1099,78 @@ class LinkTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    fun `KDoc link should be unresolved to non-existed property with the name of Kotlin getter #3681`() {
+        testInline(
+            """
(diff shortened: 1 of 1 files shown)


dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
```diff
@@ -1099,6 +1099,78 @@ class LinkTest : BaseAbstractTest() {
...
+    @Test
+    fun `KDoc link should be unresolved to non-existed property with the name of Kotlin getter #3681`() {
+        testInline(
+            """
+            |/src/main/kotlin/Testing.kt
+            |fun getProperty() = 0
+            |/**
+            |* [property] is unresolved
+            | */
+            | usage() = 0
```
whyoleg, line 1111: shouldn't we have `fun usage()` ?
curious why the test is successful

whyoleg (changes requested):

dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
vmishenev: We should. See the fix.

PR #3973, opened 2024-12-22
Filter non-directory and hidden files from version directories

### LINKED ISSUE #3963: Versioning plugin + Dokka v2 Gradle plugin false .DS_Store warning
**Describe the bug**
Using the versioning plugin with the Dokka v2 Gradle plugin produces the warning:
> [:dokkaGeneratePublicationHtml] Failed to find versions file named version.json in /{olderVersionsDir}/.DS_Store

**Expected behaviour**
No warning should be output. .DS_Store (and probably any non-directory file) should be ignored.

**To Reproduce**
Run the `:dokkaGenerate` task with a .DS_Store file in the `olderVersionsDir` directory.

**Dokka configuration**

```Kotlin
dependencies {
}

        olderVersionsDir = projectDir.resolve("docs/older")
        // ...
    }
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v8.11.1
- Dokka version: 2.0.0

**Additional context**
Reproducible with [my library, Kotbase](https://github.com/jeffdgr8/kotbase/pull/32).

jeffdgr8:
Fixes #3963

### DIFF
--- a/dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
+++ b/dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
@@ -23,7 +23,9 @@ public data class VersioningConfiguration(
         if (olderVersionsDir != null)
             assert(olderVersionsDir!!.isDirectory) { "Supplied previous version $olderVersionsDir is not a directory!" }
 
-        return olderVersionsDir?.listFiles()?.toList().orEmpty() + olderVersions.orEmpty()
+        return olderVersionsDir?.listFiles()
+            ?.filter { it.isDirectory && !it.isHidden }
+            ?.toList().orEmpty() + olderVersions.orEmpty()
     }
 
     public companion object {


dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
```diff
@@ -23,7 +23,9 @@ public data class VersioningConfiguration(
         if (olderVersionsDir != null)
             assert(olderVersionsDir!!.isDirectory) { "Supplied previous version $olderVersionsDir is not a directory!" }
 
-        return olderVersionsDir?.listFiles()?.toList().orEmpty() + olderVersions.orEmpty()
+        return olderVersionsDir?.listFiles()
+            ?.filter { it.isDirectory && !it.isHidden }
```
whyoleg, line 27: Do we really need to filter out `hidden` directories here? Their definition is a bit vague, and I think that getting just directories should be enough.

If you think, that we should still filter out `hidden` directories for some reason, then it would be nice to add a comment for reasoning there.

whyoleg (changes requested):
Thanks!
Could you please check the comment?

dokka-subprojects/plugin-versioning/src/main/kotlin/org/jetbrains/dokka/versioning/VersioningConfiguration.kt
jeffdgr8: My reasoning was that typically a hidden directory wouldn't be used to store a versioned docs archive, but could be some system created or metadata directory. While .DS_Store was the cause of the issue I ran into and would be filtered out by just ignoring non-directory files, I was thinking there could be other hidden system or metadata files that are directories. An example might be .git or less likely .idea.

If you think it's more likely valid version directories actually are hidden, we could remove the `!it.isHidden` check. I'm fine either way.

whyoleg (approved):
Thanks!

PR #3974, opened 2024-12-22
Drop last additional newline character added when rewriting parsed HTML

### LINKED ISSUE #3965: Old versions get additional new line added
**Describe the bug**
Using the versioning plugin, older version HTML files receive an extra new line whenever they're processed with `:dokkaGenerate`. When a project replaces the older version HTML files every time they're reprocessed when API docs are regenerated, this will continuously grow the size of these older version files.

**Expected behaviour**
Older version files shouldn't receive an extra new line character, only necessary changes should be made to the files.

**To Reproduce**
Run `:dokkaGenerate` on a project with older version docs and compare the processed files with the original to see the additional new line character in every HTML file.

**Dokka configuration**

```Kotlin
dependencies {
}

        olderVersionsDir = projectDir.resolve("docs/older")
        // ...
    }
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v8.11.1
- Dokka version: 2.0.0 (also 1.9.20)

**Additional context**
Reproducible with [my library, Kotbase](https://github.com/jeffdgr8/kotbase/pull/32). Just run `:dokkaGenerate` and observe the Git changes. The only changes will be the additional new line characters to the older version HTML files. Every subsequent run will add an additional new line.

jeffdgr8:
Fixes #3965

### DIFF
--- a/dokka-subprojects/plugin-templating/src/main/kotlin/org/jetbrains/dokka/templates/DirectiveBasedTemplateProcessing.kt
+++ b/dokka-subprojects/plugin-templating/src/main/kotlin/org/jetbrains/dokka/templates/DirectiveBasedTemplateProcessing.kt
@@ -41,7 +41,8 @@ public class DirectiveBasedHtmlTemplateProcessingStrategy(private val context: D
                 handleCommandAsComment(command, bodyTrimed, input, output)
             }
 
-            Files.write(output.toPath(), listOf(document.outerHtml()))
+            // drop last additional \n character that's added
+            Files.write(output.toPath(), listOf(document.outerHtml().dropLast(1)))
             true
         } else false
 


dokka-subprojects/plugin-templating/src/main/kotlin/org/jetbrains/dokka/templates/DirectiveBasedTemplateProcessing.kt
```diff
@@ -41,7 +41,8 @@ public class DirectiveBasedHtmlTemplateProcessingStrategy(private val context: D
                 handleCommandAsComment(command, bodyTrimed, input, output)
             }
 
-            Files.write(output.toPath(), listOf(document.outerHtml()))
+            // drop last additional \n character that's added
+            Files.write(output.toPath(), listOf(document.outerHtml().dropLast(1)))
```
whyoleg, line 45: It might not be correct to drop just the last character here, as theoretically it can be a wrong character (not a new line).

The real issue here is the wrong usage of `Files.write`, which accepts a list of lines. However, we provide the whole HTML in one line, so inside `Files.write`, it appends an additional new line at the end.
So probably better fix would be to not use `Files.write` but do something like:

```suggestion
            output.writeText(document.outerHtml())
```

whyoleg (changes requested):
Thanks for the PR and sorry for the long delay with review!
Could you take a look on the comment and adapt the code if possible?

dokka-subprojects/plugin-templating/src/main/kotlin/org/jetbrains/dokka/templates/DirectiveBasedTemplateProcessing.kt
jeffdgr8: Ah, yes! Great observation. This should resolve the real issue. Thank you!

whyoleg (approved):
Thanks, LGTM!

PR #3978, opened 2024-12-28
Versioning with Dokka Gradle plugin v2

jeffdgr8:
Update versioning readme for Dokka Gradle plugin v2


whyoleg:
Hey! Thanks for the PR, but we would like to wait until DGPv2 will be stable and enabled by default and only then update all documentation.

PR #3989, opened 2025-01-10
Replace Potentially Scam Url from Readme projects using dokka [Needs Priority]

apolostudio:
The url is wrong (the correct one https://hexagontk.com/stable/api/) and redirects to a scammy-like slots site. Corection: Hexagon still uses Dokka in its Api docs

Note that clicking on [Hexagon Old Url](https://hexagonkt.com/api/index.html) redirects to below image website
Old Url redirecting to Site
Actual correct api docs use dokka

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Some libraries that use Dokka for their API reference documentation:
 
 * [kotlinx.coroutines](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/)
 * [Bitmovin](https://cdn.bitmovin.com/player/android/3/docs/index.html)
+* [Hexagon](https://hexagontk.com/stable/api/)
 * [Ktor](https://api.ktor.io/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,6 @@ Some libraries that use Dokka for their API reference documentation:
 
 * [kotlinx.coroutines](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/)
 * [Bitmovin](https://cdn.bitmovin.com/player/android/3/docs/index.html)
-* [Hexagon](https://hexagonkt.com/api/index.html)
```
MGaetan89, line 23: Hexagon API still uses Dokka: https://hexagontk.com/stable/api/
Maybe the link could be updated to point to that URL instead?
```md
* [Hexagon](https://hexagontk.com/stable/api/)
```

apolostudio: yeah sorry my fault, i was checking normal docs  (which mkmaterial docs is used), the api docs do use dokka, changing it now!

vmishenev (approved):
Thank you for noticing it

PR #3990, opened 2025-01-10
Update expected data in integration tests

vmishenev:


adam-enko (approved):
I see the tests are passing, so I'll approve. But how did they get out of sync? 

I added these tests quickly, so maybe we can take some time to improve them.

whyoleg (approved):

Issue #3996, opened 2025-01-15
Supertypes for java source use incorrect generics


juliamcclellan:
Sample java source:
```java
class Bar<T> {}
public class Foo extends Bar<String> {}
```

The super class type of `Foo` should be `Bar<String>`, but the `supertypes` on the `DClass` for `Foo` instead includes the type `Bar<T>`, using the type parameter from the definition of `Bar` instead of the concrete type used in `Foo`'s definition.

(This is handled correctly for kotlin source.)

PR #3997, opened 2025-01-15
Update java supertype processing

juliamcclellan:
The processing of supertypes for java source would convert `PsiClassType`s to `PsiClass`es while creating `AncestryNode`s. This meant that the type parameters used for the supertypes were from the `PsiClass` definition instead of the `PsiClassType` usage like they should be.

For example, in the following java code:
```
class Bar<T> {}
public class Foo extends Bar<String> {}
```

The super class type of `Foo` should be `Bar<String>`. When processing the supertypes of `Foo`, the `PsiClassType` representing `Bar<String>` was converted to a `PsiClass`. The `PsiClass` reflects the definition of `Bar`, so `Bar<T>` became the super class type instead of `Bar<String>`.

Fixes: #3996


vmishenev (approved):

vmishenev (approved):

vmishenev:
Thank you for your contribution.

PR #4009, opened 2025-01-27
Cache `packages` field, which is used by `DefaultExternalLocationProvider.resolve`

whyoleg:
Found and tested on https://github.com/rjaros/kilua (results are reproducible even after several runs).
Reduces allocations inside [DefaultExternalLocationProvider.resolve](https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/external/DefaultExternalLocationProvider.kt#L22) from 11 GB to 1 GB. Previously `packages` were recreated for each `resolve`.
The changes don't really affect kotlinx.io and kotlinx.coroutines.

For future: probably it might make sense to also cache results of `DefaultExternalLocationProvider.resolve`?
In this specific project there are a lot of references to `@Composable` annotation passing through this function call.

### DIFF
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/shared/PackageList.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/shared/PackageList.kt
@@ -14,8 +14,7 @@ public data class PackageList(
     val locations: Map<String, String>,
     val url: URL
 ) {
-    val packages: Set<String>
-        get() = modules.values.flatten().toSet()
+    val packages: Set<String> by lazy { modules.values.flatten().toSet() }
 
     public fun moduleFor(packageName: String): Module? {
         return modules.asSequence()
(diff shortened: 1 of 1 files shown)


dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/shared/PackageList.kt
```diff
@@ -14,8 +14,7 @@ public data class PackageList(
     val locations: Map<String, String>,
     val url: URL
 ) {
-    val packages: Set<String>
-        get() = modules.values.flatten().toSet()
+    val packages: Set<String> by lazy { modules.values.flatten().toSet() }
 
     public fun moduleFor(packageName: String): Module? {
         return modules.asSequence()
```
vmishenev, line 20: Also, it seems this could be rewritten a bit better:
e.g. `modules.entries.firstOrNull { it.value.contains(packageName) }?.key`

vmishenev (approved):
>probably it might make sense to also cache results of DefaultExternalLocationProvider.resolve?

It might, but I'm afraid it could increase `maxHeapSize`. Anyway, we should assess the potential performance.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/shared/PackageList.kt
whyoleg: Thanks!
Keeping in mind this function (which I initially just ignored as there were no hot-spot in it), I've rewritten implementation so that `moduleFor` will be just map entry access.
I've also rerun perf tests on the same projects and the result is roughly the same.

PR #4010, opened 2025-01-27
chore: improved pre commit checks in UI

berezinant:
* fixed missed linting errors 
* improved pre commit checks
those checks should only affect commits when there are staged files from plugin-base-frontend


whyoleg (approved):
Nice improvement !

PR #4014, opened 2025-01-31
Update KGP to 2.1.10

whyoleg:
Depends on #3896


.github/workflows/preview-publish-ga.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         with:
           develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
       - name: Document coroutines
-        run: ./gradlew :dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines --stacktrace "-Dorg.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=500m"
+        run: ./gradlew :dokka-integration-tests:gradle:testExternalProjectKotlinxCoroutines --stacktrace
```
adam-enko, line 27: I see that `-Xmx2` is set both for the integration tests and in the root `gradle.properties`. Is setting Xmx in two places necessary? I think setting in the `gradle.properties` should also affect the integration tests.

Just setting it in one place would avoid the risk of the values getting out of sync. And being able to find the best value and using as a default would help with developers who run integration tests locally.

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```diff
@@ -16,9 +17,9 @@ fun Project.configureGradleKotlinCompatibility() {
     if (!dokkaBuild.enforceGradleKotlinCompatibility.get()) return
 
     extensions.configure<KotlinJvmProjectExtension>("kotlin") {
-        @OptIn(ExperimentalKotlinGradlePluginApi::class)
-        compilerVersion.set("2.0.20")
-        coreLibrariesVersion = "2.0.20"
+        @OptIn(ExperimentalKotlinGradlePluginApi::class, ExperimentalBuildToolsApi::class)
+        compilerVersion.set("2.0.21")
+        coreLibrariesVersion = "2.0.21"
```
adam-enko, line 22: Just double checking: Should the Kotlin versions here be 2.1.21, like it is in the version catalog, or should it be 2.0.21?

adam-enko:
JsIRGradleIntegrationTest is failing, I wonder what the cause is...

```
  java.lang.IllegalStateException: Unspecified version of react for kotlin 2.1.21
    at org.jetbrains.dokka.it.gradle.JsIRGradleIntegrationTest.execute(JsIRGradleIntegrationTest.kt:33)
```

.github/workflows/preview-publish-ga.yml
whyoleg: > I see that -Xmx2 is set both for the integration tests

Could you expand on that? Do you mean when we run integration tests, we have `-Xmx2g` for the spawned in Gradle IT daemons? I don't think that I understand why we should keep those values in sync.

In this PR, we reduce the memory provided for the main Gradle process and remove any unnecessary custom memory settings on GHA CI.

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
whyoleg: It should be 2.0.21.
Kotlin 2.1.x doesn't support AV/LV = 1.4 anymore

whyoleg: I could (try to) migrate those values to the Version Catalog (with the changes from https://github.com/Kotlin/dokka/pull/3896#discussion_r2099886624)

gradle/libs.versions.toml
```diff
@@ -3,10 +3,9 @@
 # We use a different compiler version to compile modules and Gradle Plugin
 # to be able to use Kotlin Language/API version 1.4 to be compatible with Gradle 7.
 # The logic leaves in build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```
MGaetan89, line 5: Not link to this PR, but just in case:
```suggestion
# The logic lives in build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
whyoleg: I've migrated the version of the compiler used via BTA to compile modules as wel as AV/LV to version catalog in https://github.com/Kotlin/dokka/pull/4143

gradle/libs.versions.toml
whyoleg: Thanks! Fixed in another PR (https://github.com/Kotlin/dokka/pull/4143)

adam-enko (approved):

PR #4016, opened 2025-01-31
Fix merging of implicit expect/actual with a single declaration

vmishenev:
The main idea of this fix is that pages with implicit expect/actual declarations and a single declaration should have the same name since they are merged by `PageMerger` based on the page name. Before the fix, pages with the same DRI/path were unmerged, which caused an exception.

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
@@ -6,8 +6,12 @@ package expectActuals
 
 import org.jetbrains.dokka.DokkaSourceSetID
 import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
+import org.jetbrains.dokka.model.DFunction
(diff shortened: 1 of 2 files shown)


dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
```diff
@@ -435,4 +439,68 @@ class ExpectActualsTest : BaseAbstractTest() {
...
+            fun isShowing(): Boolean {
+                return false
+            }
+        }
+        """.trimMargin(),
+        multiplatformConfiguration
+    ) {
+        pagesGenerationStage = { root ->
+            val cl = root.dfs { it.name == "Skiko" && it is ClasslikePageNode } ?: throw IllegalStateException()
+            assertEquals(2, cl.children.count { it.name == "[jvm]isShowing" })
```
vmishenev, line 468: Before the fix, there were 2 pages: `[jvm]isShowing` and `isShowing` with the different names, but the same DRI.

whyoleg (approved):
Nice!

Issue #3951, opened 2024-12-09
Update Annalysis API to `2.1.20-dev-5634` or later


vmishenev:
Blocked by https://youtrack.jetbrains.com/issue/KT-73776

vmishenev:
Updating AA to `2.2.0-dev-745` is currently blocked by https://youtrack.jetbrains.com/issue/KT-74567/Analysis-API-Standalone-AbstractMethodError-PluginStructureProviderResourceDataLoader.loadLjava-lang-StringLjava-io-InputStream

whyoleg:
> Updating AA to 2.2.0-dev-745 is currently blocked by updating the IJ platform version to 241.19416.19.

@vmishenev Let's synchronize version for K1 and K2 compiler versions (and so update IJ platform).
IJ platform update will happen only in Kotlin 2.2.0, so I don't think that we should wait until then.

Issue #3701, opened 2024-07-19
[K2] KMP: Support intersecting source roots


atyrin:
Sometimes projects may use the same folder as source root in different modules. Like the same dir for several targets in KMP.
It can be done with such syntax:
```Kotlin
kotlin {
  sourceSets {
    jsMain {
      kotlin.srcDir('src/browserMain/kotlin')
    }
    wasmJsMain {
      kotlin.srcDir('src/browserMain/kotlin')
    }
  }
}
```

Possible scenarios for that: Android and JVM source sets or Js and WasmJs. Sharing source sources for them using KMP hierarchies is just painful.

Projects that already use the pattern:
* https://github.com/cashapp/molecule/blob/34e57f70b1f7e7b08e9cd2ce6923e2f2b3be54b9/molecule-runtime/build.gradle#L77
* https://github.com/JakeWharton/mosaic/blob/eb76d4c1baf4605614f2346d081340c8d574320a/samples/counter/build.gradle#L18
* https://github.com/f-droid/fdroidclient/blob/d38b952e6dddef3d3baef9007e88d238e675be46/libs/download/build.gradle#L57
* https://github.com/GitLiveApp/firebase-kotlin-sdk/blob/1abbb9b64fb2c9e77aaee22f74a3fe9a2f548b7d/firebase-config/build.gradle.kts#L145

Kotlin with K2 works fine for them. 
IJ with Kotlin K2 Analysis also works: it just uses one of the platforms for analysis. And report warning: 
```
Path [/Users/Andrei.Tyrin/IdeaProjects/_external/molecule/molecule-runtime/src/browserMain/kotlin] of module [molecule.molecule-runtime.jsMain] was removed from modules [molecule.molecule-runtime.wasmJsMain]
```

Dokka at the same time has the special treatment for that (#3239): 
```
Pre-generation validity check failed: Source sets 'android' and 'jvm' have the common source roots: /Users/Andrei.Tyrin/IdeaProjects/_gates/firebase-kotlin-sdk_mpp/firebase-config/src/androidMain/kotlin. Every Kotlin source file should belong to only one source set (module).
```

----

There is a ticket to forbid such setup in KGP: https://youtrack.jetbrains.com/issue/KT-62829/Add-a-check-that-two-source-sets-cannot-have-the-same-file-system-directory. But the priority is low.

vmishenev:
Related: [#3373](https://github.com/Kotlin/dokka/issues/3373)

[...]

PR #4027, opened 2025-02-11
[K2] Support generating documentation for stdlib

vmishenev:
StdLib should copy some source and sample files to avoid [#3701](https://github.com/Kotlin/dokka/issues/3701).
Also, generating documentation works well with the fix for https://youtrack.jetbrains.com/issue/KT-74740. However, to have the fix, Dokka needs to update AA - #3951, which is currently blocked.


whyoleg (approved):
Nice!

PR #4037, opened 2025-02-24
test: exclude content of UI assets from comparison in html output tests

berezinant:


dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
```diff
@@ -64,6 +64,12 @@ class AndroidComposeIT {
         withClue("expect the same HTML is generated") {
             val expectedHtml = project.projectDir.resolve("expectedData/html")
             val actualHtmlDir = project.projectDir.resolve("build/dokka/html")
+            val excludedFiles = listOf(
+                "ui-kit/ui-kit.min.js",
+                "ui-kit/ui-kit.min.css",
+                "styles/main.css",
+                "scripts/main.js",
+            )
```
adam-enko, line 72: Might it be a good idea to create a global variable `dokkaHtmlStyleFiles: List<String>`,  so if they need to be changed they can be updated in all tests?

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/kotestFiles.kt
```diff
@@ -20,8 +20,8 @@ import kotlin.io.path.*
  *
  * Only files will be compared, directories are ignored.
  */
-infix fun Path.shouldBeADirectoryWithSameContentAs(path: Path) {
-    val differences = describeFileDifferences(this, path)
+fun Path.shouldBeADirectoryWithSameContentAs(path: Path, excludeFiles: List<String> = emptyList()) {
```
adam-enko, line 23: `excludeFiles` will completely exclude the files from the comparison. Maybe it would be better to exclude checking the file's content instead? That way we can still check that the files exist, but the actual styling isn't relevant?

berezinant: Moved the check inside the contents comparison part

dokka-integration-tests/gradle/src/test/kotlin/AndroidComposeIT.kt
berezinant: Done

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/kotestFiles.kt
adam-enko: Thanks! I think it'd be a good idea to update the variable name too, to be more descriptive. wdyt?

E.g. `filesExcludedFromContentCheck`

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/TestConstants.kt
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.it.gradle
+
+object TestConstants {
+    val DokkaHtmlAssetsFiles = listOf(
```
adam-enko, line 8:
```suggestion
object TestConstants {
    /**
     * Dokka HTML files that are only used for styling the generated HTML content.
     *
     * These files should be excluded from file-content assertion checks, to avoid
     * minor style changes affecting the tests.
     */
    val DokkaHtmlAssetsFiles = listOf(
```

suggestion: add a small comment to explain. I would find it useful. Feel free to reword it.

adam-enko (approved):

Issue #3661, opened 2024-06-28
kdoc parser fails when there is a newline inside link/img text.


owengray-google:
[If you have link text that is very long and splits onto multiple lines, 
like this, the kdoc parser does not recognize it as a link. 
We have run into this because ktfmt (https://github.com/facebook/ktfmt) is now formatting our kdoc syntax, producing the diff in this link. 
In the general case, like this link itself, I don't think it is viable to always require the link be on one line.](https://android-review.googlesource.com/c/platform/frameworks/support/+/3110218/4/compose/material/material/src/commonMain/kotlin/androidx/compose/material/Chip.kt)

The CommonMark spec for markdown [explicitly allows line breaks inside link text](https://spec.commonmark.org/0.31.2/#example-541), and specifies that img tags use a superset of the link syntax.

![unnamed (1)](https://github.com/Kotlin/dokka/assets/56644404/9b2b428f-24a9-41a0-a4f0-6411afca0544)

owengray-google:
Curiously, it appears that inline doc comments on property parameters use a different system, in which this somewhat works:  
(However, note that the hoverover text on the 'working' picture is `assist chip * image`. This is probably a separate bug.)
<img width="1241" alt="Screenshot 2024-07-11 at 4 42 02 PM" src="https://github.com/user-attachments/assets/ca1e5cd8-ea3c-41ee-afc9-06bb3baa6d5c">
<img width="1307" alt="Screenshot 2024-07-11 at 4 23 26 PM" src="https://github.com/user-attachments/assets/11a4cff4-a109-499b-b6b8-4dcfdf89bef4">

vmishenev:
1. A breakline is supported for an inline markdown link to an element in IDE and Dokka:
 ```kt
 /**
 * Text
 * [some 
 * link](https://www.google.com/images/branding/googlelogo/2x/googlelogo_color_272x92dp.png)
 *
 * Text:
 * [some
 *  text][f]
 */
fun f() = 0
 ```
 IDE:
![image](https://github.com/user-attachments/assets/1e737125-c75e-49ca-9263-c129c3d03011)

 2. External links/image links do not support a breakline due to the lexer grammar [here](https://github.com/JetBrains/kotlin/blob/fd7d542c945a0dfb0a18bc65de7ea33bb2425612/compiler/psi/src/org/jetbrains/kotlin/kdoc/lexer/KDoc.flex#L159). 
[...]

PR #4058, opened 2025-03-11
Update Analysis API to 2.2.0-dev-8822

vmishenev:
The AA 2.2.0-dev-8822  contains the fix of https://youtrack.jetbrains.com/issue/KT-74555

Presumably, the BCV plugin fails because `kotlinx-metadata-jvm` is unable to read Kotlin 2.2 metadata (e.g. see https://github.com/jetbrains/kotlin/commit/1a45888d50e0e8599d3763a5c21a888f34ff213f ). To avoid this, inline functions were copied.


whyoleg (approved):
may be it makes sense to add tests for https://youtrack.jetbrains.com/issue/KT-74555 / https://github.com/Kotlin/dokka/issues/3661 in this PR?
Still, it's fine of course to do it separately

PR #4059, opened 2025-03-12
Add test for handling newlines in KDoc link text

vmishenev:


dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
```diff
@@ -1118,6 +1119,37 @@ class LinkTest : BaseAbstractTest() {
...
+            |fun usage() = 0
+            |}
+        """.trimMargin(),
+            configuration
+        ) {
+            documentablesMergingStage = { module ->
+                val doc = module.dfs { it.name == "usage" }?.documentation?.values?.single()
+                    ?: throw IllegalStateException("Can't find documentation for declaration 'usage'")
+                val externalLink = doc.firstMemberOfType<A>()
+                assertEquals("some link", externalLink.children.first().text())
```
whyoleg, line 1145: it would be nice to also check, that we parsed the full link correctly, including url (`https://www.google.com/`)
Same for the `docLink` below

dokka-subprojects/plugin-base/src/test/kotlin/markdown/ParserTest.kt
```diff
@@ -1302,6 +1313,33 @@ class ParserTest : KDocTest() {
...
+                Description(
+                    CustomDocTag(
+                        listOf(
+                            P(
+                                listOf(
+                                    Img(
+                                        emptyList(),
+                                        mapOf(
+                                            "href" to "https://www.google.pl/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png",
+                                            "alt" to "Sample\nimage"
```
whyoleg, line 1330: why do we have new line here preserved while we have just space/concatenation in cases above?
I would expect no `\n` here

whyoleg (changes requested):
Nice, but I'm curious about the image link case

dokka-subprojects/plugin-base/src/test/kotlin/markdown/ParserTest.kt
vmishenev: Note that `alt` is an attribute where Dokka does not normalize space and so on.
Meanwhile, Dokka normalizes spaces/newlines in HTML text (`Text` ).
For example, it means
`![&nbsp;\n]` is stored as `alt="&nbsp;\n"` in `Img`
`[&nbsp;\n]]` is stored as `"  "` in `Text`

whyoleg: > Note that alt is an attribute where Dokka does not normalize space and so on.

just curious, do you know the reason for this?

If we will look on original reported issue - they have a problem with images too.
markdown spec also says that it should behave like ordinary `link text`: https://spec.commonmark.org/0.31.2/#image-description

So in this case I would really expect for behavior to be the same.

vmishenev: > just curious, do you know the reason for this?


TBH I don't know exactly

> If we will look on original reported issue - they have a problem with images too.

Yes, the issue was that we didn't recognize image links at all. Now, I see no problem in replacing `\n` with a space (or unescaping) on a render side if needed.

> markdown spec also says that it should behave like ordinary link text: 


It also has the hint:
> Though this spec is concerned with parsing, not rendering, it is recommended that in rendering to HTML, only the plain string content of the [image description](https://spec.commonmark.org/0.31.2/#image-description) be used.

For example,
`b **b**` renders as `foo ar` without formatting.
`[b **b**](/url)` renders as `<a href="/url">b <strong>b</strong></a>`.


So, how we should keep it in `DocumentableModel` or how we should render remains an open question.

As for me, the consistency between the `alt` attribute and the text raises some questions (What about the other attributes? Breaking changes? Should we do this even for text? and so on) and should be addressed in a separate issue if needed.

whyoleg: Thanks for additional examples!
Sure, let's discuss this in a separate issue.

whyoleg (approved):

PR #4071, opened 2025-03-21
Incorrect links being used for the Dokka plugins page

StillLutto:
It's as simple as that. I'm guessing that the default branch used to be the latest version, but that that's recently changed. I don't know more about this codebase, so this change might not make sense, if so, simply just tell me.

Have a wonderful day!


adam-enko:
Hi, the versions are updated when they are uploaded to the kotlinlang.org website.

https://kotlinlang.org/docs/dokka-plugins.html#notable-plugins

PR #4077, opened 2025-03-31
UI fixes for table of contents

berezinant:


dokka-subprojects/plugin-base-frontend/src/main/ui-kit/dropdown/index.ts
```diff
@@ -79,6 +99,10 @@ function addKeyboardNavigation(dropdown: HTMLElement): void {
       (dropdown.querySelector(DROPDOWN_TOGGLE) as HTMLElement)?.focus();
     }
   });
+  const dropdownOptions = dropdown.querySelectorAll('.dropdown--option');
+  dropdownOptions.forEach((option: Element) => {
+    option.addEventListener('keydown', preventScrollBySpaceKey);
+  });
```
nikpachoo, line 105: I'd add the event listener via event delegation to reduce listeners (but, if the number of options is low we can leave it as is).

dokka-subprojects/plugin-base-frontend/src/main/ui-kit/utils.ts
```diff
@@ -34,6 +34,20 @@ export function getScreenType(): ScreenType {
   return isMobile() ? 'mobile' : isTablet() ? 'tablet' : 'desktop';
 }
 
+export function isFocusableElement(element: HTMLElement): boolean {
+  let currentElement: HTMLElement | null = element;
+  if (currentElement.tabIndex === -1) {
+    return false;
+  }
+  while (currentElement) {
+    if (currentElement.style.display === 'none' || getComputedStyle(currentElement).display === 'none') {
```
nikpachoo, line 43: it does not account for elements that are hidden via `visibility: hidden`.
Do you think it make sense to support the `visibility` case?

nikpachoo (approved):

dokka-subprojects/plugin-base-frontend/src/main/ui-kit/utils.ts
berezinant: supported this case also ✅

dokka-subprojects/plugin-base-frontend/src/main/ui-kit/dropdown/index.ts
berezinant: the amount of the option is not too high but I changed it anyway

PR #4080, opened 2025-04-04
Support context parameters

vmishenev:
Key moments:
1. There are 2 ways to extend DocumentableModel in data classes:
    - via `extra`, e.g.
    ```kt
    internal class ContexrParametersExtra
    public val DFunction.contextParameters = extra[ContexrParametersExtra]
    ```
    - via an additional [constructor and a copy method](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api) to keep binary compatibility. 
    
The latter is implemented in this PR.
    
  2. `ExperimentalDokkaApi` has been introduced since context parameters have not been released officially and the AA has experimental API for them.
  3. The AA supports context parameters without extra flags by default.  In Dokka, this feature is under the flag `org.jetbrains.dokka.analysis.experimental.contextParameters`, but it can be removed **(???)** (The stability of the ContextParameters API should be clarified; see point 2 above)
  
 4. Dynamical formatting does not work well. Currently, it works only for either context parameters or value parameters, but not both.
 
image

I'll create a separate issue for it, as this part primarily concerns the JS script. (see #2659, #3116)

 UPD: https://github.com/Kotlin/dokka/issues/4084


dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -139,6 +167,11 @@ public data class PointingToCallableParameters(val parameterIndex: Int) : DriTar
     override fun toString(): String = "PointingToCallableParameters($parameterIndex)"
 }
 
+@ExperimentalDokkaApi
+public data class PointingToContextParameters(val parameterIndex: Int) : DriTarget() {
```
vmishenev, line 171: This is used only for merging context parameters in [DefaultDocumentableMerger.kt](https://github.com/Kotlin/dokka/pull/4080/files#diff-00add2e0f152afbda9fba4e4e7e8d666bcfb619525d9d73bb211e7776e7bd1a9)

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -493,6 +504,8 @@ internal class DokkaSymbolVisitor(
                         dri
                     )
                 },
+                contextParameters = if (!contextParametersEnabled) emptyList() else @OptIn(KaExperimentalApi::class) propertySymbol.contextParameters
```
whyoleg, line 507: Question related to LanguageFeatures: what could be the reason for us to guard calls to `contextParameters`?
AFAIU AA API will just return `emptyList` anyway

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -84,9 +85,36 @@ public val DRI.sureClassNames: String
...
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    public fun copy(
+        name: String = this.name,
+        receiver: TypeReference? = this.receiver,
+        params: List<TypeReference> = this.params
+    ): Callable = Callable(name, receiver, params)
+
+    public fun signature(): String {
+        /**
+         * Compatibility with [signature] without context parameters must be preserved,
```
whyoleg, line 108: Nice catch!

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -139,6 +167,11 @@ public data class PointingToCallableParameters(val parameterIndex: Int) : DriTar
...
 }
 
+@ExperimentalDokkaApi
+public data class PointingToContextParameters(val parameterIndex: Int) : DriTarget() {
+    override fun toString(): String = "PointingToContextParameters($parameterIndex)"
+}
+
 public fun DriTarget.nextTarget(): DriTarget = when (this) {
     is PointingToGenericParameters -> PointingToGenericParameters(this.parameterIndex + 1)
     is PointingToCallableParameters -> PointingToCallableParameters(this.parameterIndex + 1)
```
whyoleg, line 177: Should we add `PointingToContextParameters` here?

dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
```diff
@@ -1365,6 +1366,59 @@ class SignatureTest : BaseAbstractTest() {
...
+        val writerPlugin = TestOutputWriterPlugin()
+
+        testInline(
+            source,
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
+                writerPlugin.writer.renderedContent("root/example/simple-fun.html").firstSignature().match(
+                    "context(", @OptIn(ExperimentalDokkaApi::class) ContextParameters(
```
whyoleg, line 1385: minor: I think it will be enough to just guard the function with OptIn, so that this part of the code will be less cluttered (same in other similar places)

dokka-subprojects/plugin-base/src/test/kotlin/model/TypesTest.kt
```diff
@@ -115,4 +118,26 @@ class TypesTest : AbstractModelTest("/src/main/kotlin/classes/Test.kt", "types")
             }
         }
     }
+
+    @Test
+    @OnlySymbols("context parameters")
+    @OptIn(ExperimentalDokkaApi::class)
+    fun `functional type with context parameters and receiver`() = withContextParametersEnabled {
```
whyoleg, line 125: I see that there is a test above involving typealias to a functional type. Maybe it makes sense to also add similar test for context parameters?

whyoleg (changes requested):
Nice!
I've left several questions/nit-picks, but mostly it looks cool!
Still, the question regarding should be guard this under a flag (and language features) or not is remaining.
My main idea is to ask AA API regarding the correct way here and compatibility of `experimental` API - I mean, what could go wrong when we use such APIs. As I understand, those are already used in IDEA (2025.1), so probably they should be stable enough to use at our  side without guarding them

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
```diff
@@ -43,17 +43,20 @@ internal fun getLanguageVersionSettings(
     languageVersionString: String?,
     apiVersionString: String?
 ): LanguageVersionSettingsImpl {
-    val languageVersion = LanguageVersion.fromVersionString(languageVersionString) ?: LanguageVersion.LATEST_STABLE
+    val languageVersion = LanguageVersion.fromVersionString(languageVersionString) ?: LanguageVersion.KOTLIN_2_1
```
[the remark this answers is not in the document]
vmishenev: leftover

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
```diff
@@ -223,12 +229,93 @@ public data class DFunction(
...
-        get() = parameters
+        get() = parameters + @OptIn(ExperimentalDokkaApi::class) contextParameters
 
     override fun withNewExtras(newExtras: PropertyContainer<DFunction>): DFunction = copy(extra = newExtras)
+
+    /**
+     * Creates a copy of DFunction without contextParameters to maintain binary compatibility
+     */
+    @OptIn(ExperimentalDokkaApi::class)
+    public fun copyWithoutContextParameters(): DFunction = copy(contextParameters = emptyList())
```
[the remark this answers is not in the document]
vmishenev: leftover after a failed attempt by AI to generate `copy`

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: I changed it just to keep the contract of the public function.

vmishenev:
The AA team has confirmed that calling `contextParameters` in AA is safe, even when the feature is disabled in a language version.  
Therefore, I will remove the flag `org.jetbrains.dokka.analysis.experimental.contextParameters` in Dokka since I do not know valid user cases for disabling the context parameters.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
```diff
@@ -255,6 +257,7 @@ public class KotlinSignatureProvider(
             ?: false
     }
 
+    @OptIn(ExperimentalDokkaApi::class)
```
whyoleg, line 260: minor: as we do opt-in on a class itself, it’s unnecessary to repeat opt-in in nested scopes

whyoleg (approved):
Good to go!
Please link the issue about context parameters formating to the PR if already created

PR #4088, opened 2025-04-25
feat: reworked markdown tables according to new design

berezinant:
This is how it looks like
image
image


dokka-subprojects/plugin-base-frontend/src/main/ui-kit/table/styles.scss
```diff
@@ -0,0 +1,69 @@
...
+  min-width: 40px;
+
+  text-align: left;
+  vertical-align: top;
+
+  border: 1px solid var(--color-b20);
+}
+
+th {
+  padding: 7.5px 15.5px;
```
nikpachoo, line 37: Could you please double check the float values? Didn't expect to have it there.

nikpachoo (approved):

nikpachoo (approved):

Issue #1407, opened 2020-08-30
Javadoc only shows first sentence of documentation for methods


GlenKPeterson:
**Describe the bug**
I expected that if I put a period in my Javadoc for a method, that anything after the period will still show up in the Method Detail section lower down the page, but it just shows the first sentence in both places (it's correct in the Method Summary section).  Seems to work fine for objects/classes, it's methods/fields that are the problem.

### kotlin source file:
```kotlin
    /**
     * Implement this using the implementations in [StringUtils] as a guide to pretty-print
     * Indented data structures for joyful debugging.
     * If you aren't then using `.toString()` for anything else, implement by calling `indentedStr(0)`.
     */
    fun indentedStr(indent:Int):String
```

### HTML output:
```html
<!-- ============ METHOD DETAIL ========== -->
<section role="region">
    <ul class="blockList">
        <li class="blockList"><a id="method.detail">
            <!--   -->
        </a>
            <h3>Method Detail</h3>

            <a id="indentedStr(Integer)">
                <!--   -->
            </a>
            <ul class="blockListLast">
                <li class="blockList">
                    <h4>indentedStr</h4>
                    <pre class="methodSignature"> abstract <a href=https://docs.oracle.com/javase/8/docs/api/java/lang/String.html>String</a> <a href=IndentedStringable.html#indentedStr(Integer)>indentedStr</a>(<a href=https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html>Integer</a> indent)</pre>
                    <div class="block">Implement this using the implementations in <a href=StringUtils.html>StringUtils</a> as a guide to pretty-print Indented data structures for joyful debugging.</div>

                    <!-- TODO missing return annotation -->
                </li>
            </ul>

        </li>
    </ul>
</section>
```
It's missing `If you aren't then using...` sentence.

**To Reproduce**
Build this project:
https://github.com/GlenKPeterson/Indented

Run with:
```text
./gradlew clean assemble dokkaJavadoc
```

**Dokka configuration**
Default configuration produces this output.

**Installation**
- Operating system: Linux
- Dokka version: 1.4.0-rc
```text
$ ./gradlew --version

------------------------------------------------------------
Gradle 6.3
------------------------------------------------------------

Build time:   2020-03-24 19:52:07 UTC
Revision:     bacd40b727b0130eeac8855ae3f9fd9a0b207c60

Kotlin:       1.3.70
Groovy:       2.5.10
[...]

PR #4105, opened 2025-05-17
Fix constructor details to show description instead of brief in Javadoc

dmitrysulman:
When generating Javadoc, the **Constructor Detail** section displays only the brief constructor description (i.e., the first sentence) instead of the full description. See line 252 in `class.korte`:

https://github.com/Kotlin/dokka/blob/540fe084aadac0edb333350210ee22ee9901a3de/dokka-subprojects/plugin-javadoc/src/main/resources/views/class.korte#L245-L264

[A similar issue for method details](https://github.com/Kotlin/dokka/issues/1407) was fixed [long time ago](https://github.com/Kotlin/dokka/commit/aeb2014eee704be377c06205d16f60562d2a8cf1#diff-bff61c124f11db527d3aaa3f9bd879c40b8a57bffec72087e1ecec7414d5bdddR273)  in version `1.4.0`.

This PR fixes the issue by replacing `constructor.brief` with `constructor.description`.

### DIFF
--- a/dokka-subprojects/plugin-javadoc/src/main/resources/views/class.korte
+++ b/dokka-subprojects/plugin-javadoc/src/main/resources/views/class.korte
@@ -249,7 +249,7 @@
                                     <li class="blockList">
                                         <h4>{{ constructor.name }}</h4>
                                         <pre>{{ constructor.name }}({{ constructor.inlineParameters|raw }})</pre>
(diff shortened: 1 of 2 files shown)


whyoleg (approved):
Thanks!
I've additionally checked all other `brief` vs `description` usage in the template, and looks like there are no other issues like this.
I've started the test on CI, after they are finished, I will merge the PR

PR #4126, opened 2025-06-01
[K2] Fix `override` modifier for generic functions

### LINKED ISSUE #3848: [K2] Missed `override` modifier for generic functions
```kt
class A<T>  : C<T>()

open class C<T> : D<T> {
    override fun f(a: T) = 1
}

interface D<T> {
    fun f(a: T) = 0
}
```
The signature of `A.f` does not have `override` modifier.
K1:
image

K2:
image

Related #3677

vmishenev:
fix #3848

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
@@ -899,7 +899,7 @@ internal class DokkaSymbolVisitor(
      */
     private fun KaSession.isDokkaOverride(symbol: KaCallableSymbol): Boolean {
         fun KaCallableSymbol.isCompilerOverride() =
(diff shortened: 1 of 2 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -894,6 +894,37 @@ internal class DokkaSymbolVisitor(
 
 
     // ----------- Translators of modifiers ----------------------------------------------------------------------------
+    /**
+     * Dokka has its own conditions for the keyword `override`
+     */
+    private fun KaSession.isDokkaOverride(symbol: KaCallableSymbol): Boolean {
+        fun KaCallableSymbol.isCompilerOverride() =
+            if (this is KaPropertySymbol) isOverride else if (this is KaNamedFunctionSymbol) isOverride else throw IllegalStateException()
```
whyoleg: minor:
```suggestion
            if (this is KaPropertySymbol) isOverride else if (this is KaNamedFunctionSymbol) isOverride else error("Should be property or function but was '${this::class}'")
```

whyoleg (approved):

PR #4133, opened 2025-06-10
Enable DGPv2 by default

### LINKED ISSUE #4097: Enable DGPv2 by default
Stabilize the new Dokka Gradle Plugin introduced in [Dokka 2.0.0](https://github.com/Kotlin/dokka/releases/tag/v2.0.0) and enable it by default, with the ability to fallback to the old Dokka Gradle Plugin with a warning.

- [ ] Switch the [PluginMode.Default](https://github.com/Kotlin/dokka/blob/69a33323b06b962e2dc29a5fd930f88b8fd568b6/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt#L243) to `V2Enabled`
- [ ] Adapt log messages to log nothing when plugin mode is `V2Enabled`
- [ ] Adapt tests and IT where needed

adam-enko:
Enable DGPv2 by default.

- Remove the V2 logging message.
- Update tests. V1 tests needed `pluginMode=V1Enabled`.

fix #4097

### DIFF
--- a/dokka-integration-tests/gradle/projects/serialization/serialization.diff
+++ b/dokka-integration-tests/gradle/projects/serialization/serialization.diff
@@ -102,7 +102,6 @@ index b450fd5e..927b34b9 100644
 \ No newline at end of file
 +}
 diff --git a/buildSrc/src/main/kotlin/dokka-conventions.gradle.kts b/buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
-index d3013727..ed3311cd 100644
 index d3013727..b58ad96a 100644
 --- a/buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
 +++ b/buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
@@ -162,13 +161,6 @@ index 93513985..42b93ac2 100644
      dokkaSourceSets {
          configureEach {
              externalDocumentationLink {
-@@ -76,4 +81,4 @@ tasks.withType<DokkaTaskPartial>().named("dokkaHtmlPartial") {
-             }
-         }
-     }
--}
-\ No newline at end of file
-+}
 diff --git a/gradle.properties b/gradle.properties
 index deb1585f..774b0161 100644
 --- a/gradle.properties


dokka-integration-tests/gradle/src/testUiShowcaseProject/kotlin/UiShowcaseIntegrationTest.kt
```diff
@@ -24,6 +24,7 @@ class UiShowcaseIntegrationTest : AbstractGradleIntegrationTest(), TestOutputCop
         val result = createGradleRunner(
             buildVersions,
             ":dokkaGenerate",
+            dgpPluginMode = "V2Enabled",
```
mazhukinevgeniy: How would you describe the v2 test coverage now, is it safe to assume that it's tested enough to merge this?

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleProjectUtils.kt
```diff
@@ -5,14 +5,25 @@ package org.jetbrains.dokka.gradle.utils
 
 import org.gradle.api.Project
 
+fun Project.enableV1Plugin(
```
mazhukinevgeniy, line 8: What's the future of this, is it going to be removed together with V1 in Dokka 2.2.0? Does it mean that all tests that use this would be removed?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -240,7 +223,7 @@ constructor(
 
         companion object {
             /** The default value, if [Params.pluginMode] is not set. */
-            val Default: PluginMode = V1Enabled
+            val Default: PluginMode = V2EnabledWithHelpers
```
whyoleg, line 226: What is the reason not to use just `V2Enabled`?
I believe setting `V2Enabled` as `Default` will be less confusing for both new users and those who migrated, so that they will be able to also drop `V2Enabled` from their `gradle.properties`

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -131,7 +131,9 @@ constructor(
         when (mode) {
             V1Enabled -> logV1PluginMessage()
             V2EnabledWithHelpers -> logV2PluginMigrationMessage()
-            V2Enabled -> logV2PluginMessage()
+            V2Enabled -> {
+                // v2 is the default, don't log messages to avoid spamming users
```
whyoleg, line 135: Not sure where the logic should be, but what do you think about logging something like:
```
Dokka Gradle Plugin V2 is enabled by default starting from Dokka 2.1.0, and so it's possible to drop the following properties from `gradle.properties`
org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
org.jetbrains.dokka.experimental.gradle.pluginMode.nowarn=true
```
In case the user explicitly sets `V2Enabled`? (The question is related to the comment regarding the `Default` being `V2Enabled` below)

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleProjectUtils.kt
whyoleg: @adam-enko, just to refresh my memory - we don't have any intersected logic which is used both in V1 and V2 plugins, and so if we remove `classicTest` (and so this util will not be needed), the test coverage for V2 will stay the same?

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -56,6 +56,7 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
          * Defaults to [gradleLogLevel], so that the Dokka logs are always produced.
          */
         dokkaLogLevel: LogLevel = gradleLogLevel,
+        dgpPluginMode: String? = "V1Enabled",
```
whyoleg: I believe it makes sense to migrate all the current IT that uses `V1` to `V2` (in a separate PR).
I don't think it should be complex, as the build setup there is rather simple.

That could be, of course, postponed until the moment we decide to drop `V1`, but still, it's something to keep in mind

dokka-integration-tests/gradle/src/testUiShowcaseProject/kotlin/UiShowcaseIntegrationTest.kt
whyoleg:
```suggestion
            dgpPluginMode = null,
```
As the project is configured to use `V2` in `gradle.properties`.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -115,6 +116,11 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
                         add("-P${TestEnvironment.TRY_K2}=true")
                     }
 
+                    if (dgpPluginMode != null) {
+                        add("-Porg.jetbrains.dokka.experimental.gradle.pluginMode=${dgpPluginMode}")
```
whyoleg: On an additional thought, I think that it will be better to inline those into projects `gradle.properties`, as those projects are periodically opened separately from the IT to test some specific build failures.
So it would be nice if this continues to work without manual work

whyoleg (changes requested):

dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/GradleProjectUtils.kt
adam-enko: > What's the future of this, is it going to be removed together with V1 in Dokka 2.2.0? Does it mean that all tests that use this would be removed?

Yes, once we remove DGPv1 in 2.2.0 we can remove `enableV1Plugin()` (and also `enableV2Plugin()`).]

> we don't have any intersected logic which is used both in V1 and V2 plugins, and so if we remove classicTest (and so this util will not be needed), the test coverage for V2 will stay the same?

Correct, there's nothing shared between v1 and v2. So when we delete `classicMain/` we can also delete `classicTest/`. V2 test coverage will be unaffected.

dokka-integration-tests/gradle/src/testUiShowcaseProject/kotlin/UiShowcaseIntegrationTest.kt
adam-enko: tl;dr this specific line just keeps the status quo, and doesn't affect test coverage. v2 has a lot of tests, but they use separate test infrastructure.

Longer answer:

I think unit and integration test coverage for v1 and v2 are at the same level.

The integration tests for v1 use individual projects defined in https://github.com/Kotlin/dokka/tree/v2.0.0/dokka-integration-tests/gradle/projects.

The v2 integration test projects are dynamically created in each test, in https://github.com/Kotlin/dokka/tree/v2.0.0/dokka-runners/dokka-gradle-plugin/src/testFunctional.

v2 also tests against the example projects, and verifies they produce the exact same generated docs (committed into the repo https://github.com/Kotlin/dokka/tree/v2.0.0/dokka-integration-tests/gradle/src/testExampleProjects/expectedData/exampleProjects)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
adam-enko: The downside of going directly to V2Enabled is: if v1 users update to DGP 2.2.0 then suddenly their buildscripts would break (because of missing task accessors).

Using V2EnabledWithHelpers as the default would keep the buildscripts working, which would allow DGP to log the deprecation migration messages.

I'm not against changing the default to V2Enabled, if you think it's better.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
adam-enko: Yeah, that sounds like a good idea.

Although, we could wait until the next release, when v1 is removed, and log `"org.jetbrains.dokka.experimental.gradle.pluginMode is unused, please remove it"`?

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
adam-enko: https://github.com/Kotlin/dokka/issues/4138

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
whyoleg: Thanks, I agree! Let's leave it like this for Dokka 2.1.0

PR #4141, opened 2025-06-12
DGPv2: Add lifecycle tasks for generating formats

### LINKED ISSUE #3960: DGPv2: missing `dokkaGenerateHtml` and `dokkaGenerateJavadoc` lifecycle tasks
The DGPv2 migration docs reference tasks that don't exist

https://github.com/Kotlin/dokka/blob/fc574c855fb75a4fe30185aa9f12b74295b095d4/docs/topics/dokka-migration.md?plain=1#L405

The tasks should exist, and the tasks names are defined [here](https://github.com/Kotlin/dokka/blob/v2.0.0/dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/TaskNames.kt#L14), but for some unknown reason they are not registered as tasks (mistake while copying Dokkatoo? Or during another PR? Or were they intentionally removed?).

### Tasks

- [ ] Find out what tasks should be used instead of `dokkaGenerateHtml` and `dokkaGenerateJavadoc`. (Probably `dokkaGeneratePublicationHtml`?)
- [ ] The docs should be updated to remove the references. Instead replace them with the above.
- [ ] Investigate and determine if the tasks should be re-added. (I'm on the fence. I think it's confusing if there's too many tasks, but on the other hand `dokkaGeneratePublicationHtml` is a mouthful, and there's no obvious difference between it and `dokkaGenerateModuleHtml`, so a shorter lifecycle helper task could help.)

### LINKED ISSUE #4165: Move module/publication generation tasks to 'other' group
Dokka has tasks that generate partial modules and finalised publications.

These tasks are not intended to be invoked by plugin users. Instead the lifecycle tasks (see #3960) should be used instead.

The module/publication terminology is not clearly explained in the documentation, nor does it need to be. (For reference, a 'publication' is the final aggregated result of creating 1 or more partial modules.)

Currently the tasks are put in the 'dokka' task group, which highlights them. However, this can be confusing as it's not clear what these tasks do or the difference between them.

### Note

Sometimes the publication generation task is used to produce some other content, like a Javadoc JAR. The lifecycle task will not have any output files, so cannot be used for this. 

If we make the 'generatePublication' task less visible, it could be harder for users to find out how to create a Javadoc JAR.

I want to consider if there's anything we can do to mitigate this.

### Proposal

- [ ] The `dokkaGenerateModule*` and `dokkaGeneratePublication*` Tasks set the group as 'other', so they are less visible to users and are less likely to be seen and be confused with the lifecycle task.

adam-enko:
fix #3960
fix #4165

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/stringUtils.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/stringUtils.kt
@@ -4,6 +4,10 @@
 package org.jetbrains.dokka.gradle.utils
 
 
+fun String.normalizeLineEndings(): String =
+    lines().joinToString("\n")
+
+
 fun String.splitToPair(delimiter: String): Pair<String, String> =
     substringBefore(delimiter) to substringAfter(delimiter)
 
--- a/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaPluginFunctionalTest.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaPluginFunctionalTest.kt
@@ -38,9 +38,9 @@ class DokkaPluginFunctionalTest : FunSpec({
                 .addArguments("tasks", "--group=dokka", "--quiet")
                 .build {
                     withClue(output) {
-                        val dokkaTasks = output
-                            .substringAfter("Dokka tasks", "<missing start of group>")
-                            .substringBefore("\n\n", "<missing end of group>")
+                        val dokkaTasks = output.normalizeLineEndings()
+                            .substringAfter("Dokka tasks")
+                            .substringBefore("\n\n")
                             .lines()
                             .filter { " - " in it }
                             .associate { it.splitToPair(" - ") }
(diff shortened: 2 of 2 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaFormatTasks.kt
```diff
@@ -73,4 +75,14 @@ class DokkaFormatTasks(
 
             applyFormatSpecificConfiguration()
         }
+
+    /**
+     * The lifecycle task that will generate a Dokka Publication for the specific [formatName] output.
+     */
+    val lifecycleGenerate: TaskProvider<DefaultTask> =
+        project.tasks.register<DefaultTask>(taskNames.generate) {
+            description = "Generate Dokka $formatName publications"
```
whyoleg:
```suggestion
            description = "Generate Dokka $formatName publication"
```
?

dokka-runners/dokka-gradle-plugin/src/test/kotlin/formats/DokkaFormatTasksTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.jetbrains.dokka.gradle.utils.enableV2Plugin
+
+class DokkaFormatTasksTest : FunSpec({
+
+    context("verify format tasks") {
+        val project = projectWithDgpV2()
+
+        val dokkaExtension = project.extensions.getByType<DokkaExtension>()
+
+        val formatName = "testFormatName"
```
ALikhachev, line 27: nit: could this `formatName` variable used in assertions related to it? What do you think if it makes sense?

ALikhachev (approved):

dokka-runners/dokka-gradle-plugin/src/test/kotlin/formats/DokkaFormatTasksTest.kt
adam-enko: It could - but some of the expected strings are pascal case `TestFormatName`, so I think it looks a bit odd using `formatName` in some of the expected strings and not in others.

e.g.

```kotlin
dokkaTasks.lifecycleGenerate.name shouldBe "dokkaGenerateTestFormatName"
dokkaTasks.lifecycleGenerate.get().description shouldBe "Generate Dokka $formatName publication"
```

PR #4154, opened 2025-06-23
Fix param tag for type parameter on implicit primary constructor

### LINKED ISSUE #4145: [K2] Missing param tag for type parameter on implicit primary constructor
```
/** @param T a type */
class Foo<T>
```

With K1, the `documentation` on the `DFunction` for the implicit primary constructor of `Foo` is
`DocumentationNode(children=[Param(root=CustomDocTag(children=[P(children=[Text(body=a type, children=[], params={})], params={})], params={}, name=MARKDOWN_FILE), name=T)])`.

With K2, the `documentation` for the constructor `DFunction` is empty.

Possibly related to #3348?

vmishenev:
fix #4145

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KDocProvider.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KDocProvider.kt
@@ -89,6 +89,8 @@ internal fun KaSession.findKDoc(symbol: KaSymbol): KDocContent? {
         val containingClass = symbol.fakeOverrideOriginal.containingSymbol as? KaClassSymbol
         if (containingClass?.origin != KaSymbolOrigin.SOURCE) return null
         val kdoc = (containingClass.psi as? KtDeclaration)?.docComment ?: return null
+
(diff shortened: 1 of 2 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KDocProvider.kt
```diff
@@ -98,6 +98,7 @@ internal fun KaSession.findKDoc(symbol: KaSymbol): KDocContent? {
             val paramSections = kdoc.findSectionsContainingTag(KDocKnownTag.PARAM)
             return KDocContent(constructorSection, paramSections)
         }
+        return KDocContent(kdoc.getDefaultSection(), kdoc.getAllSections())
```
vmishenev, line 103: This simply duplicates the logic of IDE's `lookupOwnedKDoc`

whyoleg: You mean the one `lookupOwnedKDoc` which is located below in this file?
In any case, it would be nice to add a comment that the logic here should be duplicated to `lookupOwnedKDoc` below if we can't extract it

whyoleg (approved):

PR #4155, opened 2025-06-24
Update olderVersionsDir to be an optional input.

### LINKED ISSUE #3966: Gradle plugin v2 with versioning plugin requires olderVersionsDir to exist
**Describe the bug**
With the Dokka Gradle plugin v2, using the versioning plugin and configuring `olderVersionsDir` to a directory path that doesn't currently exist, the task `:dokkaGeneratePublicationHtml` fails with the error:

> A problem was found with the configuration of task ':dokkaGeneratePublicationHtml' (type 'DokkaGeneratePublicationTask').
>   - In plugin 'org.jetbrains.dokka.gradle.DokkaBasePlugin_Decorated' type 'org.jetbrains.dokka.gradle.tasks.DokkaGeneratePublicationTask' property 'generator.pluginsConfiguration.versioning$1.olderVersionsDir' specifies directory '.../docs/older' which doesn't exist.

This was not an error before the v2 Gradle plugin.

**Expected behaviour**
My project configures the `:dokkaGeneratePublicationHtml` task with a `doFirst` lambda that moves the older version directory to the configured path. So the files are present when the task runs, but not before it runs.

It should not be an error to configure this path to a currently non-existent directory. It's undesirable to have the empty directory exist in the file system when the code expects it to only exist temporarily while the API docs are being generated.

**To Reproduce**
Configure a project with the versioning plugin and set `olderVersionsDir` to a non-existent path. Then run `:dokkaGenerate` to see the error.

**Dokka configuration**

```Kotlin
dependencies {
}

        olderVersionsDir = projectDir.resolve("some/non-existent/dir")
        // ...
    }
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v8.11.1
- Dokka version: 2.0.0

**Additional context**
Reproducible with [my library, Kotbase](https://github.com/jeffdgr8/kotbase/pull/32) by removing [this line of code](https://github.com/jeffdgr8/kotbase/blob/af4ec9fd22012f51751d8e6d4e21d06d5c056699/build.gradle.kts#L30): `tempOlderDir.mkdir()` (check out commit [af4ec9f](https://github.com/jeffdgr8/kotbase/tree/af4ec9fd22012f51751d8e6d4e21d06d5c056699)).

adam-enko:
If an input is annotated with `@InputDirectory` then Gradle requires the directory exists on disk. Instead, use `@InputFiles`, which means Gradle will treat olderVersionsDir as optional. (see https://github.com/gradle/gradle/issues/2016#issuecomment-965126604)

Fix #3966

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/plugins/DokkaVersioningPluginParameters.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/plugins/DokkaVersioningPluginParameters.kt
@@ -16,8 +16,11 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity.RELATIVE
 import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
 import org.jetbrains.dokka.gradle.internal.addAll
+import org.jetbrains.dokka.gradle.internal.describeType
 import org.jetbrains.dokka.gradle.internal.putIfNotNull
 import javax.inject.Inject
+import kotlin.io.path.exists
+import kotlin.io.path.isDirectory
 
 
 /**
@@ -92,6 +95,8 @@ constructor(
     abstract val renderVersionsNavigationOnAllPages: Property<Boolean>
 
     override fun jsonEncode(): String {
+        validate()
+
         val versionsOrdering = versionsOrdering.orNull.orEmpty()
 
         return buildJsonObject {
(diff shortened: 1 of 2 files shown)


whyoleg (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/plugins/DokkaVersioningPluginParameters.kt
```diff
@@ -65,7 +68,7 @@ constructor(
      * For more information, see
      * [Directory structure](https://github.com/Kotlin/dokka/blob/master/plugins/versioning/README.md#directory-structure).
      */
-    @get:InputDirectory
+    @get:InputFiles
```
ALikhachev, line 74: Shall we add an explicit check that it is a directory?
From the mentioned Gradle issue: 
> What Gradle doesn't have is then some additional validation that if the file exists that it is actual a file/directory.

ALikhachev (changes requested):

ALikhachev (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/plugins/DokkaVersioningPluginParameters.kt
adam-enko: Thanks, good idea. I've added a check.

Issue #3979, opened 2024-12-29
Dokka GP v2 excludes the local module if submodules are added as dependencies


DanySK:
**Describe the bug**

I have a project with many subproject. The root project hosts the entrypoint of the entire application.
When generating documentation for the root project, if no subproject is added through `dependency`, only the local module's documentation is produced (as expected).
When, instead,
```
dependencies {
    subprojects.forEach { dokka(it) }
}
```
is used, the documentation for all modules **but** the current is produced.

If
```
dependencies {
    allprojects.forEach { dokka(it) }
}
```
is used instead, the local module is included, but the left menu has broken links

**Expected behaviour**

Either the local module is always included, or it can be included in the dependencies without breaking the website

**Screenshots**

This is the documentation generated with 
```
dependencies {
    allprojects.forEach { dokka(it) }
}
```

![Peek 2024-12-29 23-21](https://github.com/user-attachments/assets/8ee09515-5feb-4d06-a955-7244013d645a)

Links on the right work correctly, links in the left bar are broken.

**To Reproduce**

1. Create a mutlimodule project
2. Apply dokka
3. add sources to the root project and to the subprojects
4. Enable dokka v2
5. run `dokkaGenerate`

**Dokka configuration**

The following is my convention plugin for dokka. Looks complicated , but all the complexity is just linking source from GitHub and linking libraries from javadoc.io.

```Kotlin
import Util.currentCommitHash
import Util.fetchJavadocIOForDependency
import org.jetbrains.dokka.gradle.AbstractDokkaLeafTask
import org.jetbrains.dokka.gradle.tasks.DokkaBaseTask
import java.time.Duration

/*
 * Copyright (C) 2010-2024, Danilo Pianini and contributors
 * listed, for each module, in the respective subproject's build.gradle.kts file.
 *
 * This file is part of Alchemist, and is distributed under the terms of the
 * GNU General Public License, with a linking exception,
 * as described in the file LICENSE in the Alchemist distribution's top directory.
 */
plugins {
    id("org.jetbrains.dokka")
}

val minJavaVersion: String by properties

dokka {
    dokkaSourceSets.configureEach {
        enableKotlinStdLibDocumentationLink.set(true)
        enableJdkDocumentationLink.set(true)
        jdkVersion.set(minJavaVersion.toInt())
        skipDeprecated.set(false)
        skipEmptyPackages.set(true)
        /*
         * Source links to GitHub
         */
        listOf("kotlin", "java")
[...]

PR #4158, opened 2025-06-24
Use project name as default `modulePath` for root project to avoid clashing outputs during aggregation

adam-enko:
Check if the modulePath of modules are valid subdirectories inside the publication directory.

Add actionable suggestion to error message (update the `modulePath` to avoid clashing).

#3979


dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MultiModuleAggregationTest.kt
```diff
@@ -0,0 +1,149 @@
...
+            }
+            project.runner
+                .addArguments(
+                    ":dokkaGeneratePublicationHtml",
+                    "--stacktrace",
+                    "--rerun",
+                )
+                .forwardOutput()
+                .build {
+                    test("expect validation failure") {
```
ALikhachev:
```suggestion
                    test("expect generation success") {
```

ALikhachev (changes requested):
I'm feeling the explanation on how it helps with https://github.com/Kotlin/dokka/issues/3979

whyoleg (commented):
The checks seem fine, but can't we make the default behaviour better?
This issue could usually occur when we have both sources and aggregation in `rootProject`. So can't we just set `modulePath.convention(if(project == rootProject) rootProject.name else ...)` as I've suggested [here](https://jetbrains.slack.com/archives/C02H0GSH474/p1749198964859819?thread_ts=1749134470.606169&cid=C02H0GSH474) (internal link) - if not, could you please provide a reason why it might be an issue?

adam-enko:
> I'm feeling missing the explanation on how it helps with #3979

Good point, I'll explain:

Dokka creates _publications_. A publication contains either a single _module_, or multiple _modules_. If there are multiple modules Dokka will generate a 'landing page' for all modules in the root output directory.

The location of each module within the output directory is controlled by `modulePath` (so users can make sure each module has a stable location, even if the name or location of the source module changes).

The convention for `modulePath` is the subproject path, converted to a file path  (e.g. `:foo-subproject` results in `$outputDir/foo-subproject`, `:x:y:z-lib` to `$outputDir/x/y/z-lib`). This works fine if there's only one project (the root project), _or_ there are multiple subprojects. But, if a project aggregates _both_ the root project _and_ several subprojects then the root project will have a blank `modulePath`. This will conflict with the multi-module 'landing page'. Dokka will overwrite files, but not give any warning, resulting in an apparently broken publication - see #3979. 

~It is surprisingly complex to define a generic default that means there's no conflicts, or doesn't affect single-module publications.~ Update: I was misremembering, see comments below.

We discussed and thought it's a rare use case that can be manually solved by manually setting a distinct `modulePath` (Also, putting sources in the root project is against Gradle's best practices https://docs.gradle.org/9.0.0/userguide/best_practices_general.html#no_source_in_root)

So, this PR adds a check: When there are multiple modules `modulePath` cannot be blank, otherwise it will conflict with the landing page.

Additionally, I realised we should also check that `modulePath` cannot be a relative path that 'escapes' the output directory (e.g. `../../../foo-module`).

adam-enko:
> So can't we just set `modulePath.convention(if(project == rootProject) rootProject.name else ...)`

Yup, that works!

I mistakenly thought there was some issue with using the root project name, but I've tested and it works as expected. I forgot that `modulePath` is only relevant for aggregated publications, I thought it would also affect single-module publications.

PR #4159, opened 2025-06-25
Fix some Javadoc links with generic parameters

vmishenev:
In K1, we register IDE's service `JavadocManager`, we could do the same in K2:
`registerProjectService(JavadocManager::class.java, JavadocManagerImpl(project))`
However, it seems unnecessary, since the link is unresolved in K1 as well.

### DIFF
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiCommentsUtils.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiCommentsUtils.kt
@@ -44,6 +44,10 @@ internal fun PsiElement.referenceElementOrSelf(): PsiElement? =
 
 internal fun PsiDocTag.linkElement(): PsiElement? =
     dataElements.firstOrNull {
+        /**
+         * * According to the [com.intellij.java.syntax.parser.JavaDocParser],
+         * no other node is expected here for the reference (link)
+         */
         it.node.elementType == JavaDocElementType.DOC_REFERENCE_HOLDER ||
                 it.node.elementType == JavaDocElementType.DOC_METHOD_OR_FIELD_REF
     }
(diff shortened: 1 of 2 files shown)


vmishenev:
Notice

According to [the specification ](https://docs.oracle.com/en/java/javase/22/docs/specs/javadoc/doc-comment-spec.html#references), Javadoc links can have type parameters,
> Parameterized types may be used in the class and member parts of the reference

In Dokka, rendered links and labels have no type parameters at all.

whyoleg:
> In K1, we register IDE's service JavadocManager; we could do the same in K2

Is there a reason not to do it in K2? I mean, if we don't register it, we may throw NPE in some cases when parsing Javadoc. But if we register it, as far as I understand, we will just receive some empty results instead of NPE.

Or maybe I'm wrong, and with the changes in this PR, we will not fall back to using `JavadocManager` at all?

vmishenev:
> Or maybe I'm wrong, and with the changes in this PR, we will not fall back to using JavadocManager at all?

I mean this, at least in the current case. However, I do know why it was registered in K1. It just seems unnecessary.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiCommentsUtils.kt
```diff
@@ -44,7 +43,10 @@ internal fun PsiElement.referenceElementOrSelf(): PsiElement? =
     } else this
 
 internal fun PsiDocTag.linkElement(): PsiElement? =
-    valueElement ?: dataElements.firstOrNull { it !is PsiWhiteSpace }
+    dataElements.firstOrNull {
+        it.node.elementType == JavaDocElementType.DOC_REFERENCE_HOLDER ||
+                it.node.elementType == JavaDocElementType.DOC_METHOD_OR_FIELD_REF
```
whyoleg, line 52: Could you please add a comment on why those two specific types should be used here?
It's not really clear for me why only those two out of all `JavaDocElementType`s we should care about

PR #4167, opened 2025-07-10
Update Analysis API to 2.2.20-dev-6511

vmishenev:
Updating to the latest 2.2.20-dev-8169  is blocked by a compilation error (https://github.com/Kotlin/dokka/issues/4178).

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
@@ -52,15 +52,31 @@ internal fun KaSession.resolveKDocTextLinkToDRI(link: String, context: PsiElemen
  */
 internal fun KaSession.resolveKDocTextLinkToSymbol(link: String, context: PsiElement? = null): KaSymbol? {
     val kDocLink = createKDocLink(link, context)
-    return kDocLink?.let { analyze(kDocLink) { resolveToSymbol(it) } }
+    return kDocLink?.let {
+        /**
+         *  Get [KaSession] is associated with [a dangling module][org.jetbrains.kotlin.analysis.api.projectStructure.KaDanglingFileModule]
+         */
+        analyze(kDocLink) { resolveToSymbol(it) }
+    }
 }
 
 private fun KaSession.createKDocLink(link: String, context: PsiElement?): KDocLink? {
-    // see https://kotlin.github.io/analysis-api/in-memory-file-analysis.html
-    val psiFactory = if(context != null)  KtPsiFactory.contextual(context)  else {
+    /**
+     * Creates a dangling file stored in-memory
+     * A such file belongs to [a separate module][org.jetbrains.kotlin.analysis.api.projectStructure.KaDanglingFileModule]
+     *
+     * Additional information: https://kotlin.github.io/analysis-api/in-memory-file-analysis.html#stand-alone-file-analysis
+     * @see [org.jetbrains.kotlin.analysis.api.projectStructure.KaDanglingFileModule]
+     */
+    val psiFactory = if (context != null) KtPsiFactory.contextual(context) else {
         val currentModule: KaSourceModule = useSiteModule as? KaSourceModule
             ?: throw IllegalStateException("Resolving KDoc links can be done only in a source module, not $useSiteModule")
-        val randomKtSourceFile: KtFile = @OptIn(KaExperimentalApi::class) currentModule.psiRoots.filterIsInstance<KtFile>().firstOrNull() ?: return null
+
+        // To pass context (dependencies) from the current source module to a dangling module,
+        // a random source file is selected as the context file
+        val randomKtSourceFile: KtFile =
+            @OptIn(KaExperimentalApi::class) currentModule.psiRoots.filterIsInstance<KtFile>().firstOrNull()
+                ?: return null
 
         KtPsiFactory.contextual(randomKtSourceFile)
     }
@@ -70,7 +86,7 @@ private fun KaSession.createKDocLink(link: String, context: PsiElement?): KDocLi
     /**
     * [$link]
     */
- """.trimIndent()
+    """.trimIndent()
     ) as? KDoc
 
     return kDoc?.getDefaultSection()?.children?.filterIsInstance<KDocLink>()?.singleOrNull()
@@ -83,6 +99,13 @@ private fun KaSession.createKDocLink(link: String, context: PsiElement?): KDocLi
  * @return [DRI] or null if the [link] is unresolved
  */
 internal fun resolveKDocLinkToDRI(kDocLink: KDocLink): DRI? {
+    /**
+     * [kDocLink] can belong to [a dangling module][org.jetbrains.kotlin.analysis.api.projectStructure.KaDanglingFileModule]
+     * or [a source module][org.jetbrains.kotlin.analysis.api.projectStructure.KaSourceModule]
+     *
+     * Since [KaSession] is associated with a specific module,
+     * [analyze] should be called to get a corresponding instance of [KaSession]
+     */
     analyze(kDocLink) {
         val linkedSymbol = resolveToSymbol(kDocLink)
         return if (linkedSymbol == null) null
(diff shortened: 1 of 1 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -70,15 +82,17 @@ private fun KaSession.createKDocLink(link: String, context: PsiElement? = null):
...
+        val linkedSymbol = resolveToSymbol(kDocLink)
+        return if (linkedSymbol == null) null
+        else getDRIFromSymbol(linkedSymbol)
+    }
 }
 
 private fun KaSession.resolveToSymbol(kDocLink: KDocLink): KaSymbol? {
-    val lastNameSegment = kDocLink.children.filterIsInstance<KDocName>().lastOrNull()
-    return lastNameSegment?.mainReference?.resolveToSymbols()?.sortedWith(linkCandidatesComparator)?.firstOrNull()
+        val lastNameSegment = kDocLink.children.filterIsInstance<KDocName>().lastOrNull()
```
whyoleg: minor: redundant indentation

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -70,15 +82,17 @@ private fun KaSession.createKDocLink(link: String, context: PsiElement? = null):
  *
  * @return [DRI] or null if the [link] is unresolved
  */
-internal fun KaSession.resolveKDocLinkToDRI(link: KDocLink): DRI? {
-    val linkedSymbol = resolveToSymbol(link)
-    return if (linkedSymbol == null) null
-    else getDRIFromSymbol(linkedSymbol)
+internal fun resolveKDocLinkToDRI(kDocLink: KDocLink): DRI? {
+    analyze(kDocLink) {
```
whyoleg, line 109: As far as I understand, this will be called on each kdoc link resolution, am I right? Can it cause a degradation in performance?
Also, it's not clear for me, why we need an additional `analyze` call here. Could you please explain why having an already established session for the module is not enough?

gradle/libs.versions.toml
```diff
@@ -22,7 +22,7 @@ javaDiffUtils = "4.12"
 
 ## Analysis
 kotlin-compiler = "2.2.0-dev-8822"
```
whyoleg, line 24: shouldn't we keep them in sync?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -48,11 +52,19 @@ internal fun KaSession.resolveKDocTextLinkToDRI(link: String, context: PsiElemen
  */
 internal fun KaSession.resolveKDocTextLinkToSymbol(link: String, context: PsiElement? = null): KaSymbol? {
     val kDocLink = createKDocLink(link, context)
-    return kDocLink?.let { resolveToSymbol(it) }
+    return kDocLink?.let { analyze(kDocLink) { resolveToSymbol(it) } }
```
whyoleg: Same question here regarding additional `analyze` call as for the `resolveKDocLinkToDRI`

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -48,11 +52,19 @@ internal fun KaSession.resolveKDocTextLinkToDRI(link: String, context: PsiElemen
...
 }
 
-private fun KaSession.createKDocLink(link: String, context: PsiElement? = null): KDocLink? {
-    val psiFactory = context?.let { KtPsiFactory.contextual(it) } ?: KtPsiFactory(this.useSiteModule.project)
+private fun KaSession.createKDocLink(link: String, context: PsiElement?): KDocLink? {
+    // see https://kotlin.github.io/analysis-api/in-memory-file-analysis.html
```
whyoleg: TBH, the link doesn't help me to understand why we take a random file here. Maybe we can always enforce the `context` to be non-nullable on use-site and use ?
It would be also nice for me to understand why `KtPsiFactory(project)` doesn't work here

vmishenev:
> Also, I see that some tests failed on CI. Could you check?

The tests are flaky. Rerunning them made them green.

gradle/libs.versions.toml
vmishenev: It is unnecessary when the both versions use [the same version of  IntelliJ SDK](https://github.com/JetBrains/kotlin/blob/master/gradle/versions.properties#L1)

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
vmishenev: > Can it cause a degradation in performance?

It should not degrade performance since [creating KaSessions is cacheable](https://github.com/JetBrains/kotlin/blob/3a2a45e311b7408a5ef2bf1a8264e946bddb4e81/analysis/analysis-api-fir/src/org/jetbrains/kotlin/analysis/api/fir/KaFirSessionProvider.kt#L111).
 

> Could you please explain why having an already established session for the module is not enough?
 
 I left a comment in code

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
vmishenev: > Maybe we can always enforce the context to be non-nullable on use-site and use ?

As for me, it does not matter:
- this is internal
- there is a range of cases where we really don't have context: links in md files, `SampleProvider`, synthetic functions `valueOf`, `values` (have  no psi).

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
vmishenev: I left a comment in code

vmishenev:
I left some comments about the implementation of the AA in code. I hope they won't be outdated.  

>  There may be some YT issues.

there is a bunch of YT issues.
I think   [KT-71135](https://youtrack.jetbrains.com/issue/KT-71135/AA-exception-from-sealed-inheritors-checker-when-analyzeCopy) is a game changer.

> we have a source use-site module and want to analyze an element in a dangling file module. At first glance, this shouldn't be legal, as dangling files shouldn't be reachable from a use-site source module.

gradle/libs.versions.toml
whyoleg: I believe that at this point we can use the same version for both (as K1 will be deprecated anyway). Hope that different versions will not affect/complicate the fix for https://github.com/Kotlin/dokka/issues/4178 if the versions will be too different

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
whyoleg: Got it, yes, I mostly agree here, probably enforcing it will not have a lot of benefits.

whyoleg:
Thanks for all the explanations!

Issue #4178, opened 2025-07-23
Build `kotlin-analysis-symbols` module with the latest compiler version


vmishenev:
The Analysis API is built with the latest compiler version (or with a bootstrap version in the  `master` branch).
According to [the forward compatibility policy](https://kotlinlang.org/docs/kotlin-evolution-principles.html#evolving-the-binary-format):
> Preferably (but we can't guarantee it), the binary format is mostly forwards compatible with the next language release, but not later ones (in the cases when new features are not used, for example, 1.9 can understand most binaries from 2.0, but not 2.1).

So  `kotlin-analysis-symbols`  shoud be built  with the latest compiler version.

Although,   2.0.21 that is the latest Kotlin compiler version, which supports Kotlin Language/API version 1.4,  and the DGP needs to support Gradle 7, and [Gradle 7 is still stuck on 1.4](https://docs.gradle.org/current/userguide/compatibility.html)

~~Bumping the language level should be possible for the DGPv2, since the Dokka dependencies are in a separate classpath that's only used in the isolated worker, but it's not possible for the DGPv1.~~

Issue #2903, opened 2023-03-07
Use Gradle Worker API


vmishenev:
**_WIP_**

### Motivation
1. Currently, Dokka reloads a big classpath for every task. Also, it has reflection magic to support different versions of plugins.
2. Dokka works parallel badly. For example, the run time on on a [project with 100 tasks](https://github.com/Kotlin/dokka/blob/master/integration-tests/gradle/projects/it-sequential-tasks-execution-stress/build.gradle.kts) is 8 minutes.

### Proposal
[ Gradle Worker API](https://docs.gradle.org/current/userguide/worker_api.html) can help Dokka to avoid it. But we have 2 options for using it:
1. Use cached classpath and [`noIsolation`](https://docs.gradle.org/current/userguide/worker_api.html#changing_the_isolation_mode) mode (to keep a classpath in a static variable that is shared between tasks). This approach is used in [Kapt](https://github.com/JetBrains/kotlin/blob/36591ab8a07444d09deec9edd1e1931fefa6d67a/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptWithoutKotlincTask.kt#L294). 
2. Use [`processIsolation`](https://docs.gradle.org/current/userguide/worker_api.html#changing_the_isolation_mode) mode. Dokka task will be executed in worker daemon processes. The running processes with the same classpath can be reused for other tasks. In this case, the classpath is loaded once per process. 

We need to choose only one option. 
We already have prototypes. @aSemy created [one for the second option](https://github.com/Kotlin/dokka/pull/2740) and there is a prototype for the cached classpath [here](https://github.com/Kotlin/dokka/tree/feat/gradle_dokka_worker_daemon).

### Pros&cons
- From a performance point of view,   [`processIsolation`](https://docs.gradle.org/current/userguide/worker_api.html#changing_the_isolation_mode) requires time to run worker processes ([default is number of CPU processors](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_performance)) and load classpath in each process.  However, if daemons are already running, the time of Dokka executing is the same as for cached classpath. 
_**My observations:**_ 
Coroutines (precompilied): ~35s (cached classpath) vs ~66s (processIsolation with 8 workers)
[...]

PR #4182, opened 2025-07-30
Build `kotlin-analysis-symbols` module with the latest compiler version

vmishenev:
Here, `kotlin-analysis-symbols` is built with the latest compiler version (#4178), and Dokka uses the shadowed stdlib from `kotlin-analysis-symbols` or `kotlin-analysis-descriptors`.

### DIFF
--- a/dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
+++ b/dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
@@ -22,7 +22,11 @@ class MultiModule0IntegrationTest : AbstractGradleIntegrationTest() {
             buildVersions,
             ":moduleA:dokkaHtmlMultiModule",
             ":moduleA:dokkaGfmMultiModule",
-            ":moduleA:dokkaJekyllMultiModule"
+            ":moduleA:dokkaJekyllMultiModule",
+           jvmArgs = listOf(
+                "-Xmx1G", "-XX:MaxMetaspaceSize=800m",
+                "-XX:SoftRefLRUPolicyMSPerMB=10" // to free up the metaspace on JVM 8, see https://youtrack.jetbrains.com/issue/KT-55831/
+            )
         ).buildRelaxed()
 
         result.shouldHaveTask(":moduleA:dokkaHtmlMultiModule").shouldHaveOutcome(SUCCESS, FROM_CACHE)
--- a/dokka-integration-tests/gradle/src/testTemplateProjectCollector/kotlin/Collector0IntegrationTest.kt
+++ b/dokka-integration-tests/gradle/src/testTemplateProjectCollector/kotlin/Collector0IntegrationTest.kt
@@ -22,6 +22,10 @@ class Collector0IntegrationTest : AbstractGradleIntegrationTest() {
             ":moduleA:dokkaJavadocCollector",
             ":moduleA:dokkaGfmCollector",
             ":moduleA:dokkaJekyllCollector",
+            jvmArgs = listOf(
+                "-Xmx1G", "-XX:MaxMetaspaceSize=800m",
+                "-XX:SoftRefLRUPolicyMSPerMB=10" // to free up the metaspace on JVM 8, see https://youtrack.jetbrains.com/issue/KT-55831/
+            )
         ).buildRelaxed()
 
         result.shouldHaveTask(":moduleA:dokkaHtmlCollector").shouldHaveOutcome(SUCCESS, FROM_CACHE)


vmishenev:
In DGPv2, a shadowed `stdlib` (in this case v2.2.0) from `kotlin-analysis-symbol` should be used, but I encountered  https://github.com/gradle/gradle/issues/34442 

cc @adam-enko @whyoleg

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorLoggingTest.kt
```diff
@@ -25,6 +25,9 @@ class DokkaGeneratorLoggingTest : FunSpec({
                 .build {
                     output.invariantNewlines() shouldContain """
                         > Task :dokkaGenerateModuleHtml
```
vmishenev, line 27: The actual order can be different:
```
e: [:dokkaGenerateModuleHtml] test error message
w: [:dokkaGenerateModuleHtml] test warn message
> Task :dokkaGenerateModuleHtml
```

gradle/libs.versions.toml
```diff
@@ -7,22 +7,28 @@
...
+
 gradlePlugin-kotlin = "2.2.0"
 # See: https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin
 gradlePlugin-android = "7.3.1"
 
 kotlinx-coroutines = "1.7.3"
 kotlinx-collections-immutable = "0.3.6"
 kotlinx-serialization = "1.6.0"
-kotlinx-bcv = "0.13.2"
+kotlinx-bcv = "0.17.0"
```
whyoleg, line 17: BTW, we might as well migrate to the BCV shipped with KGP like in https://github.com/Kotlin/kotlinx-io/pull/466 (in a separate PR)

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```diff
@@ -16,11 +16,15 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 // and to do that we need to use old Kotlin version, which has support for such an old Kotlin version (e.g., 2.0.20)
 fun Project.configureGradleKotlinCompatibility() {
     if (!dokkaBuild.enforceGradleKotlinCompatibility.get()) return
-
+    val analysisK2Projects = listOf("analysis-kotlin-symbols")
```
whyoleg, line 23: I was almost going to write a comment asking to remove the probably unnecessary compiler arguments (`-Xsuppress-version-warnings` and `-Xskip-metadata-version-check`) as we now compile it with a compatible version, but!
I think that we could, instead of those changes, create `gradle.properties` inside `analysis-kotlin-symbols` and put `org.jetbrains.dokka.enforceGradleKotlinCompatibility=false` in there
This way, we will use the current KGP version (2.2.0), and we will have two fewer versions to maintain in the version catalog :)

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
```diff
@@ -23,6 +23,7 @@ class MultiModule0IntegrationTest : AbstractGradleIntegrationTest() {
             ":moduleA:dokkaHtmlMultiModule",
             ":moduleA:dokkaGfmMultiModule",
             ":moduleA:dokkaJekyllMultiModule",
+            jvmArgs = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=1400m")
```
whyoleg: What is the reason we need to adapt those? The logic inside DGPv1 hasn't changed, AFAIU.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
```diff
@@ -145,13 +145,9 @@ constructor(
         logger.info("[$path] running with workerIsolation $isolation")
         val workQueue = when (isolation) {
             is ClassLoaderIsolation ->
-                workers.classLoaderIsolation {
-                    classpath.from(runtimeClasspath)
-                }
-
+                workers.classLoaderIsolation()
```
whyoleg, line 148: I'm really curious, though: what are the benefits of workers at all now for Dokka?
- We do really have a need to run one task per Gradle project, so no parallelization benefits
  - Gradle will still parallelize tasks across projects with `--parallel` or `org.gradle.parallel=true`
- We don't need ClassLoader isolation, as we create our own

For example, [KSP](https://github.com/google/ksp/blob/4df562275f9a5bf74709919a747f43df7db688fc/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspAATask.kt#L146) uses `noIsolation` + custom ClassLoader caching. But at least they may run KSP tasks in parallel per target

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -787,7 +788,7 @@ internal class DokkaSymbolVisitor(
 
     private fun KtExpression.toDefaultValueExpression(): Expression? = when (node?.elementType) {
         KtNodeTypes.INTEGER_CONSTANT -> PsiLiteralUtil.parseLong(node?.text)?.let { IntegerConstant(it) }
-        KtNodeTypes.FLOAT_CONSTANT -> if (node?.text?.toLowerCase()?.endsWith('f') == true)
+        KtNodeTypes.FLOAT_CONSTANT -> if (node?.text?.lowercase(getDefault())?.endsWith('f') == true)
```
whyoleg:
```suggestion
        KtNodeTypes.FLOAT_CONSTANT -> if (node?.text?.lowercase()?.endsWith('f') == true)
```

Probably could be just?

gradle/libs.versions.toml
vmishenev: I thought about it as bcv 0.18 does not produce the same dump locally, but it says
>  it is recommended to continue using this plugin in production projects until KGP API stabilization.

It seems that kotlinx-io is part of dogfooding, so presumably, we can also participate.

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
vmishenev: The version of the AA has been changed. I suppose the size of using classes has increased.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
vmishenev: In theory we can change the default isolation mode to `noIsolation`, but I thing it is a big change for one PR and it requires some valuation. 

I'm not sure what you mean here: 
> We do really have a need to run one task per Gradle project, so no parallelization benefits
> Gradle will still parallelize tasks across projects with --parallel or org.gradle.parallel=true

~~Do you consider all tasks including non-Dokka ones!?~~

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
vmishenev: I thought `enforceGradleKotlinCompatibility` did something more.
Although `enforceGradleKotlinCompatibility=false` is less explicit, it seems okay.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
vmishenev: > We do really have a need to run one task per Gradle project, so no parallelization benefits
Gradle will still parallelize tasks across projects with --parallel or org.gradle.parallel=true

Is your question a kind of [this one](https://github.com/Kotlin/dokka/issues/2903#issuecomment-1649539284)?

Or do  you mean that we have only one item (unit) of work per Gradle task? 

>  But at least they may run KSP tasks in parallel per target

Doesn't `--parallel ` help them?

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
whyoleg: BTW: 2.3.0-dev-2555 contains a fix for https://youtrack.jetbrains.com/issue/KT-71706/Analysis-API-Standalone-StandaloneProjectFactory.createSearchScopeByLibraryRoots-creates-inefficient-file-based-search-scopes

It would be nice to update it separately and recheck how it affects the memory

gradle/libs.versions.toml
whyoleg: https://github.com/Kotlin/dokka/issues/4199

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
whyoleg: `enforceGradleKotlinCompatibility` is used only in that file, and it only configures how we compile Kotlin code.
There is additional setup needed only for the `dokka-gradle-plugin` module itself.

> Although enforceGradleKotlinCompatibility=false is less explicit, it seems okay.

I feel like it's what we do here — we remove compatibility enforcement of `analysis-kotlin-symbols` with Gradle/Kotlin runtimes of specific versions, and just compile it with the compiler of the current KGP version, like if it were in an ordinary Kotlin project :)
Yeah, it's less explicit about the intent, but the idea is the same.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
whyoleg: > In theory we can change the default isolation mode to noIsolation, but I thing it is a big change for one PR and it requires some valuation.

Yeah, let's not change it now, as it will also affect the fact that we will run Dokka inside the Gradle process (in the same way how `classLoader` isolation works).

> Is your question a kind of https://github.com/Kotlin/dokka/issues/2903#issuecomment-1649539284?
> Or do you mean that we have only one item (unit) of work per Gradle task?
> Doesn't --parallel  help them?

In short:
- `--parallel` or `org.gradle.parallel=true` only allows to [execute tasks from different subprojects in parallel](https://docs.gradle.org/current/userguide/performance.html#sec:enable_parallel_execution).
- workers, on the other hand, allow to [execute tasks inside one project in parallel](https://docs.gradle.org/current/userguide/worker_api.html), and, in addition, allow to start multiple parallel units of work per Gradle task. Also, it's parallel by default; no configuration flag is needed, and it doesn't even require the project to have `--parallel` enabled at all. Amount of parallelity in Gradle is controlled by `org.gradle.workers.max=N`, where N is the number of parallel units of work that could be done across the whole project. AFAIK, it includes both worker-related tasks (or specifically units of work) and just parallel tasks across projects.

Coming back to Dokka, for each Gradle Project, Dokka needs to run just one task. The `--parallel` flag nicely handles this.
While in the case of KSP, they do run multiple tasks **even inside one project** (per target), so workers are required to run them in parallel.

In any case, let's change nothing now.

whyoleg (approved):

whyoleg:
Please check CI :)

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
whyoleg: https://github.com/Kotlin/dokka/issues/4200

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
vmishenev: >  create gradle.properties inside analysis-kotlin-symbols and put org.jetbrains.dokka.enforceGradleKotlinCompatibility=false in there

The Gradle doesn't read `gradle.properties ` from subprojects(

https://github.com/gradle/gradle/issues/23572

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
vmishenev: _JIC I'm not sure that help us with the metaspace. The fix is for the heap._

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
whyoleg: Looks like the comment about why we don't configure compatibility for the K2 module is now lost - could you restore it?  :)

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```diff
@@ -15,7 +15,13 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 // We must use Kotlin language/api version 1.4 (DEPRECATED) to support Gradle 7,
 // and to do that we need to use old Kotlin version, which has support for such an old Kotlin version (e.g., 2.0.20)
 fun Project.configureGradleKotlinCompatibility() {
-    if (!dokkaBuild.enforceGradleKotlinCompatibility.get()) return
+    /**
+     * The AA is built with the latest compiler version (a bootstrap compiler)
+     * To be compatible with the AA, Dokka analysis should be compiled with approximately the same version
+     * See https://kotlinlang.org/docs/kotlin-evolution-principles.html#evolving-the-binary-format
+     */
+    val analysisK2Projects = listOf("analysis-kotlin-symbols")
```
adam-enko, line 23: I'd really rather avoid referring to subprojects by name in convention plugins. It makes the conventions more fragile, and it's harder to reason about.

Instead, I'd suggest updating DokkaBuildProperties to either add `btaCompilerVersion` and `btaLanguageVersion`, or a a single boolean on/off flag and zip it with the properties. Then the behaviour can be controlled per subproject more idiomatically.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/dokkaBootstrapFactory.kt
```diff
@@ -0,0 +1,54 @@
...
+    dokkaClasspath: Set<File>,
+    dokkaConfiguration: DokkaConfiguration,
+    logger: BiConsumer<String, String>
+) {
+    DokkaBootstrap(dokkaClasspath, DokkaBootstrapImpl::class).apply {
+        configure(dokkaConfiguration.toCompactJsonString(), logger)
+        val uncaughtExceptionHolder = AtomicReference<Throwable?>()
+        /**
+         * Run in a new thread to avoid memory leaks that are related to ThreadLocal (that keeps `URLCLassLoader`)
+         * Currently, all `ThreadLocal`s leaking are in the compiler/IDE codebase.
```
adam-enko: As written this comment implies it will cause leaks - is this re-write closer to what you mean?

```suggestion
         * Run in a new thread to avoid memory leaks that are related to ThreadLocal (that keeps `URLCLassLoader`)
         * Without a new thread all `ThreadLocal in the compiler/IDE codebase will leak.
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
adam-enko: Agreed, there's no benefit to classloader isolated workers now.

I'm a bit concerned about how this might affect performance, but given the default mode is process isolation it's probably fine. KGP has uses a buildservice to cache the classloaders, DGP should look into implementing something similar (eventually, not in this PR. I'm just mentioning it for future reference.) https://github.com/JetBrains/kotlin/blob/9a39a84c33c37436e4b0ae8b8102af02d27d2bf3/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/internal/ClassLoadersCachingBuildService.kt#L31-L34

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/dokkaBootstrapFactory.kt
```diff
@@ -0,0 +1,54 @@
...
+
+internal fun DokkaBootstrap(classpath: Set<File>, bootstrapClass: KClass<out DokkaBootstrap>): DokkaBootstrap {
+    val runtimeClassLoader = URLClassLoader(
+        classpath.map { it.toURI().toURL() }.toTypedArray(),
+        ClassLoader.getSystemClassLoader().parent
+    )
+
+    val runtimeClassloaderBootstrapClass = runtimeClassLoader.loadClass(bootstrapClass.qualifiedName)
+    val runtimeClassloaderBootstrapInstance = runtimeClassloaderBootstrapClass.constructors.first().newInstance()
+    return automagicTypedProxy(
```
adam-enko: Is it possible to avoid using `automagicTypedProxy`? I'd prefer to avoid bringing v1 code into v2.

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
vmishenev: > Instead, I'd suggest updating DokkaBuildProperties to either add btaCompilerVersion and btaLanguageVersion, or a a single boolean on/off flag and zip it with the properties. 

This is `enforceGradleKotlinCompatibility`, not an extra flag, but I cannot set it per project due to [gradle/gradle#23572](https://github.com/gradle/gradle/issues/23572). (see [the discussion](https://github.com/Kotlin/dokka/pull/4182#discussion_r2262939406) above)
Although, I could use a workaround from there. WDYT?

adam-enko:
CI OOMs during DGPv1, Jekyll, and Jackson?

```

00:03:38       * Exception is:
00:03:38       org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':moduleA:dokkaJekyllCollector'.
00:03:38         at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:142)
00:03:38         at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
00:03:38         at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:140)
00:03:38         at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:128)
00:03:38         at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)
00:03:38         at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
00:03:38         at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
00:03:38         at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
00:03:38         at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:57)
00:03:38         at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:204)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:199)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:53)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationExecutor.call(DefaultBuildOperationExecutor.java:73)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter.execute(EventFiringTaskExecuter.java:52)
00:03:38         at org.gradle.execution.plan.LocalTaskNodeExecutor.execute(LocalTaskNodeExecutor.java:69)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$InvokeNodeExecutorsAction.execute(DefaultTaskExecutionGraph.java:322)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$InvokeNodeExecutorsAction.execute(DefaultTaskExecutionGraph.java:309)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$BuildOperationAwareExecutionAction.execute(DefaultTaskExecutionGraph.java:302)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$BuildOperationAwareExecutionAction.execute(DefaultTaskExecutionGraph.java:288)
00:03:38         at org.gradle.execution.plan.DefaultPlanExecutor$ExecutorWorker.execute(DefaultPlanExecutor.java:462)
00:03:38         at org.gradle.execution.plan.DefaultPlanExecutor$ExecutorWorker.run(DefaultPlanExecutor.java:379)
00:03:38         at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
00:03:38         at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:49)
00:03:38       Caused by: org.gradle.internal.UncheckedException: java.lang.reflect.InvocationTargetException
00:03:38         at org.gradle.internal.UncheckedException.throwAsUncheckedException(UncheckedException.java:68)
00:03:38         at org.gradle.internal.UncheckedException.throwAsUncheckedException(UncheckedException.java:41)
00:03:38         at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:128)
00:03:38         at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:58)
00:03:38         at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:51)
00:03:38         at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:29)
00:03:38         at org.gradle.api.internal.tasks.execution.TaskExecution$3.run(TaskExecution.java:236)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.run(DefaultBuildOperationRunner.java:47)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:68)
00:03:38         at org.gradle.api.internal.tasks.execution.TaskExecution.executeAction(TaskExecution.java:221)
00:03:38         at org.gradle.api.internal.tasks.execution.TaskExecution.executeActions(TaskExecution.java:204)
00:03:38         at org.gradle.api.internal.tasks.execution.TaskExecution.executeWithPreviousOutputFiles(TaskExecution.java:187)
00:03:38         at org.gradle.api.internal.tasks.execution.TaskExecution.execute(TaskExecution.java:165)
00:03:38         at org.gradle.internal.execution.steps.ExecuteStep.executeInternal(ExecuteStep.java:89)
00:03:38         at org.gradle.internal.execution.steps.ExecuteStep.access$000(ExecuteStep.java:40)
00:03:38         at org.gradle.internal.execution.steps.ExecuteStep$1.call(ExecuteStep.java:53)
00:03:38         at org.gradle.internal.execution.steps.ExecuteStep$1.call(ExecuteStep.java:50)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:204)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:199)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:53)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationExecutor.call(DefaultBuildOperationExecutor.java:73)
00:03:38         at org.gradle.internal.execution.steps.ExecuteStep.execute(ExecuteStep.java:50)
00:03:38         at org.gradle.internal.execution.steps.ExecuteStep.execute(ExecuteStep.java:40)
00:03:38         at org.gradle.internal.execution.steps.RemovePreviousOutputsStep.execute(RemovePreviousOutputsStep.java:68)
00:03:38         at org.gradle.internal.execution.steps.RemovePreviousOutputsStep.execute(RemovePreviousOutputsStep.java:38)
00:03:38         at org.gradle.internal.execution.steps.CancelExecutionStep.execute(CancelExecutionStep.java:41)
00:03:38         at org.gradle.internal.execution.steps.TimeoutStep.executeWithoutTimeout(TimeoutStep.java:74)
00:03:38         at org.gradle.internal.execution.steps.TimeoutStep.execute(TimeoutStep.java:55)
00:03:38         at org.gradle.internal.execution.steps.CreateOutputsStep.execute(CreateOutputsStep.java:51)
00:03:38         at org.gradle.internal.execution.steps.CreateOutputsStep.execute(CreateOutputsStep.java:29)
00:03:38         at org.gradle.internal.execution.steps.CaptureStateAfterExecutionStep.executeDelegateBroadcastingChanges(CaptureStateAfterExecutionStep.java:124)
00:03:38         at org.gradle.internal.execution.steps.CaptureStateAfterExecutionStep.execute(CaptureStateAfterExecutionStep.java:80)
00:03:38         at org.gradle.internal.execution.steps.CaptureStateAfterExecutionStep.execute(CaptureStateAfterExecutionStep.java:58)
00:03:38         at org.gradle.internal.execution.steps.ResolveInputChangesStep.execute(ResolveInputChangesStep.java:48)
00:03:38         at org.gradle.internal.execution.steps.ResolveInputChangesStep.execute(ResolveInputChangesStep.java:36)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.executeWithoutCache(BuildCacheStep.java:181)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.executeAndStoreInCache(BuildCacheStep.java:154)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.lambda$executeWithCache$4(BuildCacheStep.java:121)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.lambda$executeWithCache$5(BuildCacheStep.java:121)
00:03:38         at org.gradle.internal.Try$Success.map(Try.java:164)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.executeWithCache(BuildCacheStep.java:81)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.lambda$execute$0(BuildCacheStep.java:70)
00:03:38         at org.gradle.internal.Either$Left.fold(Either.java:115)
00:03:38         at org.gradle.internal.execution.caching.CachingState.fold(CachingState.java:59)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.execute(BuildCacheStep.java:69)
00:03:38         at org.gradle.internal.execution.steps.BuildCacheStep.execute(BuildCacheStep.java:47)
00:03:38         at org.gradle.internal.execution.steps.StoreExecutionStateStep.execute(StoreExecutionStateStep.java:36)
00:03:38         at org.gradle.internal.execution.steps.StoreExecutionStateStep.execute(StoreExecutionStateStep.java:25)
00:03:38         at org.gradle.internal.execution.steps.RecordOutputsStep.execute(RecordOutputsStep.java:36)
00:03:38         at org.gradle.internal.execution.steps.RecordOutputsStep.execute(RecordOutputsStep.java:22)
00:03:38         at org.gradle.internal.execution.steps.SkipUpToDateStep.executeBecause(SkipUpToDateStep.java:110)
00:03:38         at org.gradle.internal.execution.steps.SkipUpToDateStep.lambda$execute$2(SkipUpToDateStep.java:56)
00:03:38         at org.gradle.internal.execution.steps.SkipUpToDateStep.execute(SkipUpToDateStep.java:56)
00:03:38         at org.gradle.internal.execution.steps.SkipUpToDateStep.execute(SkipUpToDateStep.java:38)
00:03:38         at org.gradle.internal.execution.steps.ResolveChangesStep.execute(ResolveChangesStep.java:73)
00:03:38         at org.gradle.internal.execution.steps.ResolveChangesStep.execute(ResolveChangesStep.java:44)
00:03:38         at org.gradle.internal.execution.steps.legacy.MarkSnapshottingInputsFinishedStep.execute(MarkSnapshottingInputsFinishedStep.java:37)
00:03:38         at org.gradle.internal.execution.steps.legacy.MarkSnapshottingInputsFinishedStep.execute(MarkSnapshottingInputsFinishedStep.java:27)
00:03:38         at org.gradle.internal.execution.steps.ResolveCachingStateStep.execute(ResolveCachingStateStep.java:89)
00:03:38         at org.gradle.internal.execution.steps.ResolveCachingStateStep.execute(ResolveCachingStateStep.java:50)
00:03:38         at org.gradle.internal.execution.steps.ValidateStep.execute(ValidateStep.java:102)
00:03:38         at org.gradle.internal.execution.steps.ValidateStep.execute(ValidateStep.java:57)
00:03:38         at org.gradle.internal.execution.steps.CaptureStateBeforeExecutionStep.execute(CaptureStateBeforeExecutionStep.java:76)
00:03:38         at org.gradle.internal.execution.steps.CaptureStateBeforeExecutionStep.execute(CaptureStateBeforeExecutionStep.java:50)
00:03:38         at org.gradle.internal.execution.steps.SkipEmptyWorkStep.executeWithNoEmptySources(SkipEmptyWorkStep.java:254)
00:03:38         at org.gradle.internal.execution.steps.SkipEmptyWorkStep.execute(SkipEmptyWorkStep.java:91)
00:03:38         at org.gradle.internal.execution.steps.SkipEmptyWorkStep.execute(SkipEmptyWorkStep.java:56)
00:03:38         at org.gradle.internal.execution.steps.RemoveUntrackedExecutionStateStep.execute(RemoveUntrackedExecutionStateStep.java:32)
00:03:38         at org.gradle.internal.execution.steps.RemoveUntrackedExecutionStateStep.execute(RemoveUntrackedExecutionStateStep.java:21)
00:03:38         at org.gradle.internal.execution.steps.legacy.MarkSnapshottingInputsStartedStep.execute(MarkSnapshottingInputsStartedStep.java:38)
00:03:38         at org.gradle.internal.execution.steps.LoadPreviousExecutionStateStep.execute(LoadPreviousExecutionStateStep.java:43)
00:03:38         at org.gradle.internal.execution.steps.LoadPreviousExecutionStateStep.execute(LoadPreviousExecutionStateStep.java:31)
00:03:38         at org.gradle.internal.execution.steps.AssignWorkspaceStep.lambda$execute$0(AssignWorkspaceStep.java:40)
00:03:38         at org.gradle.api.internal.tasks.execution.TaskExecution$4.withWorkspace(TaskExecution.java:281)
00:03:38         at org.gradle.internal.execution.steps.AssignWorkspaceStep.execute(AssignWorkspaceStep.java:40)
00:03:38         at org.gradle.internal.execution.steps.AssignWorkspaceStep.execute(AssignWorkspaceStep.java:30)
00:03:38         at org.gradle.internal.execution.steps.IdentityCacheStep.execute(IdentityCacheStep.java:37)
00:03:38         at org.gradle.internal.execution.steps.IdentityCacheStep.execute(IdentityCacheStep.java:27)
00:03:38         at org.gradle.internal.execution.steps.IdentifyStep.execute(IdentifyStep.java:44)
00:03:38         at org.gradle.internal.execution.steps.IdentifyStep.execute(IdentifyStep.java:33)
00:03:38         at org.gradle.internal.execution.impl.DefaultExecutionEngine$1.execute(DefaultExecutionEngine.java:76)
00:03:38         at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:139)
00:03:38         at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:128)
00:03:38         at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)
00:03:38         at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
00:03:38         at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
00:03:38         at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
00:03:38         at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:57)
00:03:38         at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:204)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:199)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:53)
00:03:38         at org.gradle.internal.operations.DefaultBuildOperationExecutor.call(DefaultBuildOperationExecutor.java:73)
00:03:38         at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter.execute(EventFiringTaskExecuter.java:52)
00:03:38         at org.gradle.execution.plan.LocalTaskNodeExecutor.execute(LocalTaskNodeExecutor.java:69)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$InvokeNodeExecutorsAction.execute(DefaultTaskExecutionGraph.java:322)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$InvokeNodeExecutorsAction.execute(DefaultTaskExecutionGraph.java:309)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$BuildOperationAwareExecutionAction.execute(DefaultTaskExecutionGraph.java:302)
00:03:38         at org.gradle.execution.taskgraph.DefaultTaskExecutionGraph$BuildOperationAwareExecutionAction.execute(DefaultTaskExecutionGraph.java:288)
00:03:38         at org.gradle.execution.plan.DefaultPlanExecutor$ExecutorWorker.execute(DefaultPlanExecutor.java:462)
00:03:38         at org.gradle.execution.plan.DefaultPlanExecutor$ExecutorWorker.run(DefaultPlanExecutor.java:379)
00:03:38         at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
00:03:38         at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:49)
00:03:38       Caused by: java.lang.reflect.InvocationTargetException
00:03:38         at org.jetbrains.dokka.gradle.internal.DokkaBootstrapFactoryKt$DokkaBootstrap$1.configure(dokkaBootstrapFactory.kt:36)
00:03:38         at org.jetbrains.dokka.gradle.internal.DokkaBootstrapFactoryKt.generateDocumentationViaDokkaBootstrap(dokkaBootstrapFactory.kt:60)
00:03:38         at org.jetbrains.dokka.gradle.AbstractDokkaTask.generateDocumentation$dokka_gradle_plugin(AbstractDokkaTask.kt:238)
00:03:38         at org.jetbrains.dokka.gradle.DokkaCollectorTask.generateDocumentation$dokka_gradle_plugin(DokkaCollectorTask.kt:18)
00:03:38         at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:125)
00:03:38         ... 117 more
00:03:38       Caused by: java.lang.OutOfMemoryError: Metaspace
00:03:38         at com.fasterxml.jackson.core.util.DefaultPrettyPrinter.<init>(DefaultPrettyPrinter.java:52)
00:03:38         at com.fasterxml.jackson.core.util.DefaultPrettyPrinter.<init>(DefaultPrettyPrinter.java:101)
00:03:38         at com.fasterxml.jackson.databind.SerializationConfig.<clinit>(SerializationConfig.java:38)
00:03:38         at com.fasterxml.jackson.databind.ObjectMapper.<init>(ObjectMapper.java:642)
00:03:38         at com.fasterxml.jackson.databind.ObjectMapper.<init>(ObjectMapper.java:567)
00:03:38         at com.fasterxml.jackson.databind.json.JsonMapper.<init>(JsonMapper.java:93)
00:03:38         at com.fasterxml.jackson.databind.json.JsonMapper.<init>(JsonMapper.java:89)
00:03:38         at com.fasterxml.jackson.databind.json.JsonMapper.builder(JsonMapper.java:114)
00:03:38         at com.fasterxml.jackson.module.kotlin.ExtensionsKt.jsonMapper(Extensions.kt:27)
00:03:38         at com.fasterxml.jackson.module.kotlin.ExtensionsKt.jacksonObjectMapper(Extensions.kt:32)
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/dokkaBootstrapFactory.kt
adam-enko: Thanks!

btw it'd be nice to use Java's ServiceLoader. KGP uses it. It's cleaner than the manual classloader hackery.

There's a good tutorial here https://www.baeldung.com/java-spi

If this PR works, let's merge it as-is, and investigate ServiceLoader later.

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
```diff
@@ -23,6 +23,7 @@ class MultiModule0IntegrationTest : AbstractGradleIntegrationTest() {
             ":moduleA:dokkaHtmlMultiModule",
             ":moduleA:dokkaGfmMultiModule",
             ":moduleA:dokkaJekyllMultiModule",
+            jvmArgs = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=1600m")
```
whyoleg: TBH, I don't really like that we increased this value from `-XX:MaxMetaspaceSize=1G` to 1.6 GB already.

WDYT about splitting the PR into two parts:

- re-setup analysis modules, building, and Dokka running
- updating the version of AA

So far, it's not clear what causes this **50% spike** in metaspace memory usage. We should probably not just ignore it by increasing the memory used.

adam-enko: I like the idea of splitting it up. We could also split the custom classloader into a separate PR (it'd help me be able to review only the parts that are related to DGP)

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
```diff
@@ -23,6 +23,10 @@ class MultiModule0IntegrationTest : AbstractGradleIntegrationTest() {
             ":moduleA:dokkaHtmlMultiModule",
             ":moduleA:dokkaGfmMultiModule",
             ":moduleA:dokkaJekyllMultiModule",
+           jvmArgs = listOf(
+                "-Xmx1G", "-XX:MaxMetaspaceSize=600m",
+                "-XX:SoftRefLRUPolicyMSPerMB=10" // to free up the metaspace on JVM 8, see https://youtrack.jetbrains.com/issue/KT-55831/
```
adam-enko, line 28: Interesting!

Do you think we should consider adding this arg to the process-isolated workers in DGPv2?

Also, perhaps `-XX:+AlwaysPreTouch` would help here too? (I'm not very confident it helps with DGPv2, but it doesn't seem to hurt.)

https://github.com/Kotlin/dokka/blob/71fa92f1ecf056486ca1cddb8b8f737b4288fc40/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt#L115-L123

vmishenev: > Do you think we should consider adding this arg to the process-isolated workers in DGPv2?

I prefer using a cached classloader if it is possible with Shared Build Services.
We can temporarily use this flag in DGPv2 until a cached classloader is implemented.
If it’s really needed, let’s do it in another PR.


> Also, perhaps -XX:+AlwaysPreTouch would help here too? 

Here "-XX:SoftRefLRUPolicyMSPerMB=10" help us very well.  I’d like to minimize the use of "magical" JVM flags.

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
vmishenev: It seems we have the same problem we encountered during the stress tests on **JVM 8** — https://github.com/Kotlin/dokka/pull/3802.
The change in the order of dependencies (and the stdlib version) in the classpath somehow affects GC that leads to flaky integration tests.
Nevertheless, there is no memory leak. 


I also noticed one misuse in Dokka runtime classpath.
Actually it has two different versions of stdlib: a shadowed version in `kotlin-analysis-*` and one from other dependencies.
For example, it has 2 different versions: 2.2 (in`kotlin-analysis-*` ) and 2.0.21
Version aligning does not work for the shadowed stdlib. 
image
I think it causes the classloader/GC to perform additional work.

vmishenev: JIC aligning the stdlib version with 2.2.0 does NOT help to avoid increasing (aggressive GC) metaspace in the tests.
 I used this hack for an  experiment - https://github.com/Kotlin/dokka/pull/4208/commits/b7fb1d2e5a2b9a44ac1986582bfabc59e0ecdf34 
(the same works for the `base-plugin`). Notice: at compile time the old stdlib is used. 
[The build scan ](https://scans.gradle.com/s/v2comr7axyph6/dependencies?toggled=W1swXSxbMCwxXSxbMCwxLFs1XV0sWzAsMSxbM11dXQ)

 An alternative solution is to build `base-plugin` with Kotlin 2.2.0 so it will have stdlib 2.2.0. In this case, we don't need to change the DGPv1. Also, we can consider to exclude the shadowed stdlib from `kotlin-analysis-*`

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
whyoleg: Both workaround and explicit project specification here are fine for me, as in reality, we don't need a general solution here, but only for this specific module, because of a specific issue

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
whyoleg: > I also noticed one misuse in Dokka runtime classpath.
Actually it has two different versions of stdlib: a shadowed version in kotlin-analysis-* and one from other dependencies.

But I believe that it should not cause any issues, as we load stdlib classes from `kotlin-analysis-*` artefacts first?
I mean, we had the same situation previously, as shadowed `kotlin-analysis-*` always contains `kotlin-stdlib` coming from `kotlin-compiler`
Or am I missing something?

> An alternative solution is to build base-plugin with Kotlin 2.2.0 so it will have stdlib 2.2.0.

I'm not sure I understand how this will help, as we will still have multiple stdlib versions (one from `kotlin-compiler-k2` and one from `base`), though they will be closer to each other (`2.2.0-dev`/`2.3.0-dev` and `2.2.0`).

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
whyoleg: > Do you think we should consider adding this arg to the process-isolated workers in DGPv2?

I think we previously decided that it's fine to use this option in tests only in https://github.com/Kotlin/dokka/pull/3802, so I suggest sticking with it.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
whyoleg: Extracted into https://github.com/Kotlin/dokka/issues/4215

build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
adam-enko: Let's leave it as is. If I have time I'll come back and tinker with it later.

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
adam-enko: We discussed briefly in KBT and we think it could be helpful, but it's a sensitive flag so let's leave it for now.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/dokkaBootstrapFactory.kt
vmishenev: JIC #4216

dokka-integration-tests/gradle/src/testTemplateProjectMultimodule0/kotlin/MultiModule0IntegrationTest.kt
vmishenev: > I mean, we had the same situation previously, as shadowed kotlin-analysis-* always contains kotlin-stdlib coming from kotlin-compiler

Sorry, I forgot  that `kotlin-compiler` already includes a shadowed stdlib, although  `kotlin-compiler-for-ide` does not have. 
So, for now, there is nothing we can do about it.

> But I believe that it should not cause any issues, as we load stdlib classes from kotlin-analysis-* artefacts first?

Yes, in theory, it could only lead to performance issues, but I’m not sure.

PR #4186, opened 2025-08-05
Suppress deprecation of AbstractDokkaTask in test

adam-enko:


adam-enko:
Closing: instead I'll fix this in https://github.com/Kotlin/dokka/pull/4187

PR #4187, opened 2025-08-06
Throw error when running V1 migration helper tasks

### LINKED ISSUE #4164: Throw error when v2 migration tasks are run when v2 is enabled
The next Dokka release will enable DGP v2 by default, but with v2 migration helpers enabled.

This could introduce confusion if DGP is automatically updated. The migration tasks will run, but not do anything. It's also likely to cause confusion if the tasks are run on CI.

To prevent mistakes, the migration tasks [registered here](https://github.com/Kotlin/dokka/blob/86e485421f320f2e28abcbba3574785b1a602f7f/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt) should throw an error when they are run.

No action is necessary if v1 is enabled, the migration tasks won't be registered.

adam-enko:
Fix #4164

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
@@ -52,16 +52,16 @@ private fun configureDokkaTaskConventions(project: Project) {
                     appendLine("Cannot run Dokka V1 tasks when V2 mode is enabled.")
                     appendLine("Dokka Gradle plugin V1 mode is deprecated, and scheduled to be removed in Dokka v2.2.0.")
                     appendLine("To finish migrating to V2 mode, please check the migration guide https://kotl.in/dokka-gradle-migration")
-                    append("Suggestion: ")
                     when {
                         "html" in this@task.name.lowercase() ->
-                            appendLine("Use `dokkaGenerate` or `dokkaGenerateHtml` tasks instead.")
+                            appendLine("Suggestion: Use `dokkaGenerate` or `dokkaGenerateHtml` tasks instead.")
 
                         "javadoc" in this@task.name.lowercase() ->
-                            appendLine("Use `dokkaGenerate` or `dokkaGenerateJavadoc` tasks instead.")
+                            appendLine("Suggestion: Use `dokkaGenerate` or `dokkaGenerateJavadoc` tasks instead.")
 
-                        else ->
-                            appendLine("Use `dokkaGenerate` task instead.")
+                        else -> {
+                            // Don't suggest alternative tasks for GFM and Jekyll, since DGPv2 does not support these formats
+                        }
                     }
                 }
             )
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
@@ -38,6 +38,7 @@ private fun configureDokkaTaskConventions(project: Project) {
         // The tasks are non-operable and should be hidden, to help nudge users to the DGPv2 tasks.
         // Setting tasks with group null will hide it when running `gradle tasks`,
         // and put it in the 'other' group in IntelliJ (which effectively hides it).
+        @Suppress("UsePropertyAccessSyntax") // property-access syntax doesn't accept `null`
         setGroup(null)
 
         notCompatibleWithConfigurationCache("Dokka V1 tasks use deprecated Gradle features. Please migrate to Dokka Plugin V2, which fully supports Configuration Cache. See https://kotl.in/dokka-gradle-migration")
@@ -144,7 +145,7 @@ private fun Configuration.deprecate(replaceWith: String) {
         if (this is DeprecatableConfiguration) {
             addDeclarationAlternatives(replaceWith)
         }
-    } catch (e: Throwable) {
+    } catch (_: Throwable) {
         // Deprecating configurations is an internal Gradle feature, so it might be unstable.
         // Because these migration helpers are temporary, just ignore all errors.
     }
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
@@ -33,7 +33,7 @@ internal fun addV2MigrationHelpers(
 }
 
 private fun configureDokkaTaskConventions(project: Project) {
-    project.tasks.withType<@Suppress("DEPRECATION") AbstractDokkaTask>().configureEach {
+    project.tasks.withType<@Suppress("DEPRECATION") AbstractDokkaTask>().configureEach task@{
         // The DGPv1 tasks are only present to prevent buildscripts with references to them from breaking.
         // The tasks are non-operable and should be hidden, to help nudge users to the DGPv2 tasks.
         // Setting tasks with group null will hide it when running `gradle tasks`,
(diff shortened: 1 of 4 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
```diff
@@ -38,13 +38,24 @@ private fun configureDokkaTaskConventions(project: Project) {
...
+        // must have an output directory, else the doFirst won't run
+        outputDirectory.set(temporaryDir)
+
+        doFirst("Disable Dokka V1 task") {
+            throw DokkaV1TaskDisabledException()
+        }
     }
 }
 
+internal class DokkaV1TaskDisabledException : UnsupportedOperationException(
```
whyoleg: Do we really need a separate exception type here? I feel like throwing `UnsupportedOperationException` will be enough

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaV1TaskDisabledTest.kt
```diff
@@ -21,12 +22,14 @@ class DokkaV1TaskDisabledTest : FunSpec({
             }
         }
 
-        test("v1 tasks should be skipped when v2 is enabled") {
+        test("v1 tasks should be fail (because v2 is enabled by default)") {
```
whyoleg:
```suggestion
        test("v1 tasks should fail (because v2 is enabled by default)") {
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
```diff
@@ -38,13 +38,24 @@ private fun configureDokkaTaskConventions(project: Project) {
...
+    "Cannot run Dokka V1 task when V2 mode is enabled. $DOKKA_V1_DEPRECATION_MESSAGE"
```
whyoleg: BTW, maybe we can mention the correct task name for the HTML format directly in the message?
Maybe just add something like: `Call 'dokkaGenerate' or 'dokkaGenerateHtml' instead`

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
```diff
@@ -38,13 +38,24 @@ private fun configureDokkaTaskConventions(project: Project) {
...
-        onlyIf("Dokka V1 tasks are disabled due to the V2 flag being enabled") { false }
-        enabled = false
-        notCompatibleWithConfigurationCache("Dokka V1 tasks use deprecated Gradle features. Please migrate to Dokka Plugin V2, which fully supports Configuration Cache.")
+
+        notCompatibleWithConfigurationCache("Dokka V1 tasks use deprecated Gradle features. Please migrate to Dokka Plugin V2, which fully supports Configuration Cache. See https://kotl.in/dokka-gradle-migration")
+
+        // must have an output directory, else the doFirst won't run
+        outputDirectory.set(temporaryDir)
+
+        doFirst("Disable Dokka V1 task") {
```
whyoleg, line 49: Minor: The tasks are no longer disabled but **unsupported**, so it might make sense to rephrase.

BTW, is that message even logged anywhere by Gradle?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
adam-enko: Not really, I just wanted to make it as blindingly obvious as possible that the task cannot run. A custom exception here is like a human-readable error code, and I think it would help users find the cause if searching online.

I'm not set on it - I can remove it if you'd like?

whyoleg: For me, it's fine both ways, it's just that:
* User can see exception class **only** when running with `--stactrace`
* The message clearly says what's wrong and gives the link to the migration guide

In any case, let's leave it there and see, it doesn't hurt to have it :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
adam-enko: Hmmm very good idea. I think it'd help users a lot. However, it's a bit tricky to map the V1 tasks onto V2 tasks. Also, the migration-helper tasks are registered for formats that aren't present in V2 (Jekyll, GFM).

whyoleg: I believe that we need to concentrate here on HTML users (the majority).
If the task name contains `html`:
```
Dokka Gradle plugin V2 mode is enabled, so calling V1 tasks is not possible anymore.
Use `dokkaGenerate` or `dokkaGenerateHtml` tasks instead of `dokkaHtml`/`dokkaHtmlMultiModule`.
For other details, please check the migration guide https://kotl.in/dokka-gradle-migration
Dokka Gradle plugin V1 mode tasks and classes are deprecated, and scheduled to be removed in Dokka v2.2.0
```
Otherwise, just:
```
Dokka Gradle Plugin V2 mode is enabled, so calling V1 tasks is not possible anymore.
Please check the migration guide https://kotl.in/dokka-gradle-migration
Dokka Gradle plugin V1 mode tasks and classes are deprecated, and scheduled to be removed in Dokka v2.2.0
```

Something like this

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
adam-enko: When the mode is `V2EnabledWithHelpers`, the tasks are disabled, no? I don't see much difference between 'unsupported' and 'disabled'.

The messages are logged at `--info` level, they're more for debugging than anything else.

whyoleg: > When the mode is V2EnabledWithHelpers, the tasks are disabled, no?

They throw an `unsupported` exception trying to run them :)

Task is disabled if it's not executed, e.g., `onlyIf { false }`, `enabled=false`, but in this case, it throws an exception

In any case, that's very minor, so never mind

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
adam-enko: I've updated the message, please take a look

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
```diff
@@ -47,14 +47,31 @@ private fun configureDokkaTaskConventions(project: Project) {
...
+                    append("Suggestion: ")
+                    when {
+                        "html" in this@task.name.lowercase() ->
+                            appendLine("Use `dokkaGenerate` or `dokkaGenerateHtml` tasks instead.")
+
+                        "javadoc" in this@task.name.lowercase() ->
+                            appendLine("Use `dokkaGenerate` or `dokkaGenerateJavadoc` tasks instead.")
+
+                        else ->
+                            appendLine("Use `dokkaGenerate` task instead.")
```
whyoleg: Minor: There is probably no reason to suggest the `dokkaGenerate` task, as it will do nothing. I mean, we don't support other formats, so users will need to migrate from them :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
whyoleg: Nice, thanks!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/v2MigrationUtils.kt
adam-enko: Done, please take a look

whyoleg: Perfect, thanks!

PR #4188, opened 2025-08-06
Migrate to `redirector.kotlinlang.org`

whyoleg:
Additionally, align cache redirector usages to be similar to kotlin.git usages

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -6,9 +6,7 @@ rootProject.name = "build-logic"
 
 pluginManagement {
     repositories {
-        maven("https://cache-redirector.jetbrains.com/repo.maven.apache.org/maven2") {
-            name = "MavenCentral-JBCache"
-        }
+        mavenCentral { setUrl("https://cache-redirector.jetbrains.com/maven-central") }
         maven("https://cache-redirector.jetbrains.com/plugins.gradle.org/m2") {
             name = "GradlePluginPortal-JBCache"
         }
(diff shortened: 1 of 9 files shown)


build-logic/settings.gradle.kts
```diff
@@ -6,9 +6,7 @@ rootProject.name = "build-logic"
 
 pluginManagement {
     repositories {
-        maven("https://cache-redirector.jetbrains.com/repo.maven.apache.org/maven2") {
-            name = "MavenCentral-JBCache"
-        }
+        mavenCentral { setUrl("https://cache-redirector.jetbrains.com/maven-central") }
```
adam-enko: Could we keep the repository name? If there are problems with the repo Gradle prints a nicer error message which helps track down the cause.

whyoleg: Sure, I can get them back. TBH, I've never seen those names appear anywhere, and `kotlin.git` doesn't provide them too :)
E.g., if the dependency can't be resolved, it will still just print the URL:
```
A problem occurred configuring project ':dokka-gradle-plugin'.
> Could not resolve all dependencies for configuration 'classpath'.
   > Could not find com.example.xxx:xxx:1.
     Searched in the following locations:
       - https://cache-redirector.jetbrains.com/maven-central/com/example/xxx/xxx/1/xxx-1.pom
       - https://cache-redirector.jetbrains.com/plugins.gradle.org/m2/com/example/xxx/xxx/1/xxx-1.pom
     Required by:
         project :dokka-gradle-plugin > project :build-logic
```

adam-enko: Ah yeah, you're right. It's not shown in the console output. I was thinking of the build scan report 

image

whyoleg: Got it, thanks!  Now I see how that might be useful
For this PR, it shows like [this](https://ge.jetbrains.com/s/pzwe3nvuczkvk/dependencies/repositories?repoToggled=W1sxXSxbOF1d):
Image

I will then bring those names back, just for consistency, so that we know that we cache-redirector everywhere.
As well as update the name of Node/Yarn repositories to not include the URL, as it's redundant there

Thanks for the explanation!

Issue #4195, opened 2025-08-09
Dokka plugin emits warnings about Unsafe on JDK 24


rnett:
**Describe the bug**

I'm using the dokka V2 plugin, and whenever I generate docs, I see a whole bunch of these warnings:
```
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by com.intellij.util.containers.Unsafe (file:/C:/Users/rnett/.gradle/caches/modules-2/files-2.1/org.jetbrains.dokka/analysis-kotlin-descriptors/2.0.0/9da12c077375be02fe09e636ae83411a4696c733/analysis-kotlin-descriptors-2.0.0.jar)
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by com.intellij.util.containers.Unsafe (file:/C:/Users/rnett/.gradle/caches/modules-2/files-2.1/org.jetbrains.dokka/analysis-kotlin-descriptors/2.0.0/9da12c077375be02fe09e636ae83411a4696c733/analysis-kotlin-descriptors-2.0.0.jar)

WARNING: Please consider reporting this to the maintainers of class com.intellij.util.containers.Unsafe
WARNING: Please consider reporting this to the maintainers of class com.intellij.util.containers.Unsafe
```

I'm using JDK 24 to run Gradle.

**Expected behaviour**

Dokka does not emit warnings about the use of Unsafe.

**To Reproduce**

Use Dokka on a project, run it with JDK 24.

**Additional context**

This may be a problem with the Kotlin compiler rather than Dokka.  Even if it is, it would be good for Dokka to provide an option to set the JDK toolchain used to run Dokka's analysis, so that when this becomes an error in future JDKs it doesn't hold the whole build back.

whyoleg:
Hey! Thanks for the report!
Unfortunately, at this moment, the only possible solution is to pass `--sun-misc-unsafe-memory-access=allow` for the worker. 

> This may be a problem with the Kotlin compiler rather than Dokka

Yes, it's a problem inside the kotlin compiler and its dependencies (specifically IntelliJ Platform), and so it will take some time to resolve those...

> it would be good for Dokka to provide an option to set the JDK toolchain used to run Dokka's analysis, so that when this becomes an error in future JDKs it doesn't hold the whole build back.

[...]

Issue #4194, opened 2025-08-07
org.jetbrains.dokka:dokka-maven-plugin not compatible with JDK 25


berezovskyi:
**Describe the bug**
The invocation of the dokka-maven-plugin fails under JDK 25-ea with the error message "25-beta"

**Expected behaviour**
The plugin works under JDK 25-ea or fails with the actual reason due to e.g. breaking change.

**Screenshots**
https://github.com/eclipse-lyo/lyo/actions/runs/16811117667/job/47616418179?pr=704#step:5:10012

**To Reproduce**
Steps to reproduce the behavior, ideally with an example project

**Dokka configuration**
Configuration of dokka used to reproduce the bug

```xml
<plugin>
  <groupId>org.jetbrains.dokka</groupId>
  <artifactId>dokka-maven-plugin</artifactId>
  <version>${v.dokka}</version>
  <executions>
    <execution>
      <phase>generate-sources</phase>
      <goals>
        <goal>javadocJar</goal>
      </goals>
    </execution>
  </executions>
  <configuration>
    <dokkaPlugins>
      <plugin>
        <groupId>org.jetbrains.dokka</groupId>
        <artifactId>kotlin-as-java-plugin</artifactId>
        <version>${v.dokka}</version>
      </plugin>
    </dokkaPlugins>
  </configuration>
</plugin>
```

https://github.com/eclipse-lyo/lyo/blob/4d58cde01583f4e4fefe1b485360646bf1f5449d/trs/client/trs-client/pom.xml#L34C1-L55C16

**Installation**
- Operating system:Linux
- Build tool: Maven 3.9.11
- Dokka version: 2.0.0

**Additional context**
Would be good to test all parts with JDK 25-ea before the LTS release next month.

whyoleg:
Hey! Thanks for the report!
Your issue is similar to "[KT-77518](https://youtrack.jetbrains.com/issue/KT-77518. Kotlin CLI: JavaVersion.parse() fails on 25-beta", as Dokka depends on the Kotlin compiler. It works fine, though, with [Oracle OpenJDK](https://jdk.java.net/25/).

After some tests, I can say that Dokka with K1 analysis works fine even now with JDK 25, and K2 analysis will work fine after #4202.

berezovskyi:
Thanks @whyoleg for looking into this. Just re-ran my build with the Oracle OpenJDK 25-EA build but it still fails, I guess due to #4202 - I will wait.

Build info for Oracle 25-EA:

[...]

PR #4202, opened 2025-08-12
Prepares Dokka for support of JDK 25

whyoleg:
This fixes https://github.com/Kotlin/dokka/issues/4194, but because of "[KT-77518](https://youtrack.jetbrains.com/issue/KT-77518) Kotlin CLI: JavaVersion.parse() fails on 25-beta," we can't properly test the changes on CI. As all GitHub Actions Java distributions use `25-java`. Still, I was able to test it locally with [JDK 25-ea from Oracle](https://jdk.java.net/25/)

This fixes https://github.com/Kotlin/dokka/issues/4195, but it's really just a workaround until we update to the latest IntelliJ platform, which doesn't use `Unsafe`. For this, the Kotlin compiler should be updated first.
So far, "[KT-74494](https://youtrack.jetbrains.com/issue/KT-74494) Update IntelliJ platform to 251.14649.2 in Kotlin" exists, but I will follow up on whether it will be enough or if we will need something newer.

JDK 25 will be released on September 16.

Note: New method `getChars` was [added to the CharSequence](https://cr.openjdk.org/~prr/8357176/api/java.base/java/lang/CharSequence.html#getChars(int,int,char%5B%5D,int)) in JDK 25 - will follow up with compiler team, about it

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -114,6 +114,10 @@ constructor(
                 debug.convention(false)
                 jvmArgs.convention(
                     listOf(
+                        // https://openjdk.org/jeps/498
+                        // suppresses: sun.misc.Unsafe::objectFieldOffset has been called by com.intellij.util.containers.Unsafe
+                        // requires IntelliJ platform update to resolve the issue
+                        "--sun-misc-unsafe-memory-access=allow",
                         //"-XX:MaxMetaspaceSize=512m",
                         "-XX:+HeapDumpOnOutOfMemoryError",
                         "-XX:+AlwaysPreTouch", // https://github.com/gradle/gradle/issues/3093#issuecomment-387259298
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -5,6 +5,7 @@ package org.jetbrains.dokka.gradle
 
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
+import org.gradle.api.JavaVersion
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Plugin
 import org.gradle.api.Project
@@ -114,16 +115,21 @@ constructor(
                 debug.convention(false)
                 jvmArgs.convention(
                     listOf(
-                        // https://openjdk.org/jeps/498
-                        // suppresses: sun.misc.Unsafe::objectFieldOffset has been called by com.intellij.util.containers.Unsafe
-                        // requires IntelliJ platform update to resolve the issue
-                        "--sun-misc-unsafe-memory-access=allow",
                         //"-XX:MaxMetaspaceSize=512m",
                         "-XX:+HeapDumpOnOutOfMemoryError",
                         "-XX:+AlwaysPreTouch", // https://github.com/gradle/gradle/issues/3093#issuecomment-387259298
                         //"-XX:StartFlightRecording=disk=true,name={path.drop(1).map { if (it.isLetterOrDigit()) it else '-' }.joinToString("")},dumponexit=true,duration=30s",
                         //"-XX:FlightRecorderOptions=repository=$baseDir/jfr,stackdepth=512",
-                    )
+                    ) + if (JavaVersion.current() >= JavaVersion.VERSION_24) {
+                        // https://openjdk.org/jeps/498
+                        // the option has been available since Java 24,
+                        // has `warn` value since Java 25,
+                        // will have `deny` value at some point after Java 26
+                        //
+                        // suppresses: sun.misc.Unsafe::objectFieldOffset has been called by com.intellij.util.containers.Unsafe
+                        // requires IntelliJ platform update to resolve the issue
+                        listOf("--sun-misc-unsafe-memory-access=allow")
+                    } else emptyList()
                 )
             }
         )
--- a/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
+++ b/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
@@ -77,6 +77,7 @@ internal class JvmMappedMethodsDocumentableFilterTransformer(context: DokkaConte
         "CharSequence",
         "codePoints()Ljava/util/stream/IntStream;",
         "chars()Ljava/util/stream/IntStream;",
+        // note: this method is not handled by the compiler (https://youtrack.jetbrains.com/issue/KT-80102)
         "getChars(II[CI)V"
     ) +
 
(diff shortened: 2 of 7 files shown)


dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
```diff
@@ -74,7 +74,10 @@ internal class JvmMappedMethodsDocumentableFilterTransformer(context: DokkaConte
 
     // copy-pasted from [here](https://github.com/JetBrains/kotlin/blob/0ef41d75b5901afea68a09cc3e762b52d8ce1e3c/core/compiler.common.jvm/src/org/jetbrains/kotlin/builtins/jvm/JvmBuiltInsSignatures.kt#L15)
     private val VISIBLE_METHOD_SIGNATURES: Set<String> = inJavaLang(
-        "CharSequence", "codePoints()Ljava/util/stream/IntStream;", "chars()Ljava/util/stream/IntStream;"
+        "CharSequence",
+        "codePoints()Ljava/util/stream/IntStream;",
+        "chars()Ljava/util/stream/IntStream;",
+        "getChars(II[CI)V"
```
adam-enko: Is there an updated link that shows where these come from?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
```diff
@@ -119,7 +120,16 @@ constructor(
...
                         //"-XX:FlightRecorderOptions=repository=$baseDir/jfr,stackdepth=512",
-                    )
+                    ) + if (JavaVersion.current() >= JavaVersion.VERSION_24) {
+                        // https://openjdk.org/jeps/498
+                        // the option has been available since Java 24,
+                        // has `warn` value since Java 25,
+                        // will have `deny` value at some point after Java 26
+                        //
+                        // suppresses: sun.misc.Unsafe::objectFieldOffset has been called by com.intellij.util.containers.Unsafe
+                        // requires IntelliJ platform update to resolve the issue
```
adam-enko, line 130: Is this the relevant IJ issue? 

https://youtrack.jetbrains.com/projects/IJPL/issues/IJPL-191435/WARNING-sun.misc.UnsafearrayBaseOffset-called-by-com.intellij.util.containers.Unsafe-on-running-intellij-code-in-Kotlin-under

If so, it'd be worth adding a link in the comment?

adam-enko (approved):

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
whyoleg: > New method getChars was [added to the CharSequence](https://cr.openjdk.org/~prr/8357176/api/java.base/java/lang/CharSequence.html#getChars(int,int,char%5B%5D,int)) in JDK 25 - will follow up with compiler team, about it

As I said, I will follow up. The Kotlin compiler does nothing here, 
I added it, as `String.getChars` was there already in the list too (`HIDDEN_METHOD_SIGNATURES`), and without it, the test fails with new members.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: Yeah, could be :)
It's hard to be sure, as it warns only once:
> JDK 24 will, by default, issue a warning on the first occasion that any memory-access method is used, whether directly or via reflection. That is, it will issue at most one warning regardless of which memory-access methods are used and how many times any particular method is used. This will alert application developers and users to the forthcoming removal of the methods, and the need to upgrade libraries.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
adam-enko: okay, thanks. Might be worth adding a comment with this info in the code?

whyoleg: https://youtrack.jetbrains.com/issue/KT-80102

whyoleg:
The initial fix with re-ordering dependencies will not work, as the shadowed Kotlin compiler doesn't have all the methods from the IntelliJ platform. Specifically, we need `com.intellij.util.concurrency.AppExecutorUtil.shutdownApplicationScheduledExecutorService`, which is not used in the Kotlin compiler and so is excluded from it.

(Dokka's analysis dependencies setup is hard...)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
vmishenev: By default, the compiler consider new unknown signatures as `NOT_CONSIDER_METHOD_SIGNATURES`  (so-called grey list).
`getChars` can be moved to `NOT_CONSIDER_METHOD_SIGNATURES` from `VISIBLE_METHOD_SIGNATURES` until [KT-80102](https://youtrack.jetbrains.com/issue/KT-80102) is implemented. Although, actually it does not matter.

vmishenev (approved):

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
whyoleg: Got it, thanks!
I will monitor [KT-80102](https://youtrack.jetbrains.com/issue/KT-80102), and will align the behavior with it when it is ready :)

berezovskyi:
Thanks! Regarding Java 25, FYI, there is a separate GH Action from Oracle that provides clean (?) builds of OpenJDK EA.

See https://github.com/eclipse-lyo/lyo/pull/704/files#diff-5dbf1a803ecc13ff945a08ed3eb09149a83615e83f15320550af8e3a91976446R75-R88 how I arranged a "dual-boot" for the JDK installation in my test matrix. You should be able to do a similar check with `if: matrix.javaVersion != '25-ea'`. Not sure how it would work with the toolchains.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/JvmMappedMethodsDocumentableFilterTransformer.kt
dzharkov: > By default, the compiler consider new unknown signatures as NOT_CONSIDER_METHOD_SIGNATURES (so-called grey list).

NB: Since [KT-65438](https://youtrack.jetbrains.com/issue/KT-65438) we don't really have a grey list. If a method is not in visible nor invisible lists, it's considered to be `WEAKLY_HIDDEN`, which means that it's impossible to call it via built-in interface, but still allowed for overrides (though, with a deprecation warning).

PR #4205, opened 2025-08-14
Update Kotlin Slack signup link to use kotl.in shortlink

adam-enko:
Minor change - the short link already redirects to the same link.

Updating it just helps with consistency and resilience, in case it ever changes.


whyoleg (approved):
Nice, thanks!

P.S. we do have so much shortcut links this days :)

PR #4210, opened 2025-08-18
Fix parsing of `>` in code blocks

### LINKED ISSUE #3775: 'greater than' symbol is not rendered in large code blocks
**Describe the bug**
When using three-backticks block in the documentation, sign `>` is not rendered properly:

```
 * It is also an established pattern to validate input in user's classes in the following manner:
 * ```
 * @Serializable
 * class User(val age: Int, val name: String) {
 *     init {
 *         require(age > 0) { ... }
 *         require(name.isNotBlank()) { ... }
 *     }
 * }
 * ```
 * While clearly being serialization error (when compromised data was deserialized),
 * 
```

image

I also tried `&gt;`, but it is rendered literally as `&gt;` and not as `>`.

**Expected behaviour**
Sign `>` should be visible in the screenshot between `age` and `0`

**To Reproduce**
Reproducer also available here: https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serialization-exception/ (source: https://github.com/Kotlin/kotlinx.serialization/blob/79ecba6e813cc0d59758c9d3f229288b6bd90e8a/core/commonMain/src/kotlinx/serialization/SerializationExceptions.kt#L27)

**Dokka configuration**

```Kotlin
dokka = "1.9.20"
gradlePlugin-dokka = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version.ref = "dokka"}
```

whyoleg:
Fixes #3775

Additionally, it could improve the performance (and memory pressure) of Markdown parsing a bit, as we don't create three `Regex`s anymore on each text processing

-----

Initially, I tested that code block and thought that the problem was inside `prism.js`, so we couldn't do anything.
But after additional checks, I was wrong, and the problem was inside our smart markdown parser, which had wrong regex replacements.
Specifically, `Regex(" >+ +")` was meant to replace only block quotes, like this one:
```
 > Blockquotes are very handy in email to emulate reply text.
 > This line is part of the same quote.
```
But in reality, it was replacing all `>` signs :)

### DIFF
--- a/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt
+++ b/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt
@@ -281,7 +281,7 @@ public open class MarkdownParser(
 
     private fun textHandler(node: ASTNode, keepAllFormatting: Boolean) = DocTagsFromIElementFactory.getInstance(
         MarkdownTokenTypes.TEXT,
-        body = text.substring(node.startOffset, node.endOffset).transform(),
+        body = preprocessTextElement(text.substring(node.startOffset, node.endOffset)),
         keepFormatting = keepAllFormatting
     )
 
@@ -524,11 +524,6 @@ public open class MarkdownParser(
         return LeafASTNode(type, startOffset, endOffset)
     }
 
-    private fun String.transform() = this
-        .replace(Regex("\n\n+"), "")        // Squashing new lines between paragraphs
-        .replace(Regex("\n"), " ")
-        .replace(Regex(" >+ +"), " ")      // Replacement used in blockquotes, get rid of garbage
-
     private fun detailedException(baseMessage: String, node: ASTNode) =
         IllegalStateException(
             baseMessage + " in ${kdocLocation ?: "unspecified location"}, element starts from offset ${node.startOffset} and ends ${node.endOffset}: ${
@@ -549,6 +544,35 @@ public open class MarkdownParser(
                 .joinToString(separator = ".")
                 .takeIf { it.isNotBlank() }
         }
+
+        // Note: Regex creation is a complex operation, and so it should be created once
+        private val blockquoteNewLineRegex = Regex("\n>+ ")
+
+        /**
+         * Performs preprocessing of the Markdown text element.
+         *
+         * First replacement: blockquotes handling
+         * Markdown parser,
+         * when handling blockquotes will handle (remove) `>` symbols only for the first line
+         * and will do nothing for further lines.
+         * So we need to do it manually, for example, when parsing the following Markdown:
+         * ```text
+         * > First line
+         * > Second line
+         * ```
+         * Text element will contain: `First line\n> Second line`.
+         *
+         * Second replacement: combining multiple lines into a single paragraph
+         * When parsing the following Markdown:
+         * ```text
+         * First line
+         * Second line
+         * ```
+         * Text element will contain: `First line\n Second line`.
+         * But it's still a single paragraph which should be concatenated into a single one.
+         */
+        private fun preprocessTextElement(text: String): String {
+            return text.replace(blockquoteNewLineRegex, "\n").replace('\n', ' ')
+        }
     }
 }
-
--- a/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt
+++ b/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt
@@ -133,12 +133,21 @@ public open class MarkdownParser(
     private fun List<ASTNode>.evaluateChildrenWithDroppedEnclosingTokens(count: Int) =
         drop(count).dropLast(count).evaluateChildren()
 
-    private fun blockquotesHandler(node: ASTNode) =
-        DocTagsFromIElementFactory.getInstance(
+    /**
+     * Tracks if we are currently processing elements inside of [MarkdownElementTypes.BLOCK_QUOTE] element.
+     * More info on the specifics and why we need it, in the documentation of [preprocessTextElement]
+     */
+    private var isInsideBlockquote = false
+    private fun blockquotesHandler(node: ASTNode): List<DocTag> {
+        isInsideBlockquote = true
+        val result = DocTagsFromIElementFactory.getInstance(
             node.type, children = node.children
                 .filterIsInstance<CompositeASTNode>()
                 .evaluateChildren()
         )
+        isInsideBlockquote = false
+        return result
+    }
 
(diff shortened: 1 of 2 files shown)


dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt
```diff
@@ -549,6 +544,35 @@ public open class MarkdownParser(
...
+         * Second replacement: combining multiple lines into a single paragraph
+         * When parsing the following Markdown:
+         * ```text
+         * First line
+         * Second line
+         * ```
+         * Text element will contain: `First line\n Second line`.
+         * But it's still a single paragraph which should be concatenated into a single one.
+         */
+        private fun preprocessTextElement(text: String): String {
```
vmishenev: Could you make it execute only within `blockquotesHandler`?
It doesn’t make sense to preprocess the entire Markdown text.
Also, it would make it clearer that the code is needed only for blockquotes.

whyoleg: It's a good idea, yes, and initially, I was thinking about it, but TBH, I failed to find a meaningful way to do it.
At the current moment, when processing `TEXT` elements, we don't know which parent tag (e.g., `BLOCK_QUOTE`) we are inside, and vice versa, when processing `BLOCK_QUOTE`, we don't expect that inside will be just `TEXT` elements.
So we can't influence the child handlers' behavior based on the parent (which could not even be the direct parent).

So, the best way here is probably to track via a flag that we are inside `BLOCK_QUOTE` and do this regex replacement. I don't feel like it will work consistently, so it will not cause other bugs. Also, it will definitely complicate the behavior for a rather specific case.

Or, maybe you have some other idea on how we can handle those?

vmishenev: The flag (like existing `keepFormating`) is the simplest way, but I think it should not be propagated.
TBH regex is not a good option for parsing. Actually, the original MD parser can differentiate between the `BLOCK_QUOTE` token and just angle brackets.
We just need to filter out such tokens e.g. in  `evaluateChildrenIgnoringBlockQuotes` if `node.parent.type == BLOCK_QUOTE`

Maybe this solution is more reliable. 

>  it will not cause other bugs

I'm not even sure that this regexp will not cause other bugs. I hope it is covered by unit tests. 

> Also, it will definitely complicate the behavior for a rather specific case.

The idea is to make it more local, without touching the parsing of every text.

whyoleg: > The flag (like existing keepFormating) is the simplest way, but I think it should not be propagated.

But I don't see how it's possible. Here is the example of how the following markdown comment is processed:
```md
 > text 1
 > text 2
 
 > text 3
```

Function calls from parser with indentation for nesting:
```
visitNode: Markdown:BLOCK_QUOTE
 blockquotesHandler: children = '[Markdown:PARAGRAPH]'
  visitNode: Markdown:PARAGRAPH
   visitNode: Markdown:TEXT
    textHandler: body = 'text 1\n> text 2'     <-- new line symbol here, and `>` is on new line
visitNode: Markdown:BLOCK_QUOTE
 blockquotesHandler: children = '[Markdown:PARAGRAPH]'
  visitNode: Markdown:PARAGRAPH
   visitNode: Markdown:TEXT
    textHandler: body = 'text 3'
```

So here, we need to propagate the flag through the `PARAGRAPH` handler and maybe even some other nodes, such as nested block quotes.

> Actually, the original MD parser can differentiate between the BLOCK_QUOTE token and just angle brackets.

Yes, it can, and it strips the first `>`. So, maybe a better fix would be to fix this inside the markdown parser itself? 

> TBH regex is not a good option for parsing.

Yeah, I agree! It's just that it was there already, and I was thinking that a _small fix_ could solve the original issue with minimal problems. I still see the current approach in the PR as a workaround, though.

> node.parent.type == BLOCK_QUOTE

For some reason, all `node.parent == null` in case of `node.type == TEXT` :)
And, as I showed above, the real parent there will be `PARAGRAPH`

> I'm not even sure that this regexp will not cause other bugs.

I feel like it should not, as previously it was a more relaxed one:
- old: `" >+ +"`
- new: `"\n>+ "`

And yes, there are tests for blockquotes and even nested blockquotes there, plus I've added the test for `>` handling - so I feel like we should be safe here?

> The idea is to make it more local, without touching the parsing of every text.

I agree, that would be nice, but I really feel like, because of the current approach with visitors, the only way is to track it via a global variable like this:
```
var isInsideBlockQuote = false

fun blockquotesHandler() {
    isInsideBlockQuote = true
    // old logic here
    isInsideBlockQuote = false
}

fun textHandler() {
    preprocessTextElement(text, handleBlockQuoteStartingSymbol = isInsideBlockQuote)
}
```

Yes, in that case, it will be more local, so we will use regex only inside of `blockquote` blocks.
WDYT about such an approach?

vmishenev:
> And yes, there are tests for blockquotes and even nested blockquotes there, plus I've added the test for > handling - so I feel like we should be safe here?

I thought about this case:
````
```md
sddsds
> sddsds
```
````
but the regexp "magically" does not work here)


My assumption:  we need to filter out `BLOCK_QUOTE` tokens only in the **first** PARAGRAPH.
image
Otherwise, you can have a look at QuickDoc.


>  WDYT about such an approach?

I think it is better than the current one)

File: dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt (hunk 1 above)
Comment (vmishenev) [reply dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt]: > And yes, there are tests for blockquotes and even nested blockquotes there, plus I've added the test for > handling - so I feel like we should be safe here?

I thought about this case:
````
```md
sddsds
> sddsds
```
````
but the regexp "magically" does not match here)


My assumption:  we need to filter out `BLOCK_QUOTE` tokens only in the **first-level** PARAGRAPH.
image
Otherwise, you can have a look at QuickDoc.


>  WDYT about such an approach?

I think it is better than the current one)

whyoleg: > but the regexp "magically" does not match here)

It looks like everything is handled correctly, and in that case, the markdown parser correctly adds `Br` elements in the code block (as shown in the test added in https://github.com/Kotlin/dokka/pull/4210/commits/92c8992c3fe4eadf490a96d37942c1857655b4a7).
It seems like the only real issue here, and that is why we need that regex: to overcome what I think is a bug in https://github.com/JetBrains/markdown.
I will create an issue there.

> I think it is better than the current one)

Got it! Then will do!

vmishenev (approved):

PR #4156, opened 2025-06-24
Test Gradle 9

adam-enko:
#4150

PR #4213, opened 2025-08-18
Fix compatibility with KGP 2.2.20

whyoleg:


adam-enko:
Is DGP not compatible with 2.2.0? I added it as a tested version in https://github.com/Kotlin/dokka/pull/4156. Or is there a mistake, and it's not actually tested?

whyoleg:
DGP is compatible with both 2.2.0 and 2.2.20.

This fix is needed for the specific case outlined in the issue mentioned in the code.
It was discovered in https://github.com/Kotlin/dokka/pull/4209: in KGP 2.2.20, `androidVariant` became a nullable and deprecated property, and so the compilation failed
This PR handles this just more gracefully.

adam-enko:
~I've tried the code in this PR in a the `android-it` test project, and I got this error~ nm I copied the wrong code.

We've had another suggestion from the AGP team, so I'll try it out.

whyoleg:
replaced by #4231

Issue #4191, opened 2025-08-06
[K2] Broken resolution of kdoc links in module/package documentation


whyoleg:
Presumably, after #4167, all kdoc links in module/package documentation become just plain text.

Before ([commit](https://github.com/Kotlin/dokka/commit/6b32cd6a04a9d392be6022538a3b710f370dd5ce)):

<img width="2188" height="1004" alt="Image" src="https://github.com/user-attachments/assets/6ce7c579-1e89-43a8-9151-1f96e9a831d3" />

After ([commit](https://github.com/Kotlin/dokka/commit/e59dc2a61b1c94e1a53c27555899392377e35f3c)):

<img width="2142" height="988" alt="Image" src="https://github.com/user-attachments/assets/e31d8362-869f-4fea-83ff-2c7e221d9c98" />

**Expected behaviour**
Links should be resolved as before

PR #4214, opened 2025-08-18
[K2] Fix resolving KDoc links in module/package documentation #4191

vmishenev:


whyoleg (approved):
Nice!
I do like, that the code flow now feels more logical :)

Issue #2933, opened 2023-03-21
Provide a separate artifact for Dokka Core API classes


aSemy:
**Is your feature request related to a problem? Please describe**

[Dokkatoo](https://github.com/adamko-dev/dokkatoo/) uses a [Gradle Worker with classpath isolation](https://docs.gradle.org/current/userguide/worker_api.html#changing_the_isolation_mode) to execute the [Dokka Generator](https://github.com/Kotlin/dokka/blob/v1.8.10/core/src/main/kotlin/DokkaGenerator.kt) in [Dokka Core](https://github.com/Kotlin/dokka/tree/v1.8.10/core).

Dokka Generator is configured via some Kotlin interfaces, that are defined in the same module.

https://github.com/Kotlin/dokka/blob/775377625994339aff47b078013f60f518d22919/core/src/main/kotlin/configuration.kt#L133-L264

Because Dokka Generator and the configuration classes are defined the same module, this tightly couples Dokkatoo to a specific Dokka version. This dependency also drags along Dokka Generator, even though it is not required during Gradle project configuration.

Since Dokkatoo uses classpath isolation to run Dokka Generator, the version of Dokka Generator can be set on the fly.

**Describe the solution you'd like**

Split the interfaces for Dokka Core configuration into a separate artifact.

* `org.jetbrains.dokka:dokka-core` - contains Dokka Generator, and depends on `org.jetbrains.dokka:dokka-core-api`
* `org.jetbrains.dokka:dokka-core-api` - contains pure configuration classes and interfaces, without logic

Declaring a feature-variant _might_ be a possible way of doing this https://docs.gradle.org/current/userguide/feature_variants.html, but a more straight-forward way would be do create a separate subproject

**Describe alternatives you've considered**

Keep on using dokka-core. 

**Additional context**

Related: 

* #2904 has a related aim of trying to make the Dokka API more organised

**Are you willing to provide a PR?**

Yes

IgnatBeresnev:
Thanks for creating the issue! I kept it in mind after our discussion, and we just had a technical breakdown meeting about it.

We've actually agreed to take a bit of a turn with the `core` artifact/module. 

The first idea was indeed to break it down into `core` and `core-api`, but after some thinking and evaluating the code that we have inside `core`, we decided to just refactor `core` instead. There's a problem now that we don't know what "core" is, and which classes should be put in core and which in base.

[...]

PR #4216, opened 2025-08-18
Add DokkaGeneratorProvider create DokkaGenerator using ServiceLoader

adam-enko:
POC - will continue after #4182 is merged


adam-enko:
This is something I'd like to continue with when DGPv1 is removed. It would help with removing unnecessary dependencies from DGP, which pollutes users buildscripts.

It's related to #2933.

Issue #4190, opened 2025-08-06
Enable K2 by default


vmishenev:
Turn the `org.jetbrains.dokka.experimental.tryK2` flag on by default. It should not print any warnings about K2, but users should still have the option to roll back to K1.

Dokka should probably display a warning for K1, stating that it does not support new language features starting from Kotlin 2.0.0.

whyoleg:
Note: we should also change the dependency inside the maven plugin: https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-runners/runner-maven-plugin/src/main/kotlin/org/jetbrains/dokka/maven/DokkaMojo.kt#L436

Issue #3369, opened 2023-11-23
[K2] Generated getter for property in kotlin-as-java doesn't inherit kdoc description


atyrin:
```Kotlin
/**
 * @property propertyWithKDoc some prop
 */
class CrossLinksSource(){
    val propertyWithKDoc: Int = 0
}
```

With `kotlin-as-java` plugin the generated getter doesn't get the description.

<img width="689" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/e967545c-7d08-41d8-a5eb-a1356e160a86">

In K1 it worked correctly:

<img width="667" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/9c6c5c4e-92f9-4aa8-aea7-02db18373d63">

**Installation**
- Dokka version: 1.9.20

Parent: https://github.com/Kotlin/dokka/issues/3328
Relates: #3367

IgnatBeresnev:
Is blocking?
* Testing: no
* Beta: seems like no

whyoleg:
Looks like this is the same as mentioned in https://github.com/Kotlin/dokka/issues/3752#issuecomment-2573659072
(cc @vmishenev)

AbdullinAM:
Related AA issue [KT-82851 Property accessors should inherit KDoc from the parent property](https://youtrack.jetbrains.com/issue/KT-82851)

Available in build [2.3.20-dev-6019](https://github.com/JetBrains/kotlin/releases/tag/build-2.3.20-dev-6019)

AbdullinAM:
Fixed in #4370

PR #4218, opened 2025-08-19
Enable K2 analysis by default

adam-enko:
Part of #4190


dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -187,28 +187,28 @@ constructor(
...
-    private fun logK2AnalysisMessage() {
+    private fun logK1AnalysisWarning() {
         if (primaryService && !parameters.k2AnalysisNoWarn.getOrElse(false)) {
-            logger.warn("warning: Dokka K2 Analysis is enabled")
+            logger.warn("warning: Dokka K1 Analysis is enabled")
             logger.lifecycle(
                 """
-                |Dokka K2 Analysis is Experimental and is still under active development.
-                |It can cause build failures or generate incorrect documentation. 
+                |Dokka K1 Analysis is deprecated. It can cause build failures or generate incorrect documentation.
```
ALikhachev: Minor: 
```suggestion
                |Dokka K1 Analysis is deprecated and going to be removed in a future Dokka release. It can cause build failures or generate incorrect documentation.
```

adam-enko:
The expected data for the example projects needs to be updated.

Instructions are here: 
https://github.com/Kotlin/dokka/tree/v2.0.0/dokka-integration-tests#example-projects

ALikhachev:
@vmishenev @whyoleg please take a look at test data changes, some of them look suspicious. For example, removal of `The name of the person to be greeted.`

vmishenev:
>  please take a look at test data changes, some of them look suspicious. For example, removal of The name of the person to be greeted.

Thank you. This is a know bug https://github.com/Kotlin/dokka/issues/3369 in the plugins `kotlin-as-java` and `javadoc` that are in Alpha.

Missing the empty `<span class="token keyword"></span>` is also expected.

dokka-integration-tests/gradle/projects/it-android/expectedData/html/it-android/it.android/-integration-test-activity/index.html
whyoleg, line 1: Those changes are related to how Java properties (`getX` + `setX`) are represented now in Dokka. With K2, they are represented as properties, not just two functions.
So the changes feel correct

PR #4221, opened 2025-08-21
Use K2 analysis in Dokka Maven plugin

vmishenev:
Part of https://github.com/Kotlin/dokka/issues/4190


dokka-runners/runner-maven-plugin/src/main/kotlin/org/jetbrains/dokka/maven/DokkaMojo.kt
```diff
@@ -433,7 +433,7 @@ public abstract class AbstractDokkaMojo(
             cacheRoot = cacheRoot?.let(::File),
             sourceSets = listOf(sourceSet),
             // TODO [beresnev] analysis switcher
```
whyoleg: we should probably remove the comment now :)

PR #4226, opened 2025-08-25
Fix expect/actual wrong way around in shouldBeADirectoryWithSameContentAs

adam-enko:
The expected/actual directories are the wrong way around, which leads to confusing test failure messages.

Thanks to @ALikhachev for spotting it.


whyoleg (approved):
Nice catch!

PR #4228, opened 2025-08-26
Build the Dokka Maven Plugin with the lastest compiler version

vmishenev:


build-logic/src/main/kotlin/dokkabuild/utils/gradleKotlinCompatibility.kt
```diff
@@ -20,7 +20,7 @@ fun Project.configureGradleKotlinCompatibility() {
      * To be compatible with the AA, Dokka analysis should be compiled with approximately the same version
      * See https://kotlinlang.org/docs/kotlin-evolution-principles.html#evolving-the-binary-format
      */
-    val analysisK2Projects = listOf("analysis-kotlin-symbols")
+    val analysisK2Projects = listOf("analysis-kotlin-symbols", "runner-maven-plugin")
```
whyoleg, line 23: As the `runner-maven-plugin` is an included build, it's possible to safely remove flags [there](https://github.com/Kotlin/dokka/blob/e56edc64823a651d278740e8ef9b74820555cc9a/dokka-runners/runner-maven-plugin/gradle.properties#L9-L10) in `gradle.properties`.

whyoleg (approved):
I beleive, it should be fine for 2.1.0-Beta.
Interesting, why we don't have similar problem with CLI...

PR #4229, opened 2025-08-26
Use the shadowed stdlib in the Dokka Maven Plugin

vmishenev:


dokka-runners/runner-maven-plugin/build.gradle.kts
```diff
@@ -15,8 +15,16 @@ plugins {
 overridePublicationArtifactId("dokka-maven-plugin")
 
 dependencies {
+    // to override jackson from the compiler in `analysis-kotlin-symbols`
```
whyoleg: I don't see jackson inside of `analysis-kotlin-symbols` artefact at all, so something is wrong here

dokka-runners/runner-maven-plugin/build.gradle.kts
```diff
@@ -15,8 +15,16 @@ plugins {
...
+    // https://github.com/JetBrains/kotlin/blob/b17a978c6165a43dfe8fdfcc2e6dc2e6856c3c9a/gradle/libs.versions.toml#L26
+    implementation(libs.jackson.kotlin)
+    implementation(libs.jackson.xml)
+
     // this version is required, so that it will be available in the POM of plugin
+    implementation("org.jetbrains.dokka:analysis-kotlin-symbols:$version")
```
whyoleg: AFAIK, it will use a non-shadowed artifact, and to use a shadowed one, we need to do something similar to what we do in [cli integration tests](https://github.com/Kotlin/dokka/blob/2ad0c1f299f3dd05f37894a5fc7efe3800f55e69/dokka-integration-tests/cli/build.gradle.kts#L36-L38).

Though I have no idea how it will affect publication

dokka-runners/runner-maven-plugin/build.gradle.kts
vmishenev: Do you see the folder `/com/fasterxml/aalto/` e.g. in https://repo1.maven.org/maven2/org/jetbrains/dokka/analysis-kotlin-symbols/2.0.0/analysis-kotlin-symbols-2.0.0.jar ?

dokka-runners/runner-maven-plugin/build.gradle.kts
vmishenev: It does not matter here
The main idea that publishing `dokka-runners/runner-maven-plugin/build/publications/jvm/pom-default.xml` has a dependency on `analysis-kotlin-symbols`

dokka-runners/runner-maven-plugin/build.gradle.kts
whyoleg: Yes, I see, but it's not a dependency of Jackson JSON or XML (which we do use in Dokka); it's a dependency of the **IntelliJ platform**. And so it should not cause any conflicts because the dependencies we use don't depend on `aalto`, and adding jackson with our versions should not affect anything here.

E.g., our `core` module, which depends on `jackson.json`, `jackson.xml`, and `jackson.databind` has the following dependencies coming from Jackson:
```
+--- com.fasterxml.jackson.module:jackson-module-kotlin:2.15.3
|    +--- com.fasterxml.jackson.core:jackson-databind:2.15.3
|    |    +--- com.fasterxml.jackson.core:jackson-annotations:2.15.3
|    |    |    \--- com.fasterxml.jackson:jackson-bom:2.15.3
|    |    |         +--- com.fasterxml.jackson.core:jackson-annotations:2.15.3 (c)
|    |    |         +--- com.fasterxml.jackson.core:jackson-core:2.15.3 (c)
|    |    |         +--- com.fasterxml.jackson.core:jackson-databind:2.15.3 (c)
|    |    |         +--- com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.15.3 (c)
|    |    |         \--- com.fasterxml.jackson.module:jackson-module-kotlin:2.15.3 (c)
|    |    +--- com.fasterxml.jackson.core:jackson-core:2.15.3
|    |    |    \--- com.fasterxml.jackson:jackson-bom:2.15.3 (*)
|    |    \--- com.fasterxml.jackson:jackson-bom:2.15.3 (*)
|    +--- com.fasterxml.jackson.core:jackson-annotations:2.15.3 (*)
|    +--- org.jetbrains.kotlin:kotlin-reflect:1.5.32 -> 2.0.21 (*)
|    \--- com.fasterxml.jackson:jackson-bom:2.15.3 (*)
+--- com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.15.3
|    +--- com.fasterxml.jackson.core:jackson-core:2.15.3 (*)
|    +--- com.fasterxml.jackson.core:jackson-annotations:2.15.3 (*)
|    +--- com.fasterxml.jackson.core:jackson-databind:2.15.3 (*)
|    +--- org.codehaus.woodstox:stax2-api:4.2.1
|    +--- com.fasterxml.woodstox:woodstox-core:6.5.1
|    |    \--- org.codehaus.woodstox:stax2-api:4.2.1
|    \--- com.fasterxml.jackson:jackson-bom:2.15.3 (*)
```

vmishenev: My bad, I mixed up the names jackson and fastexml because I was in a rus
Yes, `aalto-xml` is also the dependency IntelliJ:
image
Although, the `kotlin-compiler` has it,  `analysis-kotlin-symbols` is magically published with stripped `aalto-xml` from `kotlin-compiler`.



Actual problem is that `jackson-dataformat-xml` starts using `aalto-xml` for parsing XML:
See [the stack trace](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Dokka_IntegrationTests/5433431)
`jackson-dataformat-xml`uses Stax API. See also https://docs.oracle.com/javase/8/docs/api/javax/xml/stream/XMLInputFactory.html#newFactory-java.lang.String-java.lang.ClassLoader- and `META-INF/services` in `analysis-kotlin-symbols`.

This line just changes the order of `XMLInputFactory`-services.
_Note: `kotlin-compiler`  does NOT have `META-INF/services/XMLInputFactory`, but IntelliJ has it_ 

The solution could be (I haven't checked it yet) - excluding `META-INF/services/XMLInputFactory` from `analysis-kotlin-symbols`.

whyoleg: Oh my, that's an interesting coincidence :)

dokka-runners/runner-maven-plugin/settings.gradle.kts
```diff
@@ -30,6 +30,12 @@ dependencyResolutionManagement {
             setUrl("https://cache-redirector.jetbrains.com/maven-central")
             name = "MavenCentral-JBCache"
         }
+        maven("https://cache-redirector.jetbrains.com/intellij-repository/releases") {
```
whyoleg: Why do we need those repositories?

dokka-runners/runner-maven-plugin/build.gradle.kts
whyoleg: I've compared artifacts before and after this change, and they feel fine, except for the changes to the manifest—I don't know how they could be used, but we have unshadowed dependencies there, and it feels strange for me.
I believe that we need to somehow provide a correct `shadowed` artifact there.

image

dokka-runners/runner-maven-plugin/settings.gradle.kts
vmishenev: They were needed for the non-shadowed  `analysis-kotlin-symbols`

dokka-runners/runner-maven-plugin/build.gradle.kts
vmishenev: Anyway, using a shadowed jar is more correct since we could modify it, e.g. excluding some files, and such changes would not affect the direct dependency (at least at compile-time).

PR #4231, opened 2025-08-27
KGP 2.2.10 + AGP: Use Android components to determine if source-sets are 'publishable'

adam-enko:
Continuation of https://github.com/Kotlin/dokka/pull/4213

DGP should only document 'publishable' (i.e. main, non-test) source sets. DGP needs to extract this information from KGP and AGP.

For AGP projects, DPG currently uses `KotlinJvmAndroidCompilation.androidVariant` to determine publishability of a source set. However, [this is deprecated in KGP 2.2.10](https://github.com/JetBrains/kotlin/blob/v2.2.10/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/targets/jvm/KotlinJvmAndroidCompilation.kt#L38-L39). It will eventually be removed. This is because AGP will implement Kotlin compilation internally (AKA Kotlin built-in).

This PR uses an alternative AGP utility: `AndroidComponentsExtension`. DGP extracts Components from this extension and stores it in a separate data structure (to avoid Configuration Cache issues with serialising too much data). If a Component has at least one Variant, it is considered 'publishable'. Each Component is associated with a single KotlinCompilation (matched by name), and each KotlinSourceSet is associated with a list of KotlinCompilations. 

While using `AndroidComponentsExtension` is required for AGP9, it also works for AGP7 and 8. To reduce complexity, this PR uses the same approach for all AGP versions.

---

This PR is the first part of supporting AGP 9 in DGP. The next PR is #4295, which will support and test AGP 9 with and without Kotlin built-in.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/findExtensionLenient.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/findExtensionLenient.kt
@@ -9,7 +9,7 @@ import org.gradle.api.Project
 /**
  * Try and get an extension from [Project.getExtensions], or `null` if it's not present.
  *
- * Handles as many errors as possible.
+ * If [T] is not accessible in the current classloader, returns `null`.
  *
  * Logs a warning if the extension is present, but the wrong type
  * (probably caused by an inconsistent buildscript classpath https://github.com/gradle/gradle/issues/27218)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -438,34 +428,61 @@ private class KotlinCompilationDetailsBuilder(
...
+     * This can be overridden manually, though.
+     *
+     * @see DokkaSourceSetSpec.suppress
+     */
+    private fun KotlinCompilation<*>.isPublished(): Provider<Boolean> {
+        return when (this) {
+            is KotlinMetadataCompilation<*> ->
+                providers.provider { true }
+
+            // Use string-based comparison, not the actual classes, because AGP has deprecated and
```
whyoleg: This comment should be moved to `isJvmAndroidPublished`

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -438,34 +428,61 @@ private class KotlinCompilationDetailsBuilder(
...
+    }
 
-                else ->
-                    name == MAIN_COMPILATION_NAME
+    private fun isJvmAndroidPublished(
+        compilation: KotlinJvmAndroidCompilation,
+    ): Provider<Boolean> {
+
+        // in KGP 2.2.10 androidVariant will be nullable KT-77023
+        if (currentKotlinToolingVersion.supportsAgpKotlinBuiltInCompilation()) {
```
whyoleg: I'm curious: Can we rely on the approach using Android components all the time, and not just in KGP 2.2.10+?
It feels a bit strange that we use some public AGP API, which should work in all cases, but for some reason, we use it only for newer versions :)

I mean, if we used those components only in case `compilation.androidVariant` is `null` - then fine, but it looks like it works all the time.
WDYT?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -618,3 +635,64 @@ private class KotlinSourceSetDetailsBuilder(
...
     }
 }
+
+
+/** Try and get [KotlinProjectExtension], or `null` if it's not present. */
+private fun Project.findKotlinExtension(): KotlinProjectExtension? =
+    findExtensionLenient<KotlinProjectExtension>("kotlin")
+
+
+private typealias AndroidComponentsExtension = com.android.build.api.variant.AndroidComponentsExtension<*, *, *>
```
whyoleg: Curious: What is the reason for this `typealias`? Just to hide those `<*,*,*>` in the function below? :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -618,3 +635,64 @@ private class KotlinSourceSetDetailsBuilder(
...
+/**
+ * Collect [AndroidVariantInfo]s for all variants in the project.
+ *
+ * Should only be called when AGP is applied (otherwise the `androidComponents` extension will be missing).
+ */
+private fun collectAndroidVariants(
+    project: Project,
+    androidVariants: SetProperty<AndroidVariantInfo>,
+) {
+    val androidComponents = project.findAndroidComponentExtension()
```
whyoleg, line 695: What should be the expected behavior for the case when there is no this extension? Shouldn't we log something at least?

adam-enko: DGP will log a warning if the extension can't be found

https://github.com/Kotlin/dokka/blob/a869239d43f39cbfd2e687fb2eef973b74d61f5a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/findExtensionLenient.kt#L17-L62

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: Ah, this is a leftover from copying some other code. It's not necessary, I'll remove it.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: Good question, I'm honestly not sure. I'll need to ask in the team.

I _think_ to use Android components, we'd require some minimum AGP version. Here, instead of checking the AGP version directly, a KGP version of >2.1.0 implies the AGP version will be high enough. 

We could replace the KGP version check with an AGP version check? But then maybe the reverse assumption isn't true - so DGP would additionally need to check the KGP version to make sure `compilation.androidVariant` isn't nullable. If we end up with two checks... why not just have one?

But you are probably right, and we can just remove the KGP version, and maybe add an AGP version check if necessary. 

I'll check with the team and get back to you...

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -312,6 +276,33 @@ private class KotlinCompilationDetailsBuilder(
...
+     * Used to determine whether a source set is published or not.
+     * See [KotlinSourceSetDetails.isPublishedSourceSet].
+     *
+     * Android variant info must be fetched eagerly,
+     * since AGP doesn't provide a lazy way of accessing component information.
+     *
+     * @see collectAndroidVariants
+     * @see supportsAgpKotlinBuiltInCompilation
+     */
+    private fun getAgpVariantInfo(
```
adam-enko, line 296: TODO: write a test for AGP/KGP 

https://github.com/Kotlin/dokka/pull/4231#pullrequestreview-3165405851

adam-enko: I've created two integration projects to test AGP with built-in Kotlin:

- `com.android.library` (9.0.0-alpha08) 
- `com.android.kotlin.multiplatform.library` (9.0.0-alpha08) + `kotlin("multiplatform")`

However, they don't work because AGP + built-in Kotlin cannot be handled with Dokka's existing KotlinAdapter and AndroidAdapter. 

It will take some time to update DGP to extract variant dependency information from the new `com.android.build.api.variant.AndroidComponentsExtension`.

whyoleg: I believe we can then tackle AGP 9 support in the next release (patch? 2.1.1?), WDYT?

adam-enko:
This change is only necessary to support AGP with Kotlin built-in, which is only supported in AGP v9+. AGP 9 isn't released yet, so we don't need to include this PR in the DGP 2.1.0 release.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: Yes, let's support AGP 9 in 2.1.1.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I've checked and we can use the 'new' approach (extract info from AGP's `AndroidComponentsExtension`) unconditionally.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -618,3 +619,52 @@ private class KotlinSourceSetDetailsBuilder(
...
+    project: Project,
+    androidVariants: SetProperty<AndroidVariantInfo>,
+) {
+    val androidComponents = project.findAndroidComponentExtension()
+
+    androidComponents?.onVariants { variant ->
+        val hasPublishedComponent =
+            variant.components.any { component ->
+                component is Variant
+            }
```
whyoleg, line 703: It would be really nice to add some comment on why this is correct, as at first sight (especially without knowledge of Android), this check looks very fishy.
And from second sight, when I opened an IDE and checked the `Variant` hierarchy, I was still a bit confused :)
Additionally, we previously explicitly checked for application/library variants, but now we simply check if it's a variant. However, I notice that there are `TestVariant` and `DynamicFeatureVariant` entries, which we also include now. And I'm fine with the second one, but the first one, again, feels a bit strange. 

Are there any AGP docs for this?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -618,3 +619,52 @@ private class KotlinSourceSetDetailsBuilder(
...
+
+        androidVariants.add(
```
whyoleg, line 705: Minor: Do we really need to store information about all variants, or can we just store `publishedVariantNames`?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I'm not very familiar with AGP either, but I'll try and add some comments.

To answer quickly, I believe each compilation should be associated with a single Component, which might be a Variant. A Variant is something that's published.

I believe `TestVariant` indicates something that's used for testing, like a library of test-utilities.

adam-enko: For reference here's the example I got from AGP devs:

```kotlin
android {
   ....
}

val variants = mutableListOf<Variant>()

androidComponents {
    onVariants {
        variants.add(it)
    }
}

afterEvaluate {
    println("================ Printing all variants")
    variants.forEach { variant ->
        println("Variant: ${variant.name}")
        println("    Variant's components: " + variant.components.joinToString { it.name + " (" + it.javaClass.simpleName + ")" })
    }

    println("================ Printing all components")
    val components = variants.flatMap { it.components }
    components.forEach { component ->
        println("Component: ${component.name}, isMainComponent = ${component is Variant}")
    }

    println("================ Printing all compilations")
    kotlin.target.compilations.forEach { compilation ->
        val component = components.single { it.name == compilation.name }
        println("Compilation: ${compilation.name}, mapped to component ${component.name}, isMainComponent = ${component is Variant}")
    }
}
```

```text
================ Printing all variants
Variant: debug
    Variant's components: debug (ApplicationVariantImpl_Decorated), debugTestFixtures (TestFixturesImpl_Decorated), debugAndroidTest (DeviceTestImpl_Decorated), debugUnitTest (HostTestImpl_Decorated), debugScreenshotTest (HostTestImpl_Decorated)
Variant: release
    Variant's components: release (ApplicationVariantImpl_Decorated), releaseTestFixtures (TestFixturesImpl_Decorated), releaseScreenshotTest (HostTestImpl_Decorated)
================ Printing all components
Component: debug, isMainComponent = true
Component: debugTestFixtures, isMainComponent = false
Component: debugAndroidTest, isMainComponent = false
Component: debugUnitTest, isMainComponent = false
Component: debugScreenshotTest, isMainComponent = false
Component: release, isMainComponent = true
Component: releaseTestFixtures, isMainComponent = false
Component: releaseScreenshotTest, isMainComponent = false
================ Printing all compilations
Compilation: debug, mapped to component debug, isMainComponent = true
Compilation: debugAndroidTest, mapped to component debugAndroidTest, isMainComponent = false
Compilation: debugUnitTest, mapped to component debugUnitTest, isMainComponent = false
Compilation: release, mapped to component release, isMainComponent = true
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: Probably not - but storing more information helps with more detailed logging which I think will be helpful if something goes wrong.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I've added a summary on `collectAndroidVariants()` - please take a look.

adam-enko:
> 1. Please change the PR name and so commit message, as this PR is about a different thing.

Sorry, I'm not following, could you write the title you want to see?

I was going to update the commit message when squash-merging to use the issue body.

> 2. It would be nice to check in which AGP version those new APIs were added - just in case.

I believe the Components extension has been available since AGP 7.

https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/extension/AndroidComponentsExtension

>  3. Is it right that this logic will stay the same when we support AGP9, but will be executed only for AGP8?

Yes, this logic will work for AGP9. It will also work for AGP7 and AGP8. (Does this help? I'm not sure I understood your question.)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: Actually, in #4295 to support AGP9 we'll have to collect more information from AGP, so `AndroidVariantInfo` will contain more information, not just the publishability.

whyoleg: Got it! Thanks!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: Nice, thanks, now it feels much better!

> I'm not very familiar with AGP either

Me too, that's why it would be really nice to have this information in the comments :)

whyoleg:
> Sorry, I'm not following, could you write the title you want to see?
> I was going to update the commit message when squash-merging to use the issue body.

I mean, the PR is not just about compatibility with KGP 2.2.10+, but mostly about the improvement of the association of kotlin compilations with AGP variants. And this helps Dokka to be compatible with KGP 2.2.10 at **compile-time**, as, as we discussed previously, at runtime we would never stumble on [KotlinJvmAndroidCompilation.androidVariant](https://github.com/JetBrains/kotlin/blob/v2.2.10/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/targets/jvm/KotlinJvmAndroidCompilation.kt#L38-L39) having `null`. And so the current PR title is really confusing.

So maybe something like this would be a nice first line for the commit message (and PR name):
```markdown
Update DGP logic for understanding if the Android compilation is `publishable` based on Android components
```
WDYT?
(the rest of the commit message could be based on the issue body)

> I believe the Components extension has been available since AGP 7.
> https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/extension/AndroidComponentsExtension

Nice, thanks! Though it's highlighted with strikethrough, is it deprecated in AGP9?
image

In any case, it's not a blocker for this PR.

> Yes, this logic will work for AGP9. It will also work for AGP7 and AGP8. (Does this help? I'm not sure I understood your question.)

Got it, sorry for the confusion. For some reason, I was expecting that AGP9 would have some more incompatible APIs.

---

**From my side, nothing is blocking the PR from merging. Thanks for answering!**

adam-enko:
> > Sorry, I'm not following, could you write the title you want to see?
> > I was going to update the commit message when squash-merging to use the issue body.
> 
> I mean, the PR is not just about compatibility with KGP 2.2.10+, but mostly about the improvement of the association of kotlin compilations with AGP variants. And this helps Dokka to be compatible with KGP 2.2.10 at **compile-time**, as, as we discussed previously, at runtime we would never stumble on [KotlinJvmAndroidCompilation.androidVariant](https://github.com/JetBrains/kotlin/blob/v2.2.10/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/targets/jvm/KotlinJvmAndroidCompilation.kt#L38-L39) having `null`. And so the current PR title is really confusing.
> 
> So maybe something like this would be a nice first line for the commit message (and PR name):
> 
> ```
> Update DGP logic for understanding if the Android compilation is `publishable` based on Android components
> ```
> 
> WDYT? (the rest of the commit message could be based on the issue body)

Thanks! That helps. I'll update it.

> 
> > I believe the Components extension has been available since AGP 7.
> > https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/extension/AndroidComponentsExtension
> 
> Nice, thanks! Though it's highlighted with strikethrough, is it deprecated in AGP9?



~Yeah, I'm not sure about that. At least in the AGP 8 docs it says 'Added in 7.0.0'~

EDIT: ah, I realised I linked to the wrong one. AGP7 has two `AndroidComponentsExtension` interfaces in different packages, and one is deprecated.

- https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/variant/AndroidComponentsExtension
- https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/extension/AndroidComponentsExtension

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/findExtensionLenient.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package org.jetbrains.dokka.gradle.internal
+
+import org.gradle.api.Project
+
+/**
+ * Try and get an extension from [Project.getExtensions], or `null` if it's not present.
+ *
+ * Handles as many errors as possible.
```
ALikhachev: Nit: This sounds too vague, doesn't it? I'd suggest either removing it or making it more specific

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -624,3 +624,89 @@ private class KotlinSourceSetDetailsBuilder(
...
+ *
+ * Next, determine if the [Variant] associated with a [KotlinCompilation] is 'publishable' by
+ * checking if it _or_ any of its [nestedComponents][Variant.nestedComponents]
+ * are 'publishable' (i.e. is an instance of [Variant]).
+ * (We can we use [Variant.components] to check both the [Variant] and its `nestedComponents` the same time.)
+ */
+private fun collectAndroidVariants(
+    project: Project,
+    androidVariants: SetProperty<AndroidVariantInfo>,
+) {
```
ALikhachev, line 694: Nit: it could be an extension function on SetProperty:
```suggestion
private fun SetProperty<AndroidVariantInfo>.collectAndroidVariantsFrom(
    project: Project,
) {
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -624,3 +624,89 @@ private class KotlinSourceSetDetailsBuilder(
...
+ * i.e. inside a `withPlugin(...) {}` block.
+ *
+ * ## How to determine publishability of AGP Variants
+ *
+ * There are several Android Gradle plugins.
+ * Each AGP has a specific associated [Variant]:
+ * - `com.android.application` - [com.android.build.api.variant.ApplicationVariant]
+ * - `com.android.library` - [com.android.build.api.variant.DynamicFeatureVariant]
+ * - `com.android.test` - [com.android.build.api.variant.LibraryVariant]
+ * - `com.android.dynamic-feature` - [com.android.build.api.variant.TestVariant]
```
ALikhachev, line 667: We trigger this function with different plugins, like
```
withPlugin(PluginId.AndroidBase) { collectAndroidVariants(project, androidVariants) }
withPlugin(PluginId.AndroidApplication) { collectAndroidVariants(project, androidVariants) }
withPlugin(PluginId.AndroidLibrary) { collectAndroidVariants(project, androidVariants) }
```

however, we do not trigger it for the listed `com.android.dynamic-feature` or `com.android.test`, for example. Does it mean we cover those by `com.android.base`?

If yes, why just having `com.android.base` is not enough and we should explicitly react on `com.android.application` and `com.android.library`? The API for fetching variants seems lazy, so from my POV this causes collecting variants multiple times, but it has only a performance effect since duplicates are automatically excluded from the set constructed via `androidVariants`

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I'd prefer to avoid an extension function, since it 'leaks' the receiver's API into the function body, which can make it more difficult to write and read.

This function will be changed in the next PR #4295 so I'd prefer to avoid changing it here.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/findExtensionLenient.kt
adam-enko: updated

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: > we do not trigger it for the listed `com.android.dynamic-feature` or `com.android.test`, for example. Does it mean we cover those by `com.android.base`?

The behaviour by AGP isn't well specified, but I think the answer is 'no'.

ALikhachev: Does it mean then that this code does not cover some of the Android family plugins?

adam-enko: Good question...

Thinking again, I _think_ `com.android.base` is non-functional here. Only the actual plugins should matter. But I'd prefer to keep it, just to be safe.

For `com.android.dynamic-feature` and `com.android.test`: I didn't add them here because we haven't had any requests to support them, we don't have any tests for them, and I'm not sure if anyone would ever want to generate Dokka docs for them.

@whyoleg Do you have any opinion on whether DGP should support projects with `com.android.dynamic-feature` or `com.android.test` plugins?

whyoleg: I see that KGP also does perform checks based on [multiple ids](https://github.com/JetBrains/kotlin/blob/08f266c255fd1dfae6c278af28575fdca8572436/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/utils/androidPluginIds.kt#L12) (not using `com.android.base`), and I haven't found any tests which use `com.android.dynamic-feature`.
So I'm leaning more towards having the same behaviour as in KGP - **react on all four plugins, even if we don't have tests for them.**

Also, AFAIU, with AGP9, users will use kotlin built into those plugins, and this will be less of an issue in the future.

adam-enko: I updated DGP to react to all Android plugins.

UPDATE: oops, I updated it in #4295, not this PR 

https://github.com/Kotlin/dokka/pull/4295/commits/9ee88a174c1a2f42ffba32a94ea9508feab829eb

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I updated it to an extension function in #4295

ALikhachev: Was the change pushed? I don't see it as an extension function in 4295 🤔

adam-enko: https://github.com/Kotlin/dokka/pull/4295/commits/614dd9e26105459028723acaca7a8164384fd3d6

ALikhachev (approved):

Issue #3719, opened 2024-07-30
Unresolved links are formatted as plain text


dkhalanskyjb:
**Describe the bug**
When a code link is not resolved, not only does it not work as a link—it also does not look like code.

**Expected behaviour**
When I write `[Dispatchers.IO]` and it's not resolved, I expect it to at least be formatted as code.

**Screenshots, To Reproduce**
See https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/

PR #4237, opened 2025-09-01
Add test `KDoc Link to a class with quoted name should be resolved`

vmishenev:
Note: a quoted name is rendered as inlined code.

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
@@ -1205,6 +1205,34 @@ class LinkTest : BaseAbstractTest() {
     }
 
     @Test
+    @OnlySymbols("#3356")
+    fun `KDoc Link to a class with quoted name should be resolved`() {
+        testInline(
+            """
+            |/src/main/kotlin/Testing.kt
+            |package example
+            |/**
+            |* Class: [Quoted Class Name] is unresolved in K2, but resolved in K1
+            |* [`Quoted Class Name`] is resolved in K2, but unresolved in K1 
+            |* [example.`Quoted Class Name`] is resolved in K2 and K1
+            |*/
+            |class `Quoted Class Name`
+        """.trimMargin(),
+            configuration
+        ) {
+            documentablesMergingStage = { module ->
+                assertEquals(
+                    listOf(
+                        "`Quoted Class Name`" to DRI("example", "Quoted Class Name"),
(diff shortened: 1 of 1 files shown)


whyoleg:
Here is how the following links are rendered on master (and Dokka 2.1.0-Beta):

```
/**
 * [`Quoted Class Name`]
 * [`test%^!@#_me`]
 * [String.length]
 * [`String.length`]
 * [String]
 * [String.`length`]
 * [`String.`length``]
 * [`hexToByteArray`]
 */
```

image

Should we create an additional issue to decide how such cases should be rendered? https://github.com/Kotlin/dokka/issues/3719 could be related.

(cc @berezinant just in case you might have discussed such cases during code-blocks rework)

whyoleg:
@vmishenev K2 tests are failing with metaspace error. Could you investigate?
I see that it fails in our new `DokkaBootstrap` implementation.

whyoleg (approved):

vmishenev:
> K2 tests are failing with metaspace error. Could you investigate?

#4244

> Should we create an additional issue to decide how such cases should be rendered?

As for me, it looks logical. Or do you have a proposal?

whyoleg:
> As for me, it looks logical. Or do you have a proposal?

I don't have a direct proposal, but it feels weird that we mix Kotlin syntax to reference quoted declarations with Markdown syntax for inline code blocks. I understand why this happens: `[xxx]` is a shorthand for `[xxx][xxx]`, and so one is resolved based on Kotlin syntax, and the other based on Markdown syntax, where we have different meanings for backticks.

Here, I would expect all of the examples to render consistently:
- All declarations rendered as `inline code` or not at all.
- In case the declaration name requires backticks (has spaces), those backticks should be rendered.

JIC, here is what it looks like in the latest IDEA with support for backticks in KDoc:

image

And I do really like how it's done there; it's consistent.

I'm still not quite understanding why this is resolved:
```
`String.length`
```
But that's a question for the KDoc parser (probably also needs an issue)

vmishenev:
> Here, I would expect all of the examples to render consistently:

Good ideas. We should create an issue

> But that's a question for the KDoc parser (probably also needs an issue)

Yes, this looks like a bug

whyoleg:
Created #4251 

> But that's a question for the KDoc parser (probably also needs an issue)

BTW, looks like it's fine that it's resolved, because we can do this:

image

Though we should not really resolve it to `String.length` - probably the issue here is on the AA side (as it's also resolved in IDEA) or in our link resolver, which is copied from IDEA :)

Created #4252

PR #4244, opened 2025-09-03
Make GC more aggressive in integration tests

vmishenev:


dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt
```diff
@@ -40,7 +40,9 @@ abstract class AbstractGradleIntegrationTest : AbstractIntegrationTest() {
     fun createGradleRunner(
         buildVersions: BuildVersions,
         vararg arguments: String,
-        jvmArgs: List<String> = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=1G"),
+        jvmArgs: List<String> = listOf("-Xmx2G", "-XX:MaxMetaspaceSize=800m",
+            "-XX:SoftRefLRUPolicyMSPerMB=10" // to free up the metaspace on JVM 8, see https://youtrack.jetbrains.com/issue/KT-55831/
```
whyoleg: TBH, I was thinking about just providing this argument explicitly to `withJvmArguments` function call below in addition to `jvmArgs`, so that we don't need to repeat it, if we need to override other `jvmArgs` for test
WDYT?

dokka-integration-tests/gradle/src/testTemplateProjectBasic/kotlin/BasicGradleIntegrationTest.kt
```diff
@@ -28,6 +28,10 @@ class BasicGradleIntegrationTest : AbstractGradleIntegrationTest() {
             "dokkaJavadoc",
             "dokkaGfm",
             "dokkaJekyll",
+            jvmArgs = listOf(
```
whyoleg: Can it be removed? And in all other places?

Issue #3680, opened 2024-07-08
[K2] Link to an inherited declaration in `@see` block replace class container to parent one  


atyrin:
Relates to #3350

The code 
```Kotlin
open class Bar { fun foo() {} }
class Foo : Bar()

/**
 * I describe class Foo and method foo here: [Foo.foo]
 * And then describe class Bar and method foo here: [Bar.foo]
 *
 * @see [Foo.foo]
 */
fun usage() {}
```

Rendered in K2 mode as:

<img width="500" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/9ecd5057-60ec-48ed-8401-98e66e9b1b02">

Here `@see [Foo.foo]` replaces with `[Bar.foo]`. That is not happened in the description body.

-----

In K1 the reference in `@see` was not resolved.
<img width="500" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/0fd5bec9-7d37-4827-ac3a-172d7c1b3462">

----
**Installation**
- Dokka version: 2.0.0

whyoleg:
Expected result: `@see [Foo.foo]` should be rendered as `Foo.foo` in `See also` section in the same way as `[Foo.foo]` in the default KDoc description section

AbdullinAM:
Consider this example here:
```kotlin
package org.example.project

import org.example.project.O.property

object O {
    val property = ""
}

open class Bar { fun foo() {} }
class Foo : Bar()

/**
 * link to [Foo.foo]
 * link to [property]
 *
 * @see [Foo.foo]
 * @see [property]
 */
fun usage() {}
```

Here is how all the links are resolved:
<img width="1057" height="405" alt="Image" src="https://github.com/user-attachments/assets/089b3359-46b7-4129-8c2f-31f826543e6e" />

Basically, the resolver uses following rules:
* All links in the regular text are rendered as they are written in KDoc
* All links in `@see` tag (and several other tags) are rendered as fully-qualified declaration names of the `DRI`

This is the root cause of this issue. 

Currently there are tests that expect `@see [property]` to be rendered as `O.property` (see `ContentForSeeAlsoTest::should prefix static function and property links with class name`). We can't keep the old behaviour and solve this issue: there is no reliable way to differentiate between cases where we would want/don't want to render the FQ name.

In my opinion, the behaviour of the links in text and in tags (not just `@see`, but most of the others as well) should be similar. Meaning, we render everything as it is written in KDoc and update the tests accordingly.

[...]

PR #4248, opened 2025-09-08
Use KDoc tag subject name when rendering `@see` tags (#3680)

AbdullinAM:


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KdocMarkdownParser.kt
```diff
@@ -74,7 +74,7 @@ internal fun parseFromKDocTag(
                         val dri = pointedLink(tag)
                         See(
                             parseStringToDocNode(tag.getContent(), externalDRIProvider),
-                            dri?.fqDeclarationName() ?: tag.getSubjectName().orEmpty(),
+                            tag.getSubjectName().orEmpty(),
```
whyoleg, line 73: We also use `dri?.fqDeclarationName()` above for the `@throws` tag: maybe it would make sense to remove it there, too?

Not sure, though, how it's important there, and how it will affect the output.

whyoleg, line 73: Also, we do use the same `DRI.fqDeclarationName` in a similar context [here](https://github.com/Kotlin/dokka/blob/0980b42523efbb08f9f112366802679f51e08018/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/MarkdownParser.kt#L57). Those are used when parsing separate markdown files or [templates](https://github.com/Kotlin/dokka/tree/8e5c63d035ef44a269b8c43430f43f5c8eebfb63/dokka-subprojects/plugin-base/src/main/resources/dokka/docs/kdoc) for enums.
While it should not affect the output, it would be nice to have consistent logic for such cases.
So I would propose to at least make the logic consistent.
(Those were introduced in https://github.com/Kotlin/dokka/pull/2627, btw)
@vmishenev, WDYT?

dokka-subprojects/plugin-base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
whyoleg, line 1: nice test name :)
(from before the k1/k2 split)

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KdocMarkdownParser.kt
AbdullinAM: My original logic with `@throws` was that having a FQ is more intuitive, as we basically know that it always links to a class. Having a FQ there will be more informative.

But I agree with your point of making the behaviour more consistent. I will see how much it will affect the tests

AbdullinAM: I agree, we should try to make behaviour for `See` consistent everywhere. I will experiment with this to see if it affects the output

dokka-subprojects/plugin-base/src/test/kotlin/content/seealso/ContentForSeeAlsoTest.kt
```diff
@@ -216,7 +216,7 @@ class ContentForSeeAlsoTest : BaseAbstractTest() {
         }
     }
 
-    @OnlyDescriptors("issue #3179")
+//    @OnlyDescriptors("issue #3179")
```
whyoleg: Nice finding!
```suggestion
```

dokka-subprojects/plugin-base/src/test/kotlin/translators/DefaultDescriptorToDocumentableTranslatorTest.kt
```diff
@@ -14,6 +14,7 @@ import org.jetbrains.dokka.links.PointingToDeclaration
 import org.jetbrains.dokka.model.*
 import org.jetbrains.dokka.model.doc.*
 import utils.OnlyDescriptors
+import utils.OnlySymbols
```
whyoleg:
```suggestion
```

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/KdocMarkdownParser.kt
whyoleg, line 1: Nice idea to make the same behavior in K1 too!

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KdocMarkdownParser.kt
whyoleg: As you've marked the PR as ready for review: have you experimented with it? :)

AbdullinAM: Oh, this part slipped my mind while I was changing the K1 behaviour.
Tested it: it does not affect anything in tests. Pushing an update

PR #4257, opened 2025-09-15
Handle `InvocationTargetException` exception explicitly in `DokkaBootstrap`

### LINKED ISSUE #4235: Error about intersected source sets is not visible by default
It is probably relevant to any error produced by the dokka engine.

If a project has samples that are configured incorrectly (#3760) then the failure in 2.1.0 looks like:

```
> Task :mpp-module:dokkaGenerateModuleHtml FAILED
w: [:single-target-jvm:dokkaGenerateModuleHtml] Couldn't resolve link for HttpClient in KJvmAnotherClass.kt/org.jetbrains.qa.common.getHttpClientFromKtor

[Incubating] Problems report is available at: file:///Users/Andrei.Tyrin/IdeaProjects/dokka-multimodule-project/build/reports/problems/problems-report.html

* What went wrong:
Execution failed for task ':mpp-module:dokkaGenerateModuleHtml'.
> A failure occurred while executing org.jetbrains.dokka.gradle.workers.DokkaGeneratorWorker

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (Powered by Develocity).
> Get more help at https://help.gradle.org.

BUILD FAILED in 20s
128 actionable tasks: 105 executed, 23 up-to-date
Configuration cache entry stored.
```

And till I call it with `--stacktrace` it is not clear what exactly happened. With stacktrace:

```
Caused by: org.jetbrains.dokka.DokkaException: Pre-generation validity check failed: Source sets 'apple' and 'appleTest' have the common sample roots: /Users/Andrei.Tyrin/IdeaProjects/dokka-multimodule-project/mpp-module/src/kotlin/commonMain/samples.kt. Every Kotlin source file should belong to only one source set (module). 
```

In 2.0.0 with K2 opt-in the error was shown by default.

whyoleg:
Handle `InvocationTargetException` exception explicitly in `DokkaBootstrap` to show better error messages in DGPv2, as Gradle will not show the cause of `InvocationTargetException` without `--stacktrace` flag

Fixes #4235

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorFailureTest.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorFailureTest.kt
@@ -11,26 +11,7 @@ import org.jetbrains.dokka.gradle.utils.*
 
 class DokkaGeneratorFailureTest : FunSpec({
     context("DokkaGenerator failure:") {
-        val project = gradleKtsProjectTest("dokka-generator-failure") {
-            buildGradleKts = """
-                |plugins {
-                |    kotlin("jvm") version embeddedKotlinVersion
-                |    id("org.jetbrains.dokka") version "$DOKKA_VERSION"
-                |}
-                |
-                |dokka {
-                |  dokkaSourceSets.configureEach {
-                |    suppress = false // to enable documentation for `test` source set
-                |    sourceRoots.from("src/shared/kotlin") // causes an error in Dokka generator checker
-                |  }
-                |}
-                |
-                """.trimMargin()
-
-            createKotlinFile("src/main/kotlin/Main.kt", "class Main")
-            createKotlinFile("src/test/kotlin/Test.kt", "class Test")
-            createKotlinFile("src/shared/kotlin/Shared.kt", "class Shared")
-        }
+        val project = createProject()
 
         test("expect failure message from checkers to be shown by default") {
             project.runner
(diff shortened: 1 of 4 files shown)


dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorFailureTest.kt
```diff
@@ -0,0 +1,57 @@
...
+package org.jetbrains.dokka.gradle
+
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.string.shouldContain
+import org.jetbrains.dokka.gradle.internal.DokkaConstants.DOKKA_VERSION
+import org.jetbrains.dokka.gradle.utils.*
+
+class DokkaGeneratorFailureTest : FunSpec({
+    context("DokkaGenerator failure:") {
+        val project = gradleKtsProjectTest("dokka-generator-failure") {
```
vmishenev: Can you make this test more straightforward? Or even add one more with a custom exception/pre-generation check?
For example, if intersecting roots are allowed in the future, this case could be lost.
See [the original test ](https://github.com/Kotlin/dokka/blob/e59dc2a61b1c94e1a53c27555899392377e35f3c/dokka-runners/dokka-gradle-plugin/src/classicTest/kotlin/AutomagicProxyTest.kt#L32C66-L32C94)

vmishenev (approved):

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/DokkaGeneratorFailureTest.kt
whyoleg: Oh, nice idea!
I will add an additional test similar to the previous one.

It would still be nice to have a functional test for DGP, so that we really show the exception from checks or just failures here.

adam-enko:
post merge review: Looks good!

I think the ServiceLoader implementation would also help with error handling https://github.com/Kotlin/dokka/pull/4216

Issue #4240, opened 2025-09-01
dokkaGenerate fails with "Couldn't resolve link" using 2.1.0-Beta


jeffdgr8:
**Describe the bug**
The `dokkaGenerate` Gradle task fails with multiple "Couldn't resolve link" warnings using Dokka 2.1.0-Beta.

**Expected behaviour**
`dokkaGenerate `succeeds in generating the API docs, as it does in Dokka 2.0.0.

**To Reproduce**
Run `./gradlew dokkaGenerate` on my library, [Kotbase](https://github.com/jeffdgr8/kotbase), after setting the `dokka` version variable to `2.1.0-Beta` in `libs.versions.toml`.

**Dokka configuration**
```Kotlin
dependencies {
    dokkaPlugin(libs.dokka.versioning)
}

dokka {
    dokkaSourceSets.configureEach {
        includes.from("README.md")

        sourceLink {
            localDirectory.set(projectDir.resolve("src"))
            remoteUrl("https://github.com/jeffdgr8/kotbase/tree/main/${project.name}/src")
            remoteLineSuffix.set("#L")
        }

        externalDocumentationLinks {
            register("kotlinx.coroutines") {
                url("https://kotlinlang.org/api/kotlinx.coroutines/")
            }
            register("kotlinx-datetime") {
                url("https://kotlinlang.org/api/kotlinx-datetime/")
                packageListUrl("https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/package-list")
            }
            register("kotlinx-io") {
                url("https://kotlinlang.org/api/kotlinx-io/")
            }
        }
    }

    pluginsConfiguration.html {
        customAssets.from(rootProject.projectDir.resolve("docs/site/assets/images/logo-icon.svg"))
        footerMessage = "© 2022-${LocalDate.now().year} Jeff Lockhart"
    }
}
```

**Installation**
- Operating system: macOS
- Build tool: Gradle v8.12.1
- Dokka version: 2.1.0-Beta

**Additional context**
Truncated Gradle logs:
```
...
w: [:couchbase-lite-ee:dokkaGenerateModuleHtml] Couldn't resolve link for DatabaseConfiguration in ConfigurationFactoriesEE.kt/kotbase.newConfig
w: [:couchbase-lite-ee:dokkaGenerateModuleHtml] Couldn't resolve link for ReplicatorConfiguration in ConfigurationFactoriesEE.kt/kotbase.newConfig
[...]

Issue #4116, opened 2025-05-21
[K2] kotlinx-io: Error classes on Wasm platfrom


vmishenev:
![Image](https://github.com/user-attachments/assets/701cbe73-66a8-4d0b-bf0e-fdf219748201)

![Image](https://github.com/user-attachments/assets/0a9775f3-e9ce-488a-9956-99b25af9d2ec)

~~Might be related to #3636~~

vmishenev:
The problem is that **the classpath** of the wasm source set does not have stdlib:
```
Wasm Source set:
    sourceRoots =   
     0 =  "/tmp/junit1478378937306345798/project/core/wasm/src"
    classpath =     
     0 = "/tmp/junit1478378937306345798/project/core/wasm/src"
     1 =  "/tmp/junit1478378937306345798/project/core/common/src"
```
(depends on common)
For example,
- `wasm-js` has `kotlin-stdlib-wasm-js-2.1.0.klib`
- `wasm-wasi` has `kotlin-stdlib-wasm-wasi-2.1.0.klib`
- `js` has `kotlin-stdlib-js-2.1.0.klib`


K2 has a stricter project structure than K1.

vmishenev:
The source set wasm is a kind of intermediate source set, although it has a wasm platform instead of common.
Should the Analysis API resolve a class from a transitive common dependency? (see [the test](https://github.com/Kotlin/dokka/blob/3acae700eda91f8dc0ba2b9d1873d7c77ecd83c8/dokka-subprojects/plugin-base/src/test/kotlin/translators/KMPTest.kt#L73))

vmishenev:
> Should the Analysis API resolve a class from a transitive common dependency?

No, K2 Dokka's classpath is not propagated (see `directRegularDependencies` in the AA).


The current problem might be related to https://github.com/Kotlin/dokka/issues/3386#issuecomment-2127026886

vmishenev:
The expected classpath should contain the `kotlin-stdlib` klib like e.g. in the `nodeFilesystemShared` source set.

atyrin:
JIC: https://youtrack.jetbrains.com/issue/KT-61988/Wasm-common-sharing-code-between-wasm-targets

vmishenev:
Blocked by https://youtrack.jetbrains.com/issue/KT-80551/compileDependencyFiles-misses-some-dependencies-for-Wasm 
but we probably can have a workaround

PR #4258, opened 2025-09-15
Use `JVM` platform if the source-set is shared between jvm and android-jvm targets

whyoleg:
Found in https://github.com/Kotlin/dokka/issues/4240#issuecomment-3292162497

[Coil](https://github.com/coil-kt/coil), and maybe some other projects, have a custom shared source set which is shared between `jvm` and `android` targets. Sometimes, in that case, JDK classes could be used in those source sets. While Kotlin doesn't officially support this use case:
- [KT-76097](https://youtrack.jetbrains.com/issue/KT-76097) Add experimental support for shared JVM and Android source set
- [KT-42466](https://youtrack.jetbrains.com/issue/KT-42466) HMPP: JVM & Android intermediate source set
- [KT-42468](https://youtrack.jetbrains.com/issue/KT-42468) HMPP: JVM and Android intermediate source set publication
- and maybe more...

Still, IDEA does support this use case, and both types and KDoc links are resolved against the JDK classes:

image

WDYT about replicating this behavior in DGPv2, so that if there is a shared source set between `jvm` and `android`, we will use `JVM` instead of `common` platform during analysis?

As mentioned in https://github.com/Kotlin/dokka/issues/4240#issuecomment-3292162497, the same could be done on the user's side, which solves this issue.


atyrin:
Could you please add a test for the scenario? At least a simple integration.

adam-enko:
I think we'll also need to adjust this code to support webMain (or, more generally, any shared WasmJS/JS source sets) - see #4116.

I'm just sharing it as it comes to mind. We don't have to update the code in this PR, but  maybe there's an elegant way of fixing the problem here too?

vmishenev:
> I think we'll also need to adjust this code to support webMain

As I remember, there is the same issue with JS+WasmJS sourceset in Kotlin Wrappers

whyoleg:
> I think we'll also need to adjust this code to support webMain (or, more generally, any shared WasmJS/JS source sets)

I don't think it's the same case — those are two different issues, and they have two underlying mechanisms/problems.

Kotlin, at this moment, has the ability to have a shared source set, which is also analyzed not as `common`, but as a platform-specific one only for `native`. But, in case of shared source set between `jvm` and `android-jvm`, we can still try to analyze it as a `jvm`, because in most of the cases, it will work, as Android SDK has most of the JDK APIs. This is probably how it works in IDEA. And it's still just a workaround, which is nice to have, because, as I said, Kotlin doesn't properly support this, and so this intermediate source-set might have problems with publication/consumption (one of the issues in the description).

The case with a shared source set between `js` and `wasm-js` is totally different because in that situation, we will have a source set with different platforms. So, the only platform with which we could analyze that source set is `common`. And it's correct, and it's how klib backends and compilers work, AFAIK.

The case with `webMain`, shared between `js` and `wasm-js` or `wasm-js` and `wasm-wasi`, should work correctly if KGP provides correct classpaths, or we will somehow propagate those manually (https://github.com/Kotlin/dokka/issues/4116#issuecomment-3161843635)

whyoleg:
> Could you please add a test for the scenario? At least a simple integration.

The test is added in https://github.com/Kotlin/dokka/pull/4258/commits/1252dc9ecb691e35a93ab34289621df4fed44033, and thanks to it, a bug was discovered (https://github.com/Kotlin/dokka/pull/4258/commits/1a205ec7e00ca33414b1e5a62333270eb18c9e03), which is required to be fixed for the original changes to work correctly, as otherwise, `allPlatforms` will always have `common` platform coming from metadata target.

whyoleg:
blocked by #4263

vmishenev (approved):

PR #4260, opened 2025-09-16
Don't add Android's `bootClasspath` to ALL source-sets in case android plugin is applied in KMP project

whyoleg:
This is not required, but because of this, even `commonMain` in a multiplatform project that has the `android` plugin applied will have `android.jar` in the classpath, and so it could theoretically resolve to declarations from there.
(spotted while writing a test for https://github.com/Kotlin/dokka/pull/4258)

@adam-enko, do you have any suggestions on how to test this?


adam-enko:
> @adam-enko, do you have any suggestions on how to test this?

We want to check that `DokkaSourceSet.classpath` passed to Dokka generator contains `android.jar`. This could change based on the AGP version, so it should be an integration test.

So, add an assertion in the AGP integration tests, e.g. `dokka-integration-tests/gradle/src/testTemplateProjectAndroid/kotlin/Android0GradleIntegrationTest.kt`, that checks the Dokka configuration contains `android.jar`.

Does that sound doable?

PR #4261, opened 2025-09-16
Support for nested typealiases

AbdullinAM:


AbdullinAM:
Let me know if you have any ideas what test cases we should also add

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -429,6 +435,8 @@ internal class DokkaSymbolVisitor(
                 syntheticJavaProperties.map { visitPropertySymbol(it, dri) } +
                 javaFields.map { visitJavaFieldSymbol(it, dri) }
 
+        val typealiases = classifiers.filterIsInstance<KaTypeAliasSymbol>()
```
whyoleg, line 439: we should probably convert `classifiers` to `List` in place of it declaration above, in the same way as for callables, to prevent re-evaluating sequence coming from AA
I remember there were performance issues related to it. I don't think, that for classifiers it's a must, but it feels logical for me to do the same as for callables

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
```diff
@@ -153,6 +154,7 @@ public data class DClass(
     override val functions: List<DFunction>,
     override val properties: List<DProperty>,
     override val classlikes: List<DClasslike>,
+    override val typealiases: List<DTypeAlias>,
```
whyoleg: Let's try not to break backward compatibility in the similar way, how we done when added context parameters (#4080)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/DocumentableReplacerTransformer.kt
```diff
@@ -55,6 +56,7 @@ public abstract class DocumentableReplacerTransformer(
                 (classlike.takeIf { !wasClassChange } ?: classlike.copy(
                     functions = functions.mapNotNull { it.target },
                     classlikes = classlikes.mapNotNull { it.target },
+                    typealiases = typealiases.mapNotNull { it.target },
```
whyoleg, line 59: what about other classlikes below? the same logic should be there

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
```diff
@@ -205,6 +208,7 @@ public data class DEnumEntry(
...
```
whyoleg: oh, that's so funny, we have to have this (and `classlikes`) in `EnumEntry` :)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/ExtensionExtractorTransformer.kt
```diff
@@ -48,21 +48,27 @@ public class ExtensionExtractorTransformer : DocumentableTransformer {
...
 
             is DClass -> copy(
                 classlikes = newClasslikes.awaitAll(),
+                typealiases = newTypealiases.awaitAll(),
                 extra = extra + findExtensions(classGraph, extensionMap)
             )
 
             is DEnum -> copy(
                 classlikes = newClasslikes.awaitAll(),
+                typealiases = newTypealiases.awaitAll(),
```
whyoleg, line 71: the same should be added for `DInterface` below

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -644,15 +648,8 @@ public open class DefaultPageCreator(
...
-            // This groupBy should probably use LocationProvider
-            .groupBy(Documentable::name)
-            .mapValues { (_, elements) ->
-                // This hacks displaying actual typealias signatures along classlike ones
-                if (elements.any { it is DClasslike }) elements.filter { it !is DTypeAlias } else elements
-            }
-
-        val groups = grouped.entries
+        // This groupBy should probably use LocationProvider
+        val groups = types.groupBy(Documentable::name).entries
```
whyoleg, line 649: Curious, why we don't have tests failures after this change?
I feel like it could be related to case of `expect class A` + `actual typealias A` - such cases exists in kotlinx (io and coroutines at least) or stdlib, so we should probably check, how this will affect them? And add test if needed

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
whyoleg: We should also update `children` property to return typealiases

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -148,18 +148,21 @@ public open class DefaultPageCreator(
...
         val entries = classlikes.flatMap { if (it is DEnum) it.entries else emptyList() }
 
         val childrenPages = constructors.map(::pageForFunction) +
                 if (mergeImplicitExpectActualDeclarations)
                     nestedClasslikes.mergeClashingDocumentable().map(::pageForClasslikes) +
                             functions.mergeClashingDocumentable().map(::pageForFunctions) +
                             props.mergeClashingDocumentable().map(::pageForProperties) +
-                            entries.mergeClashingDocumentable().map(::pageForEnumEntries)
+                            entries.mergeClashingDocumentable().map(::pageForEnumEntries) +
+                            typealiases.mergeClashingDocumentable().map(::pageForClasslikes)
```
whyoleg: should we here do the same as in packages?
I mean do `(nestedClasslikes + typealiases).mergeClashingDocumentable()...`?

Maybe in case of something like this:
```kotlin
//common
expect class A
//jvm
actual class A {
    class B
}
//native
actual class A {
    typealias B = String
}
```

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -148,18 +148,21 @@ public open class DefaultPageCreator(
...
                 else
                     nestedClasslikes.renameClashingDocumentable().map(::pageForClasslike) +
-                            (functions + props).renameClashingDocumentable().mapNotNull(::pageForMember)  +
-                            entries.renameClashingDocumentable().map(::pageForEnumEntry)
+                            (functions + props).renameClashingDocumentable().mapNotNull(::pageForMember) +
+                            entries.renameClashingDocumentable().map(::pageForEnumEntry) +
+                            typealiases.mergeClashingDocumentable().map(::pageForClasslikes)
```
whyoleg: `renameClashingDocumentable` should be used here (also, comment above)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -397,6 +400,7 @@ public open class DefaultPageCreator(
         dri = @Suppress("UNCHECKED_CAST") (scopes as List<Documentable>).dri,
         sourceSets = sourceSets,
         types = scopes.flatMap { it.classlikes } +
+                scopes.flatMap { it.typealiases } +
                 scopes.filterIsInstance<DPackage>().flatMap { it.typealiases },
```
whyoleg: this line should be dropped

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
AbdullinAM: Not anymore :)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
AbdullinAM: This is actually related to the example from below:
```
//common
expect class A
//jvm
actual class A {
    class B
}
//native
actual class A {
    typealias B = String
}
```

Before the change class `B` and typealias `B` are not merged together. After the change --- they are merged. Which is the same behaviour we expect if both of them are regular classes

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -436,15 +438,15 @@ internal class DokkaSymbolVisitor(
...
 
-        fun Sequence<KaNamedClassSymbol>.filterOutCompanion() =
+        fun List<KaNamedClassSymbol>.filterOutCompanion() =
                 filterNot {
                     it.classKind == KaClassKind.COMPANION_OBJECT
                 }
 
         val classlikes = classifiers.filterIsInstance<KaNamedClassSymbol>()
-            .filterOutCompanion() // also, this is a hack to filter out companion for enum
+            .filterOutCompanion()
```
whyoleg, line 448: minor: below in `classlikesWithoutCompanion = classlikes.toList()` there is now a redundant `toList()` call :)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/pages/annotations/SinceKotlinTransformer.kt
```diff
@@ -104,35 +104,39 @@ public class SinceKotlinTransformer(
...
+                typealiases = typealiases.map { it.transform() as DTypeAlias },
             )
 
             is DEnum -> copy(
                 documentation = appendSinceKotlin(versions),
                 classlikes = classlikes.map { it.transform(versions) as DClasslike },
                 functions = functions.map { it.transform(versions) as DFunction },
-                properties = properties.map { it.transform(versions) as DProperty }
+                properties = properties.map { it.transform(versions) as DProperty },
+                typealiases = typealiases.map { it.transform() as DTypeAlias },
```
vmishenev: `it.transform(versions)`

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/SuppressedByConditionDocumentableFilterTransformer.kt
```diff
@@ -48,6 +48,7 @@ public abstract class SuppressedByConditionDocumentableFilterTransformer(
 
         val functions = classlike.functions.map { processMember(it) }
         val classlikes = classlike.classlikes.map { processClassLike(it) }
+        val typealiases = (classlike as? WithTypealiases)?.typealiases?.map { processMember(it) }.orEmpty()
```
vmishenev, line 51: `wasClasslikeChanged` should be changed according to   `typealiases`:
```kt
val wasClasslikeChanged = (functions + classlikes + properties + typealiases)...
```

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
```diff
@@ -328,8 +335,10 @@ public data class DInterface(
     override val modifier: SourceSetDependent<Modifier>,
     override val sourceSets: Set<DokkaSourceSet>,
     override val isExpectActual: Boolean,
-    override val extra: PropertyContainer<DInterface> = PropertyContainer.empty()
-) : DClasslike(), WithAbstraction, WithCompanion, WithGenerics, WithSupertypes, WithExtraProperties<DInterface> {
+    override val extra: PropertyContainer<DInterface> = PropertyContainer.empty(),
+    override val typealiases: List<DTypeAlias> = emptyList(),
+) : DClasslike(), WithAbstraction, WithCompanion, WithGenerics, WithSupertypes, WithExtraProperties<DInterface>,
+    WithTypealiases {
     override val children: List<Documentable>
```
vmishenev, line 498: Should we put `typealiases` into `children` here and below?

vmishenev (commented):
`@whyoleg`  provided a good review.
It would be nice to check not only the Content model, but also the Documentable model.

Also, some tests can be edited/added for each modified transformer, e.g. https://github.com/Kotlin/dokka/blob/f524273591ee9d360238e0f03ee63e7b9c5002b0/dokka-subprojects/plugin-base/src/test/kotlin/filter/KotlinArrayDocumentableReplacerTest.kt#L118

dokka-subprojects/plugin-base/src/test/kotlin/content/typealiases/NestedTypealiasTest.kt
```diff
@@ -0,0 +1,631 @@
...
+                                }
+                            }
+                        }
+
+                        table2("Types") {
+                            element("A") {
+                                divergentInstance {
+                                    group4 {
+                                        +"typealias "
+                                        group { groupedLink { +"Foo.A" } }
```
vmishenev: Sorry, I haven't noticed.
Should we render a nested typealias as `A` or `Foo.A`?
cc @whyoleg

AbdullinAM: I think that to be more consistent, we need to show only `A` in the typealias declaration itself, and `Foo.A` in the use-cases. This will be similar to how nested classes are displayed currently

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
```diff
@@ -413,7 +413,7 @@ public class KotlinSignatureProvider(
                 processExtraModifiers(t)
                 keyword("typealias ")
                 group(styles = mainStyles + t.stylesIfDeprecated(sourceSet)) {
-                    signatureForProjection(t.type)
+                    signatureForProjection(t.type, linkText = t.name)
```
whyoleg: TBH, this feels a bit like a hack (hope that it will not backfire on us)
Maybe, it will be better to call this parameter somehow `typealias` related, as it's really safe to use, only when we know that `p` will be a `GenericTypeConstructor` - as in `DTypealias`.

Or we can even not use `signatureForProjection` here, as it feels like an abuse, as it's not really a projection here, but a declaration itself (IMO). In that case, it might be possible to copy or reuse the logic that is used for other class-like declarations [here](https://github.com/Kotlin/dokka/blob/e76721618b9c12c84154b16d680aa2c800c00659/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt#L188-L196).

Still, feel free to ignore. I'm not insisting here.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt
```diff
@@ -163,12 +168,13 @@ public data class DClass(
     override val modifier: SourceSetDependent<Modifier>,
     override val sourceSets: Set<DokkaSourceSet>,
     override val isExpectActual: Boolean,
-    override val extra: PropertyContainer<DClass> = PropertyContainer.empty()
+    override val extra: PropertyContainer<DClass> = PropertyContainer.empty(),
+    override val typealiases: List<DTypeAlias> = emptyList(),
```
whyoleg, line 172: Please add a deprecated constructor and copy it as with `contextParameters` in `DFunction` to maintain the ABI.

whyoleg (changes requested):
Nice!
Please, fix the ABI, similar to how it's done in #4080, and we are good to go

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
AbdullinAM: Replaced the use of `signatureForProjection` with the same logic that is used for other class-likes

whyoleg:
The TeamCity IT setup is now fixed, and the CI is green. PR could be merged :)

PR #4263, opened 2025-09-17
DGPv2: Fix `isMetadata` flag

whyoleg:
Contains IT from https://github.com/Kotlin/dokka/pull/4268 (reproducer for [KT-81108](https://youtrack.jetbrains.com/issue/KT-81108)) - working correctly now.

As shown in the changed IT expectedData, we now treat source sets like `linuxMain` as `native` and analyze them using the `native` analysis platform.


whyoleg:
AA issue: [KT-81108](https://youtrack.jetbrains.com/issue/KT-81108) AA: java.lang.ClassCastException: class org.jetbrains.kotlin.fir.FirBinaryDependenciesModuleData cannot be cast to class org.jetbrains.kotlin.analysis.low.level.api.fir.projectStructure.LLFirModuleData

vmishenev:
> @vmishenev, have you seen similar errors before? I

No, I don't remember.
Can you create a unit test only for this class https://github.com/Kotlin/kotlinx-io/blob/b1db70bb9822b0095b13626d955ddc8ebc1e7f17/core/androidNative/src/files/FileSystemAndroid.kt#L31  ?
It would be useful for Dokka as well

whyoleg:
> Can you create a unit test

I can't, as it depends on K/N platform libraries :(
But I will try to create an integration test and check if it reproducible there

whyoleg:
Reproduced the issue from kotlinx.io in https://github.com/Kotlin/dokka/commit/59286fe4a9915c211594e2a20131f00a196b7fe7 - I will create an issue in Dokka to track this

(also, I will extract test changes into one more PR)

vmishenev:
> I can't, as it depends on K/N platform libraries :( 

You can copy the required subset of them for tests.

whyoleg:
> You can copy the required subset of them for tests.

But it fails specifically because of `cnames.structs.DIR` in the constructor, which is a bit of a special case in K/N and comes from platform libraries (or cinterop). If I use anything else in the constructor, it will not fail.
So I'm not sure that we could do a unit test for it...

adam-enko:
`:dokka-integration-tests:cli:` is failing on TC - is that unrelated to this PR?

whyoleg:
> :dokka-integration-tests:cli: is failing on TC - is that unrelated to this PR?

@adam-enko, no, it should be fixed by #4324. Rebased the PR.

adam-enko (approved):

PR #4267, opened 2025-09-18
Re-enable kotlin multiplatform test and add even more tests for assertions!

whyoleg:
For some reason, the test was disabled, while in reality, everything is resolved correctly :)

Additionally, `assertNoUnknownClassErrorsInHtml` is now used in all tests in `test` folder


dokka-integration-tests/gradle/projects/it-kotlin-multiplatform/src/iosArm64Main/kotlin/iosX64Fn.kt
```diff
@@ -4,7 +4,7 @@ package a.b.c
  * A `iosX64Main` function that uses:
  * - [CommonMainCls] from `commonMain`
  * - [NativeMainCls] from `nativeMain`
- * - [LinuxMainCls] from `linuxMain`
+ * - [IosMainCls] from `iosMain`
```
whyoleg, line 7: This is the only link that wasn't resolved previously - and that's right! You can't resolve function from `linuxMain` in `ios*Main`
(same change for `iosX64Main` source set below)

adam-enko (approved):

PR #4268, opened 2025-09-18
Reproduce kotlinx.io issue with `cnames` usage

whyoleg:
Depends on #4263
Reproducer for [KT-81108](https://youtrack.jetbrains.com/issue/KT-81108) AA: java.lang.ClassCastException: class org.jetbrains.kotlin.fir.FirBinaryDependenciesModuleData cannot be cast to class org.jetbrains.kotlin.analysis.low.level.api.fir.projectStructure.LLFirModuleData


whyoleg:
Incorporated into #4263

PR #4271, opened 2025-09-19
Drop usages of APIs deprecated in KGP

whyoleg:
[Internal thread](https://jetbrains.slack.com/archives/C06QB5UHN9Y/p1758277314200869)

This PR makes DGPv1 compatible with KGP 1.9 and later. That's not an issue, because DGPv2 has the same compatibility policy, and DGPv1 supported those only for historical reasons.


whyoleg:
@atyrin FYI

PR #4276, opened 2025-09-20
Fix extension properties generating fake field

### LINKED ISSUE #4173: `kotlin-as-java-plugin` not rendering extension properties correctly
Given the code 
```kotlin
val String.test: String
    get() = this + " test"
```
the `kotlin-as-java-plugin` produces 

Image

A spurious, nonexistent field is created, as well as messing up the signature for the function: it's supposed to be `getTest(String $self)`, not `String getTest()`

Seggan:
Fixes #4173


Seggan:
Only fixes the specific extension property case, not the more general non-backed property case

PR #4280, opened 2025-09-24
Fix backwards interface method modifiers

### LINKED ISSUE #4279: Default methods in interfaces are backwards in the Javadoc format
For the code 
```kotlin
interface Test {
    fun foo()

    fun bar() {}
}
```
The Javadoc format generates the abstract method as `abstract` and the default method as just a normal method:

Image

However, in real Javadoc, it should be the opposite; i.e., the abstract method should render normally and the default method have a `default`:

Image

Seggan:
Fixes #4279

### DIFF
--- a/dokka-subprojects/core/api/dokka-core.api
+++ b/dokka-subprojects/core/api/dokka-core.api
@@ -1590,6 +1590,10 @@ public final class org/jetbrains/dokka/model/ExtraModifiers$JavaOnlyModifiers$Vo
 	public static final field INSTANCE Lorg/jetbrains/dokka/model/ExtraModifiers$JavaOnlyModifiers$Volatile;
 }
 
+public final class org/jetbrains/dokka/model/ExtraModifiers$JavaOnlyModifiers$Default : org/jetbrains/dokka/model/ExtraModifiers$JavaOnlyModifiers {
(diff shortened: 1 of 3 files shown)


Seggan:
I cannot figure out how to write the test as 
```
    @Test
    fun `interface methods should have default when needed`() {
        val source = source(
            """
            interface A {
                fun b() {}
                fun c()
            }
            """
        )
        val writerPlugin = TestOutputWriterPlugin()

        testInline(
            source,
            configuration,
            pluginOverrides = listOf(writerPlugin)
        ) {
            renderingStage = { _, _ ->
                val signatures = writerPlugin.writer
                    .renderedContent("root/kotlinAsJavaPlugin/-a/index.html")
                    .signature()

                val methodWithBody = signatures[1]
                methodWithBody.match(
                    "public default ", A("Unit"), A("b"), "()"
                )

                val methodWithoutBody = signatures[2]
                methodWithoutBody.match(
                    "public ", A("Unit"), A("c"), "()"
                )
            }
        }
    }
```
just throws `"public default " expected but found: <a href="https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-unit/index.html">Unit</a>`

PR #4283, opened 2025-09-26
Fix property docs not being propagated to generated accessors

### LINKED ISSUE #3369: [K2] Generated getter for property in kotlin-as-java doesn't inherit kdoc description

Seggan:
Fixes #3369


AbdullinAM (commented):
Hi! Thank you for the PR!


Overall the fix looks good. However, when discussing this topic in a separate PR (#4370) we decided that in would be more suitable and consistent to handle this 'inheritance' for property accessors in Analysis API. 

We created an [issue](https://youtrack.jetbrains.com/issue/KT-82851/Property-accessors-should-inherit-KDoc-from-the-parent-property) explaining the change and it was implemented already, we're waiting for the next release of AA to have it available in Dokka.

Seggan:
Alright makes sense, should I close this then?

AbdullinAM:
@Seggan Yes, I think we can close this PR. Thank you once again for creating it, and sorry for a (very) late reply

PR #4286, opened 2025-09-27
Rework the Gradle attributes to make them integrate better with the ecosystem

martinbonnin:
See https://github.com/martinbonnin/dokka/blob/496313e42aac29664eeb730ab77dc96a93e214ec/dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaFormatPlugin.kt#L153 for a description of this PR.

tldr; this introduces a specific `dokka-java-runtime` usage to avoid clashing with other tools in the ecosystem.

Closes https://github.com/adamko-dev/dokkatoo/issues/165


Vampire:
Before considering this PR, I'd like to suggest reviewing my analysis of the problem just posted to https://github.com/adamko-dev/dokkatoo/issues/165#issuecomment-3358646296 :-)

adam-enko:
Thanks very much for the contribution @martinbonnin. 

I've copied your commits into a branch in the repo so we can run the tests, and made some modifications. We can keep progressing in https://github.com/Kotlin/dokka/pull/4334

PR #4290, opened 2025-10-06
Replace createTempDir with createTempDirectory

fzhinkin:
KUP start failing once `createTempDir` was deprecated with error (https://github.com/JetBrains/kotlin/commit/852ea9f86e76dc1442cdab05a222a2e945e7ea21).

### DIFF
--- a/dokka-subprojects/core-test-api/src/main/kotlin/org/jetbrains/dokka/testApi/testRunner/TestRunner.kt
+++ b/dokka-subprojects/core-test-api/src/main/kotlin/org/jetbrains/dokka/testApi/testRunner/TestRunner.kt
@@ -21,6 +21,7 @@ import java.nio.file.Files
 import java.nio.file.InvalidPathException
 import java.nio.file.Path
 import java.nio.file.Paths
+import kotlin.io.path.ExperimentalPathApi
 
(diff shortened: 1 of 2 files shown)


dokka-subprojects/core-test-api/src/main/kotlin/org/jetbrains/dokka/testApi/testRunner/TestRunner.kt
```diff
@@ -170,8 +171,8 @@ public abstract class AbstractTest<M : TestMethods, T : TestBuilder<M>, D : Dokk
         Files.write(file, content.toByteArray(charset))
     }
 
-    @Suppress("DEPRECATION") // TODO migrate to kotlin.io.path.createTempDirectory with languageVersion >= 1.5
-    private fun createTempDir(): File = kotlin.io.createTempDir()
+    @OptIn(ExperimentalPathApi::class)
+    private fun createTempDir(): File = kotlin.io.path.createTempDirectory().toFile()
```
whyoleg: minor: it feels like we can easily avoid using experimental APIs here and just write `Files.createTempDirectory("dokka-test").toFile()` :)
(minor, because it's a test source-set)

PR #4295, opened 2025-10-14
Support AGP with Kotlin Built-in

### LINKED ISSUE #4256: AGP 9 support
**Describe the bug**
Android Gradle Plugin 9 (`9.0.0-alpha05`) doesn't seem to be supported. All projects using Dokka are logging this:
> &gt; Configure project :my-project
> class org.jetbrains.dokka.gradle.adapters.AndroidExtensionWrapper could not get Android Extension for project :my-project

**Expected behaviour**
Dokka should behave properly on AGP 9, just like it works fine on AGP `8.13.0`.

**To Reproduce**
In an Android project using Dokka v2, change the AGP dependencies to use (the latest preview of) version 9.

**Installation**
- Operating system: macOS
- Build tool: Gradle 9.0.0
- Dokka version: 2.0.0

**Additional context**
I'm aware that AGP 9 is in preview status right now, but I figured it's a good idea to open an issue early for the [breaking API changes](https://developer.android.com/build/releases/agp-preview) done in AGP.

adam-enko:
When using AGP 9, KGP is no longer required. AGP will be responsible for configuring Kotlin compilation. DGP needs to support Kotlin Built-in by extracting information from AGP instead of KGP.

Fix #4256

### Summary of changes

- Update KotlinAdapter and AndroidAdapter to handle AGP with Kotlin built-in.
  - Capture classpaths of variants from `AndroidComponentsExtension` instead of the Android extension.
  - Update captured AGP variant info to include classpaths.
  - Only apply KotlinAdapter and AndroidAdapter once per project (Remove the `exec()` function, only apply the plugin once using `KotlinAdapter.applyTo()`).
- Add tests for AGP 9.
  - `it-android-kotlin-jvm-builtin` - AGP & kotlin-built-in (no KGP plugin).
  - `it-android-kotlin-mp-builtin` - AGP & kotlin-multiplatform.
- Fix AGP <9 tests & update test data.

#### Integration tests: support AGP 9 testing and version filtering 

- Add min/max version to `@TestsAndroid` and `@TestsAndroidCompose` filter AGP versions.
- Add SemVerRange to help with filtering.
- Add AGP9 properties.

### DIFF
--- a/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
+++ b/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
@@ -34,10 +34,6 @@ class AndroidKotlinMultiplatformBuiltInTest {
         testedVersions: TestedVersions.Android,
     ) {
         assumeTrue(
-            testedVersions.agp.major < 9,
-            "TODO Support AGP with kotlin-built-in https://github.com/Kotlin/dokka/pull/4295"
-        )
-        assumeTrue(
             testedVersions.agp >= SemVer("8.10.0"),
             "The com.android.kotlin.multiplatform.library plugin is only available in AGP 8.10.0+ " +
                     "(current AGP ${testedVersions.agp}). " +
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginIds.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginIds.kt
@@ -21,5 +21,6 @@ internal object PluginIds {
         "com.android.library",
         "com.android.test",
         "com.android.dynamic-feature",
+        "com.android.kotlin.multiplatform.library",
     )
 }


dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
```diff
@@ -116,51 +124,48 @@ private interface AndroidExtensionWrapper {
...
-                        }
+                private val androidVariants: SetProperty<AndroidVariantInfo> =
+                    objects.setProperty(AndroidVariantInfo::class)
+
+                /**
+                 * Get the `android.jar` for the current project.
+                 *
+                 * Need to double-wrap with [Provider] because AGP will only
+                 * compute the boot classpath after the compilation options have been finalized.
+                 * Otherwise, AGP throws an exception.
```
ALikhachev: Does it mean you can't even acquire the Provider object until some point in time? Isn't it a bug?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
```diff
@@ -116,51 +124,48 @@ private interface AndroidExtensionWrapper {
...
-                            }
-                            .artifacts
-                            .resolvedArtifacts
-                            .map { artifacts -> artifacts.map(ResolvedArtifactResult::getFile) }
-
-                        androidComponentsCompileClasspath.from(artifactTypeFiles)
-                    }
+                init {
+                    collectAndroidVariants(androidComponents, androidVariants)
+                }
```
ALikhachev: nit: IMO, it would be clearer if you called it in the function body instead of the init block.

Like
```
val androidVariants = objects.setProperty(AndroidVariantInfo::class)
collectAndroidVariants(androidComponents, androidVariants)
return object : AndroidExtensionWrapper {
...
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -210,6 +191,73 @@ abstract class KotlinAdapter @Inject constructor(
...
+        /**
+         * Check all plugins to see if they are a subtype of [KotlinBasePlugin].
+         * If any are, log a warning.
+         *
+         * ##### Motivation
+         *
+         * If the buildscript classpath is inconsistent, it might not be possible for DGP
+         * to react to KGP because the [KotlinBasePlugin] class can't be loaded.
+         * If so, DGP will be lenient and not cause errors,
+         * but it must display a prominent warning to help users find the problem.
```
ALikhachev, line 246: but what would be my actions as a user if I see this warning? Shall I report it? Shall I do something with the project?

adam-enko: Users can try and fix the problem, workaround it, or report it.

There are too many potential causes and solutions to detail them in an error message.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
adam-enko: updated

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
adam-enko: I think it's intentional AGP behaviour. The provider depends on `compileSdkVersion`, which does not use the Provider API.

https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/BaseExtension.kt;l=472?q=BaseExtension.kt

adam-enko: Actually, I think that comment is based on the old method of getting the bootClasspath from the Android extension.

https://github.com/Kotlin/dokka/blob/79034952d5cb247cf04f6a83356cbbe55d808d60/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt#L163

It doesn't seem to be applicable to AndroidComponentsExtension.

adam-enko: I removed the double-provider and the AndroidProjectIT test failed:

```
  AndroidExtensionWrapper: could not find AndroidComponentsExtension
  java.lang.IllegalStateException: targetCompatibility is not yet finalized
    at com.android.build.gradle.internal.CompileOptions.getTargetCompatibility(CompileOptions.kt:52)
    at com.android.build.gradle.internal.CompileOptions$AgpDecorated_Decorated.getTargetCompatibility(Unknown Source)
    at com.android.build.gradle.internal.tasks.factory.BootClasspathConfigImpl$1.invoke(BootClasspathConfigImpl.kt:57)
    at com.android.build.gradle.internal.tasks.factory.BootClasspathConfigImpl$1.invoke(BootClasspathConfigImpl.kt:57)
    at com.android.build.gradle.internal.tasks.factory.BootClasspathConfigImpl$bootClasspath$2.invoke(BootClasspathConfigImpl.kt:173)
    at com.android.build.gradle.internal.tasks.factory.BootClasspathConfigImpl$bootClasspath$2.invoke(BootClasspathConfigImpl.kt:166)
    at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
    at com.android.build.gradle.internal.tasks.factory.BootClasspathConfigImpl.getBootClasspath(BootClasspathConfigImpl.kt:166)
    at org.gradle.api.internal.provider.DefaultProvider.calculateOwnValue(DefaultProvider.java:73)
    at org.gradle.api.internal.provider.AbstractMinimalProvider.calculateOwnPresentValue(AbstractMinimalProvider.java:82)
    at org.gradle.api.internal.provider.AbstractMinimalProvider.get(AbstractMinimalProvider.java:102)
    at com.android.build.gradle.internal.dsl.SdkComponentsImpl.getBootClasspath(SdkComponentsImpl.kt:63)
    at com.android.build.gradle.internal.dsl.SdkComponentsImpl_Decorated.getBootClasspath(Unknown Source)
    at org.jetbrains.dokka.gradle.adapters.AndroidExtensionWrapper$Companion$forAndroidComponents$1.<init>(AndroidAdapter.kt:143)
    at org.jetbrains.dokka.gradle.adapters.AndroidExtensionWrapper$Companion.forAndroidComponents(AndroidAdapter.kt:135)
    at org.jetbrains.dokka.gradle.adapters.AndroidAdapterKt.AndroidExtensionWrapper(AndroidAdapter.kt:101)
    at org.jetbrains.dokka.gradle.adapters.AndroidAdapterKt.access$AndroidExtensionWrapper(AndroidAdapter.kt:1)
    at org.jetbrains.dokka.gradle.adapters.AndroidAdapter.apply(AndroidAdapter.kt:45)
    at org.jetbrains.dokka.gradle.adapters.AndroidAdapter.apply(AndroidAdapter.kt:35)
```

I'll add the double-provider back.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidVariantInfo.kt
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @param[name] [Variant.name].
+ * @param[hasPublishedComponent] `true` if any component of the variant is 'published',
+ * i.e. it is an instance of [Variant].
+ */
+internal data class AndroidVariantInfo(
+    val name: String,
+    val hasPublishedComponent: Boolean,
+    val compileClasspath: FileCollection,
+    val compileConfiguration: Configuration,
```
whyoleg: looks like it's unused

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/JavaAdapter.kt
whyoleg: I believe we can also extract those into a property similar to `PluginId.kgpPlugins` and then use it in `AndroidAdapter.applyTo` and `KotlinCompilationDetailsBuilder.getAgpVariantInfo` so that we have those IDs in a single place.
Ideally, in this case, we should expose only `kgpPlugins` and `agpPlugins` properties from `PluginId`, so that we don't accidentally use other ones :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -210,6 +191,73 @@ abstract class KotlinAdapter @Inject constructor(
...
+
+        /**
+         * Applies [KotlinAdapter] to the current project when any plugin of type [KotlinBasePlugin]
+         * is applied.
+         *
+         * [KotlinBasePlugin] is the parent type for the Kotlin/JVM, Kotlin/Multiplatform, Kotlin/JS plugins,
+         * as well as AGP's kotlin-built-in plugin.
+         */
+        internal fun applyTo(project: Project) {
+            findKotlinBasePlugins(project)?.all {
```
whyoleg, line 204: Question: `AndroidAdapter` additionally has `project.plugins.withType<DokkaBasePlugin>().all {` - should we have it here?
IMO, we shouldn't have it at all, as those are internal plugins which we apply from `DokkaBasePlugin`. Others shouldn't apply them anyway.
I'm just trying to be consistent, no more.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -210,6 +191,73 @@ abstract class KotlinAdapter @Inject constructor(
             val kgpVersion = getKotlinPluginVersion(logger)
             KotlinToolingVersion(kgpVersion)
         }
+
+        /**
+         * Applies [KotlinAdapter] to the current project when any plugin of type [KotlinBasePlugin]
+         * is applied.
+         *
+         * [KotlinBasePlugin] is the parent type for the Kotlin/JVM, Kotlin/Multiplatform, Kotlin/JS plugins,
+         * as well as AGP's kotlin-built-in plugin.
```
whyoleg, line 201: So, correct me if I'm wrong: AGP9 comes with dependency on `kotlin-gradle-plugin-api` of some version (e.g., 2.2.0) and so for the `KotlinAdapter` to be usable with AGP9 we can use only types from that artefact, right?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaFormatPlugin.kt
```diff
@@ -65,9 +65,9 @@ abstract class DokkaFormatPlugin(
         target.pluginManager.apply(DokkaBasePlugin::class)
 
         // apply the plugin that will autoconfigure Dokka to use the sources of a Kotlin project
-        target.pluginManager.apply(type = KotlinAdapter::class)
+        KotlinAdapter.applyTo(target)
+        AndroidAdapter.applyTo(target)
         target.pluginManager.apply(type = JavaAdapter::class)
```
whyoleg: minor: it would be better if `JavaAdapter` followed the same apply convention as others

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MultiModuleFunctionalTest.kt
```diff
@@ -518,19 +518,33 @@ class MultiModuleFunctionalTest : FunSpec({
...
                     "clean",
                     "--stacktrace",
+                    "--info",
                 )
-                .forwardOutput()
                 .build {
-                    // the root project doesn't have the KGP applied, so KotlinProjectExtension shouldn't be applied
-                    output shouldNotContain "Dokka Gradle plugin failed to get extension kotlin org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension in :\n"
+                    // the root project doesn't have KGP applied, so KotlinAdapter shouldn't be applied
+                    output shouldContain "Dokka Gradle Plugin could not load KotlinBasePlugin in project root project 'kpe-warning'"
```
whyoleg: is `project root project` expected?

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MultiModuleFunctionalTest.kt
```diff
@@ -518,19 +518,33 @@ class MultiModuleFunctionalTest : FunSpec({
...
+                .addArguments(
+                    "clean",
+                    "--stacktrace",
+                    "--warn",
+                )
+                .build {
+                    // the subprojects should have KotlinAdapter applied, but the extension should be unavailable
+                    // because the buildscript classpath is inconsistent.
+                    // (DGP is applied to the root project, but KGP is not.)
+                    output shouldContain "Dokka Gradle Plugin could not load KotlinBasePlugin in project project ':subproject-hello', but plugin org.jetbrains.kotlin.jvm is applied"
```
whyoleg: is `project project` expected?

dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
```diff
@@ -8,7 +8,11 @@ android {
     namespace = "org.jetbrains.dokka.it.android"
     defaultConfig {
         minSdkVersion(21)
-        setCompileSdkVersion(29)
+        setCompileSdkVersion(30)
```
whyoleg, line 11: Curious: why this change? is it why we have test data update? Or because of some other change?
Ideally, if test data changed because of it, it would be better to extract this into separate PR, as it's feels unrelated to AGP9 update for me, and can hide some underlying changes in the PR itself

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+    defaultConfig {
+        minSdk = 21
+    }
+}
+
+dependencies {
+    implementation("androidx.appcompat:appcompat:1.1.0")
+}
+
+// WORKAROUND https://github.com/Kotlin/dokka/issues/3701
```
whyoleg: Could you expand on why this happens?
We don't have this issue with a KGP+AGP8 combination AFAIK

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/gradle.properties
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+#
+
+android.useAndroidX=true
+#android.builtInKotlin=true
```
whyoleg: Drop?

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/gradle.properties
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+#
+
+android.useAndroidX=true
+#android.builtInKotlin=true
```
whyoleg: drop?

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+kotlin {
+    androidLibrary {
+        namespace = "com.example.kmpfirstlib"
+        compileSdk = 33
+        minSdk = 24
+
+        withHostTestBuilder {}.configure {}
+        withDeviceTestBuilder {
+            sourceSetTreeName = "test"
+        }
```
whyoleg: feels redundant

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+        getByName("androidHostTest") {
+            dependencies {
+            }
+        }
+
+        getByName("androidDeviceTest") {
+            dependencies {
+            }
+        }
```
whyoleg: feels redundant

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.jetbrains.dokka.it.gradle.junit.*
+import kotlin.io.path.name
+import kotlin.io.path.readText
+
+/**
+ * Integration test for the `it-android-kotlin-mp-builtin` project.
+ */
+@TestsDGPv2
+@TestsAndroid(
+    minAgpVersion = "9.0.0",
```
whyoleg: Documentation says, that it should be supported from 8.10.0 (https://developer.android.com/kotlin/multiplatform/plugin#:~:text=kotlin.,better%20integration%20with%20Android%20Studio.).
Can we try to test it also on AGP 8.13.0 (latest AGP8)?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidVariantInfo.kt
adam-enko: removed

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/JavaAdapter.kt
adam-enko: done

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I think in theory, yes, that's right. It's a little more tricky in practice however, since Gradle bundles Kotlin Gradle plugin.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaFormatPlugin.kt
adam-enko: done

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/gradle.properties
adam-enko: done

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/gradle.properties
adam-enko: done

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
adam-enko: Makes sense - removed

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
adam-enko: removed

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MultiModuleFunctionalTest.kt
adam-enko: good spot! fixed

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MultiModuleFunctionalTest.kt
adam-enko: good spot! fixed

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
adam-enko: Without it when I run `gradle :dokka-integration-tests:gradle:test` I see a failure for `org.jetbrains.dokka.it.gradle.AndroidKotlinJvmBuiltInTest`.

```
Execution failed for task ':dokkaGeneratePublicationHtml'.
> A failure occurred while executing org.jetbrains.dokka.gradle.workers.DokkaGeneratorWorker
   > Pre-generation validity check failed: Source sets 'debug' and 'release' have the common source roots: /private/var/folders/_3/ps18c9f56bx2j4fyxt0z7sn40000gp/T/gradle-test17356012250814683624/generate-dokka-HTML-DokkaGradleProjectRunner-/agp-9.0.0-alpha13_dgp-2.1.0-SNAPSHOT_gradle-9.1.0_kgp-2.1.21/src/main/java. Every Kotlin source file should belong to only one source set (module). 
     Also, please consider reporting your user case: https://github.com/Kotlin/dokka/issues/3701
```

dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
adam-enko: Good question. I think it's not necessary here, but it is for the `it-multiplatform-android-jvm` project. Removing it there results in `MultiplatformAndroidJvmProjectIT` failing:

```
> Inconsistent JVM-target compatibility detected for tasks 'compileDebugJavaWithJavac' (1.8) and 'compileDebugKotlinAndroid' (17).
  
  Consider using JVM Toolchain: https://kotl.in/gradle/jvm/toolchain
  Learn more about JVM-target validation: https://kotl.in/gradle/jvm/target-validation 
```

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
adam-enko: Okay, I see your point now! DGP's logic for determining the publishability is flawed. Currently `debug` has `hasPublishedComponent=true`, but this is incorrect.

I'll try and update DGP's logic...

adam-enko: Updated.

DGP also checks to see if `Variant.buildType == "release"`. This means the debug source sets are filtered out.

(Side note: in DGP 2.1.0 they _aren't_ filtered out, but fortunately they don't have any extant source roots, so they're ignored!)

I think this is the best choice. Users might add custom buildTypes, but it's unlikely they'll want them documented. Even if they do, they can just document them manually.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: > since Gradle bundles Kotlin Gradle plugin.

I don't think, though, it leaks into the classpath. E.g., without applying KGP, I can't use KGP classes.
It might leak when using `kotlin-dsl`, but only to the build script.

So I'm not sure what you mean here.
I'm only asking about it because I'm curious whether we can now depend solely on `kotlin-gradle-plugin-api` instead of `kotlin-gradle-plugin` in DGP, ensuring compatibility with both AGP and KGP. It's not really a big deal, as we have an IT that works fine, and we should test this.

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
whyoleg: > Currently debug has hasPublishedComponent=true, but this is incorrect.

This might be correct, I see no problems here, as AGP/KGP allow to publish both of them. :)

The only issue I see is that those AGP-related source sets have intersected source roots by default.
And if those `debug` and `release` related sources have the same source roots coming from AGP, it feels like this is a bug in AGP9 with built-in kotlin.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/JavaAdapter.kt
```diff
@@ -147,5 +143,9 @@ abstract class JavaAdapter @Inject constructor(
         fun SourceSet.isPublished(): Boolean =
             name != TEST_SOURCE_SET_NAME
                     && name.startsWith(MAIN_SOURCE_SET_NAME)
+
+        internal fun applyTo(project: Project) {
+            project.pluginManager.apply(type = JavaAdapter::class)
```
whyoleg: minor (again, about consistency): we should probably react on java plugins here, and not just apply it uncoditionally. E.g. use `project.plugins.withType<JavaBasePlugin>()` as with kotlin/android adapters

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
```diff
@@ -9,12 +9,29 @@ package org.jetbrains.dokka.gradle.internal
...
+    private const val AndroidTest = "com.android.test"
+    private const val AndroidDynamicFeature = "com.android.dynamic-feature"
+    private const val AndroidKmpLibrary = "com.android.kotlin.multiplatform.library"
+    val androidPlugins: Set<String> = setOf(
+        AndroidBase,
+        AndroidApplication,
+        AndroidLibrary,
+        AndroidTest,
+        AndroidDynamicFeature,
+        AndroidKmpLibrary,
```
whyoleg: Curious: do we need it here? I mean, I thought, that with that plugin, we should have enough information coming from KGP, and so we should not apply `AndroidAdapter` in that case.

dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
whyoleg: Looks like this is because of some new requirements coming from AGP of newer version...

Does it also requires to set `setCompileSdkVersion`?

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Integration test for the `it-android-kotlin-mp-builtin` project.
+ *
+ * `com.android.kotlin.multiplatform.library` was added in 8.10.0
+ * https://developer.android.com/kotlin/multiplatform/plugin#prerequisites
+ * Let's test the latest AGP 8.x version and above.
+ */
+@TestsDGPv2
+@TestsAndroid(
+    minAgpVersion = "8.13.0",
```
whyoleg: according to `allAgpVersions` for those tests, we don't test 8.13.0 - could you fix it please?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/JavaAdapter.kt
adam-enko: True - but let's not do it in this PR, since it's not related to AGP support.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
adam-enko: You mean `AndroidKmpLibrary`?

See the test project:
https://github.com/Kotlin/dokka/blob/adam/support-agp-kotlin-builtin/dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts

The only AGP plugin is `com.android.kotlin.multiplatform.library`. If that plugin is present then AndroidAdapter is required.

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
adam-enko: thanks, done

dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
adam-enko: yes, compileSdkVersion must be at least 30 to support Java 17

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
adam-enko: iiuc the release and debug variants are typically two different compilations of the same source code, so it's usually redundant to document both of them.

I asked the AGP devs:

> Suppose a project has build type `debug`, `release` and flavors `demo`, `full`
https://developer.android.com/build/build-variants#product-flavors. Then we will have 4 main components (or variants): `demoDebug`, `demoRelease`, `fullDebug`, `fullRelease`. Each variant has a corresponding KotlinCompilation.
> Meanwhile, the source sets would be: `main`, `debug`, `release`, `demo`, `full`, `demoDebug`, `demoRelease`, `fullDebug`, `fullRelease`.
> Variant `demoDebug` uses source sets: `main`, `debug`, `demo`, `demoDebug`. Similarly for other variants.
> So any 2 variants will always have a shared source set and cause the above error.


Just to clarify: we're discussing setting a sensible default here. If users want to document non-release source sets, then it's easy to enable them in DGP. But then they'll need to determine how to handle the shared source sets.

If we wanted to make it complicated, then DGP could try and filter out the duplicate source roots automatically? But this sounds like a separate feature request. I'd like to find out if it's a common request, and what would be required.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
whyoleg: > You mean AndroidKmpLibrary?

Yeah, I mean specifically this one.

Got it, for some reason, I thought that `compilationClasspath` from the kotlin compilation itself should be enough.

dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
whyoleg: Got it, thanks!

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
whyoleg: Here is what I see in `it-android-kotlin-jvm-builtin` with AGP 9.0.0-alpha14:
```kotlin
// added in build script
android.sourceSets.all {
    println(name)
    println("  kotlin: ${kotlin.directories}")
    println("  java: ${java.directories}")
}
```
output:
```
androidTest
  kotlin: [src/androidTest/java, src/androidTest/kotlin]
  java: [src/androidTest/java]
androidTestDebug
  kotlin: [src/androidTestDebug/java, src/androidTestDebug/kotlin]
  java: [src/androidTestDebug/java]
androidTestRelease
  kotlin: [src/androidTestRelease/java, src/androidTestRelease/kotlin]
  java: [src/androidTestRelease/java]
debug
  kotlin: [src/debug/java, src/debug/kotlin]
  java: [src/debug/java]
main
  kotlin: [src/main/java, src/main/kotlin]
  java: [src/main/java]
release
  kotlin: [src/release/java, src/release/kotlin]
  java: [src/release/java]
test
  kotlin: [src/test/java, src/test/kotlin]
  java: [src/test/java]
testDebug
  kotlin: [src/testDebug/java, src/testDebug/kotlin]
  java: [src/testDebug/java]
testFixtures
  kotlin: [src/testFixtures/java, src/testFixtures/kotlin]
  java: [src/testFixtures/java]
testFixturesDebug
  kotlin: [src/testFixturesDebug/java, src/testFixturesDebug/kotlin]
  java: [src/testFixturesDebug/java]
testFixturesRelease
  kotlin: [src/testFixturesRelease/java, src/testFixturesRelease/kotlin]
  java: [src/testFixturesRelease/java]
testRelease
  kotlin: [src/testRelease/java, src/testRelease/kotlin]
  java: [src/testRelease/java]
```
I don't see intersected source roots here by default - this is what I would expect for Dokka.
The same happens with `kotlin.sourceSets` in that project:
```kotlin
kotlin.sourceSets.all {
    println(name)
    println("  kotlin: ${kotlin.srcDirs.map { it.relativeTo(projectDir) }}")
}
```
output:
```
androidTest
  kotlin: [src/androidTest/kotlin]
androidTestDebug
  kotlin: [src/androidTestDebug/kotlin]
androidTestRelease
  kotlin: [src/androidTestRelease/kotlin]
debug
  kotlin: [src/debug/kotlin]
main
  kotlin: [src/main/kotlin]
release
  kotlin: [src/release/kotlin]
test
  kotlin: [src/test/kotlin]
testDebug
  kotlin: [src/testDebug/kotlin]
testFixtures
  kotlin: [src/testFixtures/kotlin]
testFixturesDebug
  kotlin: [src/testFixturesDebug/kotlin]
testFixturesRelease
  kotlin: [src/testFixturesRelease/kotlin]
testRelease
  kotlin: [src/testRelease/kotlin]
debugAndroidTest
  kotlin: [src/debugAndroidTest/kotlin]
debugUnitTest
  kotlin: [src/debugUnitTest/kotlin]
```

And so there should be no problems with intersected source roots, as the Dokka source sets should have the same data there.
So for me, it feels like we have a problem in our adapter logic.

Dokka might not work well with variants; that's true, but it should not be an issue here.

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
adam-enko: I now see failures for `AndroidKotlinMultiplatformBuiltInTest`:

```

* Where:
Build file '/private/var/folders/_3/ps18c9f56bx2j4fyxt0z7sn40000gp/T/gradle-test5023736801955594378/generate-dokka-HTML-DokkaGradleProjectRunner-/agp-8.13.0_dgp-2.1.0-SNAPSHOT_gradle-8.14.3_kgp-1.9.25/build.gradle.kts' line: 5

* What went wrong:
An exception occurred applying plugin request [id: 'com.android.kotlin.multiplatform.library', version: '8.13.0']
> Failed to apply plugin 'com.android.internal.kotlin.multiplatform.library'.
   > Could not create plugin of type 'KotlinMultiplatformAndroidPlugin'.
      > Could not generate a decorated class for type KotlinMultiplatformAndroidPlugin.
         > org/jetbrains/kotlin/gradle/dsl/HasConfigurableKotlinCompilerOptions
```

I guess `com.android.kotlin.multiplatform.library` v8.13 doesn't work...

adam-enko: Oh, it should, but it also needs KGP to be applied (unless built-in-kotlin is enabled)

https://developer.android.com/kotlin/multiplatform/plugin#prerequisites

whyoleg: > (unless built-in-kotlin is enabled)

But, AFAIK, it's an addon plugin for `kotlin.multiplatform` KGP plugin to add android target there - TBH, I don't see how it may work without KGP

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
ALikhachev: For me it looks like a bug in AGP. `bootClasspath` is a `Provider` which should be lazy, and until we get the value, it should not cause any early initialization and such problems :think\

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
ALikhachev: IMO it's hardly actionable at this point. Maybe there should be some doc page describing potential causes and solutions and we could link it here.
Also, we can describe the consequences of this – the Kotlin autoconfiguration won't work

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
ALikhachev: From what I can see, AGP brings the entire KGP (i.e. has dependency on `org.jetbrains.kotlin:kotlin-gradle-plugin`)

> Gradle bundles Kotlin Gradle plugin.

It does not :)
As Oleg mentioned, `kotlin-dsl` brings KGP as a transitive dependency similarly to what AGP does

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
ALikhachev: > I now see failures for AndroidKotlinMultiplatformBuiltInTest:

The failure is because `com.android.kotlin.multiplatform.library` is incompatible with Kotlin 1.9, it requires 2.0.0 or higher. Also, note that the test already applies `kotlin("multiplatform")`

Though, I see in local run there's another failure in `agp: 8.13.0, dgp: 2.1.0-SNAPSHOT, gradle: 8.14.3, kgp: 2.0.21` with differences in test data, like
```
-                  <div class="platform-hinted " data-platform-hinted="data-platform-hinted"><div class="content sourceset-dependent-content" data-active="" data-togglable=":/androidMain"><div class="symbol monospace"><span class="token keyword">class </span><a href="-integration-test-activity/index.html">IntegrationTestActivity</a> : <a href="https://developer.android.com/reference/kotlin/androidx/appcompat/app/AppCompatActivity.html">AppCompatActivity</a></div><div class="brief "><p class="paragraph">Some Activity implementing <a href="https://developer.android.com/reference/kotlin/androidx/appcompat/app/AppCompatActivity.html">AppCompatActivity</a> from android x</p></div></div></div>
+                  <div class="platform-hinted " data-platform-hinted="data-platform-hinted"><div class="content sourceset-dependent-content" data-active="" data-togglable=":/androidMain"><div class="symbol monospace"><span class="token keyword">class </span><a href="-integration-test-activity/index.html">IntegrationTestActivity</a> : <span data-unresolved-link="androidx.appcompat.app/AppCompatActivity///PointingToDeclaration/">AppCompatActivity</span></div><div class="brief "><p class="paragraph">Some Activity implementing <span data-unresolved-link="androidx.appcompat.app/AppCompatActivity///PointingToDeclaration/">AppCompatActivity</span> from android x</p></div></div></div>
```

> But, AFAIK, it's an addon plugin for kotlin.multiplatform KGP plugin to add android target there - TBH, I don't see how it may work without KGP

True

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
adam-enko: I'll check with AGP devs...

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
ALikhachev: Does it make sense to extract anything besides `androidExt.bootClasspath()` in this case in AndroidAdapter?
Don't we get the classpath from Kotlin compilations?

In which cases classpath from Kotlin compilation is not enough?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
```diff
@@ -112,55 +118,50 @@ private interface AndroidExtensionWrapper {
...
-                            .resolvedArtifacts
-                            .map { artifacts -> artifacts.map(ResolvedArtifactResult::getFile) }
-
-                        androidComponentsCompileClasspath.from(artifactTypeFiles)
-                    }
-
-                    variants.all {
-                        compileConfiguration.collect("jar")
-                        //runtimeConfiguration.collect("jar")
+                /** Fetch all configuration names used by all variants. */
```
ALikhachev: Nit: they are not configuration names

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
adam-enko: > In which cases classpath from Kotlin compilation is not enough?

If built-in-kotlin isn't enabled, or it's a non-Kotlin project (although DGP doesn't have tests for this).

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
adam-enko: Oleg and I discussed. It's strange if a source set has shared roots. It's a change in behaviour in AGP 9 compared to 8. It might affect other aspects, like IDE import.

DGP is a little bit protected because DGP filters out directories that don't exist on disk. But we should wait for a response from AGP developers.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
adam-enko: Oleg and I looked into it more, applying `AndroidAdapter` when `AndroidKmpLibrary` is present doesn't appear to be necessary. `compilationClasspath` already contains `android.jar` and the classpath from Android's Components.

In fact, for kotlin-built-in & AGP9 `AndroidAdapter` shouldn't be necessary at all, since all information should be exposed via 'normal' KGP API (KotlinSourceSets and KotlinCompilations).

I'll try and remove `AndroidKmpLibrary` and report back if there are issues. If there are missing classpath JARs, we should probably report this as an AGP bug.

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
adam-enko: Thanks @ALikhachev.

I've updated the test tooling to support filtering out KGP versions, so we can require KGP is 2.0.0+ for this test.

The failure for `agp: 8.13.0, dgp: 2.1.0-SNAPSHOT, gradle: 8.14.3, kgp: 2.0.21` is because AGP 8 sets the platformType to JVM instead of AndroidJVM. This will disable the `enableAndroidDocumentationLink` option, which means DGP won't generate links to the external AGP docs. This is only a problem in AGP8. 

I think we could add a workaround: if a project has `com.android.kotlin.multiplatform.library` then do `enableAndroidDocumentationLink.set(true)`. For AGP8 this should fix it. For AGP9 it shouldn't make a difference. We can remove it eventually when we drop AGP8 support.

Maybe there's a better workaround?

adam-enko: So, `com.android.kotlin.multiplatform.library` v8 sets Kotlin target to `JVM`. v9 sets it to `AndroidJVM` (because using `JVM` causes confusion in other KMP contexts).

It's not a breaking bug, it just means DGP doesn't automatically enable `enableAndroidDocumentationLink`, so the links to AGP types aren't clickable in the HTML.

So, this PR hasn't changed anything. DGP 2.1.0 also didn't automatically enable the Android doc link when a project has `com.android.kotlin.multiplatform.library`.

If we want to fix it, I see 2 options:

1. Consider it an AGP bug. We don't fix it. Users can manually call `enableAndroidDocumentationLink.set(true)` to fix it.
2. Add a workaround in DGP. Some options:
   1. If an AGP plugin is present then run `enableAndroidDocumentationLink.convention(true)` for JVM source sets. Remove the existing `if analysisPlatform == AndroidJVM` logic.
   2. Detect if a DokkaSourceSet has `android.jar` on the classpath during task execution: if so, enable AndroidDocumentationLink.

@whyoleg @ALikhachev thoughts?

adam-enko:
Converted to a draft because I'm going to split this PR into smaller pieces.

- Update test utils to support version filtering.
- Updates to 'find extension lenient'.
- Support AGP9.
- Misc fixes.
- Adding test projects.

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
whyoleg: I'm leaning towards the first option - consider it an AGP bug and mention this limitation for AGP 8+ and `com.android.kotlin.multiplatform.library`.

adam-enko: I agree.

In any case, supporting `com.android.kotlin.multiplatform.library` v8 is not related to this PR. I'll add a workaround to the test.

dokka-integration-tests/gradle/projects/it-android-kotlin-jvm-builtin/build.gradle.kts
adam-enko: Created an AGP issue https://issuetracker.google.com/issues/461736989

However, there's some confusion about how best to proceed. Previously the AGP devs were told _not_ to use KotlinSourceSet at all https://issuetracker.google.com/issues/386221070. Registering all srcDirs was done to help compiler plugins, but considering the previous request they thought the best solution is to stop registering any srcDirs and support compiler plugins another way.

Discussions are ongoing...

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
adam-enko: Reported to AGP https://issuetracker.google.com/issues/461382865

I think for now we'll have to keep the double-provider, until the minimum-supported AGP version is fixed.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: resolved in https://github.com/Kotlin/dokka/pull/4362

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/JavaAdapter.kt
adam-enko: Fixed in https://github.com/Kotlin/dokka/pull/4362

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
adam-enko: Resolved in https://github.com/Kotlin/dokka/pull/4359

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
adam-enko: thanks, fixed

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: thanks, fixed in https://github.com/Kotlin/dokka/pull/4362

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
ALikhachev: Yeah, I agree with Oleg

adam-enko:
Failing tests are due to the year change, and are fixed by https://github.com/Kotlin/dokka/pull/4406

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinJvmBuiltInTest.kt
```diff
@@ -0,0 +1,87 @@
...
+                """.trimIndent()
+            ) {
+                val expectedFileTree = expectedHtml.toTreeString()
+                val actualFileTree = actualHtmlDir.toTreeString()
+                withClue((actualFileTree to expectedFileTree).sideBySide()) {
+                    actualFileTree shouldBe expectedFileTree
+
+                    actualHtmlDir.shouldBeADirectoryWithSameContentAs(expectedHtml, TestConstants.DokkaHtmlAssetsFiles)
+                }
+            }
```
whyoleg, line 84: Could you please add `assertNoUnknownClassErrorsInHtml(actualHtmlDir)` in the end as in https://github.com/Kotlin/dokka/blob/afe18fe3e648762ebfe6dbd834507616c16a9c0f/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt#L102

whyoleg:
@adam-enko, I now see that we still have `assumeTrue` here:
https://github.com/Kotlin/dokka/blob/3241ee4c22eb675bbdf8991a3db403decf9fae67/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt#L36-L39
Should we remove it in this PR?

adam-enko:
> @adam-enko, I now see that we still have `assumeTrue` here:
> 
> https://github.com/Kotlin/dokka/blob/3241ee4c22eb675bbdf8991a3db403decf9fae67/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt#L36-L39
> 
> Should we remove it in this PR?

argh, yes, good spot! It should be removed. And the tests fail for AGP9 so I'll try and fix them...

PR #4296, opened 2025-10-14
Fix classpath when source set has no metadata compilations

### LINKED ISSUE #3386: [K2], KMP: Extension function miss platform tab for shared source set with one target

adam-enko:
This PR contains a fix for two situations where a source set's classpath does not contain the required KMP klibs for Dokka to resolve symbols in the source set's code.

### Fix 'bamboo' source sets.

If an intermediate source set only has one actual KMP target this is informally known as a 'bamboo' source set (because, unlike a 'tree', a bamboo does not have branches).

For example, `windowsMain` and `mingwMain` only have one target: `mingwX64Main`.

In `KotlinAdapter` an intermediate source set does not have any `primaryCompilations`.
If an intermediate source set only has a single 'associated compilation' then it will be considered as a 'bamboo' source set. The single 'associated compilation' is the actual target compilation. and DGP will use the classpath of actual target.

Fixes #3386

### Workaround no metadata compilations for shared Wasm source sets (workaround for KT-80551)

KGP registers metadata compilations for source sets that have multiple native compilations, but _not_ for multiple wasm source sets.

This PR works around this by using an `internal` KGP function to register the metadata compilation for shared wasm source sets.

Workaround for: 

- https://youtrack.jetbrains.com/issue/KT-80551
- https://github.com/Kotlin/dokka/issues/4116

### Stop adding source directories to classpath

For some reason DGPv2 adds source directories to the classpath. I didn't dig into the reason, but it seems to have no purpose. So I just removed this:

https://github.com/Kotlin/dokka/blob/v2.1.0/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt#L204-L206

### Fix and update kotlin-multiplatform-example

- Update `kotlin-multiplatform-example` to be compilable (it wasn't compilable before!)
- Add 'webMain' test.

### DIFF
--- a/dokka-integration-tests/gradle/projects/it-android-compose/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-android-compose/build.gradle.kts
@@ -14,3 +14,5 @@ dependencies {
     dokka(project(":core"))
     dokka(project(":material3"))
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/projects/it-android-compose/core/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-android-compose/core/build.gradle.kts
@@ -43,3 +43,5 @@ kotlin {
         }
     }
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/projects/it-android-compose/material3/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-android-compose/material3/build.gradle.kts
@@ -42,3 +42,5 @@ kotlin {
         }
     }
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
@@ -37,3 +37,5 @@ if (agpMajorVersion < 9) {
         enableAndroidDocumentationLink.set(true)
     }
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-android/build.gradle.kts
@@ -15,3 +15,5 @@ android {
 dependencies {
     implementation("androidx.appcompat:appcompat:1.1.0")
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/projects/it-kotlin-multiplatform/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-kotlin-multiplatform/build.gradle.kts
@@ -15,3 +15,5 @@ kotlin {
     androidNativeX64()
     androidNativeX86()
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/projects/it-multiplatform-android-jvm/build.gradle.kts
+++ b/dokka-integration-tests/gradle/projects/it-multiplatform-android-jvm/build.gradle.kts
@@ -26,3 +26,5 @@ kotlin {
         }
     }
 }
+
+dokka.pluginsConfiguration.html.footerMessage.set("© 2025 Copyright")
--- a/dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
+++ b/dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
@@ -96,6 +96,7 @@ class ExampleProjectsTest {
                     sourceProjectDir.copyToRecursively(projectDir, overwrite = true, followLinks = false)
                     updateSettingsRepositories()
                     updateDokkaVersion()
+                    updateHtmlFooterMessage()
                 }
             }
 
@@ -124,6 +125,32 @@ class ExampleProjectsTest {
                     """(\("org\.jetbrains\.dokka:[-\w]+:).+("\))""".toRegex(),
                     """$1${dokkaVersion}$2""",
                 )
+
+            /**
+             * By default, the footer message uses the current year.
+             * Set it to a constant year so the integration tests don't fail every year.
+             */
+            private fun GradleProjectTest.updateHtmlFooterMessage() {
+                projectDir.walk()
+                    .filter { it.name.endsWith(".gradle.kts") }
+                    .filter { "id(\"org.jetbrains.dokka\")" in it.readText() }
+                    .filter { "footerMessage" !in it.readText() }
+                    .forEach { buildFile ->
+                        buildFile.writeText(
+                            """
+                            ${buildFile.readText()}
+                            
+                            pluginManager.withPlugin("org.jetbrains.dokka") {
+                              val dokka = extensions.getByType<org.jetbrains.dokka.gradle.DokkaExtension>()
+                              dokka.pluginsConfiguration.withType<org.jetbrains.dokka.gradle.engine.plugins.DokkaHtmlPluginParameters>().configureEach {
+                                footerMessage.set("© 2025 Copyright")
+                              }
+                            }
+                            """.trimIndent()
+
+                        )
+                    }
+            }
         }
     }
 


whyoleg:
JIC: my attempt to fix `isMetadata` failed because of an issue in AA - https://github.com/Kotlin/dokka/pull/4263

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: The main change of this PR is in `KotlinAdapter`.

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
```diff
@@ -9,7 +9,12 @@ version = "1.0-SNAPSHOT"
 kotlin {
     jvm()
 
-    js(IR) {
+    js {
+        browser()
+    }
+
+    @OptIn(org.jetbrains.kotlin.gradle.ExperimentalWasmDsl::class)
+    wasmJs {
```
whyoleg, line 17: Could you please add `wasmWasi` target and declarations in `wasmMain`, so that we can be sure, that we test the original issue from kotlinx-io?
`webMain` and `wasmMain` might behave differently I believe.
Note: that `wasmMain` is not created automatically.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -167,18 +167,33 @@ abstract class KotlinAdapter @Inject constructor(
...
                     .from(details.sourceDirectoriesOfDependents)
+
+                // This KotlinSourceSet has no primary compilations, therefore it must be an intermediate source set
+                // e.g. mingwMain or webMain.
+                // If there are no metadata compilations then we must include the classpaths of the actual targets,
+                // otherwise Dokka will be unable to analyse dependencies used in this intermediate source set.
+                // See KT-80551, https://github.com/Kotlin/dokka/issues/4116
+                if (allCompilations.none { it.isMetadata }) {
+                    logger.info("[$dkaName] No metadata compilation found for ${details.name}. The classpath of all compilations will also be used.")
+                    classpath.from(allCompilations.map { it.compilationClasspath })
```
whyoleg: Redundant line, as we do the same below?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -167,18 +167,33 @@ abstract class KotlinAdapter @Inject constructor(
...
                     acc.from(compilation.compilationClasspath)
                 }
             } else {
                 classpath
                     .from(details.sourceDirectories)
                     .from(details.sourceDirectoriesOfDependents)
+
+                // This KotlinSourceSet has no primary compilations, therefore it must be an intermediate source set
+                // e.g. mingwMain or webMain.
+                // If there are no metadata compilations then we must include the classpaths of the actual targets,
```
whyoleg: So, in case of `webMain` or `wasmMain`, which compilations will be there?

For some reason, I think that at least in the case of `wasmMain` we will have both `wasmJs` and `wasmWasi` compilations, and so we will add classpath from both of them, potentially having conflicting klibs.

For `webMain`, on the other hand, I believe that a shared compilation already exists, but I might be wrong.

All of these points indicate that this PR will address #3386, and not #4116.

**I do think that we need a functional test here, which will check the classpaths of source sets.**

examples/gradle-v2/kotlin-multiplatform-example/build.gradle.kts
adam-enko: I'll add a functional test that adds a custom shared WasmJS/WasmWASI source set. I think it's best to avoid adding custom source sets into the example projects, so as to keep them simpler.

There should be no differences between a shared source set that's provided by KGP and one that's defined by users.

adam-enko:
I've discussed more with @abdulowork and we think we've found a better workaround. We can use IdeMultiplatformImport to get the same data that IJ uses.

I'll have another attempt...

adam-enko:
Unfortunately while `IdeMultiplatformImport` returns the required klibs, it doesn't seem to be very stable. The KGP task that uses `IdeMultiplatformImport` disables CC and caching. In DGP it triggers lots of warnings when used with AGP:

```
Configuration 'jvmCompileClasspath' was resolved during configuration time.
This is a build performance and scalability issue.
See https://github.com/gradle/gradle/issues/2298
```

Investigations are ongoing...

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -165,21 +167,32 @@ abstract class KotlinAdapter @Inject constructor(
...
-                compilations.fold(classpath) { acc, compilation ->
-                    acc.from(compilation.compilationClasspath)
+                if (primaryCompilations.isNotEmpty()) {
+                    primaryCompilations.forEach { compilation ->
+                        classpath.from(compilation.compilationClasspath)
+                    }
+                } else {
+                    // handle a 'bamboo' source sets
+                    // (intermediate source set with only one actual target,
+                    // e.g. windowsMain and mingwMain only have one target: mingwX64Main)
```
whyoleg: So, this fixes https://github.com/Kotlin/dokka/issues/3386, right?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -165,21 +167,32 @@ abstract class KotlinAdapter @Inject constructor(
...
+                    allAssociatedCompilations.singleOrNull()?.let { compilation ->
```
whyoleg: is there a reason, why we use `singleOrNull` here?
When it could have more than one compilation? Shouldn't we iterate over all compilations here, as with primary ones?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -592,15 +605,22 @@ private abstract class KotlinSourceSetDetails @Inject constructor(
      * For example, the compilation for `commonMain` will also participate in compiling
      * the leaf `linuxX64`, as well as the intermediate compilations of `nativeMain`, `linuxMain`, etc.
      */
-    abstract val allCompilations: ListProperty<KotlinCompilationDetails>
+    abstract val allAssociatedCompilations: ListProperty<KotlinCompilationDetails>
```
whyoleg, line 614: minor: it's a bit strange for me to see `associated` here, as Kotlin compilations already have a concept of associated compilations, but it's about friend modules (e.g. `test` can see internal `main` declarations, because those are associated). There is even `KotlinCompilation.associatedCompilations` :)

Maybe leave it's better to use `allCompilations` as it was before? And drop `Associated` from the comment itself?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -676,20 +706,31 @@ private class KotlinSourceSetDetailsBuilder(
...
+            .map { allDependsOns ->
+                allDependsOns.fold(objects.fileCollection()) { acc, sourceSet ->
                     acc.from(sourceSet.kotlin.sourceDirectories)
                 }
-        }
+            }
+
+        val transformedMetadataDependencies =
+            allAssociatedCompilations.map { associated ->
+                if (associated.all { it.kotlinPlatform == KotlinPlatform.Wasm }) {
```
whyoleg, line 725: Is it really true only for the case with `wasmJs` + `wasmWasi` compilations?
What will be in the case we have a source-set between `wasmWasi` and `native` targets?

whyoleg: Oh, it looks like in the case with `wasmWasi` + `native`, KGP does create an intermediate compilation, and so we have a correct classpath

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/TransformedMetadataDependencyProvider.kt
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
```
whyoleg: please, could you make those scary suppress usages localized, and not affect the whole file?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/TransformedMetadataDependencyProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        return files
+    }
+
+    private fun TaskProvider<MetadataDependencyTransformationTask>.allTransformedLibraries(): Provider<Provider<List<File>>> {
+        return map { task ->
+            val taskPath = task.path
+            task.allTransformedLibraries().map { libs ->
```
whyoleg: Am I correct, that this will be invoked at some point, and not inside the `try-catch` block above, and so if KGP will change this `internal` API, we will fail at runtime on user side?
Can we avoid it?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: Can't KGP just do the same for `wasmJs` + `wasmWasi` as a start? :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: yes

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: If we're in this branch of the if statement, then this source set has no primary compilations.

If there is only one associated compilation, then it's a 'bamboo' source set: `singleOrNull()` will return the single associated compilation, and Dokka will use it's classpath.

Suggestions welcomed! I can adjust the code to break up the steps a bit, if it would help make it clearer?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I found the `allCompilations` confusing because it appeared to mean "all compilations that are related to the current build", while actually it means "all compilations that use this Kotlin Source Set".

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: Yeah, this will be fixed eventually in KGP https://youtrack.jetbrains.com/issue/KT-80551/compileDependencyFiles-misses-some-dependencies-for-Wasm

This workaround will fix the issue for older KGP versions - or are you saying Dokka shouldn't workaround the issue, and should wait for the KGP fix?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/TransformedMetadataDependencyProvider.kt
adam-enko: good catch - fixed

adam-enko:
Failing tests are due to the year change, and are fixed by https://github.com/Kotlin/dokka/pull/4406

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: Ah! Now I get it! Could you please add this to the comment?
I mean, expand on `how we handle bamboo source sets`, as it's not directly clear from the comment why using `allAssociatedCompilations.singleOrNull()` handles the bamboo case.

Also, now I see that the PR _description describes_ this in a nice way, so maybe moving it to the comment would be nice:
```
If an intermediate source set only has a single 'associated compilation' then it will be considered as a 'bamboo' source set. The single 'associated compilation' is the actual target compilation. and DGP will use the classpath of actual target.
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: I don't see how `allCompilations` here differs from `KotlinCompilationDetails.allSourceSetNames` - both can be confused solely by their names, but as it's a property of `KotlinSourceSetDetails`, so it feels fine for me.

Maybe something like `usedInCompilations` or some other name?

I'm fine with the `associated` wording, but I don't want it to be confused with a completely different KGP concept...

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: done

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/TransformedMetadataDependencyProvider.kt
adam-enko: done

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: Nope, let's have a workaround right now :)

It's just that, as I stated in the review comment, it feels like the fix on the KGP side might have been easier. Wasm (and wasm-wasi) is still in Alpha/Beta states, so it would have been fine to support this only in newer versions of KGP.

But the ship has already sailed, and we do have a workaround in place.

whyoleg (approved):
Nice! Thanks for the hard work on finding how to fix this on Dokka side!

PR #4301, opened 2025-10-17
Add olderVersionsDirName to DokkaVersioningPluginParameters

osipxd:
The field `olderVersionsDirName` was missing from `DokkaVersioningPluginParameters`, while it is listed in the [versioning plugin docs](https://github.com/Kotlin/dokka/blob/v2.1.0/dokka-subprojects/plugin-versioning/README.md#configuration-options).


dokka-runners/dokka-gradle-plugin/src/main/kotlin/formats/DokkaHtmlPlugin.kt
```diff
@@ -61,6 +61,7 @@ constructor(
             )
             register<DokkaVersioningPluginParameters>(DOKKA_VERSIONING_PLUGIN_PARAMETERS_NAME)
             withType<DokkaVersioningPluginParameters>().configureEach {
+                olderVersionsDirName.convention("older")
```
osipxd, line 64: I'm not sure about this. Maybe keep it unset by default and rely on the default value specified in the plugin?

whyoleg: I think that it's fine to have a `convention` here, in the same way we have it for `renderVersionsNavigationOnAllPages`

Issue #3958, opened 2024-12-17
Javadoc exposes absolute paths in dokka-configuration.json


TWiStErRob:
**Describe the bug**

Just migrated to Dokka 2.0.0, and I was diffing the artifacts of publishToMavenLocal. I found that the new javadoc jar contains the absolute paths of the source files and classpaths.
 
This means the resulting artifacts are [not reproducible](https://reproducible-builds.org/docs/env-variations/).
 
There might be also a security risk as some companies use Dokka to generate javadoc for public APIs and publish them online. These could leak internal implementation details of their systems, and the paths could be used to learn about their build machines. This could in theory aid attackers to exploit vulnerabilities more easily.

**Expected behaviour**

Relative paths or paths with env vars or no paths at all?

(Note re relative paths: my Gradle and Project paths are on separate Windows drive letters, so there's no way to navigate between them with `../`. Even if that was possible the depth of the folders would still make it not reproducible.)

**Screenshots**
I replaced my local absolute paths with `GRADLE_USER_HOME` and `PROJECT_CHECKOUT_DIR`.

```json
{
  "moduleName": "my-module",
  "moduleVersion": "<version>-SNAPSHOT",
  "outputDir": "%PROJECT_CHECKOUT_DIR%\\my\\module\\build\\dokka\\javadoc",
  "cacheRoot": null,
  "offlineMode": false,
  "sourceSets": [
    {
      "displayName": "jvm",
      "sourceSetID": {
        "scopeId": ":my:module",
        "sourceSetName": "main"
      },
      "classpath": [
        "%GRADLE_USER_HOME%\\caches\\8.8\\generated-gradle-jars\\gradle-api-8.8.jar",
        "%GRADLE_USER_HOME%\\wrapper\\dists\\gradle-8.8-all\\6gdy1pgp427xkqcjbxw3ylt6h\\gradle-8.8\\lib\\groovy-3.0.21.jar",
        "%GRADLE_USER_HOME%\\wrapper\\dists\\gradle-8.8-all\\6gdy1pgp427xkqcjbxw3ylt6h\\gradle-8.8\\lib\\groovy-ant-3.0.21.jar",
        "%GRADLE_USER_HOME%\\wrapper\\dists\\gradle-8.8-all\\6gdy1pgp427xkqcjbxw3ylt6h\\gradle-8.8\\lib\\groovy-astbuilder-3.0.21.jar",
[...]

PR #4309, opened 2025-10-22
Update DGPv2 examples

adam-enko:
- Update DGP to use latest 2.1.0 release
- Bump KGP version to 2.2.0
- Remove `flatMap { it.outputDirectory }` (which was a workaround for #3958)
- Remove unnecessary `org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled`, because v2 is now the default. 
- Override the Dokka version when testing the example projects.

### DIFF
--- a/dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
+++ b/dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
@@ -114,14 +114,14 @@ class ExampleProjectsTest {
             /** Replace the version of any Dokka plugin with [dokkaVersion]. */
             private fun String.replaceDokkaPluginsVersion(): String =
                 replace(
-                    """(id\("org\.jetbrains\.dokka(?:-\w\d)*"\) version ").+(")""".toRegex(),
+                    """(id\("org\.jetbrains\.dokka[-\w]*"\) version ").+(")""".toRegex(),
                     """$1${dokkaVersion}$2""",
                 )
 
             /** Replace the version of any Dokka dependency coord with [dokkaVersion]. */
             private fun String.replaceDokkaDependencyCoords(): String =
                 replace(
-                    """(\("org\.jetbrains\.dokka:(?:-\w\d)*:).+("\))""".toRegex(),
+                    """(\("org\.jetbrains\.dokka:[-\w]+:).+("\))""".toRegex(),
                     """$1${dokkaVersion}$2""",
                 )
         }


whyoleg:
@adam-enko, could you also tackle https://github.com/Kotlin/dokka/issues/4316 here?

examples/gradle-v2/custom-dokka-plugin-example/dokka-plugin-hide-internal-api/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
 }
 
 dependencies {
-    val dokkaVersion = "2.1.0-SNAPSHOT"
+    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0")
```
whyoleg: Could you please explain why we need to have this `providers.gradleProperty("dokkaVersion")`?
I don't really understand why we can't have just `val dokkaVersion = "2.1.0"` here.

examples/gradle-v2/java-example/buildSrc/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.1.0")
-    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0-SNAPSHOT")
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.2.0")
+    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0")
```
whyoleg: I have the same question as above. But at this time, I'm curious, why did it have `providers.gradleProperty("dokkaVersion")` before? :)

In case of `buildSrc` usage, I would expect the version to be inlined, in the same way as for KGP:
```kotlin
implementation("org.jetbrains.dokka:dokka-gradle-plugin:2.1.0")
```

examples/gradle-v2/multimodule-example/buildSrc/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.1.0")
-    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0-SNAPSHOT")
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.2.0")
+    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0")
```
whyoleg: same - `implementation("org.jetbrains.dokka:dokka-gradle-plugin:2.1.0")`

examples/gradle-v2/versioning-multimodule-example/buildSrc/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.1.0")
-    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0-SNAPSHOT")
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.2.0")
+    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0")
```
whyoleg: same - `implementation("org.jetbrains.dokka:dokka-gradle-plugin:2.1.0")`

examples/gradle-v2/composite-build-example/build-logic/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.1.0")
-    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0-SNAPSHOT")
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:2.2.0")
+    val dokkaVersion = providers.gradleProperty("dokkaVersion").getOrElse("2.1.0")
```
whyoleg: same - `implementation("org.jetbrains.dokka:dokka-gradle-plugin:2.1.0")`

whyoleg (approved):
Thanks! It really looks much better when we don't have `SNAPSHOT` versions in examples!

I do have mostly minor remarks, regarding usage of `providers.gradleProperty("dokkaVersion")`. IMO, it feels redundant, as just having the version there will be much easier for users. For more complex projects, version catalogs will/should be used, not a property in `gradle.properties`.

examples/gradle-v2/custom-dokka-plugin-example/dokka-plugin-hide-internal-api/build.gradle.kts
adam-enko: I'm honestly not sure. I think it's a left-over from the v1 examples.

https://github.com/Kotlin/dokka/blob/v2.1.0/examples/gradle/dokka-multimodule-example/settings.gradle.kts#L6-L7

I think it makes sense to simplify it.

Issue #4049, opened 2025-03-03
[K2] Duplicate source links for overloaded with vararg functions


atyrin:
```Kotlin
fun overloadWithVararg(novararg: String){}

fun overloadWithVararg(vararg elements: String){}
```

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/60d0bc40-a5fc-46fc-a32e-52565135c830" />

The issue doesn't reproduce with K1.
But there is similar for K1/K2: https://github.com/Kotlin/dokka/issues/3612

**Installation**
- Dokka version: 2.0.0

atyrin:
Also noticed: with K2 pages.json (responsible for search) `vararg` option is missed.

AbdullinAM:
Fixed by #4322

Issue #3502, opened 2024-02-23
Incorrect link to Javadoc's functions with type parameters


vmishenev:
```Kotlin
    /**
     * [java.util.Arrays.asList]
     */
```

**Expected behaviour**
The Dokka generates https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#asList-kotlin.Array[TypeParam(bounds=[kotlin.Any])]-

**Expected behaviour**
The correct link should be https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#asList-T...-

**Installation**
- Dokka version: 1.9.20

AbdullinAM:
Fixed in #4344

Issue #3558, opened 2024-04-09
[K2] Different PageId in K1 and K2 for varargs


IgnatBeresnev:
## Context

This issue was extracted from #3388 as a separate case, because it is tricker to fix due to both Dokka and Analysis API not handling varargs very well (see #2881).

## Reproducer

```Kotlin
fun withVararg(i: Int, vararg s: String){}
```

K1: `pageids="module:://withVararg/#kotlin.Int#kotlin.Array[kotlin.String]/PointingToDeclaration//-1486420497"`
K2: `pageids="module:://withVararg/#kotlin.Int#kotlin.String/PointingToDeclaration//-1486420497"`

**Installation**
- Dokka version: 1.9.20

vmishenev:
It is blocked by https://github.com/Kotlin/dokka/issues/2881

whyoleg:
[KT-64340](https://youtrack.jetbrains.com/issue/KT-64340) is now available in 2.3.0-dev-3168, so it's now only blocked by the design/implementation of #2881

AbdullinAM:
Fixed by #4322

Issue #4045, opened 2025-02-28
[K2] Unresolved method references with argument


atyrin:
```Kotlin
interface CacheManager {
    fun getCache(name: String?): String
}

/**
 * Extension for [CacheManager.getCache(name)]
 */
operator fun CacheManager.get(name: String): String = getCache(name)
```

In K1 reference from K1 worked, For K2 the link disappeared. 

In IJ the reference is also resolved only without a name: `[CacheManager.getCache]`

**Installation**
- Dokka version: 2.0.0

vmishenev:
Similar case: https://github.com/Kotlin/dokka/issues/3234#issuecomment-2092590106

atyrin:
Then it might be a duplicate of #3629

PR #4322, opened 2025-10-29
Add separate `Vararg` type reference for DRI

AbdullinAM:
AA has [added](https://youtrack.jetbrains.com/issue/KT-64340/Analysis-API-no-way-to-get-a-type-of-vararg-parameter) a new experimental `varargArrayType` property for `KaValueParameterSymbol` which can be used to correctly construct type references for `vararg` parameters.

This fixes both #3558 and #4049. I added test for #3558, but I'm not sure if we have the test infrastructure to test #4049.


AbdullinAM:
> #3558 only affects the HTML diff between Dokka K1 and K2.
> 
> This does not really fix #4049 since
> 
> ```kotlin
> fun f(a: Array<String>) = 0
> fun f(vararg a: String) = 0
> ```
> 
> will have the same DRI. DRI should be extended somehow.

Got it, thanks! I suggest we also add this context to #4049, because the current reproducing example presented there describes a related, but different problem.

The example you provided highlights that there is currently no difference between `Array<String>` and `vararg String` because DRI can't differentiate between arrays and varargs
> ```kotlin
> fun f(a: Array<String>) = 0
> fun f(vararg a: String) = 0
> ```


Example from #4045 highlights that `vararg String` is incorrectly parsed as just `String` in K2 currently
> ```kotlin
> fun overloadWithVararg(novararg: String){}
> fun overloadWithVararg(vararg elements: String){}
> ```

vmishenev:
> I suggest we also add this context to https://github.com/Kotlin/dokka/issues/4049, because the current reproducing example presented there describes a related, but different problem.


It is a long chain of issues) Something was lost.
 This case is in https://github.com/Kotlin/dokka/issues/2881#issuecomment-1961768017 that blocks #3558

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DRIFactory.kt
```diff
@@ -57,22 +57,24 @@ internal fun KaSession.getDRIFromConstructor(symbol: KaConstructorSymbol): DRI {
...
         )
     )
 }
 
+@OptIn(KaExperimentalApi::class)
 internal fun KaSession.getDRIFromVariable(symbol: KaVariableSymbol): DRI {
     val callableId = symbol.callableId ?: throw IllegalStateException("Can not get callable Id due to it is local")
     val receiver = symbol.receiverType?.let(::getTypeReferenceFrom)
-    val contextParams = @OptIn(KaExperimentalApi::class) symbol.contextParameters.map { getTypeReferenceFrom(it.returnType) }
+    val contextParams = symbol.contextParameters.map { getTypeReferenceFrom(it.returnType) }
```
vmishenev: In this case, I prefer having more local `OptIn`.  It helps to control the usage of the experimental API.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -146,6 +146,10 @@ public data class Nullable(val wrapped: TypeReference) : TypeReference() {
     override fun toString(): String = "$wrapped?"
 }
 
+public data class Vararg(val elementType: TypeReference) : TypeReference() {
```
vmishenev, line 149: Are the links in KDoc `[java.util.Arrays.asList]` and Javadoc `{@link java.util.Arrays#asList}` consistent?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -146,6 +146,10 @@ public data class Nullable(val wrapped: TypeReference) : TypeReference() {
...
+    override fun toString(): String = "Vararg[$elementType]"
```
vmishenev: Unfortunately, this is used in `package-list` somehow

https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/external/DefaultExternalLocationProvider.kt#L20C53-L20C67 


This may be left as it was previously, but I'm not sure.

vmishenev (commented):
I think https://github.com/Kotlin/dokka/issues/3502 also can be closed. Can you add a test for it?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/external/javadoc/JavadocExternalLocationProvider.kt
whyoleg, line 66: JIC: Here is the TODO for #3502 mentioned by @vmishenev.
So I'm not sure that it will be fixed by this PR, and it has a different fix (we need to store the type parameter name in `TypeParam`)

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
whyoleg: I believe it's our extension to `package-list`, like [here](https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/package-list) or [here](https://kotlinlang.org/api/kotlinx.coroutines/package-list), we do print DRIs for some reason (extensions?).

No matter the reason, we probably should have some `toString` here, so that `DRI.toString` will be useful.
I'm not sure about the square brackets there, as they are used in `TypeConstructor` for parameters, so maybe something like `vararg(xxx)` or as in the Javadoc `xxx...` or in kotlin `*xxx` (but it might clash with `StarProjection`)?
IMO, `vararg(xxx)` feels fine?

dokka-subprojects/plugin-base/src/test/kotlin/basic/DRITest.kt
```diff
@@ -393,4 +381,99 @@ class DRITest : BaseAbstractTest() {
             )
         }
     }
+
+    @Test
+    fun `#3558 DRI for function with varags parameter`() {
```
whyoleg, line 386: Based on the discussion above regarding the `toString` of `Vararg`, and what I see, that in some other tests we also validate `DRI.toString`, maybe it makes sense to also add tests for `toString` representation here, in addition to structural comparison, so that we also test `toString` contract?
WDYT?

whyoleg (commented):
Nice! Hope that no one is relying on the fact that we are adding one more case to the sealed `TypeReference`.

Maybe only Dackka?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/resolvers/external/javadoc/JavadocExternalLocationProvider.kt
AbdullinAM: Yes, the fix is not directly related. But it is pretty simple and it is somewhat related to the javadoc tests that I already added.

I will fix #3502 as a separate PR after this will be merged (already have a patch for it)

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
AbdullinAM: Yes

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: > No matter the reason,

It is not just used for printing, but also for resolving _somehow_ (see the link above).. It will _probably_ break the already generated `package-list`.

> This may be left as it was previously, but I'm not sure.

I mean `override fun toString(): String = "kotlin.Array[$elementType]"`, but I do not have a strong opinion here.
In the end, we should not rely on `DRI.toString()`.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: And DRI?
I had supposed the PSI parser to DRI should be changed as well.  https://github.com/Kotlin/dokka/blob/60236e9c6497ef00aa060df1c77d41e0d74b18f9/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiUtil.kt#L20

At least, it seems to affect Javadoc's links to Kotlin functions with vararg.

AbdullinAM: Actually, there is a small difference in DRI and the resulting links:

* Javadoc: `java.nio.file/Files/createDirectories/#java.nio.file.Path#java.nio.file.attribute.FileAttribute<?>.../PointingToDeclaration/`

* KDoc: `java.nio.file/Files/createDirectories/#java.nio.file.Path#vararg(java.nio.file.attribute.FileAttribute[*])/PointingToDeclaration/`

The `<?>` part comes from PSI and therefore the link is not fully valid. 

It seems like the psi parser indeed needs reworking. But I'm not sure that it should be the part of this PR

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
AbdullinAM: If you are talking about `package-list`s that are currently in the integration tests, they seem to work

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: Indeed, `JavaClassReference` can store arbitrary text. Nothing has broken so far.

> But I'm not sure that it should be the part of this PR

I think it’s better to leave it as is for now.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: > If you are talking about package-lists that are currently in the integration tests, they seem to work

Which tests?

One of the cases when we have a non-stantard location is having a platform tag in a page name, e.g. https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/[wasm-js]promise.html


Such tags are added when we have name clashing from different platforms - see the option `mergeImplicitExpectActualDeclarations=false`. 

For example, https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-local-date/[jvm]equals.html - this link cannot be constructed using only the DRI so the `package-list` has a 
```$dokka.location:kotlinx.datetime/LocalDateTime/equals/#kotlin.Any?/PointingToDeclaration/kotlinx-datetime/kotlinx.datetime/-local-date-time/[jvm]equals.html```

It will be broken for vararg functions.

However, I have just realised it is already broken for vararg by Dokka 2.1.0(

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
whyoleg: Just to confirm, is this difference covered by #3502?
Or should we have a separate issue for this case with `Files.createDirectories`?

AbdullinAM: I don't think that it is covered by #3502, but I will check and create a new issue if necessary

AbdullinAM: #4342

PR #4334, opened 2025-11-04
Rework the Gradle attributes to make them integrate better with the ecosystem

adam-enko:
Continuation of #4286

Fixes [KT-72019](https://youtrack.jetbrains.com/issue/KT-72019) and https://github.com/adamko-dev/dokkatoo/issues/165

Credit goes to @martinbonnin!

### Summary

Unfortunately, Gradle's matching algorithm prioritises the number of attributes matched over the actual attribute values. (Note that if a candidate does not have a requested attribute this still counts as a match!) This behaviour makes it impossible to strictly match Configurations, and leads to situations like this where two 'ecosystems' step on each other's toes.

The solution is in two parts:

1. Reduce the number of attributes DGP uses.

   This prevents the matching algorithm from over-matching based on attributes alone. The Bundling attribute is removed, since this is not important for DGP classpaths.
4. (As suggested by @martinbonnin) DGP configurations use a custom value for the `Usage` attribute, and use a one-way `AttributeCompatibilityRule` so DGP can still resolve dependencies.
   (A one-way rule should minimise ugly issues [where rules can leak](https://github.com/gradle/gradle/issues/30921).)

See the KDoc of `DokkaJavaRuntimeUsageCompatibilityRule` for details.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/dependencies/DokkaJavaRuntimeUsageCompatibilityRule.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/dependencies/DokkaJavaRuntimeUsageCompatibilityRule.kt
@@ -64,7 +64,7 @@ import org.gradle.api.attributes.Usage
  * - Traditional consumers disambiguate the Dokka plugins variants because the
  *   compatibility rule is one way.
  *   If the consumer asks for [Usage.JAVA_RUNTIME],
- *   then [DokkaAttribute.DokkaJavaRuntimeUsage] the rule expresses no opinion.
+ *   then the rule expresses no opinion.
  */
 internal abstract class DokkaJavaRuntimeUsageCompatibilityRule : AttributeCompatibilityRule<Usage> {
     override fun execute(details: CompatibilityCheckDetails<Usage>) {


martinbonnin:
Wowowowow! That's a lot of tests 👀 
I'm not super familiar with the test suite but let me know if I can help with anything.

Vampire:
Just in case you missed it on the copied PR.

Before considering this PR, I'd like to suggest reviewing my analysis of the problem just posted to https://github.com/adamko-dev/dokkatoo/issues/165#issuecomment-3358646296 :-)

Because I think the proper solution is not to do what is suggested here.

adam-enko:
> I'm not super familiar with the test suite but let me know if I can help with anything.

I think we're good! There was a flaky test but it should be fixed now #4335

adam-enko:
> Just in case you missed it on the copied PR.
> 
> Before considering this PR, I'd like to suggest reviewing my analysis of the problem just posted to [adamko-dev/dokkatoo#165 (comment)](https://github.com/adamko-dev/dokkatoo/issues/165#issuecomment-3358646296) :-)
> 
> Because I think the proper solution is not to do what is suggested here.

Thanks for the link. Unfortunately, it wouldn't help.

It's not possible to modify the attributes of KGP Configurations. Additionally, any fix needs to work with older KGP versions (such as those bundled into older Gradle versions). Also, itt's not appropriate for DGP to use the KotlinPlatformType, which is only relevant for KMP. Adding it won't help with non-KMP consumers.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dependencies/DokkaAttribute.kt
```diff
@@ -39,6 +39,7 @@ interface DokkaAttribute {
     }
 
     @InternalDokkaGradlePluginApi
+    @Suppress("ConstPropertyName")
```
ALikhachev, line 42: Nit: why suppress instead of following the code convention from kotlinlang.org?
https://kotlinlang.org/docs/coding-conventions.html#property-names recommends using SCREAMING_SNAKE_CASE

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dependencies/DokkaJavaRuntimeUsageCompatibilityRule.kt
```diff
@@ -0,0 +1,81 @@
...
+ * Copyright 2014 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package org.jetbrains.dokka.gradle.dependencies
+
+import org.gradle.api.attributes.AttributeCompatibilityRule
+import org.gradle.api.attributes.CompatibilityCheckDetails
+import org.gradle.api.attributes.Usage
+
+/**
+ * # Dokka Java Runtime Usage Compatibility Rule
```
ALikhachev, line 11: Thanks for providing a comprehensive description right in the code!

ALikhachev (approved):
LGTM

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dependencies/DokkaJavaRuntimeUsageCompatibilityRule.kt
adam-enko: Most of the credit goes to @martinbonnin for that!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/dependencies/DokkaAttribute.kt
adam-enko: personal preference, and it aligns with the other properties (even though the type is different)

Vampire:
>>Just in case you missed it on the copied PR.
>>
>>Before considering this PR, I'd like to suggest reviewing my analysis of the problem just posted to https://github.com/adamko-dev/dokkatoo/issues/165#issuecomment-3358646296 :-)

>>
>>Because I think the proper solution is not to do what is suggested here.
>
>Thanks for the link. Unfortunately, it wouldn't help.
>
>It's not possible to modify the attributes of KGP Configurations. Additionally, any fix needs to work with older KGP versions (such as those bundled into older Gradle versions). Also, itt's not appropriate for DGP to use the KotlinPlatformType, which is only relevant for KMP. Adding it won't help with non-KMP consumers.

I of course did not suggest that Dokka modifies the attributes of KGP configurations.
KGP should be fixed to provide the proper attributes.

The problem I see is, that attribute- / variant-aware resolution is a quite complex topic.
And throwing some more attributes or values to it, to work-around bugs in other parts of the ecosystem just makes it even more complex and fragile.
So it might be better to fix the underlying problems instead of doing symptom treatment in the dokka part. 🤷‍♂️

martinbonnin:
> attribute- / variant-aware resolution is a quite complex topic.

Candidly, this is the part that needs to be fixed. 

Both the `"Candidates that lack extra attributes are preferred"` and `"missing attributes are compatible"` rules have tripped many colleagues/community members. It probably took me several days of stepping through the Gradle source code to get a surface understanding of what's happening in the algorithm. And even though I consider myself as one of the most non-Gradle folks experts there, I still need many hours to debug a dependency resolution error. 

I understand this is complicated but the current algorithm doesn't scale. 

At least the proposed solution shows `dokka-java-runtime` in the logs so it hints that the `usage` attribute is involved there. Having the `bundling` attribute decide the resolution and relying on every possible plugin to set it correctly sounds a lot more fragile to me.

Vampire:
> Candidly, this is the part that needs to be fixed.

Feel free to convince the Gradle folks about that :-D

martinbonnin:
This is my attempt at convinving people we need to simplify all of this: https://mbonnin.net/2025-11-29_the_curse_of_turing_complete_dependency_resolution/

Issue #4326, opened 2025-10-31
[K2] Redundant `?` on properties with type of typealias to nullable type


whyoleg:
Simplified example based on kotlinx.coroutines [MDCContext](https://github.com/Kotlin/kotlinx.coroutines/blob/8c27d51025d56a7b8de8eec2fb234b0094ce84f1/integration/kotlinx-coroutines-slf4j/src/MDCContext.kt#L77C1-L83C78) implementation
```kotlin
public typealias MDCContextMap = Map<String, String>? // note - nullable type
public class MDCContext(
    public val contextMap: MDCContextMap = MDC.getCopyOfContextMap()
) : ThreadContextElement<MDCContextMap>, AbstractCoroutineContextElement(Key)
```
K1 output:

<img width="1424" height="404" alt="Image" src="https://github.com/user-attachments/assets/f1174375-e3d1-404b-9582-127d6439a6a1" />

K2 output:

<img width="1460" height="398" alt="Image" src="https://github.com/user-attachments/assets/7940e4c7-9b03-4af7-9543-0039145764a5" />

Note: additional `?` symbol

## Expected behavior

As `typealias MDCContextMap` already includes nullability, we should not show `?` on `MDCContextMap` declarations

AbdullinAM:
Fixed in #4337

PR #4337, opened 2025-11-04
Fix 'Redundant ? on properties with type of typealias to nullable type '

AbdullinAM:
Fix for #4326 

Change `KaSession#toBoundFrom` to correctly parse typealiases:

* Use `KaSession#toBoundFrom` recursively to parse reference types
* Introduce a new `ignoreTypealiases` parameter to `KaSession#toBoundFrom` to prevent infinite recursion
* Wrap bound to `Nullable` only if typealias abbreviation itself is marked nullable

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/model/TypesTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/model/TypesTest.kt
@@ -10,6 +10,7 @@ import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.model.*
 import utils.AbstractModelTest
 import utils.OnlySymbols
+import utils.assertIsInstance
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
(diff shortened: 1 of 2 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/TypeTranslator.kt
```diff
@@ -85,12 +85,15 @@ internal class TypeTranslator(
...
-    fun KaSession.toBoundFrom(type: KaType): Bound =
+    fun KaSession.toBoundFrom(type: KaType, ignoreTypealiases: Boolean = false): Bound =
         when (type) {
             is KaUsualClassType -> {
                 // after KT-66996, [type] is an expanded type
                 val abbreviation = type.abbreviation
-                if (abbreviation != null) toBoundFromTypeAliased(abbreviation, toTypeConstructorFrom(type))
+                if (abbreviation != null && !ignoreTypealiases) toBoundFromTypeAliased(
+                    abbreviation,
+                    toBoundFrom(type, ignoreTypealiases = true)
```
vmishenev: _minor:_ It seems the flag `ignoreTypealiases` is redundant. TBH, I prefer the logic without a flag. It takes mental effort to understand what the flag actually does here. 
`toTypeConstructorFrom(type).asNullableIfMarked(type)`


Otherwise, if you want to use the flag, you could move the following expression out of the `when`:
```
                val abbreviation = type.abbreviation
                if (abbreviation != null && !ignoreTypealiases) toBoundFromTypeAliased(
                    abbreviation,
                    toBoundFrom(type, ignoreTypealiases = true)
                )
```
 as `KaDynamicType`, `KaDefinitelyNotNullType`, and so on can not have an abbreviation.

PR #4340, opened 2025-11-05
Fix missing default parameter value for fake functions

### LINKED ISSUE #4320: [K2] Missing default parameter value for inherited (not overriden) members
Project: https://github.com/whyoleg/cryptography-kotlin (Dokka 2.1.0)
The project uses `suppressInheritedMembers = false`, so that inherited members are explicitly displayed. But in this case, some declarations miss default parameter values.

The function in question ([keyGenerator](https://github.com/whyoleg/cryptography-kotlin/blob/972562bd704a9aba419111132ceb00b11922333d/cryptography-core/src/commonMain/kotlin/algorithms/AES.kt#L18)) is declared inside of an `AES` interface, and there is [AES.CBC](https://github.com/whyoleg/cryptography-kotlin/blob/972562bd704a9aba419111132ceb00b11922333d/cryptography-core/src/commonMain/kotlin/algorithms/AES.kt#L52) interface, which extends `AES` but doesn't override the function.

Here is the HTML output when looking at the `AES.CBC` page:

K1 output:

Image

K2 output:

Image

In both cases, clicking on the `keyGenerator` function will redirect to the page for `AES.keyGenerator` (that's correct behavior), which also correctly shows the default parameter:

Image

The only issue is that with K2, no default parameter is shown.

vmishenev:
Fixes #4320

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
@@ -794,18 +794,17 @@ internal class DokkaSymbolVisitor(
      * It may be `null` if the owner function comes from a non-source file.
      */
     private fun KaSession.getDefaultValue(symbol: KaValueParameterSymbol, parameterIndex: Int): Expression? {
-        fun KaValueParameterSymbol.getDefaultValue(): Expression? =
+        fun KaValueParameterSymbol.getExplicitDefaultValue(): Expression? =
             if (origin == KaSymbolOrigin.SOURCE) (psi as? KtParameter)?.defaultValue?.toDefaultValueExpression() else null
+        fun KaDeclarationSymbol.findMatchingParameterWithDefaultValue(): Expression? =
+            (this as? KaFunctionSymbol)?.valueParameters?.getOrNull(parameterIndex)?.getExplicitDefaultValue()
 
-        val result = symbol.getDefaultValue()
+        val result = symbol.getExplicitDefaultValue()
         return if (result != null)
             result
         else { // in the case of fake declarations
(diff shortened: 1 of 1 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -788,9 +788,30 @@ internal class DokkaSymbolVisitor(
...
+     */
+    private fun KaSession.getDefaultValue(symbol: KaValueParameterSymbol, parameterIndex: Int): Expression? {
+        fun KaValueParameterSymbol.getDefaultValue(): Expression? =
+            if (origin == KaSymbolOrigin.SOURCE) (psi as? KtParameter)?.defaultValue?.toDefaultValueExpression() else null
+
+        val result = symbol.getDefaultValue()
+        return if (result != null)
+            result
+        else { // in the case of fake declarations
+            @OptIn(KaContextParameterApi::class)
```
whyoleg: Curious: Why do we need an opt-in here? It looks unused.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -788,9 +788,30 @@ internal class DokkaSymbolVisitor(
...
+            val ownerFunction = symbol.containingDeclaration as? KaNamedFunctionSymbol ?: return null
+
+            fun KaDeclarationSymbol.findMatchingParameterWithDefaultValue(): Expression? =
```
whyoleg: Do you mind moving this function just below `getDefaultValue`, as it's a bit hard to follow the logic with the function in the middle of the expression?

AbdullinAM (approved):

PR #4346, opened 2025-11-10
Use language-java class for <pre> and <code> tags generated from Java sources

### LINKED ISSUE #4345: Incorrect language-kotlin class used in <pre> and <code> tags for Java source
**Describe the bug**
 When generating html documentation for
```java
    /**
     * <pre>
     * {@code MyJavaFeatureUtil.doFeature()}
     * </pre>
     */
    public static void doFeature1() {
        System.out.println("Do feature1");
    }
```
from java source file, output html for this section is

Image

`<pre>` and `<code>` tags are using `language-kotlin` class.

**Expected behaviour**
When generating documentation from java source files, the class for html tags should be `language-java` instead of `language-kotlin`.

**To Reproduce**
Use [java-example](https://github.com/Kotlin/dokka/tree/master/examples/gradle-v2/java-example) project and add code from **Describe the bug** section.

yuriioliinyk4:
fix #4345


whyoleg (approved):
Nice!

Issue #4311, opened 2025-10-24
KMP: `Multiple pages associated with key` occurs when there are both actual and non-actual declarations with the same name


vmishenev:
> I am getting this error as well with my library, [Kotbase](https://github.com/jeffdgr8/kotbase). Set `dokka = 2.1.0` in `libs.versions.toml` and run `./gradlew dokkaGenerate`:
> 
> <details>
> 
> <summary>Full Gradle output</summary>
> 
> ```
> w: [:couchbase-lite:dokkaGenerateModuleHtml] Couldn't resolve link for init in CouchbaseLite.jvm.kt/kotbase.internalInit
> w: [:couchbase-lite-ee:dokkaGenerateModuleHtml] Couldn't resolve link for init in CouchbaseLite.jvm.kt/kotbase.internalInit
> 
> Task :couchbase-lite-ktx:logLinkDokkaGeneratePublicationHtml
> Generated Dokka HTML publication: http://localhost:63342/kotbase/couchbase-lite-ktx/build/dokka/html/index.html
> 
> Task :couchbase-lite-kermit:logLinkDokkaGeneratePublicationHtml
> Generated Dokka HTML publication: http://localhost:63342/kotbase/couchbase-lite-kermit/build/dokka/html/index.html
> 
> Task :couchbase-lite-paging:logLinkDokkaGeneratePublicationHtml
> Generated Dokka HTML publication: http://localhost:63342/kotbase/couchbase-lite-paging/build/dokka/html/index.html
> w: [:couchbase-lite:dokkaGeneratePublicationHtml] Couldn't resolve link for init in CouchbaseLite.jvm.kt/kotbase.internalInit
> 
> Task :couchbase-lite:dokkaGenerateModuleHtml FAILED
> w: [:couchbase-lite-ee:dokkaGeneratePublicationHtml] Couldn't resolve link for init in CouchbaseLite.jvm.kt/kotbase.internalInit
> Task :couchbase-lite-ee:dokkaGenerateModuleHtml FAILED
> Task :couchbase-lite:dokkaGeneratePublicationHtml FAILED
> 
> Task :couchbase-lite:logLinkDokkaGeneratePublicationHtml
> Generated Dokka HTML publication: http://localhost:63342/kotbase/couchbase-lite/build/dokka/html/index.html
> 
> Task :couchbase-lite-ee-kermit:logLinkDokkaGeneratePublicationHtml
[...]

Issue #3351, opened 2023-11-16
Incorrect link for properties in java stdlib


atyrin:
```Kotlin
/**
 * @see  [java.io.StringReader.lock] see property
 * @see  [java.io.StringReader.mark] see method
 */
class ClassKDocLocation {
}
```

The links for the class and method are correct. But 
1. for the property, we add a bracket `()` at the end of the line, and it searches a function instead of a property on the page: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/Reader.html#lock()
2. For method with parameter like `[java.io.StringReader.mark]` we navigates to `StringReader.html#mark(kotlin.Int)` but the expected link is `StringReader.html#mark(int)`

**Installation**
- Dokka version: 1.9.20

vmishenev:
The first case has been fixed in [#4360](https://github.com/Kotlin/dokka/pull/4360)

Issue #4338, opened 2025-11-04
Duplicate source links for the same DRI


vmishenev:
```kt
val ff = 0
fun ff() = 0
```

<img width="2310" height="402" alt="Image" src="https://github.com/user-attachments/assets/ee1b40af-285a-43db-bdbc-9e54e243bc64" />

Related: #4049, #3351

PR #4347, opened 2025-11-10
Introduce `isProperty` to `DRI` and fix merging property and function with the same name

vmishenev:
Introduces `isProperty` in `DRI` to distinguish between a property and a parameterless function with the same name, e.g.,  `val f` and `fun f()`. Meanwhile, DRI allows differentiating overloads.

Fixes https://github.com/Kotlin/dokka/issues/4311, containing the corner case because a property and a function have the same `DRI` in the same source set.

Potentially, Introducing `isProperty` could also help get https://github.com/Kotlin/dokka/issues/3351 and https://github.com/Kotlin/dokka/issues/4338 fixed

### DIFF
--- a/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
+++ b/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
@@ -92,7 +92,10 @@ public data class Callable(
     val isProperty: Boolean = false
 ) {
     init {
-        if(isProperty) require(params.isEmpty())
+        if (isProperty) require(
+            params.isEmpty(),
+            { "Callable $name is property, but has the following parameters: $params" })
+
     }
 
     @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
@@ -125,22 +128,17 @@ public data class Callable(
         contextParameters: List<TypeReference>
     ): Callable = Callable(name, receiver, params, contextParameters)
 
-    public fun signature(): String = buildString {
+    public fun signature(): String {
         /**
          * Compatibility with [signature] without context parameters must be preserved,
          * as package-list (e.g. `DefaultExternalLocationProvider` in the base plugin) and unit tests
          * rely on `dri.toString`
          */
-        @OptIn(ExperimentalDokkaApi::class)
-        if (contextParameters.isNotEmpty()) {
-            contextParameters.joinTo(this, separator = "#", postfix = "#")
-        }
-        append(receiver?.toString().orEmpty()).append("#")
-        if (isProperty) {
-            append("=") // `params` are replaced with `=`
-        } else {
-            params.joinTo(this, separator = "#")
-        }
+        val contextParameters = @OptIn(ExperimentalDokkaApi::class) contextParameters
+        return if (contextParameters.isNotEmpty())
+            "${contextParameters.joinToString("#")}#${receiver?.toString().orEmpty()}#${params.joinToString("#")}"
+        else
+            "${receiver?.toString().orEmpty()}#${params.joinToString("#")}"
     }
 
     public companion object
--- a/dokka-subprojects/plugin-base/src/test/kotlin/basic/DRITest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/basic/DRITest.kt
@@ -544,7 +544,7 @@ class DRITest : BaseAbstractTest() {
                     propDRI
                 )
                 assertEquals(
-                    "test//prop/#=/PointingToDeclaration/",
+                    "test//prop/#/PointingToDeclaration/",
                     propDRI.toString()
                 )
             }
--- a/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
@@ -1068,7 +1068,7 @@ class LinkTest : BaseAbstractTest() {
                         "String.length" to DRI(
                             "kotlin",
                             "String",
-                            Callable(name = "length", receiver = null, params = emptyList())
+                            Callable(name = "length", receiver = null, params = emptyList(), isProperty = true)
                         )
                     ),
                     module.getAllLinkDRIFrom("usage")


dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocAccessorNamingTest.kt
```diff
@@ -80,7 +80,7 @@ internal class JavadocAccessorNamingTest : AbstractJavadocTemplateMapTest() {
 
                 assertEquals(setOf(
                     "issuesFetched",
-                    "isFoo()",
+                    "isFoo",
```
vmishenev, line 83: Before the fix, it was wrong because, in this case, the getter and the backing field have the same name `isFoo`.

dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
```diff
@@ -445,9 +446,15 @@ class ExpectActualsTest : BaseAbstractTest() {
...
-    fun `should merge an implicit-expectActual function with a single property #3685`() = testInline(
-        """
+    /**
+     * This is a corner case because a property and a function have the same DRI in the same source set
+     */
+    @Nested
+    inner class MergingPropertyAndFunction {
+
+        @Test
+        fun `should merge an implicit-expectActual function with a single property #3685`() = testInline(
```
vmishenev, line 456: The behaviour has been changed, but this case is rare and used to cause an exception.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -87,7 +88,8 @@ public data class Callable(
     val receiver: TypeReference? = null,
     val params: List<TypeReference>,
     @property:ExperimentalDokkaApi
-    val contextParameters: List<TypeReference> = emptyList<TypeReference>()
+    val contextParameters: List<TypeReference> = emptyList(),
+    val isProperty: Boolean = false
```
whyoleg, line 92: Overall, the changes look fine.
The only concern I have is about invariant, that we can have non-empty `params` and `isProperty=true`.

But I don't have a good suggestion here:
1. or we can make `params` nullable (`null` meaning property) - but then `null` vs `emptyList` might be confusing a bit
2. or we can have a check in initializer, like `if(isProperty) require(params.isEmpty())`
3. or we can do nothing?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -21,7 +22,7 @@ public data class DRI(
     val extra: String? = null
 ) {
     override fun toString(): String =
-        "${packageName.orEmpty()}/${classNames.orEmpty()}/${callable?.name.orEmpty()}/${callable?.signature()
+        "${packageName.orEmpty()}/${classNames.orEmpty()}/${callable?.name.orEmpty() + if(callable?.isProperty == true) "=" else ""}/${callable?.signature()
```
whyoleg: Curious: why not embed this change into the `signature()` call?

vmishenev: because `callable?.name` is not there.
We can move it to `signature()`

whyoleg: Ah, I see, you added `=` just right after the `name`, not as part of the `signature`. Is there a reason for this?
I would expect `isProperty` to affect only `signature` output (meaning in the last part after `/`)

whyoleg:
BTW, it would be nice to add at least one test which tests `DRI.toString` with `isProperty=true`

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: `Callable. signature()` prints only a receiver and a param list and no name. I guess you would expect that the `name` also affects the `signature` output. This is probably not a good API - #3365.
For the sake of readability, the trait `isProperty` was bound to the name.
Anyway, I do not have good ideas on how to use `isProperty` in `Callable.signature()`. If you have any, I can implement one of them.

whyoleg: I would really just use `=` instead of parameters in `signature`.

Here is an example of the implementation for `signature` (uses `buildString` for readability):
```kotlin
public fun signature(): String = buildString {
    /**
     * Compatibility with [signature] without context parameters must be preserved,
     * as package-list (e.g. `DefaultExternalLocationProvider` in the base plugin) and unit tests
     * rely on `dri.toString`
     */
    @OptIn(ExperimentalDokkaApi::class)
    if (contextParameters.isNotEmpty()) {
        contextParameters.joinTo(this, separator = "#", postfix = "#")
    }
    append(receiver?.toString().orEmpty()).append("#")
    if (isProperty) {
        append("=") // `params` are replaced with `=`
    } else {
        params.joinTo(this, separator = "#")
    }
}
```

Here are some examples:
```
 com.example//function/#kotlin.Int - fun function(int: Int)
 com.example//function/#           - fun function()
 com.example//property/#=          - val property: Unit
```
(first `#` is because of the empty receiver - nothing changed here)

I feel like this way it's a bit more consistent, and we have `name` as a separate segment.
WDYT?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -87,8 +88,13 @@ public data class Callable(
     val receiver: TypeReference? = null,
     val params: List<TypeReference>,
     @property:ExperimentalDokkaApi
-    val contextParameters: List<TypeReference> = emptyList<TypeReference>()
+    val contextParameters: List<TypeReference> = emptyList(),
+    val isProperty: Boolean = false
 ) {
+    init {
+        if(isProperty) require(params.isEmpty())
```
whyoleg: minor: it might make sense to emit some message, even if it's mostly an assertion, something like `Callable $name is property, but has the following parameters: $params`

dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocAccessorNamingTest.kt
whyoleg: This means that this PR also fixes the first point from #3351, right?

whyoleg:
One last question, I promise :)

> Potentially, it could also fix https://github.com/Kotlin/dokka/issues/3351 and https://github.com/Kotlin/dokka/issues/4338

Will you add tests for those in a separate PR then?

vmishenev:
As we decided not to break `package-list` https://github.com/Kotlin/dokka/pull/4344#discussion_r2501226860, the old version of `DRI.toString`/`Callable.toString` has been restored. 

> Will you add tests for those in a separate PR then?

Sure

dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocAccessorNamingTest.kt
vmishenev: No, this test is about the Javadoc format.
To fix #3351, `JavadocExternalLocationProvider.kt` should be fixed as well. 

I updated the PR description to "Introducing isProperty could also help get https://github.com/Kotlin/dokka/issues/3351 and https://github.com/Kotlin/dokka/issues/4338 fixed" to avoid GitHub's auto-linking.

whyoleg: Thanks, now I see!

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -87,8 +88,16 @@ public data class Callable(
...
     val params: List<TypeReference>,
     @property:ExperimentalDokkaApi
-    val contextParameters: List<TypeReference> = emptyList<TypeReference>()
+    val contextParameters: List<TypeReference> = emptyList(),
+    val isProperty: Boolean = false
 ) {
+    init {
+        if (isProperty) require(
+            params.isEmpty(),
+            { "Callable \$name is property, but has the following parameters: $params" })
```
whyoleg:
```suggestion
            { "Callable $name is property, but has the following parameters: $params" })
```

Issue #4106, opened 2025-05-18
Allow regex setup for suppressedFiles


Den-Rimus:
**Question**
Or can this be achieved somehow with v2.0.0?

**Installation**
- Gradle 8.6, AGP 8.3.2
- Dokka version: 2.0.0

**Additional context**
We are maintaining quite a large SDK and are using Dokka to generate API reference page content for the Android part of it.
Some kinds of files are better to be omitted during documentation generation. However, they don't belong to some particular package (can't use the `perPackage` option). Also, some of them might share a naming scheme that could be utilized for regex suppressedFiles config. I might also mention that there are too many of those files to just plain enumerate them, plus - this list would then need to be maintained which is quite an error-prone idea from the start.

Before version 2.0.0 (actually, from version 1.4.x and up to 1.8.20) I just used your forked repo with a modified publishing config (replaced credentials to publish the gradle plugin to our repo) and adjusted the `isSuppressed` method (posting the original version):

```kotlin
    private fun isSuppressed(documentable: Documentable): Boolean {
        if (documentable !is WithSources) return false
        val sourceFile = File(source(documentable).path).absoluteFile
        return sourceSet(documentable).suppressedFiles.any { suppressedFile ->
            sourceFile.startsWith(suppressedFile.absoluteFile)
        }
    }
```

Should I continue with this approach to achieve the same behavior with v2.0.0 or is there a more adequate solution?
Could you register this as a feature request for future releases?

mtinnes:
I would agree that more granular control over what is included/excluded would be nice. The ability to selectively include specific classes/interfaces without necessarily having to include an entire package.

whyoleg:
I agree that it could be useful, but there are just too many possible options for filters, and not sure if Dokka will support all of them. But it's possible to write a small Dokka plugin similar to the [one we have in examples](https://github.com/Kotlin/dokka/blob/master/examples/plugin/hide-internal-api/src/main/kotlin/org/example/dokka/plugin/HideInternalApiPlugin.kt#L24-L33) - you just need to override one function and specify which declarations to include.

Regarding filtering only files, it's also possible to do it on the Gradle side:
1. Using `suppressedFiles` - is there a reason why you can't just provide all files to suppress there?
2. It should be possible to copy files, which need to be documented, to a temporary directory and then set `sourceRoots` to this directory.

PR #4348, opened 2025-11-14
Fix `suppressGeneratedFiles` was unused

### LINKED ISSUE #4217: Dokka 2.0.0 documenting files generated by Room
**Describe the bug**
Dokka is documenting _impl files generated by Room even when enabling suppressGeneratedFiles

**Expected behaviour**
_impl files not documented

**Screenshots**
Image

**To Reproduce**
- Create an Android library project using Room
- Generate dokka

**Dokka configuration**
```Kotlin
dokka {
    moduleName.set("Hydra core")
    dokkaSourceSets.main {
        includes.from("${rootProject.projectDir}/core/README.md")
        enableAndroidDocumentationLink = false
        suppressGeneratedFiles = true
        sourceLink {
            localDirectory.set(file("src/main/java"))
            remoteUrl("https://github.com/hydra-app/core/tree/master/core/src/main/java")
            remoteLineSuffix.set("#L")
        }
    }
    pluginsConfiguration.html {
        customAssets.from("${rootProject.projectDir}/logo-icon.svg")
        footerMessage = "© 2021-2025 Copyright KNF Apps"
    }
    dokkaPublications.html {
        suppressInheritedMembers.set(true)
        outputDirectory.set(layout.buildDirectory.dir("$rootDir/docs"))
    }
    dokkaPublications.javadoc {
        suppressInheritedMembers.set(true)
        outputDirectory.set(layout.buildDirectory.dir("$rootDir/javadoc"))
    }
}
```

**Installation**
- Operating system: Windows
- Build tool: Gradle v8.5
- Dokka version: 2.0.0

whyoleg:
Fixes #4217

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
@@ -5,6 +5,7 @@ package org.jetbrains.dokka.gradle.engine.parameters
 
 import org.gradle.api.*
 import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.FileCollection
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.plugins.ExtensionAware
 import org.gradle.api.provider.Property
@@ -16,7 +17,6 @@ import org.jetbrains.dokka.gradle.engine.parameters.SourceSetIdSpec.Companion.do
 import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
 import org.jetbrains.dokka.gradle.internal.adding
 import org.jetbrains.dokka.gradle.internal.domainObjectContainer
-import java.io.File
 import java.io.Serializable
 import javax.inject.Inject
 import kotlin.DeprecationLevel.ERROR
@@ -448,12 +448,12 @@ constructor(
     @get:InputFiles
     @get:IgnoreEmptyDirectories
     @get:PathSensitive(PathSensitivity.RELATIVE)
-    internal val inputSourceFiles: List<File>
+    internal val inputSourceFiles: FileCollection
         get() = sourceRoots.asFileTree.filter { sourceFile ->
             suppressedFiles.none { suppressedFile ->
                 sourceFile.startsWith(suppressedFile)
             }
-        }.sortedBy { it.invariantSeparatorsPath } // FileTrees have an unstable order (even on the same machine)
+        }
 
     companion object {
 
--- a/dokka-runners/dokka-gradle-plugin/src/test/kotlin/engine/parameters/DokkaSourceSetSpecTest.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/test/kotlin/engine/parameters/DokkaSourceSetSpecTest.kt
@@ -4,8 +4,12 @@
 package org.jetbrains.dokka.gradle.engine.parameters
 
 import io.kotest.core.spec.style.FunSpec
+import io.kotest.core.test.TestScope
+import io.kotest.inspectors.shouldForAll
 import io.kotest.matchers.collections.shouldBeEmpty
 import io.kotest.matchers.collections.shouldContainExactly
+import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
+import io.kotest.matchers.file.shouldBeAFile
 import io.kotest.matchers.shouldBe
 import org.gradle.api.Project
 import org.gradle.kotlin.dsl.apply
@@ -16,6 +20,11 @@ import org.jetbrains.dokka.gradle.DokkaPlugin
 import org.jetbrains.dokka.gradle.engine.parameters.VisibilityModifier.Public
 import org.jetbrains.dokka.gradle.utils.create_
 import org.jetbrains.dokka.gradle.utils.enableV2Plugin
+import java.nio.file.Path
+import kotlin.io.path.Path
+import kotlin.io.path.createDirectories
+import kotlin.io.path.relativeTo
+import kotlin.io.path.writeText
 
 class DokkaSourceSetSpecTest : FunSpec({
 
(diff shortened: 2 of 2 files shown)


whyoleg:
@adam-enko, `MultiModuleFunctionalTest` fails for some reason because of `should have outcome FROM_CACHE, but was SUCCESS` on CI.
Locally, everything is fine. Do you know what the reason might be?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
```diff
@@ -265,10 +265,16 @@ constructor(
...
-    @get:PathSensitive(PathSensitivity.RELATIVE)
+    @get:Internal
     abstract val suppressedFiles: ConfigurableFileCollection
 
+    /**
+     * We don't care about the content of those [suppressedFiles] as we use those only as paths to filter sources.
+     */
+    @get:Input
+    internal val suppressedFilesInput: List<String>
+        get() = suppressedFiles.map { it.path }.sorted()
```
adam-enko: Using the paths as an input will break relocatable build-cache. This could be the cause of the CI failure.

```
org.jetbrains.dokka.gradle.MultiModuleFunctionalTest > Gradle caching > Gradle caching build cache relocation > Gradle caching build cache relocation relocated project > org.jetbrains.dokka.gradle.MultiModuleFunctionalTest.Gradle caching build cache relocation relocated project should load all generation tasks from cache FAILED
    io.kotest.assertions.MultiAssertionError at MultiModuleFunctionalTest.kt:279
```

whyoleg: Hm, what would be a better way than to rely on paths, but not to rely on the content of the files? Relativize them?

The issue with relying on the content, apart from the fact that we don't care about the content here, is that in that case, our test starts to fail because we now declare an implicit dependency on some files generated by AGP. And that's correct.

adam-enko:
I think this needs to be fixed another way.

It doesn't make sense for `suppressedFiles` to be `@InputFiles`. This will make the Dokka tasks sensitive to changes in files that don't affect the output. Changes in suppressed files should be ignored. If only suppressed files change then the Dokka tasks should be loaded from cache.

It's similar for `sourceRoots` - it will track the suppressed files, which causes cache misses.

Instead of users supplying the files to exclude, I think it makes sense for them to pass in glob patterns for files to suppress.

Something like this:


```kotlin
abstract class DokkaSourceSetSpec {

    // ...

    @get:Internal // tracked by `actualFiles()`
    abstract val sourceRoots: ConfigurableFileCollection
    
    @get:Internal
    @Deprecated("Use `suppressedFilesPatterns` instead")
    abstract val suppressedFiles: ConfigurableFileCollection

    /** Glob patterns, relative to each source root. */
    @get:Internal // tracked by `actualFiles()`
    abstract val suppressedFilesPatterns: SetProperty<String>

    // this is just for task input tracking, it's not used anywhere
    @InputFiles
    @IgnoreEmptyDirectories
    @PathSensitive(PathSensitivity.RELATIVE)
    fun actualFiles(): FileCollection {
        return objects.fileCollection().from(
            sourceRoots
                .asFileTree
                .matching {
                    suppressedFilesPatterns.orNull?.forEach { pattern ->
                        exclude(pattern)
                    }
                }
                .files
                .sorted()
        )
    }

    // TODO Use this in DokkaSourceSetBuilder instead of `val suppressedFiles`
    @Internal
    fun suppressedFiles(): FileCollection {
        return objects.fileCollection().from(
            sourceRoots
                .asFileTree
                .matching {
                    suppressedFilesPatterns.orNull?.forEach { pattern ->
                        include(pattern)
                    }
                }
                .files
                .sorted()
        )
    }
```


There's also a new 'generated sources' option in KGP. DGP could integrate with that somehow...

https://youtrack.jetbrains.com/issue/KT-45161/Gradle-Support-registering-generated-sources-with-the-Kotlin-model#focus=Comments-27-12807139.0-0

whyoleg:
Hm, yes, that looks like a nice idea. The only problem I see is that in that case, we will provide all files separately to Dokka instead of just `roots`, which might affect performance inside Dokka, and will, for sure, make the configuration (JSON) much more verbose.

At this moment, I would really avoid such a significant change.

The new KGP API is available only in 2.3.0 (which does not even have a stable version), and is not yet adopted by any third-party plugins, but yes, we can integrate it with the `suppressGeneratedFiles` flag.

For patterns (or regex), there is #4106.

adam-enko:
> Hm, yes, that looks like a nice idea. The only problem I see is that in that case, we will provide all files separately to Dokka instead of just `roots`, which might affect performance inside Dokka, and will, for sure, make the configuration (JSON) much more verbose.

I don't think my proposal would affect that. `sourceRoots` would still be converted directly. The only change would be exposing the Gradle inputs.

whyoleg:
> I don't think my proposal would affect that. sourceRoots would still be converted directly. The only change would be exposing the Gradle inputs.

Oh, sorry, my bad! Looks like I misinterpreted that. Thanks!
I will try to implement the proposed solution then!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
```diff
@@ -444,6 +444,18 @@ constructor(
...
+    @Suppress("unused")
+    @get:InputFiles
+    @get:IgnoreEmptyDirectories
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    internal val inputSourceFiles: FileCollection
+        get() = sourceRoots.asFileTree.filter { sourceFile ->
+            suppressedFiles.none { suppressedFile ->
+                sourceFile.startsWith(suppressedFile)
+            }
+        }
```
adam-enko: Can you add a functional test (i.e. not using TestKit) to verify the contents?

A few cases to consider:

- multiple different files
- files and directories

Once done, you can remove the unused suppression.

Also, please check the order of files. FileTree is documented as having [an unstable order](https://github.com/gradle/gradle/blob/v9.2.1/subprojects/core-api/src/main/java/org/gradle/api/file/FileTree.java#L36), so I'm nervous about using it. A unit test would hopeful verify this.

whyoleg: I've added some tests; feel free to suggest more cases if you think I've missed something, as I could have. :)

Additionally, although I was unable to reproduce or understand how to introduce ordering problems, I decided to add sorting at the end, just in case. :)

dokka-runners/dokka-gradle-plugin/src/test/kotlin/engine/parameters/DokkaSourceSetSpecTest.kt
```diff
@@ -141,6 +144,158 @@ class DokkaSourceSetSpecTest : FunSpec({
             }
         }
     }
+
+    context("DokkaSourceSetSpec inputSourceFiles ->") {
```
adam-enko, line 157: Optional suggestion for these tests:

Instead of creating separate srcDir structures for each test, instead create a `Project.setupTestSrcDirs()` function that creates a srcDir structure that has all directories for all the tests.

At the moment it's a bit hard to read, since each test is testing separate directory structure and different options. If the dirs were consistent, it'd be easier to differentiate them based on the options they're testing.

This is just optional, I'll leave it up to you.

adam-enko (approved):
Nice!

Side note: I think it's quite surprising this bug was never found in Dokkatoo or Dokka, until recently!

whyoleg:
> Side note: I think it's quite surprising this bug was never found in Dokkatoo or Dokka, until recently!

In Dokka, it was reported on [Aug 19, 2025](https://github.com/Kotlin/dokka/issues/4217#issue-3332466681), it's just slipped from our attention.

But, I think, it's just because Dokka is more useful (and so used more) in libraries, and usage of code generation (like KSP) there is less popular.

dokka-runners/dokka-gradle-plugin/src/test/kotlin/engine/parameters/DokkaSourceSetSpecTest.kt
whyoleg: Hm, I agree, let me see what I can do

whyoleg: TBH, I failed to come up with a nice way here to test different combinations of suppresses (directories, files, flags) with a single structure.
It will still require manual file creation or manipulation, and so it will still have a similar structure...
I mean, what is passed in `sourceRoots`/`suppressedFiles` will vary from case to case, and what is compared in `inputSourceFiles` will also differ.

I thought about creating a class with pre-created directories, but it would be roughly the same, just hiding the paths...

Could you share your idea in "pseudo-code"? I feel, like I totally missing some obvious way to make this test better :)

adam-enko: Like this:

(I can just commit this directly if you like)

```kotlin

    context("DokkaSourceSetSpec inputSourceFiles ->") {

        val project = createProject()

        val projectDir = project.projectDir.toPath()
        fun Path.createClass(name: String) {
            createDirectories()
            resolve("$name.kt").writeText("class $name")
        }

        val mainSourceDir =
            projectDir.resolve("src/main/kotlin").apply {
                resolve("com/example").apply {
                    createClass("MainCls")
                    resolve("sub").apply {
                        createClass("Sub1MainCls")
                        createClass("Sub2MainCls")
                    }
                }
            }
        val generatedSrcDir =
            projectDir.resolve("build/generated/kotlin").apply {
                createClass("BuildGenCls")
                resolve("sub").apply {
                    createClass("Sub1BuildGenCls")
                    createClass("Sub2BuildGenCls")
                }
            }
        val customGeneratedSrcDir =
            projectDir.resolve("src/customGenerated/kotlin").apply {
                createClass("CustomGenCls")
                resolve("sub").apply {
                    createClass("Sub1CustomGenCls")
                    createClass("Sub2CustomGenCls")
                }
            }

        fun TestScope.createDss(): DokkaSourceSetSpec {
            val dss = project.createDokkaSourceSetSpec(testCase.name.testName)
            dss.sourceRoots.from(mainSourceDir, generatedSrcDir, customGeneratedSrcDir)
            return dss
        }

        fun DokkaSourceSetSpec.inputSourceFilesToRelativePaths(): List<Path> =
            inputSourceFiles.map { it.toPath().relativeTo(projectDir) }

        test("build/generated should be excluded by default") {
            val dss = createDss()
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("src/customGenerated/kotlin/CustomGenCls.kt"),
                Path("src/customGenerated/kotlin/sub/Sub1CustomGenCls.kt"),
                Path("src/customGenerated/kotlin/sub/Sub2CustomGenCls.kt"),
                Path("src/main/kotlin/com/example/MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub1MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub2MainCls.kt"),
            )
        }
        test("when suppressGeneratedFiles is set to false, expect all source files are included") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(false)
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("build/generated/kotlin/BuildGenCls.kt"),
                Path("build/generated/kotlin/sub/Sub1BuildGenCls.kt"),
                Path("build/generated/kotlin/sub/Sub2BuildGenCls.kt"),
                Path("src/customGenerated/kotlin/CustomGenCls.kt"),
                Path("src/customGenerated/kotlin/sub/Sub1CustomGenCls.kt"),
                Path("src/customGenerated/kotlin/sub/Sub2CustomGenCls.kt"),
                Path("src/main/kotlin/com/example/MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub1MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub2MainCls.kt"),
            )
        }
        test("input source files should only contain files, not directories") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(false)
            dss.inputSourceFiles.shouldForAll { it.shouldBeAFile() }
        }
        test("expect files can be excluded by exact path") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(false)
            dss.suppressedFiles.from(
                "build/generated/kotlin/sub/Sub1BuildGenCls.kt",
                "src/customGenerated/kotlin/sub/Sub1CustomGenCls.kt",
                "src/main/kotlin/com/example/sub/Sub1MainCls.kt",
            )
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("build/generated/kotlin/BuildGenCls.kt"),
                Path("build/generated/kotlin/sub/Sub2BuildGenCls.kt"),
                Path("src/customGenerated/kotlin/CustomGenCls.kt"),
                Path("src/customGenerated/kotlin/sub/Sub2CustomGenCls.kt"),
                Path("src/main/kotlin/com/example/MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub2MainCls.kt"),
            )
        }
        test("expect files can be excluded by base directories") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(false)
            dss.suppressedFiles.from(
                "build/",
                "src/",
            )
            dss.inputSourceFilesToRelativePaths().shouldBeEmpty()
        }
        test("expect files can be excluded by sub directories") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(false)
            dss.suppressedFiles.from(
                "build/generated/kotlin/sub/",
                "src/customGenerated/kotlin/sub/",
                "src/main/kotlin/com/example/sub/",
            )
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("build/generated/kotlin/BuildGenCls.kt"),
                Path("src/customGenerated/kotlin/CustomGenCls.kt"),
                Path("src/main/kotlin/com/example/MainCls.kt"),
            )
        }
        test("expect nested files can be excluded") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(false)
            dss.suppressedFiles.from(
                "build/generated/kotlin/sub/",
                "src/customGenerated/kotlin/sub/",
                "src/main/kotlin/com/example/sub/",
            )
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("build/generated/kotlin/BuildGenCls.kt"),
                Path("src/customGenerated/kotlin/CustomGenCls.kt"),
                Path("src/main/kotlin/com/example/MainCls.kt"),
            )
        }
        test("expect suppressGeneratedFiles and suppressedFiles (by specific files) exclude all generated and specified files") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(true)
            dss.suppressedFiles.from(
                "src/customGenerated/kotlin/sub/Sub1CustomGenCls.kt",
                "src/customGenerated/kotlin/sub/Sub2CustomGenCls.kt",
                "src/main/kotlin/com/example/sub/Sub1MainCls.kt",
                "src/main/kotlin/com/example/sub/Sub2MainCls.kt",
            )
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("src/customGenerated/kotlin/CustomGenCls.kt"),
                Path("src/main/kotlin/com/example/MainCls.kt"),
            )
        }
        test("expect suppressGeneratedFiles and suppressedFiles (by directories) exclude all generated and specified files") {
            val dss = createDss()
            dss.suppressGeneratedFiles.set(true)
            dss.suppressedFiles.from(
                "src/customGenerated/kotlin/",
            )
            dss.inputSourceFilesToRelativePaths() shouldContainExactly listOf(
                Path("src/main/kotlin/com/example/MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub1MainCls.kt"),
                Path("src/main/kotlin/com/example/sub/Sub2MainCls.kt"),
            )
        }
```

whyoleg: Oh, yeah, that really looks much nicer! Thanks!

> (I can just commit this directly if you like)

Yes please!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
adam-enko: whoops, I was wrong about the file ordering being important.

https://github.com/gradle/gradle/issues/36251#issuecomment-3744282351

I reverted it when updating the tests.

PR #4356, opened 2025-11-18
DGP integration tests: Add `@TestsKotlinJvm` tag

adam-enko:
Tag tests that use Kotlin JVM plugin.

(This will become more relevant when we add a test for AGP9 with Kotlin-built-in.)


adam-enko:
Merging despite the failing tests - they're caused by the current Cloudflare outage. The integration tests are passing.

PR #4359, opened 2025-11-19
DPG IT: Test `com.android.kotlin.multiplatform.library` with Kotlin Multiplatform

### LINKED ISSUE #4211: Add IT for a project with `com.android.kotlin.multiplatform.library` plugin
Google recently introduced a new plugin that provides external Android support for Kotlin Multiplatform instead of `androidTarget` coming from KGP
Migration guide: https://developer.android.com/kotlin/multiplatform/plugin

We should add an IT for this kind of project, as this will be the future way of supporting Android in Kotlin Multiplatform projects, so Dokka should work properly with them.

adam-enko:
Fix #4211

### DIFF
--- a/dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/gradle.properties
+++ b/dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/gradle.properties
@@ -3,6 +3,3 @@
 #
 
 android.useAndroidX=true
-
-org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
-org.jetbrains.dokka.experimental.gradle.pluginMode.noWarn=true
--- a/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
+++ b/dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
@@ -15,6 +15,7 @@ import org.jetbrains.dokka.gradle.utils.addArguments
 import org.jetbrains.dokka.gradle.utils.build
 import org.jetbrains.dokka.it.gradle.junit.*
 import org.jetbrains.dokka.it.gradle.junit.TestedVersions
+import org.jetbrains.dokka.it.gradle.utils.SemVer
 import org.junit.jupiter.api.Assumptions.assumeTrue
 import kotlin.io.path.name
 import kotlin.io.path.readText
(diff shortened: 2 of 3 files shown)


dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
```diff
@@ -0,0 +1,109 @@
...
+                """.trimIndent()
+            ) {
+                val expectedFileTree = expectedHtml.toTreeString()
+                val actualFileTree = actualHtmlDir.toTreeString()
+                withClue((actualFileTree to expectedFileTree).sideBySide()) {
+                    actualFileTree shouldBe expectedFileTree
+
+                    actualHtmlDir.shouldBeADirectoryWithSameContentAs(expectedHtml, TestConstants.DokkaHtmlAssetsFiles)
+                }
+            }
```
whyoleg, line 96: There should be a check in the end, like in other IT: `assertNoUnknownClassErrorsInHtml(actualHtmlDir)`

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
```diff
@@ -0,0 +1,109 @@
...
+@TestsKotlinMultiplatform
+class AndroidKotlinMultiplatformBuiltInTest {
+
+    @DokkaGradlePluginTest(sourceProjectName = "it-android-kotlin-mp-builtin")
+    fun `generate dokka HTML`(
+        project: DokkaGradleProjectRunner,
+        testedVersions: TestedVersions.Android,
+    ) {
+        assumeTrue(
+            testedVersions.agp.major >= 8,
```
whyoleg: Should we check for 8.10+, as stated in the docs?

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
```diff
@@ -0,0 +1,109 @@
...
+                    "(current KGP ${testedVersions.kgp}). " +
+                    "See https://developer.android.com/kotlin/multiplatform/plugin#prerequisites",
+        )
+
+        if (testedVersions.agp.major < 9) {
+            // AGP 8 uses KotlinTarget JVM instead of AndroidJVM.
+            // This is an AGP 8 bug - it's fixed in AGP 9.
+            // For AGP <9, workaround the bug by manually enabling the Android documentation link.
+            project.buildGradleKts += """
+            |dokka.dokkaSourceSets.configureEach {
```
whyoleg: The `workaround` is applied both here and in `build.gradle.kts` of the test project.
Probably, I would vote for having it in the project only, so that if the project is   opened separately (e.g. inside IDE), it will be there

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/gradle.properties
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+#
+
+android.useAndroidX=true
+
+org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
```
whyoleg: Is it still needed here?

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
adam-enko: thanks, added

dokka-integration-tests/gradle/src/test/kotlin/AndroidKotlinMultiplatformBuiltInTest.kt
adam-enko: good idea, done

PR #4361, opened 2025-11-19
Add tests for #4338 and #4049

vmishenev:


dokka-subprojects/plugin-base/src/test/kotlin/transformers/SourceLinkTransformerTest.kt
```diff
@@ -128,4 +128,54 @@ class SourceLinkTransformerTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Test
+    fun `each declaration should include exactly one source link in the case #4338 and #4049`() {
```
whyoleg: Would you mind splitting the test into two parts, one for #4338 and one for #4049? 
While the reason is the same (DRI), those are caused by two different parts of the DRI.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/SourceLinkTransformerTest.kt
```diff
@@ -128,4 +128,54 @@ class SourceLinkTransformerTest : BaseAbstractTest() {
...
+            |fun overloadWithVararg(vararg elements: String){}
+        """.trimMargin(),
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
+                val pageFf = writerPlugin.writer.renderedContent("root/testpackage/ff.html")
+                fun Element.getAllSourceElements() = select(".symbol .source-link")
+                assertEquals(
+                    2,
```
whyoleg, line 166: It would be nice to check not just the number of source links, but also their content; otherwise, the test feels a bit odd.

whyoleg (approved):
Just minor suggestions

PR #4362, opened 2025-11-19
Only apply 'Adapter' plugins once, and group PluginIds

adam-enko:
The AndroidAdapter, JavaAdapter, and KotlinAdapters apply their configuration multiple times.

Currently, this is acceptable because a project cannot have multiple Android/Kotlin/Java plugins applied.
But the situation gets more complicated with AGP9 and kotlin-built-in.
The kotlin-built-in plugin uses an implementation of `KotlinBasePlugin`,
so the Android and Kotlin adapters could get applied multiple times, and some of their operations (registering DokkaSourceSets) aren't idempotent.

I also updated the JavaAdapter to use the same approach for consistency (it's not strictly necessary).

---

Also, group the plugin IDs into Sets, so they can be accessed at once without needing to remember which IDs should be accessed.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
@@ -18,8 +18,6 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.provider.ProviderFactory
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.getByType
-import org.gradle.kotlin.dsl.withType
-import org.jetbrains.dokka.gradle.DokkaBasePlugin
 import org.jetbrains.dokka.gradle.DokkaExtension
 import org.jetbrains.dokka.gradle.engine.parameters.KotlinPlatform
 import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
@@ -76,11 +74,9 @@ abstract class AndroidAdapter @Inject constructor(
          * Apply [AndroidAdapter] a single time to [project], regardless of how many AGP plugins are applied.
          */
         internal fun applyTo(project: Project) {
-            project.plugins.withType<DokkaBasePlugin>().all {
-                PluginIds.android.forEach { pluginId ->
-                    project.pluginManager.withPlugin(pluginId) {
-                        project.pluginManager.apply(AndroidAdapter::class)
-                    }
+            PluginIds.android.forEach { pluginId ->
+                project.pluginManager.withPlugin(pluginId) {
+                    project.pluginManager.apply(AndroidAdapter::class)
                 }
             }
         }


dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
```diff
@@ -9,12 +9,23 @@ package org.jetbrains.dokka.gradle.internal
...
-    const val AndroidBase = "com.android.base"
-    const val AndroidApplication = "com.android.application"
-    const val AndroidLibrary = "com.android.library"
+    private const val AndroidBase = "com.android.base"
+    private const val AndroidApplication = "com.android.application"
+    private const val AndroidLibrary = "com.android.library"
+    val androidPlugins: Set<String> = setOf(
+        AndroidBase,
+        AndroidApplication,
+        AndroidLibrary,
```
whyoleg: We should probably add `test` and `dynamic-feature`, right?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
```diff
@@ -9,12 +9,23 @@ package org.jetbrains.dokka.gradle.internal
...
 internal object PluginId {
 
-    const val KotlinAndroid = "org.jetbrains.kotlin.android"
-    const val KotlinJs = "org.jetbrains.kotlin.js"
-    const val KotlinJvm = "org.jetbrains.kotlin.jvm"
-    const val KotlinMultiplatform = "org.jetbrains.kotlin.multiplatform"
+    private const val KotlinAndroid = "org.jetbrains.kotlin.android"
+    private const val KotlinJs = "org.jetbrains.kotlin.js"
+    private const val KotlinJvm = "org.jetbrains.kotlin.jvm"
+    private const val KotlinMultiplatform = "org.jetbrains.kotlin.multiplatform"
```
whyoleg: Minor: It feels like we can just inline those into `kgpPlugins` (the same applies to the AGP ones).

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
```diff
@@ -9,12 +9,23 @@ package org.jetbrains.dokka.gradle.internal
...
+    val kgpPlugins: Set<String> = setOf(
```
whyoleg: nitpick:
```suggestion
    val kotlinPlugins: Set<String> = setOf(
```
(or the other way around, `agpPlugins` instead of `androidPlugins`)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginId.kt
adam-enko: Could do, up to you.

I wanted to keep this PR focused and not add more plugins, and just keep it as an internal change.

whyoleg: I believe it should be acceptable to include this in this PR, as we have decided to add them without additional tests.
No need to have a separate PR for this, and it will be even more strange to include them in AGP9 PR, IMO.

Issue #3752, opened 2024-08-19
[K2] Documentation on extension properties is not propagated to the generated extension getters


vmishenev:
From Google:
> If you create `val FooClass.extensionProperty` with `/** docs for extensionProperty */`, those docs will never appear for Java consumers. Java consumers never use (and don't get documented) extensionProperty, they only can use `FooClassKt.getExtensionProperty(FooClass)`, which is an autogenerated getter for extensionProperty. With the K1 backend, dackka is putting /** docs for extensionProperty */ on getExtensionProperty, with the K2 backend there are no docs for getExtensionProperty, which means the documentation for extensionProperty is never surfaced for java consumers.

> Kotlin K2 IDE plugin (which uses AA) works fine in this case and shows documentation correctly
![image](https://github.com/user-attachments/assets/b7742490-2c1f-442c-89fd-424140eb2492)
![image](https://github.com/user-attachments/assets/7ef33be5-b70c-4c9e-a076-06b3b08b6585)

owengray-google:
Actually, it seems like this is happening to non-extension properties in (latest) K2 dokka.
The test we have (failing at the `fooDemocracy` java-only assertion) is:
```
@Test
fun `Class component inherits docs from same language 4x test`() {
    val pagesJ =
        """
        |public class foo {
        |    /** dew it */
        |    public void doit() {}
        |}
        |public class bar extends foo {
        |    /** {@inheritDoc} */
        |    @Override
        |    public void doit() {}
        |}
        |public class baz extends foo {
        |    /** overriding function docs */
        |    @Override
        |    public void doit() {}
        |}
    """
            .render(java = true) // Java {@inheritdoc} does not support classes & properties
    val pagesK =
        """
        | /** docs for foo */
        |class foo() {
        |    /** dew it */
        |    fun doit() {}
        |    /** thunderous applause */
        |    open val democracy = false
        |}
        |class bar() : foo {
        |    override fun doit() {}
        |    override val democracy = true
        |}
        | /** overriding docs for baz */
        |class baz() : foo {
        |    /** overriding function docs */
        |    override fun doit() {}
        |    /** KotOR */
        |    override val democracy = true
        |}
        |class maz() : foo {
        |   /** {@inheritDoc} */
        |   override fun doit() {}
        |}        """
            .render()

    for (pages in listOf(pagesJ, pagesK)) {
        val fooClass = pages.page("foo").data.content
        val barClass = pages.page("bar").data.content
[...]

Issue #2493, opened 2022-05-09
Allow actual declarations to automatically "inherit" their documentation from expect counterparts in multiplatform


mikeholler:
**Is your feature request related to a problem? Please describe**

src/kotlin/commonMain
```kotlin
/**
 * Documentation for X.
 **/
expect fun x(): Unit
```

src/kotlin/jvmMain:
```kotlin
actual fun x(): Unit {}
```

I would like to see "Documentation for X." show up in the documentation for the `actual` `jvmMain` function in the generated Dokka documentation. Currently, it does not. @IgnatBeresnev suggested I create a ticket here from `kotlinlang` Slack group. You can find that thread here: https://kotlinlang.slack.com/archives/C0F4UNJET/p1651761386753379
You can also paste a link to the relevant issue

**Describe the solution you'd like**

No changes to the above code, but a behavior change that if an `actual` signature does not have doc comments, then Dokka should check for comments in the `expect` function, and use them instead.

**Describe alternatives you've considered**

This behavior could also be opt-in, by using a documentation directive like `@includeFromExpect`, but I don't like this as much because it would lead to many duplicated doc blocks.

src/kotlin/jvmMain:
```kotlin
/**
 * @includeFromExpect
 **/
actual fun x(): Unit {}
```

**Additional context**

@IgnatBeresnev suggested that giving this issue a `kdoc-spec` label may be appropriate.

**Are you willing to provide a PR?**
I'm not.

IgnatBeresnev:
Thank you so much for creating the issue!

Indeed, it's unclear at the moment whether it should or should not be inherited and how (automatically or via additional tags), so this needs to be discussed.

I believe @owengray-google is also interested in this question/enhancement for android documentation reference, so adding the google tag

joffrey-bion:
> This behavior could also be opt-in, by using a documentation directive like @includeFromExpect, but I don't like this as much because it would lead to many duplicated doc blocks.

To draw a parallel, Java's `@inheritDoc` directive is NOT an opt-in mechanism for inheritance of Javadoc. Its purpose is to allow injecting the inherited doc somewhere in the child doc when the author wants to write child-specific stuff but still reuse doc from the parent.

The doc is inherited by default (from the inheritance hierarchy) if no Javadoc is present on the overriding member.

> Indeed, it's unclear at the moment whether it should or should not be inherited and how

Could someone please clarify the exact problematic use cases for inheritance of `expect` doc in the `actual` entities?

[...]

PR #4363, opened 2025-11-21
New KDoc api

AbdullinAM:
* Update to the latest AA
* Enable 'LanguageFeature.MultiPlatformProjects` in AA configuration
* Switch to the new KDoc resolve API
* Add tests for #3752, #3369
* Split several old tests to separate K1/K2 variants because of the difference in behaviour of the new KDoc resolve

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
@@ -67,9 +67,9 @@ internal fun getLanguageVersionSettings(
             // force to resolve light classes (lazily by default)
             AnalysisFlags.eagerResolveOfLightClasses to true,
         ),
-        specificFeatures = hashMapOf(
-            LanguageFeature.MultiPlatformProjects to if (isMultiplatformProject) LanguageFeature.State.ENABLED else LanguageFeature.State.DISABLED
-        )
+        specificFeatures = listOfNotNull(
+            if (isMultiplatformProject) LanguageFeature.MultiPlatformProjects to LanguageFeature.State.ENABLED else null
+        ).toMap()
     )
 }
 
@@ -80,7 +80,7 @@ internal fun createAnalysisSession(
     isSampleProject: Boolean = false
 ): KotlinAnalysis {
     val sourcesModule = mutableMapOf<DokkaConfiguration.DokkaSourceSet, KaSourceModule>()
-    val isMultiplatformProject = sourceSets.mapTo(mutableSetOf()) { it.analysisPlatform }.size > 1
+    val isMultiplatformProject = sourceSets.any { it.analysisPlatform != Platform.jvm }
 
     val analysisSession = buildStandaloneAnalysisAPISession(
         projectDisposable = projectDisposable,


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
```diff
@@ -62,6 +63,9 @@ internal fun getLanguageVersionSettings(
             // special flag for Dokka
             // force to resolve light classes (lazily by default)
             AnalysisFlags.eagerResolveOfLightClasses to true
+        ),
+        specificFeatures = mapOf(
+            LanguageFeature.MultiPlatformProjects to LanguageFeature.State.ENABLED,
```
whyoleg: Should we enable it only if we have a truly multi-platform project? For example, in case we have at least one source set with a non-JVM platform?

It would be nice to ask the AA/FE teams regarding the consequences of enabling this flag.

dokka-subprojects/plugin-base/src/test/kotlin/content/seealso/ContentForSeeAlsoTest.kt
```diff
@@ -824,7 +824,115 @@ class ContentForSeeAlsoTest : BaseAbstractTest() {
...
+                                group {
+                                    link { +"Unit" }
+                                    check {
+                                        sourceSets.assertSourceSet("linuxX64")
+                                    }
+                                }
+                                group {
+                                    link { +"Unit" }
+                                    check {
+                                        sourceSets.assertSourceSet("jvm")
```
whyoleg, line 903: Hmm, curious, why do we have `see` propagated here to `jvm` source set, if it's not declared in the `jvm` source set, while for `throws` and `sample` we don't propagate it.

dokka-subprojects/plugin-base/src/test/kotlin/content/exceptions/ContentForExceptions.kt
```diff
@@ -257,7 +257,100 @@ class ContentForExceptions : BaseAbstractTest() {
     }
 
     @Test
-    fun `multiplatofrm class with throws in few platforms`() {
+    @OnlySymbols("K2 inherits KDoc from expect")
```
whyoleg: Do you mind adding a link to a Dokka issue at the beginning of messages to conditional annotations, like this:
```suggestion
    @OnlySymbols("#4245: K2 inherits KDoc from expect")
```

So that it will be easier to get the context later?

dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
```diff
@@ -732,4 +734,56 @@ class ExpectActualsTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Test
+    @OnlySymbols("New KDoc resolve")
+    @OptIn(ExperimentalDokkaApi::class)
```
whyoleg: Minor: opt-in looks redundant.

dokka-subprojects/plugin-base/src/test/kotlin/expectActuals/ExpectActualsTest.kt
```diff
@@ -732,4 +734,56 @@ class ExpectActualsTest : BaseAbstractTest() {
...
+        """.trimMargin(),
+            multiplatformConfiguration
+        ) {
+            documentablesTransformationStage = { module ->
+                val a = module.findClasslike("", "A") as DClass
+                val constructor = a.constructors.single()
+                val foo = a.functions.first { it.name == "foo" }
+                val bar = a.functions.first { it.name == "bar" }
+
+                assertEquals(2, constructor.documentation.size)
```
whyoleg, line 781: It would be nice to check not just sizes, but some other content, like documentation override in case of `bar` on JVM, or source-sets, otherwise it's not clear what the test is really tests :)

dokka-subprojects/plugin-kotlin-as-java/src/main/kotlin/org/jetbrains/dokka/kotlinAsJava/converters/KotlinToJavaConverter.kt
whyoleg, line 1: Could you please extract those changes (related to `kotlin-as-java`) into a separate PR? I do have some additional questions for them, and they seem a bit unrelated to all the other changes in the PR.

AbdullinAM: Makes sense! Will do

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
AbdullinAM: According to the usages in the compiler, this feature is only used to skip some resolve and check operations related to expect/actual. So the downside of enabling it unconditionally is a potential performance hit.

I will change the behaviour to be only enabled for multiplatform projects

whyoleg: > this feature is only used to skip some resolve and check operations related to expect/actual

Yeah, but it's only how it works **now**. It might change at any moment.

dokka-subprojects/plugin-base/src/test/kotlin/content/seealso/ContentForSeeAlsoTest.kt
AbdullinAM: That is how the content page creation for these tags is implemented in [DescriptionSections.kt](https://github.com/Kotlin/dokka/blob/d5ad8dd4fe3bf6789c8487004b1a1d8809ac0b38/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DescriptionSections.kt#L155). E.g. see also is [implemented](https://github.com/Kotlin/dokka/blob/d5ad8dd4fe3bf6789c8487004b1a1d8809ac0b38/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DescriptionSections.kt#L155) in a way that inherits all tags from the 'parent' actual

AbdullinAM: Do you think this should not be intended behaviour?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
AbdullinAM: Confirmed with FE team that this flag will not change anything else, now or in the future. The only potential change is that may be moved from `LanguageFeature` to `AnalysisFlag`

whyoleg: Nice, thanks!

dokka-subprojects/plugin-base/src/test/kotlin/content/seealso/ContentForSeeAlsoTest.kt
whyoleg: I'm not sure, to be honest, but let's not change it.
I followed the git blame, and it was introduced a long time ago (https://github.com/Kotlin/dokka/commit/ca4622e32326f804a1dd03e03f36e874657442d4), and then moved to a separate file in https://github.com/Kotlin/dokka/pull/2749.

I would say that's one more case related to #2493. While in this PR we do provide a solution for a case, where we have an `actual` without documentation, there are still cases, where both `expect` and `actual` could have documentation (explicitly or via `typealias`), where we will need to specify some "merging strategies" for both documentation and tags.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
```diff
@@ -50,18 +51,24 @@ private fun getJdkHomeFromSystemProperty(logger: DokkaLogger): File? {
...
+        apiVersion = apiVersion,
+        analysisFlags = hashMapOf(
             AnalysisFlags.allowKotlinPackage to InternalConfiguration.allowKotlinPackage,
             // special flag for Dokka
             // force to resolve light classes (lazily by default)
-            AnalysisFlags.eagerResolveOfLightClasses to true
+            AnalysisFlags.eagerResolveOfLightClasses to true,
+        ),
+        specificFeatures = hashMapOf(
+            LanguageFeature.MultiPlatformProjects to if (isMultiplatformProject) LanguageFeature.State.ENABLED else LanguageFeature.State.DISABLED
```
whyoleg: Minor: I think it would be better not to pass the feature state at all, in case of a non-multiplatform project.
For example, just in case, the feature could be enabled by default in the compiler for some reason (or inside of AA based on configuration).
But it's probably just an additional safe-guard :)

PR #4366, opened 2025-11-24
KGP IT: Support testing AGP with Kotlin-built-in

adam-enko:
Add support for testing AGP projects that use kotlin-built-in.

I converted the existing `requiredAgpMajorVersions` check into a unit test. This will help verify that TestedVersionsSource doesn't accidentally filter out any of the tested versions. It could be a bit of a pain to maintain? But this test is quick and easy, and if it prevents a bug where a version is accidentally excluded, it'd be worth it. (We could use test coverage support on TeamCity? But I'm not sure how to add that, and it could be fragile or easy to ignore?)

Currently there are no test-projects for kotlin-built-in. These will be added in #4295.

### DIFF
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/KotlinBuiltInCompatibility.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/KotlinBuiltInCompatibility.kt
@@ -5,7 +5,8 @@
 package org.jetbrains.dokka.it.gradle.junit
 
 /**
- * Indicates whether an Android test project can run with or without Kotlin built-in.
+ * Indicates whether [DokkaGradlePluginTestExtension] will run
+ * an Android test project with or without Kotlin built-in.
  */
 enum class KotlinBuiltInCompatibility {
     /**
--- a/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/testTags.kt
+++ b/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/testTags.kt
@@ -54,6 +54,9 @@ annotation class TestsKotlinJvm
  * If a test is annotated with [TestsAndroid] then
  * [DokkaGradlePluginTestExtension] will run the test multiple times,
  * and provide a [DokkaGradleProjectRunner] using [TestedVersions.Android].
+ *
+ * [DokkaGradlePluginTestExtension] will use [kotlinBuiltIn]
+ * to filter the AGP versions used to test the project.
  */
 @Tag("Android")
 @Target(FUNCTION, CLASS)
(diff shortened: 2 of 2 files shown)


whyoleg:
TBH, for me it looks like complicating our Android-related IT infrastructure even more...
I will wait for the KBT team's review, but I feel that simply having `min/max` AGP version in test annotation, or even different version sources (and corresponding annotations) for AGP9 and AGP<9, will be sufficient.

adam-enko:
> TBH, for me it looks like complicating our Android-related IT infrastructure even more... I will wait for the KBT team's review, but I feel that simply having `min/max` AGP version in test annotation, or even different version sources (and corresponding annotations) for AGP9 and AGP<9, will be sufficient.

if it helps, we can remove the conditional logic when we stop supporting AGP <9.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersionsSource.kt
```diff
@@ -94,31 +93,50 @@ fun interface TestedVersionsSource<T : TestedVersions> {
      *
      * The test must be tagged with [TestsAndroid].
      */
-    object Android : TestedVersionsSource<TestedVersions.Android> {
+    class Android(
+        kotlinBuiltIn: KotlinBuiltInCompatibility = KotlinBuiltInCompatibility.Supported,
```
ALikhachev, line 97: Minor: I'd suggest clarifying in the KDoc that `kotlinBuiltIn` is used to filter out some versions based on the support for the built-in Kotlin.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/KotlinBuiltInCompatibility.kt
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.it.gradle.junit
+
+/**
+ * Indicates whether an Android test project can run with or without Kotlin built-in.
+ */
```
ALikhachev, line 10: Minor: The usage is more about AGP version rather than about test project setup

ALikhachev:
> TBH, for me it looks like complicating our Android-related IT infrastructure even more... I will wait for the KBT team's review, but I feel that simply having `min/max` AGP version in test annotation, or even different version sources (and corresponding annotations) for AGP9 and AGP<9, will be sufficient.

I feel it's fine and suitable to have it if it helps to describe the current situation clearly. It's about test infrastructure, we can easily simplify it later on if required, since there're no public guarantees to anyone about the test infra.

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersionsSource.kt
adam-enko: I updated the kdoc in the more visible location (`@TestsAndroid` and `@TestsAndroidCompose`)

dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/KotlinBuiltInCompatibility.kt
adam-enko: Tweaked a bit

PR #4368, opened 2025-11-24
KGP IT: test example projects can be assembled

adam-enko:
While trying to add Wasm in the example I found that DGP still produced output even when the code could not be compiled. To avoid creating examples that can't be compiled, verify the examples can be compiled.


adam-enko:
It's not possible to fix the tests without the fix from #4296, so I'll merge these changes into that PR.

PR #4370, opened 2025-11-25
Propagate KDocs of extension properties to generated getters/setters in Kotlin-as-Java

AbdullinAM:


dokka-subprojects/core-content-matcher-test-utils/src/main/kotlin/org/jetbrains/dokka/test/tools/matchers/content/ContentMatchersDsl.kt
```diff
@@ -75,6 +75,19 @@ public fun ContentMatcherBuilder<*>.group(block: ContentMatcherBuilder<ContentGr
     composite(block)
 }
 
+public fun ContentMatcherBuilder<*>.group2(block: ContentMatcherBuilder<ContentGroup>.() -> Unit) {
```
AbdullinAM, line 78: We can remove these from public APIs, but i found these util functions allow writing more simplified code in tests

dokka-subprojects/plugin-kotlin-as-java/src/main/kotlin/org/jetbrains/dokka/kotlinAsJava/converters/KotlinToJavaConverter.kt
```diff
@@ -307,7 +313,12 @@ public class KotlinToJavaConverter(
     internal fun DClass.functionsInJava(): List<DFunction> =
         properties
             .filter { !it.isJvmField && !it.hasJvmSynthetic() }
-            .flatMap { property -> listOfNotNull(property.getter, property.setter) }
+            .flatMap { property ->
```
whyoleg: The main question from my side is: why do we need to copy documentation from property to accessor ourselves (and only in KotlinAsJava), when it should probably be handled inside `findKDoc` in the Analysis API?

E.g., I would expect that when we call [getDocumentation(propertyAccessorSymbol)](https://github.com/Kotlin/dokka/blob/012ac6d10cfe21685408ea0db5a4836c3e10dafe/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt#L629) we should get the documentation from:
- if it's a Java synthetic property -> we get it from `getX` or `setX`. (and this might already work?)
- if it's a kotlin property -> we get it from the `property` itself.

Apart from the fact that, to me, it seems logical, I might imagine a case where we would allow documenting setter/getter in Kotlin separately (I don't have a reason for this thought :) ).
Additionally, the current solution will not really fix #3752, as Google Dackka works without kotlinAsJava.


Perhaps it would be best to ask the AA team about it?

AbdullinAM: > if it's a kotlin property -> we get it from the property itself.

I think in general this makes sense. I had the idea to implement this in the `findKDoc` API, my concern is to make sure that this will not break anything on the KTIJ side. I will talk to KTIJ and AA teams to get their opinion

> if it's a Java synthetic property -> we get it from getX or setX. (and this might already work?)

Documentation for `getX` and `setX` is correctly resolved in this case, but documentation for the synthetic property itself is not. Which is whole another question, that I don't want to target in this PR :)

AbdullinAM:
Related AA issue [KT-82851 Property accessors should inherit KDoc from the parent property](https://youtrack.jetbrains.com/issue/KT-82851)

Available in build [2.3.20-dev-6019](https://github.com/JetBrains/kotlin/releases/tag/build-2.3.20-dev-6019)

dokka-subprojects/plugin-kotlin-as-java/src/main/kotlin/org/jetbrains/dokka/kotlinAsJava/converters/KotlinToJavaConverter.kt
whyoleg, line 1: The changes in this file seem redundant to me: we should already have documentation on getters and setters, so there is no reason to copy it from the property itself.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -908,13 +908,30 @@ internal class DokkaSymbolVisitor(
...
+     * Otherwise, documentation will not be rendered
+     */
+    private fun DocumentationNode.removePropertyTag(): DocumentationNode =
+        DocumentationNode(children.map { it.removePropertyTag() })
+
+    private fun TagWrapper.removePropertyTag(): TagWrapper = when (this) {
+        is Property -> Description(root)
+        else -> this
+    }
+
```
whyoleg, line 934: Just an observation, unrelated to this PR: I think we could have similar problems in other places (not even related to kotlin-as-java), because we have the ability to document a property both via `@property` (translated to `Property`) and via a direct comment on it (which will be translated to `Description` AFAIK). Though, maybe those are somehow handled on the render side :)

BTW, it appears that we also have similar logic in [K1 analysis](https://github.com/Kotlin/dokka/blob/b49856b08a8e56cf7d272cab6a6b16a47b71afb6/dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/translator/DefaultDescriptorToDocumentableTranslator.kt#L826-L843).

dokka-subprojects/plugin-kotlin-as-java/src/main/kotlin/org/jetbrains/dokka/kotlinAsJava/converters/KotlinToJavaConverter.kt
AbdullinAM: Yes, this is a leftover from the previous commits

Issue #3005, opened 2023-05-21
Add support for @snippet tags in Javadoc.


romangraef:
**Is your feature request related to a problem? Please describe**
Java historically struggles with embedding code in Javadoc, and introduced the `@snippet` tag as an alternative to `@code` and `<pre>` blocks in Java 18 (and can be used in older versions of Java by using a Java 18 compiler with `targetCompatibility` set to 8). Common problems are annotations within code blocks, use of generics (as `<>` clash with html elements), and mismatched indentation and block comments (that naturally break out of the comment).

 
**Describe the solution you'd like**
The snippet tag ([JEP-413](https://openjdk.org/jeps/413)) bypasses a lot of the complexities of escaping annotations inside the code snippet block, as well as making proper indentation easier and allowing generics to be used normally (although block comments still don't work). Additionally the snippet tag allows for usage of external source files and regions within those source files, which makes verification / compilation of example snippets easier to keep.

**Describe alternatives you've considered**
The current alternative is to use `<pre>{@code }</pre>` which is a lot worse to use with normal Javadoc, and also has some parser mismatches between Dokka and the regular Javadoc parser (in particular with `@literal{@}`), as well as including direct HTML, which goes against the Dokka principle of not being bound to a specific output format.

**Additional context**

 - https://openjdk.org/jeps/413

**Are you willing to provide a PR?**
I am unsure if i would be able to implement this on my own (especially some of the complexities with `@highlight`, since that would probably end up needing a more complex CodeBlock doctag), but i would probably be willing to provide a PR for parts of the snippet functionality.

PR #4392, opened 2025-12-15
JEP 413: Add support for `@snippet` tag in Javadoc

yuriioliinyk4:
Impl for #3005

Useful links:
https://bugs.openjdk.org/browse/JDK-8201533
https://blogs.oracle.com/javamagazine/java-javadoc-snippet/
https://docs.oracle.com/en/java/javase/21/javadoc/snippets.html
https://belief-driven-design.com/better-code-snippets-in-javadoc-b27f5
https://docs.oracle.com/en/java/javase/18/docs/specs/javadoc/doc-comment-spec.html

### DIFF
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
@@ -95,6 +95,7 @@ internal class DefaultSnippetToHtmlConverter(
 
         val lang = attributeList.getAttribute(PsiSnippetAttribute.LANG_ATTRIBUTE)?.value?.value
 
+        // Note: each line from inline/external snippet includes its trailing '\n'
         val inlineSnippetLines = value.body?.lines()?.takeIf { it.isNotEmpty() }
 
         val externalSnippetLines = run {
@@ -158,14 +159,14 @@ internal class DefaultSnippetToHtmlConverter(
     /**
      * Parses markup for inline and external snippets. For external snippets, the snippet body is first extracted from the snippet file.
      *
-     * @param lines lines of the snippet body for inline snippets, or of the snippet file for external snippets
+     * @param linesWithNewlines lines (with newlines) of the snippet body for inline snippets, or of the snippet file for external snippets
      * @param context the PSI element providing the resolution context for link targets
      * @param externalSnippetRegionName name of the region to extract external snippet body from; null for inline snippets
      *
      * @return parsed snippet with applied markup
      */
     private fun parseSnippet(
-        lines: List<String>,
+        linesWithNewlines: List<String>,
         context: PsiElement,
         logger: SnippetLogger,
         externalSnippetRegionName: String? = null
@@ -176,7 +177,7 @@ internal class DefaultSnippetToHtmlConverter(
         // externalSnippetRegionName is not null in external snippets with a specified region, then we need firstly to find start of the snippet body (`@start` with the appropriate region name)
         var snippetBodyStarted = externalSnippetRegionName == null
 
-        val result = mutableListOf<String>()
+        val result = StringBuilder()
 
         data class Region(
             val regionName: String?, // can be null for anonymous regions
@@ -188,7 +189,7 @@ internal class DefaultSnippetToHtmlConverter(
 
         val nextLineMarkupTags = mutableListOf<String>()
 
-        main@ for (line in lines) {
+        main@ for (line in linesWithNewlines) {
             val currentLineOperations = regions.mapNotNull { it.operation }.toMutableList()
 
             val markupSpec = line.findMarkupSpec()
@@ -237,7 +238,7 @@ internal class DefaultSnippetToHtmlConverter(
                                 }
 
                                 if (toRemove.isSnippetBodyRegion) {
-                                    result.addIfNotBlank(line.clearMarkupSpecAndApplyMarkup(currentLineOperations))
+                                    result.appendIfNotBlank(line.clearMarkupSpecAndApplyMarkup(currentLineOperations))
                                     snippetBodyStarted = false
                                     break@main
                                 }
@@ -252,7 +253,7 @@ internal class DefaultSnippetToHtmlConverter(
                                 }
 
                                 if (lastRegion.isSnippetBodyRegion) {
-                                    result.addIfNotBlank(line.clearMarkupSpecAndApplyMarkup(currentLineOperations))
+                                    result.appendIfNotBlank(line.clearMarkupSpecAndApplyMarkup(currentLineOperations))
                                     snippetBodyStarted = false
                                     break@main
                                 } else {
@@ -285,9 +286,9 @@ internal class DefaultSnippetToHtmlConverter(
 
             if (snippetBodyStarted) {
                 if (markupSpec != null) {
-                    result.addIfNotBlank(line.clearMarkupSpecAndApplyMarkup(currentLineOperations))
+                    result.appendIfNotBlank(line.clearMarkupSpecAndApplyMarkup(currentLineOperations))
                 } else {
-                    result.add(line.applyMarkup(currentLineOperations))
+                    result.append(line.applyMarkup(currentLineOperations))
                 }
             }
         }
@@ -296,7 +297,7 @@ internal class DefaultSnippetToHtmlConverter(
             logger.warn("external snippet doesn't contain closing @end tag")
         }
 
-        return result.joinToString("").trimIndent()
+        return result.toString().trimIndent()
     }
 
     private fun createReplaceOperation(
@@ -403,8 +404,8 @@ internal class DefaultSnippetToHtmlConverter(
         }
     }
 
-    private fun MutableList<String>.addIfNotBlank(element: String) {
-        if (element.isNotBlank()) this.add(element)
+    private fun StringBuilder.appendIfNotBlank(element: String) {
+        if (element.isNotBlank()) this.append(element)
     }
 
     // Copied from https://github.com/JetBrains/intellij-community/blob/4a0ea4a70a7d2c1a14318c3d88ca632bcbe27e2f/java/java-impl/src/com/intellij/codeInsight/javadoc/SnippetMarkup.java#L402


dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
```diff
@@ -0,0 +1,1064 @@
...
+            documentablesCreationStage = { modules ->
+                val root = getFirstClassDocRoot(modules)
+
+                assertEquals(
+                    listOf(
+                        Pre(
+                            children = listOf(
+                                Text(
+                                    """
+                                    |/*
```
yuriioliinyk4, line 326: In this test and the test below, _WholeFileSnippet.java_ is used from resources, which contain copyright. The output of the test is whole file content including the copyright, which is expected. I’m concerned about whether this is okay, since the copyright header may be updated by a script in the future, which could break those tests. Should I change something here, or is it better to leave it as is?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,455 @@
...
+            "italic" -> "<i>" to "</i>"
+            "highlighted" -> "<mark>" to "</mark>"
+            else -> {
+                logger.warn("@snippet: invalid argument for `@highlight` type $type")
+                return null
+            }
+        }
+
+        val substring = attributes["substring"]?.htmlEscape()
+        val regex = attributes["regex"]?.htmlEscape()?.toRegex()
```
yuriioliinyk4: `htmlEscape` is needed to apply here to match literal symbols such as `[&, <, >, "]`, because `htmlEscape` function is applied to the raw line before applying any markup. But `<` and `>` can be used in regex constructs, for instance lookbehind. Should we adjust the logic to handle this, or leave it as is, given that such cases are quite rare and unlikely to occur?

Mentioned function:
`public fun String.htmlEscape(): String = replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;").replace("\"", "&quot;")`

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/DefaultPsiToDocumentableTranslator.kt
```diff
@@ -37,7 +37,7 @@ internal class DefaultPsiToDocumentableTranslator : AsyncSourceToDocumentableTra
             val localFileSystem = VirtualFileManager.getInstance().getFileSystem("file")
 
             val psiFiles = sourceRoots.parallelMap { sourceRoot ->
-                sourceRoot.absoluteFile.walkTopDown().mapNotNull {
+                sourceRoot.absoluteFile.walkTopDown().filter { !it.path.contains("/snippet-files") }.mapNotNull {
```
vmishenev: minor: if `sourceRoot` contains `"snippet-files/"`, the whole source root will be filter out. Can you fix it?
nitpick: `filterNot` is better for readability.

dokka-subprojects/analysis-java-psi/api/analysis-java-psi.api
```diff
@@ -143,6 +143,10 @@ public abstract interface class org/jetbrains/dokka/analysis/java/parsers/doctag
 	public abstract fun convertToHtml (Lorg/jetbrains/dokka/analysis/java/doccomment/DocumentationContent;Lorg/jetbrains/dokka/analysis/java/parsers/doctag/DocTagParserContext;)Ljava/lang/String;
 }
 
+public abstract interface class org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter {
```
vmishenev: Would you mind making this private, since users do not have the ability to use it and it exposes internal PSI?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,455 @@
...
+        private val ATTRIBUTE = Regex("(\\w+)\\s*(=\\s*('([^']*)'|\"([^\"]*)\"|(\\S*)))?\\s*")
+
+        private const val SNIPPET_NOT_RESOLVED = "// snippet not resolved"
+    }
+
+    override fun convertSnippet(
+        snippet: PsiSnippetDocTag
+    ): String {
+        val value = snippet.valueElement ?: run {
+            logger.error("@snippet: unable to resolve snippet")
```
vmishenev: `logger.warn` is more appropriate here. `logger.error` is intended for more fatal conditions. See the KDoc for `logger.error`.

Also, could you add some information about the snippet, for example, its location? (and in the messages below)

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,455 @@
...
+import org.jetbrains.dokka.analysis.java.util.lowercase
+import org.jetbrains.dokka.links.DRI
+import org.jetbrains.dokka.utilities.DokkaLogger
+import org.jetbrains.dokka.utilities.htmlEscape
+import java.io.File
+import kotlin.collections.fold
+
+private typealias MarkupOperation = (String) -> String
+
+public interface SnippetToHtmlConverter {
```
vmishenev: It would be nice to have some KDoc

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,455 @@
...
+            }
+
+            else -> {
+                logger.error("@snippet: unable to resolve snippet")
+                SNIPPET_NOT_RESOLVED
+            }
+        }
+
+        if (parsedSnippet.isBlank()) {
+            logger.error("@snippet: unable to resolve snippet")
```
vmishenev: nitpick:
```kt          
logger.error("@snippet: unable to resolve snippet")
SNIPPET_NOT_RESOLVED
```

looks like a code duplication. Maybe it’s better to extract it into a function.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,455 @@
...
+    private val logger: DokkaLogger
+) : SnippetToHtmlConverter {
+
+    private val sampleFiles by lazy {
+        sourceSet.samples.flatMap {
+            it.walkTopDown().filter { file -> file.isFile }
+        }
+    }
+
+    private companion object {
```
vmishenev: _style:_ In Kotlin, the recommended convention is to place a companion object at the end of its class declaration.   https://kotlinlang.org/docs/coding-conventions.html#class-layout

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
vmishenev: Do you really need `htmlEscape` here? It seems unnecessary, since the result is `result.replace(regex) { match -> match.value.wrapInTag() }` that is, the escaped value is not used in the returned value.

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
```diff
@@ -0,0 +1,1064 @@
...
+    fun `replace markup tag`() {
+        testInline(
+            """
+            |/src/main/java/example/Test.java
+            |package example;
+            |
+            | /**
+            | * {@snippet :
+            | * public class Example {
+            | *     String config = "oldValue"; // @replace substring="oldValue" replacement="newValue"
```
vmishenev, line 710: Could you add the following example to the test?
```kt
        System.out.println("Hello World!");  // @replace regex='"(.*)"' replacement='"I said, $1"'
```
(from https://belief-driven-design.com/better-code-snippets-in-javadoc-b27f5/)

vmishenev (commented):
:+1:

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,455 @@
...
+
+        val inlineSnippetLines = value.body?.lines()?.takeIf { it.isNotEmpty() }
+
+        val externalSnippetLines = run {
+            val fileAttr = attributeList.getAttribute(PsiSnippetAttribute.FILE_ATTRIBUTE)?.value
+            val classAttr = attributeList.getAttribute(PsiSnippetAttribute.CLASS_ATTRIBUTE)?.value
+
+            when {
+                fileAttr != null -> readExternalSnippetLines(fileAttr, fileAttr.value)
+                classAttr != null -> readExternalSnippetLines(classAttr, "${classAttr.value}.java")
```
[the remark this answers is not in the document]
yuriioliinyk4: JavaDoc has the [same logic](https://github.com/openjdk/jdk/blob/f786e2a22fca7c9785b2daa5667271f418c414f4/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/SnippetTaglet.java#L207C22-L207C94).

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
yuriioliinyk4: added

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: added

dokka-subprojects/analysis-java-psi/api/analysis-java-psi.api
yuriioliinyk4: I think you meant internal, done

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/DefaultPsiToDocumentableTranslator.kt
yuriioliinyk4: see https://github.com/Kotlin/dokka/pull/4392/commits/61f298c864024899c9a0cb7367f96b4c064d1fad

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: see https://github.com/Kotlin/dokka/pull/4392/commits/85d2ecc3912ae6a05641a73dc6b3673eee65ad84

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: htmlEscape is needed to match literal symbol such as `[&, <, >, "]`, because htmlEscape function is applied to the raw line before applying any markup.

Without htmlEscape on regex following examples wont work:
`// @replace regex='".*"' replacement="..."`
or 
`// @replace regex='"(.*)"' replacement='"I said, $1"'`

because `"` in regex pattern wont be matched, since `"` in the line were changed to the `&quot;`

vmishenev: I have changed the code (remove  `htmlEscape()`):
```kt
    private fun String.applyMarkup(markupOperations: List<MarkupOperation>): String =
        markupOperations.fold(this) { acc, op -> op(acc) }
        
        ///......
        val substring = attributes["substring"]
        val regex = attributes["regex"]?.toRegex()
        val replacement = attributes["replacement"] ?: run {
            logger.warn("specify 'replacement' attribute in @replace markup tag")
            return null
        }
```
and your test `SnippetTest.replace markup tag` has been passed. 

Anyway, `htmlEscape()` should be applied to the final result after all markup operations are done.

yuriioliinyk4: > Anyway, htmlEscape() should be applied to the final result after all markup operations are done.

We can’t do this because HTML tags injected during the processing of `@highlight` or `@link` will not be rendered.

For example:
`String example = "something" // @hightlight substring="example"`

We wrap `example` in `<b>` during processing, but if we apply `htmlEscape()` at the end, `<b>example</b>` will be shown as plain text instead of being rendered.

> and your test SnippetTest.replace markup tag has been passed.

but others fail, due to the problem I described above.

vmishenev: In this case, `@replace` can  unintentionally replace some html code for `@hightlight` and `@link`.
Is that possible?

yuriioliinyk4: Yes, that can happen. One possible solution can be to run all the replace operations first and then apply the highlighting and linking afterward.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/HtmlToDocTagConverter.kt
```diff
@@ -56,14 +56,17 @@ internal class HtmlToDocTagConverter(
...
             "mark" -> ifChildrenPresent { Mark(children) }
             "code" -> ifChildrenPresent { if(keepFormatting) CodeBlock(children, mapOf("lang" to "java")) else CodeInline(children, mapOf("lang" to "java")) }
-            "pre" -> if(children.size == 1) {
-                when(children.first()) {
-                    is CodeInline -> listOf(CodeBlock(children.first().children, mapOf("lang" to "java")))
-                    is CodeBlock -> listOf(children.first())
-                    else -> listOf(Pre(children))
+            "pre" -> {
+                val lang = element.attr("lang").trim().takeIf { it.isNotBlank() } ?: "java"
+                if (children.size == 1) {
```
whyoleg: minor: it might make sense to refactor `if` + `when` to use single `when` with `children.singleOrNull()`:
```kotlin
when(val child = children.singleOrNull()) {
    is CodeInline -> ...
    is CodeBlock -> ...
    else -> ...
}
```
(this will de-duplicate both `else` branches in current impl)

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/java/DescriptorKotlinDocCommentParser.kt
```diff
@@ -27,7 +28,7 @@ internal class DescriptorKotlinDocCommentParser(
         return docComment is DescriptorKotlinDocComment
     }
 
-    override fun parse(docComment: DocComment, context: PsiNamedElement): DocumentationNode {
+    override fun parse(docComment: DocComment, context: PsiNamedElement, sourceSet: DokkaSourceSet): DocumentationNode {
```
whyoleg, line 31: As we now provide `sourceSet` here, we can now path it into `parseDocumentation` call and remove the logic to determine the `sourceSet` from it.
Might also make sense to check other places

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/java/KotlinDocCommentParser.kt
```diff
@@ -27,7 +28,7 @@ internal class KotlinDocCommentParser(
         return docComment is KotlinDocComment
     }
 
-    override fun parse(docComment: DocComment, context: PsiNamedElement): DocumentationNode {
+    override fun parse(docComment: DocComment, context: PsiNamedElement, sourceSet: DokkaSourceSet): DocumentationNode {
```
whyoleg, line 29: same here (as in the comment above in `descriptors`)

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
```diff
@@ -0,0 +1,1097 @@
...
+                        )
+                    ),
+                    root.children
+                )
+            }
+        }
+    }
+
+    @Test
+    fun `highlight markup tag`() {
```
whyoleg, line 644: Curious: how this Javadoc comment will be rendered by our HTML renderer?

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
```diff
@@ -0,0 +1,1097 @@
...
+                        )
+                    ),
+                    root.children
+                )
+            }
+        }
+    }
+
+    @Test
+    fun `link markup tag`() {
```
whyoleg, line 756: Same question here: how this will look with our HTML renderer? Does it correctly supports links in code blocks?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+        }
+    }
+
+    override fun convertSnippet(
+        snippet: PsiSnippetDocTag
+    ): String {
+        val logger = SnippetLogger(dokkaLogger, snippet.containingFile?.name)
+
+        val value = snippet.valueElement ?: run {
+            return "<pre>${logAndReturnUnresolvedSnippet(logger)}</pre>"
```
whyoleg: I think it makes sense to add at least one test for the unresolved snippets

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+                }
+            }
+        }
+
+        return output
+    }
+
+    private fun readExternalSnippetLines(attribute: PsiElement, fileName: String): List<String>? {
+        /**
+         * Try to resolve the snippet file through PSI reference resolution (for files in `snippet-files`)
```
whyoleg, line 447: Just to understand this better: `snippet-files` directory is handled automatically by intellij Java analysis, right?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+
+internal class DefaultSnippetToHtmlConverter(
+    private val sourceSet: DokkaConfiguration.DokkaSourceSet,
+    private val docTagParserContext: DocTagParserContext,
+    private val dokkaLogger: DokkaLogger
+) : SnippetToHtmlConverter {
+
+    private val sampleFiles by lazy {
+        sourceSet.samples.flatMap {
+            it.walkTopDown().filter { file -> file.isFile }
```
whyoleg, line 53: Does `walkTopDown` follows symbolic links? (e.g. reminded me about https://github.com/Kotlin/dokka/pull/4399)

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+        val psiFile = attribute.reference?.resolve() as? PsiFile
+
+        return if (psiFile != null) {
+            File(psiFile.virtualFile.path).readLinesWithNewlines()
+        } else {
+            sampleFiles.singleOrNull { it.name == fileName }?.readLinesWithNewlines()
+        }
+    }
+
+    private fun File.readLinesWithNewlines() = this.readLines().map { it + "\n" }
```
whyoleg, line 459: why do we need to add a `\n` at the end of each line here? If it's really required, please add a comment explaining why, as it's not clear.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        val region = attributeList.getAttribute(PsiSnippetAttribute.REGION_ATTRIBUTE)?.value?.value
+
+        var parsedSnippet = when {
+            inlineSnippetLines != null && externalSnippetLines != null -> {
+                val parsedInlineSnippet = parseSnippet(inlineSnippetLines, snippet, logger)
+                val parsedExternalSnippet = parseSnippet(externalSnippetLines, snippet, logger, region)
+
+                if (parsedInlineSnippet != parsedExternalSnippet) {
```
whyoleg, line 119: nice to have a test for this

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+                classAttr != null -> readExternalSnippetLines(classAttr, "${classAttr.value}.java")
+                else -> null
+            }
+        }
+
+        val region = attributeList.getAttribute(PsiSnippetAttribute.REGION_ATTRIBUTE)?.value?.value
+
+        var parsedSnippet = when {
+            inlineSnippetLines != null && externalSnippetLines != null -> {
+                val parsedInlineSnippet = parseSnippet(inlineSnippetLines, snippet, logger)
```
whyoleg: minor: it feels like `parseSnippet` calls could be extracted from `when`, as they are used in all branches.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+            val operation: MarkupOperation?, // can be null for non-markup tags - `@start` (regionName in this case is specified)
+            val isSnippetBodyRegion: Boolean = false // indicates whether this region represents the actual snippet body that is processed
+        )
+        // Ordered list of snippet regions
+        val regions = mutableListOf<Region>()
+
+        var nextLineMarkupTags = mutableListOf<String>()
+
+        main@ for (line in lines) {
+            val currentLineOperations = regions.mapNotNull { it.operation }.toMutableList()
```
whyoleg, line 193: minor: `mapNotNullTo(mutableListOf()){...}` can be used instead of  `mapNotNull{...}.toMutableList()`

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+        logger: SnippetLogger,
+        externalSnippetRegionName: String? = null
+    ): String {
+        // externalSnippetRegionName is null in 2 cases:
+        // case 1: inline snippet, then we already snippetBody
+        // case 2: external snippet where the region is not specified, then we take a whole file
+        // externalSnippetRegionName is not null in external snippets with a specified region, then we need firstly to find start of the snippet body (`@start` with the appropriate region name)
+        var snippetBodyStarted = externalSnippetRegionName == null
+
+        val result = mutableListOf<String>()
```
whyoleg: looks like this could be replaced with `StringBuilder`

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+     * @param context the PSI element providing the resolution context for link targets
+     * @param externalSnippetRegionName name of the region to extract external snippet body from; null for inline snippets
+     *
+     * @return parsed snippet with applied markup
+     */
+    private fun parseSnippet(
+        lines: List<String>,
+        context: PsiElement,
+        logger: SnippetLogger,
+        externalSnippetRegionName: String? = null
```
whyoleg, line 172: Can we handle external snippet region separately from everything else before calling this function? E.g., in case of external snippets, first extract snippet in the region, and process what's inside?

I could be wrong, but it feels like because we handle both internal snippet regions (highlight/replace/link/etc) and external snippet region (start/end) in the same code flow it's harder to understand the whole pipeline here.
I would like to hear your opinions here.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+
+        data class Region(
+            val regionName: String?, // can be null for anonymous regions
+            val operation: MarkupOperation?, // can be null for non-markup tags - `@start` (regionName in this case is specified)
+            val isSnippetBodyRegion: Boolean = false // indicates whether this region represents the actual snippet body that is processed
+        )
+        // Ordered list of snippet regions
+        val regions = mutableListOf<Region>()
+
+        var nextLineMarkupTags = mutableListOf<String>()
```
whyoleg: let's stick with one:
- it's a `var`
- it's a `MutableList`

Otherwise, it's too hard to follow what is happening to it.
In this case, it feels like it could be just `var nextLineMarkupTags: List<String>`

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+
+        main@ for (line in lines) {
+            val currentLineOperations = regions.mapNotNull { it.operation }.toMutableList()
+
+            val markupSpecMatch = MARKUP_SPEC.find(line)
+            val markupSpec = markupSpecMatch?.groupValues?.get(2)
```
whyoleg: suggestion: let's extract the logic of getting `markupSpec` from line into a function, similar to `clearMarkupSpec`, and put them both near `MARKUP_SPEC` regex, so that it will be both easier to understand in what cases this regex used, and make `parseSnippet` easier to read.

It probably makes sense to do the same with other regex usages

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+            if (substring == null && regex == null) result = result.wrapInTag()
+
+            result
+        }
+    }
+
+    private fun createReplaceOperation(attributes: Map<String, String?>, logger: SnippetLogger): MarkupOperation? {
+        val substring = attributes["substring"]?.htmlEscape()
+        val regex = attributes["regex"]?.htmlEscape()?.toRegex()
+        if (!attributes.containsKey("replacement")) {
```
whyoleg: minor: I think it makes sense to handle it in the same way as `attributes["target"]` is handled in `createLinkOperation` - on `replacement` creation via `?: run {}` - just to use the same pattern

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
```diff
@@ -0,0 +1,505 @@
...
+
+    private fun String.clearMarkupSpec() = this.replace(MARKUP_SPEC, "").trimEnd() + "\n"
+
+    private fun String.applyMarkup(markupOperations: List<MarkupOperation>): String =
+        markupOperations.fold(this.htmlEscape()) { acc, op -> op(acc) }
+
+    private fun String.clearMarkupSpecAndApplyMarkup(markupOperations: List<MarkupOperation>) =
+        this.clearMarkupSpec().applyMarkup(markupOperations)
+
+    private fun String.parseMarkupTag(logger: SnippetLogger): Pair<String, Map<String, String?>>? {
```
whyoleg: suggestion: it might make sense to move the function to be near `ATTRIBUTE` regex, so that the group indices are visible - without need to go back and forth :)

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
yuriioliinyk4: image

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
yuriioliinyk4: Yes, it works correctly.
image

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: Yes

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/HtmlToDocTagConverter.kt
yuriioliinyk4: done

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/java/DescriptorKotlinDocCommentParser.kt
yuriioliinyk4: `parseDocumentation` is also used [here](https://github.com/Kotlin/dokka/blob/bf3b5eea8fb6ec50d43420c13e44f5a83fdc4a77/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/java/KotlinInheritDocTagContentProvider.kt#L28), where we dont have `sourceSet`

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: I am not sure. Anyway, it is also used [here](https://github.com/Kotlin/dokka/blob/bf3b5eea8fb6ec50d43420c13e44f5a83fdc4a77/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/DefaultPsiToDocumentableTranslator.kt#L40) to get `psiFiles` from `sourceRoots`.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: 27af89d677f6badca9c0339b0a32411ad030c9d5

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: `readLines()` returns a list of lines without `\n` at the end. This is needed only to preserve the structure of the snippet, the lines are then joined using `joinToString("").trimIndent()`.

**Why not simply use joinToString("\n")?** 
The reason for this is that whitespace is ignored between HTML tags, so when the end of the previous line and the beginning of the current line both contain HTML tags, the newline will not be rendered and the lines will collapse into a single line.

Is it clear now? Should I add comment?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: What should test assert? Is it good practice to assert logger output?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: > first extract snippet in the region, and process what's inside

Firstly I have written the same logic, but we need to keep track of markup tags above the external snippet region (before needed `@start` tag), because they can create region with markup operations and that operations must be applied to the external snippet. 

See `snippet with overlapping regions` test.

```
public class OverlappingRegions {
    public void example() {
        // @replace region="replacePassword" regex="secret\d+" replacement="***"
        String password1 = "secret123";
        // @start region="snippetBody"
        System.out.println("Processing:");
        // @highlight region="highlightData" substring="data"
        String password2 = "secret123";
        String data = "important data";
        // @end region="replacePassword"
        String password3 = "secret123"; // @end region="highlightData"
        System.out.println(data);
        // @end region="snippetBody"
    }
}
```

Replace region is created here before `@start region="snippetBody"`, but this replace operation also must be applied inside the external snippet.

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
whyoleg: For me, both having and removing the copyright header is fine. So up to you.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: I do agree with what @vmishenev says - we should not use `htmlEscape` for regex.

> One possible solution can be to run all the replace operations first and then apply the highlighting and linking afterward.

For me, this feels like the correct way: we should probably control the order of operations. Note: In the case of multiple replace regions, we should execute them in the order they appear.
It feels like we should execute those operations in a specific order: replace, link, highlight.

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/java/DescriptorKotlinDocCommentParser.kt
whyoleg: We can have `parseDocumentation` without an argument there (as it's an `internal class` anyway).
Let's create an issue for this and tackle it separately (PR is already big)

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
whyoleg: Nice, thanks!

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
whyoleg: Oh my, that's cool!

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: Is there a way to also handle files from `sourceSet.samples` this way?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: Let's create an issue at least to test that we work fine with symbolic links for Java sources.
Test for Kotlin sources is in https://github.com/Kotlin/dokka/blob/894899b251c296ab6b6605840879ebf043bf4753/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KotlinAnalysisSymlinksTest.kt

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: I get that we concatenate the string afterwards. That's why I'm curious.
For me, when I see `lines: List<String>` or `line: String`, I do not expect them to have a line ending.

And I don't really understand the reasoning here:
> The reason for this is that whitespace is ignored between HTML tags, so when the end of the previous line and the beginning of the current line both contain HTML tags, the newline will not be rendered and the lines will collapse into a single line.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: > What should test assert?

1. That warning is emitted
2. That the warning message is reasonable - in this case, we check that `diff` is correct.

Otherwise, this code flow is not thoroughly covered by tests and can do anything, including throwing an error, which may go uncaught.

> Is it good practice to assert logger output? 

In this case, yes, it is fine, as we do want to check the reported warning. In some way, it's similar to asserting a thrown exception.
Of course, there is no need to assert every log info.

Ideally, we can extract this functionality into a small class and write a unit test for it - this is an alternative.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: Oh, okay, now I see the problem. Thanks!
Then let's leave it like it is now.

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/java/DescriptorKotlinDocCommentParser.kt
yuriioliinyk4: > We can have parseDocumentation without an argument there

I didn't get it.

> Let's create an issue for this and tackle it separately (PR is already big)

So leave as it is and create issue?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: As far as i know - no. Intellij quick doc does not support specifying `snippet-path` and it is not supported on PSI level.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: #4419

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: > And I don't really understand the reasoning

Suppose,
current line ends with `</b>` because of `@highlight` applied.
next line start with `<b>` because of `@highlight`.

Then after joining those 2 lines by `joinToString("\n")`, the `\n` is between html tags. Afterwards, the '\n' is ignored, and in the output 2 lines are collapsed into single line. 

image

Output:
image

Thats why line include `\n` and then they are just joined by `""`. In this case, when applying `@highlight` to the whole line, `\n` is inside the html tag and is rendered correctly. 

Actually, it is a problem how whitespaces are processed between html tags, I got a lot of bugs due to this issue and probably there are edge cases when snippet structure is not correctly preserved.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: [9d24936](https://github.com/Kotlin/dokka/pull/4392/commits/9d24936f49ba0afab07af3c5da67570e47c221a2)

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: You are right, decided to stick with `val` and `mutableList`.
[b8acf5c](https://github.com/Kotlin/dokka/pull/4392/commits/b8acf5c2b49a7fcb3c5110daf3916fee7ed0a8d3)

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: see [f4c4d67](https://github.com/Kotlin/dokka/pull/4392/commits/f4c4d674f6c7ca702d0ac4f833f5d0d5a730893f),
I found out that html escaped snippet will be logged, e.g. `&quot;` instead of `"` will be used and ...

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: just alternative, I will leave as it is.

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/java/DescriptorKotlinDocCommentParser.kt
whyoleg: > So leave as it is and create issue?

Yes, please.

> I didn't get it.

I mean, we can have `parseDocumentation` overloaded with and without a source set. Or we can move the logic to find a source set in [KotlinInheritDocTagContentProvider](https://github.com/Kotlin/dokka/blob/bf3b5eea8fb6ec50d43420c13e44f5a83fdc4a77/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/java/KotlinInheritDocTagContentProvider.kt#L28), where no source set is provided, while using the source set provided from `parse` in `KotlinDocCommentParser`.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: Got it, thanks!

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: I have extracted usages of regex into separate functions and placed them near regex declarations.
See [3b916dd](https://github.com/Kotlin/dokka/pull/4392/commits/3b916dd09314156740f630d3534f71065652aebf)

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: 3b916dd

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: When parameter is provided as `replacement=""`, the pair `"replacement" to null` is returned, this means the deletion of specified part and it is valid.

While in link this is invalid, since target should be always specified, this is why logic is a bit different.

whyoleg: Ah! Okay, that's a bit unexpected (`""` -> `null` conversion) :)

Is this the only case for attributes having a `null` value? Overall, I think that it should not be `null` there, but an empty string. 
I think that we should make `attributes` be `Map<String, String>`, e.g., containing only non-nullable values. This will make the logic more consistent.

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
```diff
@@ -1092,6 +1092,78 @@ class SnippetTest : BaseAbstractTest() {
...
+
+                assertFalse("first line is the same") { warnMessage.contains("String first = &quot;first line is the same&quot;") }
+
+                assertTrue("second line is distinct") {
+                    warnMessage.contains("System.out.println(&quot;second line is different&quot;)")
+                            && warnMessage.contains("String second = &quot;second line is different&quot;")
+                            && warnMessage.contains("line 2")
+                }
+
+                assertFalse("third line is the same") { warnMessage.contains("System.out.println(&quot;third line is the same&quot;)") }
```
whyoleg: I think that it makes more sense to just assert the whole message:
```
assertEquals(
  warnMessage,
  """
  full warning message here
  """.trimIndent()
)
```
Otherwise, it's really hard to undestand what is checked here

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: I believe it depends on the resolution of https://github.com/Kotlin/dokka/pull/4392#discussion_r2638344891, but overall, it might be fine to have it like this (if it's hard to do otherwise), but mention in the error message that the output is after applying formatting:
```
Inline and external snippets are not the same in the hybrid snippet (after formatting and escaping are applied). diff:
// diff goes here
```

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
whyoleg: Thanks, now I get it!
Still, I think we shouldn't append a new line character here, as it's more of a hack.
As we output HTML, we can use `<br>` instead of `\n` in `joinToString`. Will this be enough?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: The issue with regex and htmlEscape is fixed in add2a2f.
Added order control of markup operations in 85cfd4e.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: > Is this the only case for attributes having a null value? Overall, I think that it should not be null there, but an empty string.
I think that we should make attributes be Map<String, String>, e.g., containing only non-nullable values. This will make the logic more consistent.

The `region` attribute can be without value (anonymous region), thats the reason for `String?` as value.

> Okay, that's a bit unexpected ("" -> null conversion)

This happens in `parseMarkupTag` function because I am using check `isNotBlank()`. We can remove those checks in `parseMarkupTag`, but then we will need to add a lot of checks for each attribute separately. So thats why I left it in one place - `parseMarkupTag.`

whyoleg: > The region attribute can be without value (anonymous region), thats the reason for String? as value.

Thanks!

> then we will need to add a lot of checks for each attribute separately.

Why? I don't think that we should do this conversion. User passes an empty string, this is a valid string value, why do we replace it with `null`? Or what is the reason?

dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/java/DescriptorKotlinDocCommentParser.kt
yuriioliinyk4: #4433

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/java/KotlinDocCommentParser.kt
yuriioliinyk4: #4433

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: Oh, I found out that we are using `takeIf { it.isNotBlank() }` , because for groups that are not captured empty string is retuned instead of null. So when empty string is returned we need to check other groups.
image

whyoleg: oh.... Fine, let's then at least document these nuances when we create the attributes map.

Also, in `createReplaceOperation`, leave a small comment on why we need to use `containsKey` and not just `get`.

yuriioliinyk4: 3843728

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
yuriioliinyk4: 2687a46

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: 2687a46

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
vmishenev: >  As we output HTML, we can use <br> instead of \n in joinToString. Will this be enough?

 I think `<br>` can just complicate the whole logic.
 
 `@yuriioliinyk4`, is there a way to store lines without breaklines?

yuriioliinyk4: >As we output HTML, we can use `<br>` instead of \n in joinToString. Will this be enough?

If we use `<br>` as line separator, do you know how we could apply `trimIndent()`? 
`trimIndent()` is must-have function here.

>is there a way to store lines without breaklines?

We can add `\n` in the `applyMarkup` before applying markup operations, but I don's see the point in this (useless refactoring), anyway in this case `result`  list will contain lines with `\n`.

FYI: the problem with whitespaces not rendered between html tags is due to [this](https://github.com/Kotlin/dokka/blob/3e2a22b16db3ac74843d96d1326dd1c50156e5a3/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/HtmlToDocTagConverter.kt#L27) line (`takeIf { it.isNotBlank() }`).

vmishenev: > but I don's see the point in this (useless refactoring)

This is more about the readability/maintainability to prevent situations like:

> when I see lines: List<String> or line: String, I do not expect them to have a line ending.

For example, if someone wants to extend your code by simply adding a new markup operation, they may not know that `line` actually contains a line break. To find this out, they would need to understand the whole logic.

I can see 2 ways to fix it:
- rename `line`/ `lines` 
- keep `line` without `\n`

> result list will contain lines with \n

As @whyoleg pointed out, `StringBuilder` is more preferable and straightforward.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: 7dfe092

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/SnippetToHtmlConverter.kt
yuriioliinyk4: 7dfe092

PR #4393, opened 2025-12-16
Prototype json to html generator

berezinant:


dokka-ir-to-html-generator/README.md
```diff
@@ -0,0 +1,51 @@
...
+    ```bash
+    ./gradlew :dokka-ir-to-html-generator:shadowJar
+    ```
+- The task bundles all dependencies and also builds the frontend bundle. The resulting JAR will be at:
+  - `dokka-ir-to-html-generator/build/libs/docgen.jar`
+
+2) Run it against any input.json
+- Provide a path to a JSON file that contains the page data (an example `input.json` is in this module’s directory):
+  ```bash
+  java -jar build/libs/docgen.jar input.json
```
berezinant, line 22: this is how to run it

PR #4396, opened 2025-12-17
Fix source links to accessors in the KotlinAsJava plugin

### LINKED ISSUE #3367: [K2] Missed source link for generated getter/setter in kotlin-as-java output
```Kotlin
class DefValue(val str: String = "defString")
```

Setup source linking and apply `kotlin-as-java` Dokka plugin.

Output in K1 for getter method will have source link
image

and for K2 it missed:

image

**Installation**
- Dokka version: 1.9.20

Parent: #3328

vmishenev:
fixes #3367

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
@@ -169,7 +169,7 @@ internal class DokkaSymbolVisitor(
             documentation = getDocumentation(typeAliasSymbol)?.toSourceSetDependent() ?: emptyMap(),
             sourceSets = setOf(sourceSet),
             generics = generics,
-            sources = typeAliasSymbol.getSource(),
+            sources = getSource(typeAliasSymbol),
             extra = PropertyContainer.withAll(
                 getDokkaAnnotationsFrom(typeAliasSymbol)?.toSourceSetDependent()?.toAnnotations(),
                 ancestryInfo.exceptionInSupertypesOrNull(),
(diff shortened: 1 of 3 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -926,7 +926,14 @@ internal class DokkaSymbolVisitor(
         else -> false
     }
 
-    private fun KaSymbol.getSource() = KtPsiDocumentableSource(psi).toSourceSetDependent()
+    private fun KaSession.getSource(symbol: KaSymbol): SourceSetDependent<DocumentableSource> {
+        val psi = when(symbol) {
+            // implicit/default accessors have no psi
+            is KaPropertyAccessorSymbol -> symbol.containingSymbol?.psi
```
vmishenev, line 949: Also, see an internal discussion in AA.

In the case of generated declarations, it returns PSI of the declaration from which this (synthetic) declaration is derived.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/SourceLinkTransformerTest.kt
```diff
@@ -128,4 +128,48 @@ class SourceLinkTransformerTest : BaseAbstractTest() {
             }
         }
     }
+
+    @Test
+    fun `synthetic enum values should have source link`() {
```
vmishenev, line 223: This test is not affected by this change. 
It has been added JIC

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
whyoleg: [(internal discussion)](https://jetbrains.slack.com/archives/C061DS4G41J/p1765816953256829)

Issue #3604, opened 2024-05-13
[K2] Inconsistent reference from constructor kdoc to class page fro K1/K2


atyrin:
```Kotlin
class A{
    /**
     * Reference to class: [A]
     */
    constructor(s: String)
}
```

In K1 the link `[A]` from kdoc leads to the class A page
In K2 the link leads to the constructor page

**Installation**
- Dokka version: 2.0.0 (2.0.0-test-49477c44dfc58e2702f4c145ff41190b39d117fb)
- KGP: 2.0.0-RC2

Previous step: #3352
Parent: #3550

vmishenev:
Presumably it a kind of https://github.com/Kotlin/dokka/issues/3451#issuecomment-1899119751

AbdullinAM:
Blocked by [KT-83152](https://youtrack.jetbrains.com/issue/KT-83152)

AbdullinAM:
Tests for the issue added in #4397, waiting for the [KT-83152](https://youtrack.jetbrains.com/issue/KT-83152)

AbdullinAM:
Fixed in #4424

PR #4397, opened 2025-12-18
New KDoc resolution

AbdullinAM:
Enable new KDoc resolution algorithm and add tests

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/content/kdoc/KDocAmbiguityResolutionTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/content/kdoc/KDocAmbiguityResolutionTest.kt
@@ -536,7 +536,7 @@ class KDocAmbiguityResolutionTest : BaseAbstractTest() {
 
 
     @Test
-    @Ignore("KT-83152 [Analysis API, KDoc] Make class name links on constructors point to the class")
+    @Ignore("#3604: blocked by KT-83152")
     fun `#3604 KDoc reference to class from constructor`() = withExperimentalKDocResolution {
         testInline(
             """
--- a/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
@@ -613,7 +613,7 @@ class LinkTest : BaseAbstractTest() {
 
     @Test
     @OnlySymbols("KEEP #389: New KDoc resolution")
-    fun `fully qualified link should lead to package K2`() = withExperimentalKDocResolution {
+    fun `fully qualified link should lead to function`() = withExperimentalKDocResolution {
         // for the test case, there is the only one link candidate in K1 and K2
         testInline(
             """
@@ -796,7 +796,7 @@ class LinkTest : BaseAbstractTest() {
 
     @Test
     @OnlySymbols("KEEP #389: New KDoc resolution")
-    fun `short link should lead to package rather than function K2`() = withExperimentalKDocResolution {
+    fun `short link should lead to function`() = withExperimentalKDocResolution {
         testInline(
             """
             |/src/main/kotlin/Testing.kt


dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/configuration/AnalysisEnvironment.kt
```diff
@@ -304,10 +305,8 @@ public class AnalysisEnvironment(
                 .filter { it.isDirectory || it.extension == KLIB_FILE_EXTENSION }
                 .forEach { libraryFile ->
                     try {
-                        val kotlinLibrary = resolveSingleFileKlib(
-                            libraryFile = KFile(libraryFile.absolutePath),
-                            strategy = ToolingSingleFileKlibResolveStrategy
-                        )
+                        val kotlinLibrary = ToolingSingleFileKlibResolveStrategy
```
AbdullinAM, line 308: Api was deprecated (hidden) in kotlin-2.3.30-dev-7064, replaced it with the code that does the same thing

dokka-subprojects/analysis-kotlin-symbols/build.gradle.kts
```diff
@@ -61,6 +60,9 @@ dependencies {
 
     // TODO [beresnev] get rid of it
     compileOnly(libs.kotlinx.coroutines.core)
+
+    // to gain access to com.intellij.diagnostic.LoadingState in the compile time
+    compileOnly(libs.intellij.util.base)
```
AbdullinAM, line 65: For some reason `LoadingState` bundled with the analysis api/compiler does not contains necessary APIs. We have the correct version of the `LoadingState` in the runtime

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt
```diff
@@ -79,6 +80,8 @@ internal fun createAnalysisSession(
     projectDisposable: Disposable = Disposer.newDisposable("StandaloneAnalysisAPISession.project"),
     isSampleProject: Boolean = false
 ): KotlinAnalysis {
+    enableExperimentalKDocResolution()
```
vmishenev: Can we make this under a flag in Dokka untill `experimentalKDocResolution` is enabled by default?

AbdullinAM: Makes sense, yes

dokka-subprojects/plugin-base/src/test/kotlin/content/typealiases/NestedTypealiasTest.kt
```diff
@@ -315,7 +315,7 @@ class NestedTypealiasTest : BaseAbstractTest() {
         |     * some text
         |     *
         |     * @see String
-        |     * @throws Unit
```
vmishenev, line 318: This case is quite unexpected without printing the reason. The KEEP only says `@exception / @throws must only be resolvable to Throwables.` Anyway, Dokka and the IDE should warn or print an error.
Maybe this should be resolved to better understand the misuse. A user should understand where the link refers, especially when it is ambiguous.
General idea:
```kt
class MyException : Exception
class Usage { 
        fun MyException()...
        /**
        @throws MyException
        */
        fun usage() {
        }
}

```

AbdullinAM: IDE currently highlights this as a warning, makes sense to add a warning to Dokka too 👍

AbdullinAM: It is already logged as a warning:
```
 /**
  * @throws String
  */
funfoo(arg: Int) {}
```
```
w: [:composeApp:dokkaGeneratePublicationHtml] Couldn't resolve link for String in main.kt/org.example.project.foo
```

AbdullinAM:
The failing tests on CI seem to be caused by external issues, master fails with same errors

whyoleg:
Note: TC IT are red because of https://github.com/Kotlin/dokka/pull/4402#discussion_r2635711945

dokka-subprojects/plugin-base/src/test/kotlin/content/typealiases/NestedTypealiasTest.kt
whyoleg: IMO, this warning should be enough.
Might make sense to add a test for unresolved links in a separate PR?

dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
```diff
@@ -707,6 +794,60 @@ class LinkTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    @OnlySymbols("KEEP #389: New KDoc resolution")
+    fun `short link should lead to package rather than function K2`() = withExperimentalKDocResolution {
```
whyoleg: test should be renamed, as with new resolution, it should lead to the function rather than package (and the test shows it)

dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
```diff
@@ -609,6 +611,90 @@ class LinkTest : BaseAbstractTest() {
         }
     }
 
+    @Test
+    @OnlySymbols("KEEP #389: New KDoc resolution")
+    fun `fully qualified link should lead to package K2`() = withExperimentalKDocResolution {
```
whyoleg: as in the test below - the name of the test is missleading now

dokka-subprojects/plugin-base/src/test/kotlin/content/kdoc/KDocAmbiguityResolutionTest.kt
```diff
@@ -0,0 +1,782 @@
...
+                    }
+                    skipAllNotMatching()
+                }
+            }
+        }
+    }
+
+
+    @Test
+    @Ignore("KT-83152 [Analysis API, KDoc] Make class name links on constructors point to the class")
```
whyoleg: Let's create an issue in Dokka to track it and put it here

whyoleg: Or we could reuse https://github.com/Kotlin/dokka/issues/3604

dokka-subprojects/plugin-base/src/test/kotlin/content/typealiases/NestedTypealiasTest.kt
vmishenev: I would make this message more straightforward about the reason: In this case, `String` should be `Throwable`.
There are two tpes of errors where the link:
- is resolved, but not Throwable
- is unresolved (e.g., because of a typo).

It seems we cannot distinguish between these two cases because the AA returns nothing in both.
Or do we agree `Couldn't resolve link for String ` is enough?

whyoleg: Yeah, it would be nice to have a better message here, but I think that we will not be able to do it without some additional information from AA.

I think that for now, this warning should be enough, and we should probably create an issue on the AA side to design how this could be supported (in both IDE and Dokka). Perhaps in this case, we need to have some diagnostics inside AA that we can propagate to users? Or still return types from the AA resolver and then additionally check those? (hope not) I don't know.

AbdullinAM: [KT-83442](https://youtrack.jetbrains.com/issue/KT-83442) [Analysis API, KDoc] Unified diagnostics for unresolved links in KDoc

whyoleg (approved):
Nice!

Could you please rebase the PR on latest main, as it should contain fixes for IT?

Issue #4065, opened 2025-03-18
Support documenting context parameters


whyoleg:
Currently it's possible to document function parameters via `@param` and receiver of function/property via `@receiver`:

```kotlin
/**
 * @receiver documentation for receiver of type [String]
 * @param number documentation for parameter with name [number]
 */
fun String.test(number: Int)
```

The same should be possible for context parameters.

There are two options on how it could work:

1. new tag `@context` - context parameters have a separate tag, similar to receiver. As context parameters have different meaning, and most likely should be represented in different section of documentation - separate from [`Parameters` section](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html), it might make sense to have separate tag. In this case receiver, function parameters and context parameters will have three different tags.

```kotlin
/**
 * @context scope documentation for context parameter
 * @receiver documentation for receiver of type [String]
 * @param number documentation for parameter with name [number]
 */
context(scope: Scope)
fun String.test(number: Int)
```

2. reusing `@param` - context parameters are treated similar to ordinary parameters. But it's still make sense to show them separately in documentation.

```kotlin
/**
 * @param scope documentation for context parameter
 * @receiver documentation for receiver of type [String]
 * @param number documentation for parameter with name [number]
 */
context(scope: Scope)
fun String.test(number: Int)
```

Bonus _(questionable)_ option: use `@param` for referencing both context parameters, function parameters and **receiver**. In KDoc it's possible to use `[this]` to reference receiver of function/property in the way it's possible to reference named parameters in documentation:

```kotlin
/**
 * when calling function, [this] should be longer than [number] symbols
 */
fun String.test(number: Int)
```

With this possibility in mind, it could be possible to replace `@receiver` tag with `@param this`:

```kotlin
/**
 * @param scope documentation for context parameter
 * @param this documentation for receiver of type [String]
 * @param number documentation for parameter with name [number]
 */
context(scope: Scope)
fun String.test(number: Int)
```

Or with brackets:

```kotlin
/**
 * @param[scope] documentation for context parameter
 * @param[this] documentation for receiver of type [String]
 * @param[number] documentation for parameter with name [number]
 */
context(scope: Scope)
fun String.test(number: Int)
```

In this case `@receiver` tag could be deprecated (or gradually phased out) and both referencing and documenting receiver will have same syntax

whyoleg:
[...]

PR #4398, opened 2025-12-18
Support context parameters in `@param` tag

AbdullinAM:
Basic implementation of #4065.

Adding a new `TagWrapper` is also an API breaking change. Another way around this is to basically have the logic from `KdocMarkdowsParser` implemented in the `DescriptorSections`. I think it makes sense to have this logic on the "documentable model" side, but I don't have a strong opinion here.

If we want to go with this implementation, probably we also need to update IDE to have similar behaviour.

### DIFF
--- a/dokka-subprojects/core/api/dokka-core.api
+++ b/dokka-subprojects/core/api/dokka-core.api
@@ -3061,9 +3061,9 @@ public final class org/jetbrains/dokka/model/doc/Param : org/jetbrains/dokka/mod
 	public final fun component1 ()Lorg/jetbrains/dokka/model/doc/DocTag;
 	public final fun component2 ()Ljava/lang/String;
 	public final fun component3 ()Lorg/jetbrains/dokka/links/DRI;
-	public final synthetic fun copy (Ljava/lang/String;)Lorg/jetbrains/dokka/model/doc/Param;
+	public final synthetic fun copy (Lorg/jetbrains/dokka/model/doc/DocTag;Ljava/lang/String;)Lorg/jetbrains/dokka/model/doc/Param;
 	public final fun copy (Lorg/jetbrains/dokka/model/doc/DocTag;Ljava/lang/String;Lorg/jetbrains/dokka/links/DRI;)Lorg/jetbrains/dokka/model/doc/Param;
-	public static synthetic fun copy$default (Lorg/jetbrains/dokka/model/doc/Param;Ljava/lang/String;ILjava/lang/Object;)Lorg/jetbrains/dokka/model/doc/Param;
+	public static synthetic fun copy$default (Lorg/jetbrains/dokka/model/doc/Param;Lorg/jetbrains/dokka/model/doc/DocTag;Ljava/lang/String;ILjava/lang/Object;)Lorg/jetbrains/dokka/model/doc/Param;
 	public static synthetic fun copy$default (Lorg/jetbrains/dokka/model/doc/Param;Lorg/jetbrains/dokka/model/doc/DocTag;Ljava/lang/String;Lorg/jetbrains/dokka/links/DRI;ILjava/lang/Object;)Lorg/jetbrains/dokka/model/doc/Param;
 	public fun equals (Ljava/lang/Object;)Z
 	public final fun getAddress ()Lorg/jetbrains/dokka/links/DRI;


dokka-subprojects/plugin-base/src/test/kotlin/content/params/ContentForParamsTest.kt
```diff
@@ -1542,6 +1542,266 @@ class ContentForParamsTest : BaseAbstractTest() {
...
+
+    @Test
+    @OnlySymbols("#4065 - context parameters")
+    fun `context receiver in param tag with brackets`() {
+        testInline(
+            """
+            |/src/main/kotlin/test/source.kt
+            |package test
+            | /**
+            |  * @param[this] receiver
```
vmishenev: Should `@param[this]` be allowed? 
[The KEEP: Streamline ambiguous KDoc links](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0389-kdoc-streamline-ambiguous-KDoc-links.md) explicitly prohibits this case.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -569,6 +569,7 @@ public open class DefaultPageCreator(
             customTagSectionContent(d, sourceSets, customTagContentProviders)
             unnamedTagSectionContent(d, sourceSets) { toHeaderString() }
 
+            contextParamsSectionContent(tags)
```
vmishenev: Have we decided to have a separate section? Should QDoc do the same?

vmishenev (commented):
> Another way around this is to basically have the logic from KdocMarkdowsParser implemented in the DescriptorSections. I think it makes sense to have this logic on the "documentable model" side

As for me, the current approach is okay. In `DescriptorSections`, we would need to calculate what a parameter is.

dokka-subprojects/plugin-base/src/test/kotlin/content/params/ContentForParamsTest.kt
AbdullinAM: I don't think we explicitly agreed on this. It was discussed in #4065, I implemented it here as a 'test' to get feedback on do we really want this behaviour

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
AbdullinAM: It was mentioned by @whyoleg during our discussion. And in general, I think it is the whole idea of #4065. Otherwise, context parameters work correctly now.

My suggestion is to sync with IDE so that QDoc shows the smae

vmishenev: In this case, should we separate the parameters of a function/constructor from the type parameters as well?

AbdullinAM: That makes sense, we can do that too. @whyoleg wdyt?

dokka-subprojects/plugin-base/src/test/kotlin/content/params/ContentForParamsTest.kt
whyoleg: Yes, we should **not** resolve `@param[this]`/`@param this`.
The content of #4065 was an exploration of different options.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/doc/TagWrapper.kt
```diff
@@ -26,6 +26,7 @@ public data class See(override val root: DocTag, override val name: String, val
 public data class Param(override val root: DocTag, override val name: String) : NamedTagWrapper()
 public data class Return(override val root: DocTag) : TagWrapper()
 public data class Receiver(override val root: DocTag) : TagWrapper()
+public data class ContextParameter(override val root: DocTag, override val name: String) : NamedTagWrapper()
```
whyoleg: I do think that we should **not** add a new `TagWrapper` class, as we don't have a tag for it :)
We reuse `@param` in the same way we use it for both value and type parameters.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
whyoleg: Yes, let's split them too (meaning parameters and type parameters) - that feels like the correct way forward.
Additionally, let's create an issue for KTIJ to do the same. I see that for Javadoc, the IDE already does this:

Image

Image

Compared to KDoc:

Image

Image

I find the Javadoc variant both more readable and logical.

Alternative: first, we can ask KTIJ about their opinion on this, and then decide here whether we should split all three kinds of parameters or not.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/doc/TagWrapper.kt
AbdullinAM: I partially agree with you. I just don't like overcomplicating the logic on the `ContentNode` side. Determining whether a link in `@param` references a type paraeter, context parameter or a regular parameter is much simpler and more reliable when we are parsing KDocs. In KDocMarkdownParser we can reuse the logic of KDoc link resolution to differenciate between them, in page creater we will have to rely on the string mapping

whyoleg: What about adding an optional DRI parameter to the `Param` class? Similar to `See` and `Throws`.

AbdullinAM: That probably works, nice 👍

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
AbdullinAM: [KTIJ-36937](https://youtrack.jetbrains.com/issue/KTIJ-36937) QuickDoc: split parameters, context parameters and type parameters into separate sections

AbdullinAM: Context: I synced with Volodya Dolzhenko and he agreed that it makes sense

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/doc/TagWrapper.kt
```diff
@@ -23,7 +24,13 @@ public data class Author(override val root: DocTag) : TagWrapper()
 public data class Version(override val root: DocTag) : TagWrapper()
 public data class Since(override val root: DocTag) : TagWrapper()
 public data class See(override val root: DocTag, override val name: String, val address: DRI?) : NamedTagWrapper()
-public data class Param(override val root: DocTag, override val name: String) : NamedTagWrapper()
+public data class Param(override val root: DocTag, override val name: String, val address: DRI? = null) : NamedTagWrapper() {
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    public constructor(root: DocTag, name: String) : this(root, name, null)
+
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    public fun copy(name: String = this.name): Param = Param(root, name, address = null)
```
whyoleg: `copy` should also contain `root` as it's first argument to make it binary compatible.

dokka-subprojects/plugin-base/src/test/kotlin/content/params/ContentForParamsTest.kt
```diff
@@ -1542,6 +1542,330 @@ class ContentForParamsTest : BaseAbstractTest() {
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    @Test
+    @OnlySymbols("#4065 - context parameters")
```
whyoleg: No context parameters here - should work for K1?
Overall, while we don't care about K1, it's somewhat unusual to see it `@OnlySymbols` here.

dokka-subprojects/plugin-base/src/test/kotlin/content/params/ContentForParamsTest.kt
```diff
@@ -1542,6 +1542,330 @@ class ContentForParamsTest : BaseAbstractTest() {
...
```
whyoleg: same here

whyoleg:
Note: TC IT are red because of https://github.com/Kotlin/dokka/pull/4402#discussion_r2635711945

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KdocMarkdownParser.kt
```diff
@@ -16,6 +17,7 @@ import org.jetbrains.kotlin.kdoc.psi.impl.KDocSection
 import org.jetbrains.kotlin.kdoc.psi.impl.KDocTag
 import org.jetbrains.kotlin.psi.psiUtil.checkDecompiledText
 
+@OptIn(ExperimentalDokkaApi::class)
```
vmishenev: I haven’t checked it in the IDE, but is this really necessary?

PR #4402, opened 2025-12-19
Update Kotlin to 2.3.0 and fixes IT

whyoleg:


dokka-integration-tests/gradle/src/test/kotlin/junit/TestedVersionsSourceTest.kt
```diff
@@ -19,7 +19,7 @@ class TestedVersionsSourceTest {
 
         actual shouldBe """
             gradle: 7.6.4, 8.14.3, 9.1.0
-            kgp: 1.9.25, 2.0.21, 2.1.21, 2.2.21, 2.3.0-Beta2
```
whyoleg, line 22: Looks like we had some merge conflict, or something, as `TestedVersionsSource.kt` uses RC, but in tests it was Beta2 - the TC IT were failing because of it :)

vmishenev: Can we get there versions from [TestedVersions.kt](https://github.com/Kotlin/dokka/pull/4402/files/b86986b412dfafa1a64979002fde36645a7b7325#diff-7395490ce468e15bd9c1153c753099819f53ea7af5a225f1a5d6421e6c4f12f8)? Or write instructions for updating them?

whyoleg: Outputs here are based on a dynamic combination of versions from [TestedVersionsSource.kt](https://github.com/Kotlin/dokka/blob/b86986b412dfafa1a64979002fde36645a7b7325/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/junit/TestedVersionsSource.kt) - which is the new test infra used only for testing DGPv2.
And this test was introduced in #4366 specifically to verify that, in various AGP-related scenarios, we test with all versions supported for a specific AGP execution mode.

Therefore, we should not obtain versions from there. However, I will add an instruction on how to update them: run the test, verify the output is correct, and replace it accordingly.

adam-enko (approved):

PR #4410, opened 2026-01-08
Use `generatedKotlin` and `allKotlinSources` for Kotlin 2.3.0 compatibility

### LINKED ISSUE #4403: KGP 2.3: Use allKotlin
Dokka does not include code wired to the new generatedKotlin SourceDirectorySet. 

Dokka uses allKotlin to include code wired to the generatedKotlin

Not using the new generatedKotlin SourceDirectorySet. 

**Additional context**
New since KGP 2.3

whyoleg:
Fixes #4403

Related to #4348: current PR contains similar test, but using the KGP API


whyoleg:
Blocked by / caused #4412 because in this PR I removed `.filter { it.exists() }`, which caused test failures as seen on CI

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/SuppressKGPGeneratedFilesFunctionalTest.kt
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+})
+
+private fun createProject(
+    suppressGeneratedFiles: Boolean
+): GradleProjectTest = gradleKtsProjectTest("suppress-kgp-generated-files-$suppressGeneratedFiles") {
+    buildGradleKts = """
+        |plugins {
+        |    kotlin("jvm") version "2.3.0" // KGP 2.3.0 is the first version that contains `generatedKotlin` API
```
adam-enko, line 63: Can we also test against older Kotlin versions, to make sure the version-check works?

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/SuppressKGPGeneratedFilesFunctionalTest.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    context("suppressGeneratedFiles") {
+
+        test("when suppressGeneratedFiles is true, generated files should be suppressed") {
+            val project = createProject(suppressGeneratedFiles = true)
+
+            project.runner
+                .addArguments(
+                    ":dokkaGeneratePublicationHtml",
+                    "--rerun-tasks",
```
adam-enko: Was there a specific reason for using `--rerun-tasks`? 

If it's necessary to prevent a specific test failure I think it'd be worth trying to fix it.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -779,6 +788,38 @@ private class KotlinSourceSetDetailsBuilder(
...
+        return if (currentKotlinToolingVersion < KotlinToolingVersion("2.3.0")) {
+            kotlin.sourceDirectories
+        } else {
+            try {
+                @OptIn(ExperimentalKotlinGradlePluginApi::class)
+                allKotlinSources
+            } catch (_: Throwable) {
+                // in case KGP will remove `allKotlinSources` in the future, as it's an experimental API
+                logger.warn { "Failed to get all Kotlin sources (`allKotlinSources`) for source set $name in project ${project.path}" }
+                kotlin.sourceDirectories
```
adam-enko: I added a new util recently to also log the stacktrace, if requested by the user. It could help with diagnosing issues.

I think it'd be useful here, but I'll leave it up to you.

```kotlin
            } catch (ex: Throwable) {
                // in case KGP will remove `allKotlinSources` in the future, as it's an experimental API
                project.logWarningWithStacktraceHint(ex) { "Failed to get all Kotlin sources (`allKotlinSources`) for source set $name in project ${project.path}" }
```

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/SuppressKGPGeneratedFilesFunctionalTest.kt
whyoleg: It's tested in integration tests.
I don't think, that we should introduce `matrix` functional tests

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/SuppressKGPGeneratedFilesFunctionalTest.kt
whyoleg: I most likely copied it from the other test, like [here](https://github.com/Kotlin/dokka/blob/b7102d45d1f4cc46da4854229743da9baedf0f9d/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KotlinMultiplatformFunctionalTest.kt#L29).
I can remove it, but will it make any difference? I thought that functional tests would be executed anyway, as we can't detect changes in assertions or other code in the file/module.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: Not sure if the stacktrace will help here, as it will probably always be the same, as it's called from inside the Dokka plugin itself, no?
Or am I missing something that could help the user in this case?

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/SuppressKGPGeneratedFilesFunctionalTest.kt
adam-enko: The functional tests use a stable directory, not a temp dir, to emulate real life scenarios. Making all tasks rerun could potentially hide issues about non-reproducible behaviour.

I'm not sure why rerun-tasks is in the other kmp tests...

Could you remove it from this PR? It should also be removed from the other tests, if it's not necessary, but that's another task for another time.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: I have no strong reasons, it's a "just in case" idea, should the stacktrace contain useful info (either for us or for users).

whyoleg: Added, "just in case" :)

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/SuppressKGPGeneratedFilesFunctionalTest.kt
whyoleg: Removed it, let's see if the test will work :)

PR #4412, opened 2026-01-09
Fix an AGP9 compatibility issue

whyoleg:
It was (most likely) introduced in https://github.com/Kotlin/dokka/pull/4295/changes/459560c2976a7b9dbfa49b17551451d9c4894b4e.

If the `com.android.kotlin.multiplatform.library` plugin is applied, we don't need to use AGP machinery to obtain classpath entries and should use `compileDependencyFiles` as with other KMP targets.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
@@ -34,6 +34,7 @@ import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation.Companion.MAIN_COMPILATION_NAME
 import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType.androidJvm
 import org.jetbrains.kotlin.gradle.plugin.mpp.AbstractKotlinNativeCompilation
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinAndroidTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJvmAndroidCompilation
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinMetadataCompilation
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinMetadataTarget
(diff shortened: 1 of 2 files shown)


dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
```diff
@@ -26,6 +28,17 @@ kotlin {
...
+    @OptIn(org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi::class)
+    applyDefaultHierarchyTemplate {
+        common {
+            group("web") {
+                withJs()
+                withWasmJs()
+            }
+        }
+    }
 }
```
adam-enko, line 42: Is this config needed? The example project doesn't use the shared web source set.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -687,7 +692,7 @@ private class KotlinSourceSetDetailsBuilder(
         allKotlinCompilationDetails: ListProperty<KotlinCompilationDetails>,
     ) {
         val extantSourceDirectories = providers.provider {
-            kotlinSourceSet.kotlin.sourceDirectories.filter { it.exists() }
+            kotlinSourceSet.kotlin.sourceDirectories
```
adam-enko, line 695: Can you say more about the reason for this change? Is there a test that will fail without it?

If it's necessary, then the name `extantSourceDirectories` should be changed, since it won't refer to existing source dirs.

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/build.gradle.kts
whyoleg: Yes, but it affects the diff (https://github.com/Kotlin/dokka/pull/4412/changes/6cc7a61dd16f7f932ad4d039e08d0afdfd4bf744)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
whyoleg: > Can you say more about the reason for this change?

Specifically because of this change, I accidentally uncovered this AGP9 related issue, and, as stated in commit message, it's a reproducer for the current issue:
> Removing this filter also reproduces the issue in AGP9 even without `common` classes because now this `common` folder is not filtered anymore

It happened in #4410, where I just removed this check, because I thought, that it is redundant. You can check TC there for the error.

So, this filter hided the problem with intersected source-sets, and I only discovered it accidentally.

> If it's necessary, then the name extantSourceDirectories should be changed

Yeah, sure, I can do it now, or we can do it #4410.

adam-enko: I re-added `.filter { it.exists() }` and re-ran AndroidKotlinMultiplatformBuiltInTest and the tests passed.

If  `.filter { it.exists() }` breaks something in #4410 then let's fix it there?

adam-enko: @whyoleg and I discussed, and the reason for removing `.filter { it.exists() }` is not to fix a bug (which was my assumption), but instead because it hides an issue with duplicate source dirs. 

It was probably added as a quick way to fix the issue, but now with AGP9 support we have a better way of registering source dirs without the need for the `.filter { it.exists() }` workaround.

PR #4430, opened 2026-01-21
Enhance logging for unresolved KDoc links

vmishenev:
Prints the location of KDoc links in an IDE-recognizable format `file:///path/to/file.ext:<line>:<column>` for navigating to a specific line and column

image

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KDocProvider.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/KDocProvider.kt
@@ -63,7 +63,7 @@ internal fun KaSession.getKDocDocumentationFrom(
 
     parseFromKDocTag(
         kDocTag = kDocContent.primaryTag,
-        externalDri = { resolveKDocLink(it, getLocation(it) ?: kdocLocation, logger, sourceSet) },
+        externalDri = { resolveKDocLink(it, logger, sourceSet) },
         kdocLocation = kdocLocation
     )
 }
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
@@ -6,6 +6,7 @@ package org.jetbrains.dokka.analysis.kotlin.symbols.kdoc
 
 import org.jetbrains.dokka.DokkaConfiguration
 import org.jetbrains.dokka.analysis.kotlin.symbols.translators.getDRIFromSymbol
+import org.jetbrains.dokka.analysis.kotlin.symbols.utils.getLocation
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.utilities.DokkaLogger
 import org.jetbrains.kotlin.analysis.api.KaExperimentalApi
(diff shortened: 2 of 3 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/utils/getURILocation.kt
```diff
@@ -0,0 +1,24 @@
...
+
+package org.jetbrains.dokka.analysis.kotlin.symbols.utils
+
+import com.intellij.psi.PsiElement
+
+/**
+ * @return a string in the format `file://file.kt:<line>:<column>`,
+ * or `null` if [psiElement] is not from a source file
+ */
+internal fun getLocation(psiElement: PsiElement): String? {
```
whyoleg, line 13: As I see, this is only really used with `KDocLink` as an argument, maybe it makes sense to just move it into (or near) `resolveKDocLink` when we log unresolved link [here](https://github.com/Kotlin/dokka/blob/8148aa2438b8b25e7cb36772c7ffa1774d390e64/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt#L36)?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/TranslatorError.kt
```diff
@@ -16,18 +16,22 @@ internal inline fun <R> KaSession.withExceptionCatcher(symbol: KaSymbol, action:
...
     } catch (e: Throwable) {
-        val file = try {
+        val filePath = try {
             symbol.psi?.containingFile?.virtualFile?.path
         } catch (e: Throwable) {
             "[$e]"
         }
-        val textRange = try {
-            symbol.psi?.textRange.toString()
+        val lineSuffix = try {
```
whyoleg, line 24: Curious: Do you remember what the reason is for having two separate `try/catch` here?

whyoleg (approved):
Nice, this is really nice!

Just small comments :)

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/utils/getURILocation.kt
vmishenev: Do you mean the call site or the declaration?
TBH  the parameter [location](https://github.com/Kotlin/dokka/blob/8148aa2438b8b25e7cb36772c7ffa1774d390e64/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt#L30C5-L30C13) can be removed in favor of `getLocation(kDocLink)`

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/TranslatorError.kt
vmishenev: If line number computation fails, the file path can still be printed, which is important for exception reporting.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/utils/getURILocation.kt
whyoleg: I mean move `getLocation` to that file, make it `private` there, and use it in `resolveKDocLink`.

> TBH the parameter [location](https://github.com/Kotlin/dokka/blob/8148aa2438b8b25e7cb36772c7ffa1774d390e64/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt#L30C5-L30C13) can be removed

Yes, that would be nice too!

vmishenev: > I mean move getLocation to that file, make it private there, and use it in resolveKDocLink.

`getLocation` is supposed to be used not only by KDoc links.

Issue #4246, opened 2025-09-04
Research analyzing Java code via Analysis API


vmishenev:
Currently, [Dokka uses PSI ](https://github.com/Kotlin/dokka/blob/master/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt)to analyze Java code and build the Documentable model.
The proposal is to use the Analysis API and symbols instead of PSI for building the Documentable model of Java code.
This change will allow Dokka to:
- Make the codebase more consistent when analyzing Java and Kotlin code
- Simplify the logic and make the Kotlin view of Java code closer to the compiler
For example, Dokka [matches accessors](https://github.com/Kotlin/dokka/blob/9914d4c03cec1bcb1f0cda91041a9a68624273ef/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt?utm_source=chatgpt.com#L677) to create a Kotlin property on its own, which is unreliable.

**Possible pitfalls:**
- The compiler ignores Java annotations (?).  Probably the symbols translator will need extra logic for Java annotations.

**Definition of done:**
- Implement a prototype under an experimental flag
- All unit tests and Dackka's test should pass with this flag enabled
- Compare the output/documentable model on some popular Java projects

Issue #4425, opened 2026-01-20
Enable new KDoc link resolution by default


whyoleg:
In #4397, we added support for enabling new KDoc link resolution under an experimental system property.
Soon, it will be enabled by default on the AA/IDEA side ([KT-83723](https://youtrack.jetbrains.com/issue/KT-83723) [Analysis API] Enable experimental KDoc resolver by default).

At some point, we should enable it by default in Dokka after additional testing on a wide range of libraries.

whyoleg:
I've tested several kotlinx libraries and all of them have unresolved links with the new resolution

kotlinx-datetime (https://github.com/Kotlin/kotlinx-datetime/commit/c8226a7ad3e1d5594a24b6c5198f66f1fa277c6c):

<details>

<summary>Dokka warnings output</summary>

```
w: [:kotlinx-datetime:dokkaGenerateModuleHtml] Couldn't resolve link: [kotlin.time.Clock] in DeprecatedClock.kt/kotlinx.datetime.toStdlibClock (:kotlinx-datetime/commonMain)
w: [:kotlinx-datetime:dokkaGenerateModuleHtml] Couldn't resolve link: [kotlin.time.Instant] in DeprecatedInstant.kt/kotlinx.datetime.toStdlibInstant (:kotlinx-datetime/commonMain)
w: [:kotlinx-datetime:dokkaGenerateModuleHtml] Couldn't resolve link: [kotlin.time.Clock] in DeprecatedClock.kt/kotlinx.datetime.Clock (:kotlinx-datetime/commonMain)
w: [:kotlinx-datetime:dokkaGenerateModuleHtml] Couldn't resolve link: [kotlin.time.Clock] in DeprecatedClock.kt/kotlinx.datetime.Clock (:kotlinx-datetime/commonMain)
w: [:kotlinx-datetime:dokkaGenerateModuleHtml] Couldn't resolve link: [kotlin.time.Instant] in DeprecatedInstant.kt/kotlinx.datetime.Instant (:kotlinx-datetime/commonMain)
w: [:kotlinx-datetime:dokkaGenerateModuleHtml] Couldn't resolve link: [kotlin.time.Instant] in DeprecatedInstant.kt/kotlinx.datetime.Instant (:kotlinx-datetime/commonMain)
[...]

PR #4445, opened 2026-01-29
Experimental Java analysis using SymbolsTranslator

yuriioliinyk4:
Partially implements #4246

New experimental flag `org.jetbrains.dokka.analysis.enableExperimentalSymbolsJavaAnalysis` is introduced to analyze java using `SymbolsTranslator` instead of `PsiTranslator`.

There are ~60 unit tests left that are failing under new experimental flag. Most of them are related to problems with annotations, type conversion in AA (e.g. java's `Object` become kotlin's `Any`, `int[]` into `IntArray`, etc), modifiers.

### DIFF
--- a/dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocClasslikeTemplateMapTest.kt
+++ b/dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocClasslikeTemplateMapTest.kt
@@ -5,6 +5,7 @@
 package org.jetbrains.dokka.javadoc
 
 import org.jetbrains.dokka.javadoc.pages.JavadocClasslikePageNode
+import org.jetbrains.dokka.javadoc.utils.OnlyJavaPsi
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
@@ -237,6 +238,7 @@ internal class JavadocClasslikeTemplateMapTest : AbstractJavadocTemplateMapTest(
         }
     }
 
+    @OnlyJavaPsi
     @Test
     fun `documented function parameters`(){
         dualTestTemplateMapInline(
--- a/dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocDeprecatedTest.kt
+++ b/dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/JavadocDeprecatedTest.kt
@@ -6,12 +6,14 @@ package org.jetbrains.dokka.javadoc
 
 import org.jetbrains.dokka.javadoc.pages.DeprecatedPage
 import org.jetbrains.dokka.javadoc.renderer.TemplateMap
+import org.jetbrains.dokka.javadoc.utils.OnlyJavaPsi
 import org.junit.jupiter.api.Tag
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
 internal class JavadocDeprecatedTest : AbstractJavadocTemplateMapTest() {
 
+    @OnlyJavaPsi
     @Test
     fun `generates correct number of sections`() {
         testDeprecatedPageTemplateMaps { templateMap ->
@@ -20,6 +22,7 @@ internal class JavadocDeprecatedTest : AbstractJavadocTemplateMapTest() {
         }
     }
 
+    @OnlyJavaPsi
     @Test
     fun `finds correct number of element for removal`() {
         testDeprecatedPageTemplateMaps { templateMap ->
@@ -69,6 +72,7 @@ internal class JavadocDeprecatedTest : AbstractJavadocTemplateMapTest() {
         }
     }
 
+    @OnlyJavaPsi
     @Test
     fun `should be sorted by position`() {
         testDeprecatedPageTemplateMaps { templateMap ->
--- a/dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/utils/TagsAnnotations.kt
+++ b/dokka-subprojects/plugin-javadoc/src/test/kotlin/org/jetbrains/dokka/javadoc/utils/TagsAnnotations.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package org.jetbrains.dokka.javadoc.utils
+
+import org.junit.jupiter.api.Tag
+
+/**
+ * Tests that pass with PSI-based Java analysis but fail with symbol-based Java analysis.
+ */
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY_GETTER,
+    AnnotationTarget.PROPERTY_SETTER
+)
+@Retention(
+    AnnotationRetention.RUNTIME
+)
+@Tag("onlyJavaPsi")
+annotation class OnlyJavaPsi(val reason: String = "")


build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
```diff
@@ -74,6 +75,22 @@ testing {
...
+                    description = "Runs tests using symbols-analysis (K2) for java (excluding tags: ${onlyJavaPsiTags + onlyDescriptorTags})"
+                    useJUnitPlatform {
+                        excludeTags.addAll(onlyJavaPsiTags + onlyDescriptorTags)
+                    }
+                    // Analysis dependencies from `symbolsTestImplementation` should precede all other dependencies
+                    // in order to use the shadowed stdlib from the analysis dependencies
+                    classpath = symbolsTestImplementationResolver.incoming.files + classpath
+
+                    // Enable experimental symbols java analysis
+                    systemProperty("org.jetbrains.dokka.analysis.enableExperimentalSymbolsJavaAnalysis", "true")
```
whyoleg, line 111: BTW, @AbdullinAM, I think it might make sense to convert the current logic regarding new KDoc resolution based on `org.jetbrains.dokka.analysis.enableExperimentalKDocResolution` to also use this approach, to check all tests with new resolution, and not just some of them, to be sure, that we will be able to safely enable the flag by default (#4425)

---

Also, for both those flags, we might want to support enabling them for IT, like we do for K2 - https://github.com/Kotlin/dokka/blob/c4a33d3412d498ad6463da47c2a0cc273b486796/build-logic/src/main/kotlin/dokkabuild.test-integration.gradle.kts#L38-L43

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/mixed/SampleMixedJvmAnalysisTest.kt
```diff
@@ -66,11 +67,13 @@ class SampleMixedJvmAnalysisTest {
         assertEquals(2, pckg.classlikes.size)
         assertEquals(2, pckg.functions.size)
 
-        val firstClasslike = pckg.classlikes[0]
-        assertEquals("MyJavaFileInKotlin", firstClasslike.name)
+        assertTrue {
+            "MyJavaFileInKotlin" in pckg.classlikes.map { it.name }
```
whyoleg, line 71: Curious: What is the reason for changing those?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -87,27 +90,39 @@ internal class DokkaSymbolVisitor(
 
     private fun <T> T.toSourceSetDependent() = if (this != null) mapOf(sourceSet to this) else emptyMap()
 
-    private fun <T : KaSymbol> Sequence<T>.filterSymbolsInSourceSet(moduleFiles: Set<KtFile>): Sequence<T> = filter {
+    private fun <T : KaSymbol> Sequence<T>.filterSymbolsInSourceSet(moduleKtFiles: Set<KtFile>, moduleJavaFiles: Set<PsiJavaFile>): Sequence<T> = filter {
         when (val file = it.psi?.containingFile) {
-            is KtFile -> moduleFiles.contains(file)
+            is KtFile -> moduleKtFiles.contains(file)
+            is PsiJavaFile -> moduleJavaFiles.contains(file) && InternalConfiguration.enableExperimentalSymbolsJavaAnalysis
```
whyoleg: I think we could just move the flag check to where we create `javaFiles` in `visitModule`, and if it's `false`, use `emptySet` there, so the other conditional logic (like in `collectPackages`) will not be needed anymore.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -135,13 +151,20 @@ internal class DokkaSymbolVisitor(
                 .map { visitClassSymbol(it, dri) }
         val typealiases = classifiers.filterIsInstance<KaTypeAliasSymbol>().map { visitTypeAliasSymbol(it, dri) }
 
+        val packageInfo = moduleJavaFiles.singleOrNull { it.name == "package-info.java" }
+        val documentation = packageInfo?.let {
+            javadocParser?.parseDocumentation(it).toSourceSetDependent()
+        }.orEmpty()
+
+        // TODO annotations from `package-info.java`
```
whyoleg: Curious: What can we do about them? Looks like we can only add them to the package model. Javadoc renders them on `package` pages.
We should at least create an issue about it and mention it here.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -918,8 +941,10 @@ internal class DokkaSymbolVisitor(
         if (symbol.origin == KaSymbolOrigin.SOURCE_MEMBER_GENERATED)
             // a primary (implicit default) constructor  can be generated, so we need KDoc from @constructor tag
             getGeneratedKDocDocumentationFrom(symbol) ?: if(symbol is KaConstructorSymbol) getKDocDocumentationFrom(symbol, logger, sourceSet) else null
+        else if (symbol.isJavaSource())
+            javadocParser?.let { getJavaDocDocumentationFrom(symbol, it) }
         else
-            getKDocDocumentationFrom(symbol, logger, sourceSet) ?: javadocParser?.let { getJavaDocDocumentationFrom(symbol, it, sourceSet) }
+            getKDocDocumentationFrom(symbol, logger, sourceSet) ?: javadocParser?.let { getJavaDocDocumentationFrom(symbol, it) }
```
whyoleg: WDYT about refactoring it to use `when(symbol.origin)`?

Also, should we remove the `javadocParser` call in the `else` branch, since we handle it explicitly based on `origin` now?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
whyoleg, line 155: Should be fixed, right?

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/mixed/SampleMixedJvmAnalysisTest.kt
vmishenev: I suppose`val firstClasslike = pckg.classlikes[0]` depends on the flag `enableExperimentalSymbolsJavaAnalysis` now. 
The order of `classlikes` has been changed.
I am not sure if it makes sense to sort `classlikes` to keep the previous order.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
AbdullinAM: Makes sense, I will create a PR

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/analysis/test/jvm/mixed/SampleMixedJvmAnalysisTest.kt
yuriioliinyk4: Yes, only the order in the list changes

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
yuriioliinyk4: > WDYT about refactoring it to use when(symbol.origin)?

0c9c59e

> Also, should we remove the javadocParser call in the else branch, since we handle it explicitly based on origin now?

[This test](https://github.com/Kotlin/dokka/blob/4108892dd129bea40eab98ad8bf91273ed7c25d9/dokka-subprojects/plugin-base/src/test/kotlin/model/MultiLanguageInheritanceTest.kt#L74) is failing after removal. So we need that call when KDoc should inherit documentation from Javadoc.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
yuriioliinyk4: We could fix this by copying the [PSI related functions](https://github.com/Kotlin/dokka/blob/4108892dd129bea40eab98ad8bf91273ed7c25d9/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt#L90) from `PsiParser`, but that would mean duplicating a lot of code in `SymbolsTranslator`. So we decided to leave it for now and fix it later together with the annotations logic.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
yuriioliinyk4: d9804f7

vmishenev:
@yuriioliinyk4 Could you resolve merge conflicts, please?

whyoleg (approved):

yuriioliinyk4:
I resolved conflicts, review pls last 3 commits.

Note that there are few failing tests outside the `plugin-base` ( `@OnlyJavaPsi` cant be applied to those tests because it is defined in the `plugin-base`)

```
> Task :dokka-subprojects:plugin-javadoc:testJavaSymbols

JavadocDeprecatedTest > should be sorted by position() FAILED
    org.opentest4j.AssertionFailedError at JavadocDeprecatedTest.kt:181

JavadocDeprecatedTest > finds correct number of element for removal() FAILED
    java.util.NoSuchElementException at JavadocDeprecatedTest.kt:195

JavadocDeprecatedTest > generates correct number of sections() FAILED
    org.opentest4j.AssertionFailedError at JavadocDeprecatedTest.kt:181

JavadocClasslikeTemplateMapTest > documented function parameters() FAILED
    java.lang.AssertionError at JavadocClasslikeTemplateMapTest.kt:242
        Caused by: org.opentest4j.AssertionFailedError at JavadocClasslikeTemplateMapTest.kt:476
        
> Task :dokka-subprojects:plugin-kotlin-as-java:testJavaSymbols

KotlinAsJavaPluginTest > Java function should keep its access modifier() FAILED
    java.lang.AssertionError at KotlinAsJavaPluginTest.kt:506

KotlinAsJavaPluginTest > java properties should keep its modifiers() FAILED
    java.lang.AssertionError at KotlinAsJavaPluginTest.kt:215
```

vmishenev:
> @OnlyJavaPsi cant be applied to those tests because it is defined in the plugin-base
 
Can you duplicate this annotation in kotlin-as-java and other plugins?

vmishenev (approved):

PR #4450, opened 2026-02-02
Update Enum.values template

fzhinkin:
Update the template to reflect changes from https://github.com/JetBrains/kotlin/commit/b56aea07e9dfbe8f37eb0ff117907e1b6fd728fa


fzhinkin:
There's also `EnumTemplatesTest`, but it assess sources of the stdlib whose version is `2.0.21`. I'm not sure how useful that test is.

whyoleg:
> There's also EnumTemplatesTest, but it assess sources of the stdlib whose version is 2.0.21. I'm not sure how useful that test is.

Previously, it was an IT test that tested the latest stdlib, but now, as we need to compile the Gradle Plugin with an older Kotlin compiler to be compatible, the test feels odd.

The idea of the test is still valid: to at least know about the changes in kotlin stdlib, such as those. :)

PR #4454, opened 2026-02-09
Add integration test target with experimental KDoc resolution enabled

AbdullinAM:


build-logic/src/main/kotlin/dokkabuild.test-integration.gradle.kts
```diff
@@ -38,6 +38,11 @@ tasks.withType<Test>().configureEach {
     val useK2 = dokkaBuild.integrationTestUseK2
     systemProperty.inputProperty("org.jetbrains.dokka.experimental.tryK2", useK2)
         .optional(true)
+
+    val enableExperimentalKDocResolution = dokkaBuild.integrationTestEnableExperimentalKDocResolution
+    systemProperty.inputProperty("org.jetbrains.dokka.analysis.enableExperimentalKDocResolution", enableExperimentalKDocResolution)
```
whyoleg, line 43: It's not enough to pass the flag here; in IT, we start a separate Gradle process, so we need to pass it to that process.
Take a look at:
https://github.com/Kotlin/dokka/blob/ef58e0297758a153b97a746406de18c62803b5da/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/AbstractGradleIntegrationTest.kt#L114

build-logic/src/main/kotlin/dokkabuild.test-integration.gradle.kts
```diff
@@ -38,6 +38,11 @@ tasks.withType<Test>().configureEach {
...
+        .optional(false)
```
whyoleg:
```suggestion
        .optional(true)
```

It's about an input, not a default value; let's be consistent with other flags.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
```diff
@@ -74,6 +74,22 @@ testing {
                 }
             }
 
+            // Create a new target for running symbols-analysis tests with enabled experimental KDoc resolution (K2).
+            val testSymbolsWithKDocResolutionTarget = targets.register("testSymbolsWithKDocResolution") {
```
whyoleg: As we now have a separate task to run tests with the flag enabled, I believe we should remove existing guarding from tests ([withExperimentalKDocResolution](https://github.com/Kotlin/dokka/blob/f9d6073a8cedec60481f44d64fce5aa1491e5c5f/dokka-subprojects/plugin-base/src/test/kotlin/utils/systemProperties.kt#L28)) and add annotations in place (`@OnlyNewKDocResolution`) where we have different behavior.

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
```diff
@@ -64,16 +66,34 @@ testing {
...
                     // Analysis dependencies from `symbolsTestImplementation` should precede all other dependencies
                     // in order  to use the shadowed stdlib from the analysis dependencies
                     classpath = symbolsTestImplementationResolver.incoming.files + classpath
                 }
             }
 
+            // Create a new target for running symbols-analysis tests with enabled experimental KDoc resolution (K2).
+            val testSymbolsWithKDocResolutionTarget = targets.register("testSymbolsWithKDocResolution") {
+                testTask.configure {
+                    val excludedTags = onlyDescriptorTags + onlySymbolsTags
```
whyoleg:
```suggestion
                    val excludedTags = onlyDescriptorTags
```

I believe we should not exclude `onlySymbolsTags` here, because otherwise we will not run K2-only tests with the new kdoc resolution.

To check this, I've executed different test tasks for `LinkTest`.
The test has:
- 41 total tests
- 4 descriptors only
- 10 symbols only
- 2 new kdoc resolution only

Then I run test tasks:
- `testDescriptors` -> 29 executed
- `testSymbols` -> 35 executed
- `testSymbolsWithKDocResolution` -> 27 executed
- `testSymbolsWithKDocResolution` with suggestion applied -> 37 executed

whyoleg (approved):
Nice! One small change and we are good to go (e.g., set up CI pipeline)

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
AbdullinAM: Got it, nice catch!

build-logic/src/main/kotlin/dokkabuild.test-k2.gradle.kts
```diff
@@ -64,16 +66,34 @@ testing {
...
+                        excludeTags.addAll(excludedTags)
                     }
                     // Analysis dependencies from `symbolsTestImplementation` should precede all other dependencies
                     // in order  to use the shadowed stdlib from the analysis dependencies
                     classpath = symbolsTestImplementationResolver.incoming.files + classpath
                 }
             }
 
+            // Create a new target for running symbols-analysis tests with enabled experimental KDoc resolution (K2).
+            val testSymbolsWithKDocResolutionTarget = targets.register("testSymbolsWithKDocResolution") {
```
vmishenev: nitpick: `testSymbolsWithNewKDocResolution`

PR #4455, opened 2026-02-09
Report about unresolved symbols

vmishenev:
The current approach have the following drawbacks: 
- reports for user-invisible symbols (e.g. private declarations)
- can report twice
- reporting logic inside translators

In any case, any unresolved symbol is abnormal behavior.

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/translators/SymbolsTranslatorTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/translators/SymbolsTranslatorTest.kt
@@ -34,7 +34,7 @@ class SymbolsTranslatorTest : BaseAbstractTest() {
                     .replace("\\","/") // for Win
 
                 assertEquals(
-                    "`UnresolvedSymbol` is unresolved in file:///PATH/Test.kt:1:7",
+                    "`UnresolvedSymbols` is unresolved in file:///PATH/Test.kt:1:7",
                     warn.replace(path, "PATH")
                 )
                 assertEquals(
--- a/dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
@@ -455,6 +455,28 @@ class SignatureTest : BaseAbstractTest() {
     }
 
     @Test
+    @OnlySymbols
+    fun `fun with unresolved parameter`() {
+        val source = source("fun simpleFun(param: UnresolvedType): Unit")
+        val writerPlugin = TestOutputWriterPlugin()
+
+        testInline(
+            source,
+            configuration,
+            pluginOverrides = listOf(writerPlugin)
+        ) {
+            renderingStage = { _, _ ->
+                writerPlugin.writer.renderedContent("root/example/simple-fun.html").firstSignature().match(
+                    "fun ", A("simpleFun"), "(", Parameters(
+                        Parameter("param: UnresolvedType"),
+                    ), ")",
+                    ignoreSpanWithTokenStyle = true
+                )
+            }
+        }
+    }
+
+    @Test
     fun `fun with vararg`() {
         val source = source("fun simpleFun(vararg params: Int): Unit")
         val writerPlugin = TestOutputWriterPlugin()


dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KotlinMultiplatformFunctionalTest.kt
```diff
@@ -148,7 +148,7 @@ private fun initKotlinMultiplatformProject(
 
         buildGradleKts = """
             |plugins {
-            |  kotlin("multiplatform") version "1.8.22"
+            |  kotlin("multiplatform") version "1.9.22"
```
vmishenev, line 151: K2 does not support a legacy JS stdlib: `println` is unresolved.

vmishenev: https://youtrack.jetbrains.com/issue/KT-64159/Analysis-API-JS-klib-do-not-work-for-Kotlin-1.9.0

whyoleg: I also think we could close the issue, since it won't be fixed, it's really not needed anymore, and it was never reported by users.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -73,19 +75,35 @@ internal class AnnotationTranslator {
...
+            dri = run {
+                val classId = annotation.classId
+                // classId might be null on a non-existing annotation call,
+                if (classId != null && classId != errorClassId) {
+                    classId.createDRI()
+                } else {
+                    val psi = annotation.psi
+                    val location = psi?.let { getLocation(it) }
+                    val text = psi?.text.orEmpty()
+                    logger.warn("Unknown annotation $text in $location")
```
whyoleg:
```suggestion
                    logger.warn("Unknown annotation '$text' in $location")
```

might make sense to add some symbols like `'` to split plain text from the message itself for better readability.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -73,19 +75,35 @@ internal class AnnotationTranslator {
...
+                    logger.debug(
+                        "Unknown annotation $text in ${location}\n" + Thread.currentThread().stackTrace.drop(1)
```
whyoleg:
```suggestion
                        "Unknown annotation '$text' in ${location}\n" + Thread.currentThread().stackTrace.drop(1)
```

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -112,22 +130,34 @@ internal class AnnotationTranslator {
...
                 is KaClassType -> ClassValue(
                     type.classId.relativeClassName.asString(),
                     type.classId.createDRI()
                 )
-                else -> ClassValue(
-                    type.toString(),
+                else -> {
+                    val psi = annotationValue.sourcePsi ?: containingAnnotation.psi
+                    val location = psi?.let { getLocation(it) }
+                    logger.warn("Unknown annotation value ${psi?.text.orEmpty()} in $location")
```
whyoleg:
```suggestion
                    logger.warn("Unknown annotation value '${psi?.text.orEmpty()}' in $location")
```

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -73,19 +75,35 @@ internal class AnnotationTranslator {
...
+                val classId = annotation.classId
+                // classId might be null on a non-existing annotation call,
+                if (classId != null && classId != errorClassId) {
+                    classId.createDRI()
+                } else {
+                    val psi = annotation.psi
+                    val location = psi?.let { getLocation(it) }
+                    val text = psi?.text.orEmpty()
+                    logger.warn("Unknown annotation $text in $location")
+                    logger.debug(
```
whyoleg, line 88: What is the reason for having `debug` here, but not in other places where just `warn` is used?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/TypeTranslator.kt
```diff
@@ -210,4 +220,20 @@ internal class TypeTranslator(
         KaClassKind.INTERFACE -> KotlinClassKindTypes.INTERFACE
         KaClassKind.ANONYMOUS_OBJECT -> KotlinClassKindTypes.OBJECT
     }
+
+    private fun KaSession.report(errorType: KaErrorType, location: Location) {
```
whyoleg: minor: `report` is too generic a name, maybe something more explicit like `logErrorType`?

dokka-subprojects/plugin-base/src/test/kotlin/translators/SymbolsTranslatorTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        sourceSets {
+            sourceSet {
+                sourceRoots = listOf("src/main/kotlin")
+            }
+        }
+    }
+
+    @Test
+    @OnlySymbols
+    fun `method overriding two documented classes picks closest class documentation`() {
```
whyoleg: test name seems unrelated

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
vmishenev: The debug mode prints a stack trace as well. I think it could be non-user-friendly to have it at the warn level.

vmishenev:
Also, it fixes https://github.com/Kotlin/dokka/issues/503 
We display a presentable name of unresolved symbol if it is available. 
For example, in the case of an implicit type `fun f() = unresolvedSymbol` it just prints `type.toString()` as before
So, yes, we need to update the IT assertions .

adam-enko:
> @adam-enko, could you maybe point to the best place for such an assertion?

In the integration tests, here:

- https://github.com/Kotlin/dokka/blob/b63e34a407185416468d7b9736cb62e4f5fd8edd/dokka-integration-tests/utilities/src/main/kotlin/org/jetbrains/dokka/it/AbstractIntegrationTest.kt#L31

- https://github.com/Kotlin/dokka/blob/d2b7db8efb9270cfe3d54fb52892ccccfba4de15/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/htmlAssertions.kt#L14

And in a few DGP tests:

https://github.com/search?q=repo%3AKotlin%2Fdokka+%2FError+class%2F+path%3A%2Fdokka-runners%2Fdokka-gradle-plugin%2F**&type=code

Issue #3633, opened 2024-05-27
Public actual declarations are not shown in docs if expect is internal


ilya-g:
**Describe the bug**
```
// common source set
internal expect fun foo() {}

// actual platform 1 (e.g. Native)
/** foo docs */
public actual fun foo() {}

// actual platform 2 (e.g. Wasm)
internal actual fun foo() {}
```

Actualizing an internal expect declaration with a public actual declarations is [allowed in KMP](https://kotlinlang.org/docs/whatsnew20.html#different-visibility-levels-of-expected-and-actual-declarations) (at least for functions, but we use it for classifiers as well in stdlib). However if such an internal expect declaration is introduced, its public actual counterparts are no longer shown in documentation.

**Expected behavior**
Public actual declarations should continue being rendered for individual platforms as if there's no expect declared.

**To Reproduce**
Kotlin branch: [rr/stdlib/docs-internal-expect-public-actual](https://github.com/JetBrains/kotlin/compare/rr/stdlib/docs-internal-expect-public-actual)
Docs build: https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_LibraryReferenceLatestDocs/509723400?buildTab=artifacts
Example declaration: `kotlin.system.exitProcess`
- before introducing expect: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.system/exit-process.html
- after introducing expect: https://buildserver-artifacts.labs.intellij.net/presignedTokenAuth/1002a5ef-c5bb-4488-b578-9156ba6e24bd/repository/download/Kotlin_KotlinDev_LibraryReferenceLatestDocs/509723400:id/latest-version.zip!/all-libs/kotlin-stdlib/kotlin.system/index.html

**Installation**
- Operating system: macOS/Windows/Linux
- Build tool: Gradle v7.6
- Dokka version: 2.0.20-dev-334

**Context**
We need to use internal expects for public declarations for proper K2 stdlib compilation of the code shared between Native and Wasm.

PR #4464, opened 2026-02-27
Add option to suppress annotated declarations

### LINKED ISSUE #1861: Option to skip things marked with an annotation

rnett:
Fixes https://github.com/Kotlin/dokka/issues/1861

Adds a `suppressedAnnotations` source set level config that causes any declarations annotated with one of the given FqNames to be omitted from the documentation.


dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/defaultConfiguration.kt
```diff
@@ -55,6 +55,7 @@ public data class DokkaSourceSetImpl(
     override val noStdlibLink: Boolean = DokkaDefaults.noStdlibLink,
     override val noJdkLink: Boolean = DokkaDefaults.noJdkLink,
     override val suppressedFiles: Set<File> = emptySet(),
+    override val suppressedAnnotations: Set<String> = DokkaDefaults.suppressedAnnotations.toMutableSet(),
```
whyoleg: New property should be added after all existing properties, as well as deprecated(hidden) constructor and copy declarations should be added to preserve ABI compatibility.
You can take a look at how it's done for nested typealiases [here](https://github.com/Kotlin/dokka/blob/b49856b08a8e56cf7d272cab6a6b16a47b71afb6/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt#L176-L254).

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/configuration.kt
```diff
@@ -109,7 +114,11 @@ public data class DokkaSourceSetID(
 public data class GlobalDokkaConfiguration(
     val perPackageOptions: List<PackageOptionsImpl>?,
     val externalDocumentationLinks: List<ExternalDocumentationLinkImpl>?,
-    val sourceLinks: List<SourceLinkDefinitionImpl>?
+    val sourceLinks: List<SourceLinkDefinitionImpl>?,
+    /**
+     * A set of annotation fully qualified names (FQNs) to suppress declarations annotated with across all source sets.
+     */
+    val suppressedAnnotations: Set<String>? = null
```
whyoleg: Can we avoid adding those here for now (I think so)?

dokka-subprojects/core-test-api/src/main/kotlin/org/jetbrains/dokka/testApi/testRunner/TestDokkaConfigurationBuilder.kt
```diff
@@ -62,6 +63,11 @@ public class TestDokkaConfigurationBuilder {
         finalizeCoroutines = false
     )
 
+    public fun applyGlobals(globals: GlobalDokkaConfiguration): DokkaConfigurationImpl {
```
whyoleg: Why do we need this?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/AnnotationSuppressionFilter.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.properties.WithExtraProperties
+import org.jetbrains.dokka.plugability.DokkaContext
+
+public class AnnotationSuppressionFilter(
+    public val dokkaContext: DokkaContext
+) : SuppressedByConditionDocumentableFilterTransformer(dokkaContext) {
+
+    private val suppressedAnnotations by lazy {
+        dokkaContext.configuration.sourceSets.flatMap { it.suppressedAnnotations }.toSet()
```
whyoleg: This isn't correct. We should not just combine configurations across source-sets.
Could you take a look at `DeprecatedDocumentableFilterTransformer` and how it works with sourceSet-specific configuration?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/AnnotationSuppressionFilter.kt
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    override fun shouldBeSuppressed(d: Documentable): Boolean {
+        if (suppressedAnnotations.isEmpty()) return false
+
+        @Suppress("UNCHECKED_CAST")
+        val annotations =
+            (d as? WithExtraProperties<AnnotationTarget>)?.extra?.get(Annotations as org.jetbrains.dokka.model.properties.ExtraProperty.Key<AnnotationTarget, Annotations>)
```
whyoleg: Could you refactor it to follow the pattern from: https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/DeprecatedDocumentableFilterTransformer.kt#L45-L49

This way, it will be a bit easier to follow the logic here.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/AnnotationSuppressionFilter.kt
```diff
@@ -0,0 +1,32 @@
...
+
+package org.jetbrains.dokka.base.transformers.documentables
+
+import org.jetbrains.dokka.model.AnnotationTarget
+import org.jetbrains.dokka.model.Annotations
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.properties.WithExtraProperties
+import org.jetbrains.dokka.plugability.DokkaContext
+
+public class AnnotationSuppressionFilter(
```
whyoleg:
```suggestion
internal class SuppressedByAnnotationsDocumentableFilterTransformer(
```

Let's follow the naming pattern here, similar to other transformers, and make it `internal` for now, since this shouldn't really be exposed to the user. Most of the transformers should be `internal`, but they are not, mostly because of historical reasons.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/DokkaBase.kt
```diff
@@ -79,6 +79,12 @@ public class DokkaBase : DokkaPlugin() {
         preMergeDocumentableTransformer providing ::SuppressedByConfigurationDocumentableFilterTransformer
     }
 
+    public val annotationSuppressionFilter: Extension<PreMergeDocumentableTransformer, *, *> by extending {
```
whyoleg:
```suggestion
    public val suppressedByAnnotationsFilter: Extension<PreMergeDocumentableTransformer, *, *> by extending {
```

whyoleg (changes requested):
Thanks for the PR!
First round of review.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/defaultConfiguration.kt
```diff
@@ -57,7 +57,111 @@ public data class DokkaSourceSetImpl(
     override val suppressedFiles: Set<File> = emptySet(),
     override val analysisPlatform: Platform = DokkaDefaults.analysisPlatform,
     override val documentedVisibilities: Set<DokkaConfiguration.Visibility> = DokkaDefaults.documentedVisibilities,
+    override val suppressedAnnotations: Set<String> = DokkaDefaults.suppressedAnnotations,
 ) : DokkaSourceSet {
+
+    @Suppress("DEPRECATION")
+    @Deprecated("Use primary constructor with suppressedAnnotations", level = DeprecationLevel.HIDDEN)
```
whyoleg:
```suggestion
    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
```

As it's `HIDDEN`, it will not be visible at all.

whyoleg (approved):
LGTM!
Let's also wait for @vmishenev's review.

docs/topics/runners/dokka-cli.md
```diff
@@ -202,6 +202,7 @@ Short summary:
 | `noStdlibLink`               | Whether to generate links to the Kotlin standard library.                                                                                                                      | 
 | `noJdkLink`                  | Whether to generate links to JDK Javadocs.                                                                                                                                     | 
 | `suppressedFiles`            | Paths to files to be suppressed. Accepts multiple paths separated by semicolons.                                                                                               |
+| `suppressedAnnotations`      | Annotation fully qualified names (FQNs) to suppress declarations annotated with. Accepts multiple values separated by semicolons.                                              |
```
vmishenev: The naming is ambiguous. 
By the name, I would expect this to work like `suppressedFiles`, meaning it should suppress annotations only, but it does not.
Maybe you have a better name. 

@whyoleg

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/SuppressedByAnnotationsDocumentableFilterTransformer.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.properties.WithExtraProperties
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.transformers.documentation.sourceSet
+
+internal class SuppressedByAnnotationsDocumentableFilterTransformer(
+    context: DokkaContext
+) : SuppressedByConditionDocumentableFilterTransformer(context) {
+
+    override fun shouldBeSuppressed(d: Documentable): Boolean {
```
vmishenev, line 17: There is a corner case in a KMP project
```
// jvm platform
@Supress 
actual fun f()

// native platform
actual fun f()
```

In this case, both declarations will be filtered out.
Meanwhile, suppressedAnnotations is one of the source set options.

As for me, this case is OK for the current PR.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/SuppressedByAnnotationsDocumentableFilterTransformerTest.kt
```diff
@@ -0,0 +1,149 @@
...
+            class Annotated
+
+            class NotAnnotated
+            """.trimIndent(),
+            configuration
+        ) {
+            documentablesMergingStage = { module ->
+                val pkg = module.packages.single { it.name == "test" }
+                val classNames = pkg.classlikes.map { it.name }
+                // SuppressMe itself should be present, but Annotated should be suppressed
```
vmishenev, line 40: Also, should we suppress the annotations themselves?
It seems that having them documented does not make sense in most cases.

vmishenev (approved):
I only have a product-related question.

docs/topics/runners/dokka-cli.md
rnett: Yeah, I agree, I just couldn't think of anything better that wasn't super verbose.  `supressAnnotatedWith` was the next closest.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/SuppressedByAnnotationsDocumentableFilterTransformerTest.kt
rnett: IMO that makes sense.  @whyoleg thoughts?  It would make `supressedAnnotations` work a bit better as the name.

docs/topics/runners/dokka-cli.md
whyoleg: Yeah, I agree too, thanks for bringing this up!
I do like the `suppressAnnotatedWith` naming

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/SuppressedByAnnotationsDocumentableFilterTransformer.kt
whyoleg: > In this case, both declarations will be filtered out.

I think we could at least have a test for this case, because it's not really obvious to me why this could happen.
We do run this filter in `preMergeDocumentableTransformer` phase, so I would expect it to be filtered out only in `jvm` source-set.
In some way, it reminds me of #3633 (though, there we filter out `expect` and not `actual`, but still), so we should handle such cases correctly during the merge.

@rnett, could you add such a test?

dokka-subprojects/plugin-base/src/test/kotlin/transformers/SuppressedByAnnotationsDocumentableFilterTransformerTest.kt
whyoleg: I don't think we should do this for multiple reasons.
One of them is that annotations could come from other libraries; e.g., there is even a mention of the compose `@Preview` annotation in the issue itself (https://github.com/Kotlin/dokka/issues/1861#issuecomment-2484406059).
The other one is composability.

We already have other mechanisms to suppress different things:
- [MustBeDocumented](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.annotation/-must-be-documented/) - to not show annotation in generated documentation.
- `@suppress` tag - to suppress declarations.
- `suppressedFiles` - to suppress all declarations in files.
- per package `suppress` flag - to suppress all declarations in the package.

So, I think that this new option (`suppressAnnotatedWith`) should also do only one thing - suppress annotated declarations, in line with all other `suppress` related mechanisms.
If annotation should also be suppressed, it could be easily done via **single** `@suppress` tag on that annotation.

Though we could mention that the `@suppress` tag can be used to hide the annotation itself in the kdoc/documentation of this new option, so it will be easier to discover.
WDYT?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/transformers/documentables/SuppressedByAnnotationsDocumentableFilterTransformer.kt
rnett: Done.

docs/topics/runners/dokka-cli.md
rnett: Done.

rnett:
The test failure is due to a artifact download failure.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/SuppressedByAnnotationsDocumentableFilterTransformerTest.kt
```diff
@@ -146,4 +143,68 @@ class SuppressedByAnnotationsDocumentableFilterTransformerTest : BaseAbstractTes
...
+                val pkg = module.packages.single { it.name == "test" }
+                val suppressMe = pkg.classlikes.find { it.name == "SuppressMe" }
+                assertNotNull(suppressMe, "SuppressMe annotation class should be present")
+
+                val allF = pkg.functions.filter { it.name == "f" }
+
+                assertTrue(pkg.classlikes.any { it.name == "SuppressMe" }, "SuppressMe annotation should be present")
+
+                // There should be at least one 'f' function (the expect from common)
+                assertTrue(allF.isNotEmpty(), "Should have at least one 'f' function (expect from common)")
```
vmishenev: Could you check the source sets of `allF`? They should be common and native.

rnett: Good idea, done.

whyoleg (approved):
Good to go from my side, thanks!

PR #4468, opened 2026-03-02
Fix DGP reading all Gradle properties

### LINKED ISSUE #4467: DGP reads all Gradle properties, which causes unnecessary CC invalidation
Replace `.provider { project.extra.properties[flag]?.toString() }` with `.provider { project.findProperty(flag)?.toString() }`

https://github.com/Kotlin/dokka/blob/v2.2.0-Beta/dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt#L341

Add a CC reuse test where (using Gradle 9.1 or 9.4-rc-1) you change a unused property but the cache is reused. (Keep in mind Gradle 9.2 and 9.3 had a regression and any Gradle property change would invalidate CC.)

Context:
DGP uses a Gradle property to toggle between v1 and v2 modes. There are a lot of unit tests for this. The tests use Gradle's ProjectBuilder, not TestKit, because it's much faster. But ProjectBuilder doesn't support mocking Gradle properties - instead the tests use extra properties.

---

Plan:

- Update Dokka's Gradle to 9.1 (currently it's 8.14.2).
- Move `dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleCCReportData.kt` to `dokka-runners/dokka-gradle-plugin/src/testFixtures/kotlin/`. 
- Create a test in `dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/`.

adam-enko:
- Replace `project.extra.properties[...]` (which reads all properties, meaning they will be CC inputs) with `project.findProperty(...)` (only reads a single property).
- Add tests to verify DGP's CC inputs.
- Add Kotlin JVM test project.

fix #4467

### DIFF
--- a/dokka-integration-tests/gradle/src/test/kotlin/KotlinJvmIT.kt
+++ b/dokka-integration-tests/gradle/src/test/kotlin/KotlinJvmIT.kt
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package org.jetbrains.dokka.it.gradle
+
+import io.kotest.assertions.withClue
+import io.kotest.matchers.string.shouldContain
+import org.gradle.testkit.runner.TaskOutcome.*
+import org.jetbrains.dokka.gradle.utils.addArguments
+import org.jetbrains.dokka.gradle.utils.build
+import org.jetbrains.dokka.gradle.utils.shouldNotHaveRunTask
+import org.jetbrains.dokka.it.gradle.junit.*
+import org.jetbrains.dokka.it.gradle.junit.TestedVersions
+import org.junit.jupiter.api.Assumptions.assumeTrue
+
+/**
+ * Integration test for the `it-kotlin-jvm` project.
+ */
+@TestsKotlinJvm
+@TestsDGPv2
+class KotlinJvmIT {
+
+    @DokkaGradlePluginTest(sourceProjectName = "it-kotlin-jvm")
+    fun `Dokka tasks should be build cacheable`(
+        project: DokkaGradleProjectRunner,
+    ) {
+        project.runner
+            .addArguments(
+                ":dokkaGenerate",
+                "--build-cache",
+            )
+            .build {
+                withClue("expect dokkaGenerate runs successfully") {
+                    shouldHaveTask(":dokkaGenerate").shouldHaveOutcome(UP_TO_DATE, SUCCESS)
+                    shouldHaveTask(":dokkaGeneratePublicationHtml").shouldHaveOutcome(FROM_CACHE, SUCCESS)
+                }
+                withClue("only one project is documented, so expect no Dokka module generation") {
+                    shouldNotHaveRunTask(":dokkaGenerateModuleHtml")
+                }
+            }
+
+        project.runner
+            .addArguments(
+                ":clean",
+                "--build-cache",
+            ).build {
+                withClue("expect clean runs successfully") {
+                    shouldHaveTask(":clean").shouldHaveOutcome(SUCCESS)
+                }
+            }
+
+        project.runner
+            .addArguments(
+                ":dokkaGenerate",
+                "--build-cache",
+            )
+            .build {
+                withClue("expect dokkaGenerate lifecycle task is up-to-date") {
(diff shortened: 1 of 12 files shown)


dokka-integration-tests/gradle/src/testExampleProjects/kotlin/ExampleProjectsTest.kt
CristianGM, line 481: you'll need to add `-Pwhatever=true` to see no one is reading it.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
```diff
@@ -338,7 +337,7 @@ constructor(
                         // (Because org.gradle.testfixtures.ProjectBuilder doesn't support mocking Gradle properties.
                         // But maybe soon! https://github.com/gradle/gradle/pull/30002)
                         project
-                            .provider { project.extra.properties[flag]?.toString() }
+                            .provider { project.findProperty(flag)?.toString() }
```
adam-enko, line 340: This change is the key fix in this PR.

adam-enko:
GitHub Actions performance is degraded :/
https://www.githubstatus.com/incidents/g5gnt5l5hf56

dokka-integration-tests/gradle/src/test/kotlin/KotlinJvmIT.kt
```diff
@@ -0,0 +1,138 @@
...
+        project: DokkaGradleProjectRunner,
+    ) {
+        project.runner
+            .addArguments(
+                ":dokkaGenerate",
+                "--build-cache",
+            )
+            .build {
+                withClue("expect dokkaGenerate runs successfully") {
+                    shouldHaveTask(":dokkaGenerate").shouldHaveOutcome(UP_TO_DATE, SUCCESS)
```
antohaby, line 34: is it UP_TO_DATE or SUCCESS? 
why cant it be just one particular state?  e.g. SUCCESS on first run, UP_TO_DATE on second?

dokka-integration-tests/gradle/src/test/kotlin/KotlinMultiplatformIT.kt
```diff
@@ -186,4 +171,53 @@ class KotlinMultiplatformIT {
...
+                shouldHaveTask(":dokkaGenerate").shouldHaveOutcome(UP_TO_DATE, SUCCESS)
+
+                output shouldContain "Configuration cache entry stored"
+
+                testConfigurationCacheResult(project, testedVersions)
+            }
+        }
+
+        withClue("second build - because sometimes KGP needs to finish installing kotlin-native-prebuilt") {
+            // without this second build the test fails on TeamCity, because the CC entry isn't reused because of:
```
antohaby, line 203: is it happening on older kotlin versions only?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/internal/PluginFeaturesService.kt
antohaby: Why not ProvidersFactory.gradleProperty() ? 

i think gradle suggests this as "recommended" https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#gradleProperty(java.lang.String)

antohaby: ah, I see above you're already using it. and this part is for unit tests. my bad :/

dokka-integration-tests/gradle/src/test/kotlin/KotlinJvmIT.kt
adam-enko: The tests run in a stable directory in the project's `build/` directory (for easier checking, and to catch caching issues.) This means if a test re-runs the tasks could be up-to-date.

dokka-integration-tests/gradle/src/test/kotlin/KotlinMultiplatformIT.kt
adam-enko: I think it's something that was introduced in newer versions, due to kotlin-native-prebuilt changes.

PR #4470, opened 2026-03-02
Remove "maven.pkg.jetbrains.space/public/p/compose/dev"

whyoleg:
It doesn't contain "org.jetbrains.compose.compiler:compiler:1.5.15" anymore, so we can't really test compose with KGP 1.9

This fixed the currently failing IT on master ([example](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Dokka_IntegrationTests/5906087)).


adam-enko:
You'll need to update the versions in `TestedVersionsSourceTest`

Issue #3385, opened 2023-11-29
[K2] KDoc references that goes after markdown blocks don't have links


atyrin:
```Kotlin
class JavaNetCookieJar

interface CookieJar {
    /**
     * Markdown syntax ```code```
     * This references doesn't work: [System.currentTimeMillis] and [JavaNetCookieJar].
     */
    fun saveFromResponse(url: String)
}
```

In K1, references are resolved:
<img width="1024" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/a6bc000c-9c0b-4d2d-bcf2-51234f217504">

---

In K2 are not:

<img width="1030" alt="image" src="https://github.com/Kotlin/dokka/assets/7689049/64a08c75-6456-49dc-9abe-b1c060c99d06">

If I put a reference before the code block in the samples, it will be resolved. Also, if it the same as reference after, like:
```
/**
     * This references works: [System.currentTimeMillis] and [JavaNetCookieJar]  
     * Markdown syntax ```code```
     * This references doesn't work: [System.currentTimeMillis] and [JavaNetCookieJar].
     */
```
The both will be resolved.

**Installation**
- Dokka version: 1.9.20

Parent: https://github.com/Kotlin/dokka/issues/3328

atyrin:
One more case. Reference works in K1 and doesn't in K2

```Kotlin
interface Call

/**
 * @param text some description with reference.
 *     But with a few lines and indent [Call]
 */
fun protocol(text: String) {}
```

vmishenev:
**Note** 
Both cases do not work in IDE K1:

![image](https://github.com/Kotlin/dokka/assets/9636937/37cccaea-7f36-4ec0-a7b8-42f135624444)
![image](https://github.com/Kotlin/dokka/assets/9636937/dfc2d2a1-94b7-4555-96e1-a92d535cebdf)

They lead to an internal error in IDE:
![image](https://github.com/Kotlin/dokka/assets/9636937/77f994e4-ce63-45bc-92e8-698144c07eb1)

vmishenev:
There is a problem in a parser of compiler (PSI).
![image](https://github.com/Kotlin/dokka/assets/9636937/1590bc46-15a9-4c87-92fc-4d33ed2cee79)

There is a workaround on Dokka's side: to resolve KDoc links like in K1.

vmishenev:
[...]

PR #4471, opened 2026-03-04
#3385 kdoc links resolution tests

AbdullinAM:


dokka-subprojects/analysis-kotlin-descriptors-compiler/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/descriptors/compiler/configuration/AnalysisEnvironment.kt
```diff
@@ -97,10 +99,14 @@ public class AnalysisEnvironment(
...
     private val kLibService: KLibService,
 ) : Disposable {
+    @OptIn(CompilerConfiguration.Internals::class)
     private val configuration = CompilerConfiguration()
 
     init {
+        @OptIn(CompilerConfiguration.Internals::class)
+        configuration.registerExtensionStorage()
         configuration.put(CommonConfigurationKeys.MESSAGE_COLLECTOR_KEY, messageCollector)
+        configuration.put(CLIConfigurationKeys.DIAGNOSTICS_COLLECTOR, BaseDiagnosticsCollector.DoNothing)
```
AbdullinAM, line 109: I'm not sure if we want a proper diagnostics collector here, so I just added a default stub to make it work. If we need a proper collector --- I'll implement it

whyoleg: I think it's fine, as it's K1.

Issue #4472, opened 2026-03-09
Documenting Android projects with flavors / multiple variants


whyoleg:
## Current state

Dokka Gradle Plugin (v2) as of Dokka 2.2.0 supports documenting three kinds of Android projects, though the underlying implementation for documenting them is different:

- KMP projects with [Android Gradle Library Plugin for KMP](https://developer.android.com/kotlin/multiplatform/plugin) (Android-KMP plugin for short).
    - Android-KMP plugin wires its sources and compilations with Kotlin Gradle Plugin via all the same machinery that's already provided for all other targets and so supported out-of-the-box in Dokka Gradle Plugin without any additional configuration. Note: Android-KMP plugin has only a single variant.
- Android projects based on AGP8 and [deprecated](https://kotlinlang.org/docs/compatibility-guide-23.html#deprecate-kotlin-android-plugin-for-agp-versions-9-0-0-and-later) `kotlin-android` Gradle Plugin.
    - `kotlin-android` plugin wires sources and classpath from Android Gradle Plugin into `KotlinSourceSet` and `KotlinCompilation`, corresponding to Android variants. Dokka Gradle Plugin can then access those sources and classpath via KGP APIs.
    - The Android Gradle Plugin creates **multiple** source roots (and so multiple `AndroidSourceSet`s), which are then combined during a **single** compilation per variant to produce a single final binary (per variant). Some of the default source roots (each corresponding to a single `AndroidSourceSet`s) created are:
        - `src/main/kotlin`
        - `src/release/kotlin`
        - `src/debug/kotlin`
        - And additional source roots for each flavor, like:
            - `src/paid/kotlin`
            - `src/paidDebug/kotlin`
            - `src/paidRelease/kotlin`
    - Dokka Gradle Plugin iterates over `KotlnSourceSet`s and creates `DokkaSourceSet`s based on them. So, e.g., with the source roots from above, we will have **six** `DokkaSourceSet`s. There is no concept of a `dependsOn` relationship, as it's not a Kotlin Multiplatform project, and the Android Gradle Plugin doesn't have one either; it just [combines sources](https://developer.android.com/build/build-variants#sourceset-build) from all source sets during variant compilations. Still, as there is only one compilation (variant) for all those source sets, it's impossible to get a correct classpath for those `DokkaSourceSet`s.
    - Because of the specifics above, there are multiple different issues when working with Android flavors with Dokka:
        - https://github.com/Kotlin/dokka/issues/2882
[...]

Issue #4458, opened 2026-02-16
AGP9 with flavors leads to intersecting source sets issue


atyrin:
The issue is the same as #3701 but this time it looks as regression.

1. Configure an Android project with AGP9 and `id("com.android.library")` plugin
2. Configure productFlavors
3. Run dokka

Dokka will fail with
```
Execution failed for task ':dokkaGeneratePublicationHtml'.
> A failure occurred while executing org.jetbrains.dokka.gradle.workers.DokkaGeneratorWorker
   > Pre-generation validity check failed: Source sets 'freeRelease' and 'paidRelease' have the common source roots: /Users/Andrei.Tyrin/IdeaProjects/dokka-repros/src/main/java, /Users/Andrei.Tyrin/IdeaProjects/dokka-repros/src/main/kotlin, /Users/Andrei.Tyrin/IdeaProjects/dokka-repros/src/release/java, /Users/Andrei.Tyrin/IdeaProjects/dokka-repros/src/release/kotlin. Every Kotlin source file should belong to only one source set (module). 
     Also, please consider reporting your user case: https://github.com/Kotlin/dokka/issues/3701
```

If I downgrade the project to AGP 8.13.2 it works fine.

**Installation**
- Dokka version: 2.2.0-Beta

[dokka-repros.zip](https://github.com/user-attachments/files/25339590/dokka-repros.zip)

radixdev:
I appear to have resolved this via:

```
dokka {
    // Set module name displayed in the final output.
    // Changing this string changes all generated URLs
    moduleName.set("...")

    dokkaSourceSets {
        // Suppress all auto-generated source sets first.
        // Dokka 2.2 + AGP 9 creates an androidJvm source set that shares source roots
        // with per-variant sets, causing a pre-generation validation failure.
        // See https://github.com/Kotlin/dokka/issues/4458
        configureEach {
            suppress.set(true)
        }

        named("release") {
            suppress.set(false)
            // ...
        }
    }
}
```

whyoleg:
For AGP8, we create Dokka source sets based on Android **source sets**. Each source set will contain only the current Android source set source roots; e.g., `release` contains only `src/release/*` sources, `paidRelease` contains only `src/paidRelease/*`.
For AGP9, we create Dokka source sets per Android **variant**. Each source set will contain source roots from all source sets related to that flavor: e.g., `paidRelease` contains `src/release/*`, `src/paidRelease/*`, and other source sets.

[...]

PR #4473, opened 2026-03-09
Detect and handle intersected source roots in Android multi-variant projects

whyoleg:
Workaround for #4458
Related to #4472

Current error output will look like this:
```
* What went wrong:
Execution failed for task ':dokkaGeneratePublicationHtml'.
> Dokka cannot generate documentation for Android projects with multiple enabled variants that have common source roots.
  Please suppress all variants except ONE via the following configuration:
  dokka {
          suppress.set(name != VARIANT_NAME)
      }
  }
  Where VARIANT_NAME could be one of the following: [freeRelease, paidRelease]
  For more information regarding the reasoning behind this restriction: https://github.com/Kotlin/dokka/issues/4472
  Common source roots:
    'freeRelease' and 'paidRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project-null/android-project-null/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project-null/android-project-null/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project-null/android-project-null/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project-null/android-project-null/src/release/kotlin
```

<details><summary>In case there are multiple dimensions</summary>
<p>

```
* What went wrong:
Execution failed for task ':dokkaGeneratePublicationHtml'.
> Dokka cannot generate documentation for Android projects with multiple enabled variants that have common source roots.
  Please suppress all variants except ONE via the following configuration:
  dokka {
          suppress.set(name != VARIANT_NAME)
      }
  }
  Where VARIANT_NAME could be one of the following: [freeHardRelease, freeLowRelease, paidHardRelease, paidLowRelease]
  For more information regarding the reasoning behind this restriction: https://github.com/Kotlin/dokka/issues/4472
  Common source roots:
    'freeHardRelease' and 'freeLowRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/free/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/free/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/kotlin
    'freeHardRelease' and 'paidHardRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/hard/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/hard/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/kotlin
    'freeHardRelease' and 'paidLowRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/kotlin
    'freeLowRelease' and 'paidHardRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/kotlin
    'freeLowRelease' and 'paidLowRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/low/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/low/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/kotlin
    'paidHardRelease' and 'paidLowRelease' have common source roots:
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/main/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/paid/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/paid/kotlin
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/java
      - /Users/Oleg.Yukhnevich/Projects/kotlin/dokka/dokka-runners/dokka-gradle-plugin/build/test-temp-dir/functional-tests/android-project/android-project/src/release/kotlin
```

</p>
</details> 

---

I've created a functional test for this to easily test both cases with an error, and when the variant is selected. Though it requires AGP9, it also requires Gradle 9.1+, which in turn requires JDK 17...
Luckily, we do run our tests on JDK 17, at least on GitHub Actions.
I don't really want to use IT here (as then it will be a bit harder to test both behaviors), but maybe it makes sense, WDYT?

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -283,7 +283,6 @@ constructor(
         }
 
         target.tasks.withType<DokkaGenerateTask>().configureEach {
-            projectDir.set(target.layout.projectDirectory)
             cacheDirectory.convention(dokkaExtension.dokkaCacheDirectory)
             workerLogFile.convention(temporaryDir.resolve("dokka-worker.log"))
             dokkaConfigurationJsonFile.convention(temporaryDir.resolve("dokka-configuration.json"))
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
@@ -13,7 +13,6 @@ import org.jetbrains.dokka.*
 import org.jetbrains.dokka.gradle.DokkaBasePlugin
 import org.jetbrains.dokka.gradle.engine.parameters.DokkaGeneratorParametersSpec
 import org.jetbrains.dokka.gradle.engine.parameters.DokkaModuleDescriptionKxs
-import org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec
 import org.jetbrains.dokka.gradle.engine.plugins.DokkaPluginParametersBaseSpec
 import org.jetbrains.dokka.gradle.formats.DokkaHtmlPlugin.Companion.extractDokkaPluginMarkers
 import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
@@ -39,10 +38,7 @@ internal class DokkaParametersBuilder(
         outputDirectory: File,
         cacheDirectory: File? = null,
         moduleDescriptorDirs: Iterable<File>,
-        projectDir: File
     ): DokkaConfiguration {
-        verifySourceRootsIntersectionsBasedOnAndroidVariants(spec.dokkaSourceSets, projectDir)
-
         val moduleName = spec.moduleName.get()
         val moduleVersion = spec.moduleVersion.orNull?.takeIf { it != Project.DEFAULT_VERSION }
         val offlineMode = spec.offlineMode.get()
@@ -154,104 +150,5 @@ internal class DokkaParametersBuilder(
             serializationFormat = DokkaConfiguration.SerializationFormat.JSON,
             values = spec.jsonEncode(),
         )
-    }
-
-    /**
-     * Throws an actionable error if the project with multiple Android variants has intersected source roots.
-     *
-     * See [org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec.basedOnAndroidVariant]
-     * for more information on why we need it.
-     */
-    private fun verifySourceRootsIntersectionsBasedOnAndroidVariants(
-        sourceSets: Set<DokkaSourceSetSpec>,
-        projectDir: File
-    ) {
-        // no check in case there is only single source-set
-        if (sourceSets.size <= 1) return
-
-        // check only source sets based on Android variants;
-        // everything else should be handled by the default checker in analysis
-        val sourceRootsIntersections = detectSourceRootsIntersections(
-            sourceSets.filter { it.basedOnAndroidVariant.getOrElse(false) }
-        )
-
-        if (sourceRootsIntersections.isEmpty()) return
-
-        val intersectedSourceSets = mutableSetOf<String>().apply {
-            sourceRootsIntersections.forEach { (s1Name, s2Name, _) ->
-                add(s1Name)
-                add(s2Name)
-            }
-        }
-
-        error(
-            """
-            |Dokka cannot generate documentation for Android projects with multiple enabled variants that have common source roots.
-            |Please suppress all variants except ONE via the following configuration:
-            |dokka {
-            |    dokkaSourceSets.configureEach {
-            |        suppress.set(name != VARIANT_NAME)
-            |    }
-            |}
-            |Where VARIANT_NAME could be one of the following: $intersectedSourceSets
-            |For more information regarding the reasoning behind this restriction: https://github.com/Kotlin/dokka/issues/4472
-            |Common source roots:
-            |${
-                sourceRootsIntersections.joinToString("\n") { (s1Name, s2Name, intersectedPaths) ->
-                    "  '$s1Name' and '$s2Name' have common source roots (relative to project directory):\n${
-                        intersectedPaths.joinToString("\n") {
-                            "    - ${it.relativeTo(projectDir).invariantSeparatorsPath}"
-                        }
-                    }"
-                }
-            }""".trimMargin()
-        )
-    }
-
-    private data class SourceRootIntersections(
-        val s1Name: String,
-        val s2Name: String,
-        val intersectedPaths: Set<File>,
-    )
-
-    // based on SourceRootIndependentChecker
-    private fun detectSourceRootsIntersections(sourceSets: List<DokkaSourceSetSpec>): List<SourceRootIntersections> {
-        fun Set<File>.normalize() = mapTo(mutableSetOf(), File::normalize)
-        fun intersectOfNormalizedPaths(normalizedPaths: Set<File>, normalizedPaths2: Set<File>): Set<File> {
-            val result = mutableSetOf<File>()
-            for (p1 in normalizedPaths) {
-                for (p2 in normalizedPaths2) {
-                    if (p1.startsWith(p2) || p2.startsWith(p1)) {
-                        result.add(p1)
-                        result.add(p2)
-                    }
-                }
-            }
-            return result
-        }
-
-        fun intersect(paths: Set<File>, paths2: Set<File>): Set<File> {
-            return intersectOfNormalizedPaths(paths.normalize(), paths2.normalize())
-        }
-
-        // no check in case there is only single source-set
-        if (sourceSets.size <= 1) return emptyList()
-
-        val intersections = mutableListOf<SourceRootIntersections>()
-        for (i in sourceSets.indices) {
-            for (j in i + 1 until sourceSets.size) {
-                val intersection = intersect(sourceSets[i].sourceRoots.files, sourceSets[j].sourceRoots.files)
-                if (intersection.isNotEmpty()) {
-                    intersections.add(
-                        SourceRootIntersections(
-                            s1Name = sourceSets[i].name,
-                            s2Name = sourceSets[j].name,
-                            intersectedPaths = intersection
-                        )
-                    )
-                }
-            }
-        }
-        return intersections
     }
 }
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt
@@ -19,6 +19,7 @@ import org.jetbrains.dokka.DokkaConfiguration.DokkaModuleDescription
 import org.jetbrains.dokka.DokkaConfigurationImpl
 import org.jetbrains.dokka.gradle.DokkaBasePlugin.Companion.jsonMapper
 import org.jetbrains.dokka.gradle.engine.parameters.DokkaGeneratorParametersSpec
+import org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec
 import org.jetbrains.dokka.gradle.engine.parameters.builders.DokkaParametersBuilder
 import org.jetbrains.dokka.gradle.internal.DokkaPluginParametersContainer
 import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
@@ -124,9 +125,8 @@ constructor(
     @InternalDokkaGradlePluginApi
     abstract val overrideJsonConfig: Property<String>
 
-    // needed to be able to relativize paths in error messages
-    @get:Internal
-    internal abstract val projectDir: DirectoryProperty
+    /** Only needed to be able to relativize paths in error messages. */
+    private val projectDir: File = project.projectDir
 
     @InternalDokkaGradlePluginApi
     enum class GeneratorMode {
@@ -139,6 +139,8 @@ constructor(
         generationType: GeneratorMode,
         outputDirectory: File,
     ) {
+        verifySourceRootsIntersectionsBasedOnAndroidVariants()
+
         val dokkaConfiguration =
             if (overrideJsonConfig.isPresent) {
                 logger.warn("w: [$path] Overriding DokkaConfiguration with overrideJsonConfig")
@@ -158,6 +160,7 @@ constructor(
         val workQueue = when (isolation) {
             is ClassLoaderIsolation ->
                 workers.classLoaderIsolation()
+
             is ProcessIsolation ->
                 workers.processIsolation {
                     forkOptions {
@@ -260,6 +263,57 @@ constructor(
     }
 
     /**
+     * Throws an actionable error if the project with multiple Android variants has intersected source roots.
+     *
+     * See [org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec.basedOnAndroidVariant]
+     * for more information on why we need it.
+     */
+    private fun verifySourceRootsIntersectionsBasedOnAndroidVariants() {
+        val sourceSets = generator.dokkaSourceSets
+
+        // no check in case there is only single source-set
+        if (sourceSets.size <= 1) return
+
+        // check only source sets based on Android variants;
+        // everything else should be handled by the default checker in analysis
+        val sourceRootsIntersections = detectSourceRootsIntersections(
+            sourceSets.filter { it.basedOnAndroidVariant.getOrElse(false) }
+        )
+
+        if (sourceRootsIntersections.isEmpty()) return
+
+        val intersectedSourceSets = mutableSetOf<String>().apply {
+            sourceRootsIntersections.forEach { (s1Name, s2Name, _) ->
+                add(s1Name)
+                add(s2Name)
+            }
+        }
+
+        error(
+            """
+            |Dokka cannot generate documentation for Android projects with multiple enabled variants that have common source roots.
+            |Please suppress all variants except ONE via the following configuration:
+            |dokka {
+            |    dokkaSourceSets.configureEach {
+            |        suppress.set(name != VARIANT_NAME)
+            |    }
+            |}
+            |Where VARIANT_NAME could be one of the following: $intersectedSourceSets
+            |For more information regarding the reasoning behind this restriction: https://github.com/Kotlin/dokka/issues/4472
+            |Common source roots:
+            |${
+                sourceRootsIntersections.joinToString("\n") { (s1Name, s2Name, intersectedPaths) ->
+                    "  '$s1Name' and '$s2Name' have common source roots (relative to project directory):\n${
+                        intersectedPaths.joinToString("\n") {
+                            "    - ${it.relativeTo(projectDir).invariantSeparatorsPath}"
+                        }
+                    }"
+                }
+            }""".trimMargin()
+        )
+    }
+
+    /**
      * Dump the [DokkaConfiguration] JSON to a file ([dokkaConfigurationJsonFile]) for debugging
      * purposes.
      */
@@ -298,7 +352,54 @@ constructor(
             outputDirectory = outputDirectory,
             moduleDescriptorDirs = moduleOutputDirectories,
             cacheDirectory = cacheDirectory.asFile.orNull,
-            projectDir = projectDir.asFile.get(),
         )
     }
+}
+
+
+private data class SourceRootIntersections(
+    val s1Name: String,
+    val s2Name: String,
+    val intersectedPaths: Set<File>,
+)
+
+// based on SourceRootIndependentChecker
+private fun detectSourceRootsIntersections(sourceSets: List<DokkaSourceSetSpec>): List<SourceRootIntersections> {
+    fun Set<File>.normalize() = mapTo(mutableSetOf(), File::normalize)
+    fun intersectOfNormalizedPaths(normalizedPaths: Set<File>, normalizedPaths2: Set<File>): Set<File> {
+        val result = mutableSetOf<File>()
+        for (p1 in normalizedPaths) {
+            for (p2 in normalizedPaths2) {
+                if (p1.startsWith(p2) || p2.startsWith(p1)) {
+                    result.add(p1)
+                    result.add(p2)
+                }
+            }
+        }
+        return result
+    }
+
+    fun intersect(paths: Set<File>, paths2: Set<File>): Set<File> {
+        return intersectOfNormalizedPaths(paths.normalize(), paths2.normalize())
+    }
+
+    // no check in case there is only single source-set
+    if (sourceSets.size <= 1) return emptyList()
+
+    val intersections = mutableListOf<SourceRootIntersections>()
+    for (i in sourceSets.indices) {
+        for (j in i + 1 until sourceSets.size) {
+            val intersection = intersect(sourceSets[i].sourceRoots.files, sourceSets[j].sourceRoots.files)
+            if (intersection.isNotEmpty()) {
+                intersections.add(
+                    SourceRootIntersections(
+                        s1Name = sourceSets[i].name,
+                        s2Name = sourceSets[j].name,
+                        intersectedPaths = intersection
+                    )
+                )
+            }
+        }
+    }
+    return intersections
 }


dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
```diff
@@ -455,6 +455,8 @@ constructor(
             }
         }
 
+    internal val isBasedOnAndroidVariant: Property<Boolean> = objects.property(Boolean::class.java)
```
whyoleg: @adam-enko, I wasn't really able to mark this property with an `@Input` annotation for some reason.
I get this, if I annotate with `@Input`:
```
* Exception is:
org.gradle.internal.execution.WorkValidationException: A problem was found with the configuration of task ':dokkaGeneratePublicationHtml' (type 'DokkaGeneratePublicationTask').
  - Type 'org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec' field 'isBasedOnAndroidVariant' without corresponding getter has been annotated with @Input.
    
    Reason: Annotations on fields are only used if there's a corresponding getter for the field.
    
    Possible solutions:
      1. Add a getter for field 'isBasedOnAndroidVariant'.
      2. Remove the annotations on 'isBasedOnAndroidVariant'.
```
And this, in case of `@get:Input`:
```
* What went wrong:
A problem was found with the configuration of task ':dokkaGeneratePublicationHtml' (type 'DokkaGeneratePublicationTask').
  - Type 'org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec' function 'isBasedOnAndroidVariant$dokka_gradle_plugin()' should not be annotated with: @Input.
    
    Reason: Input/Output annotations are ignored if they are placed on something else than a getter.
    
    Possible solutions:
      1. Remove the annotations.
      2. Rename the method.
```
In case I make it `abstract` (instead of `objects.property`), I get this:
```
* What went wrong:
Could not determine the dependencies of task ':dokkaGeneratePublicationHtml'.
> Could not create domain object 'paidDebug' (DokkaSourceSetSpec) in DokkaSourceSetSpec container
   > Could not create an instance of type org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec.
      > Could not generate a decorated class for type DokkaSourceSetSpec.
         > Cannot have abstract method DokkaSourceSetSpec.isBasedOnAndroidVariant$dokka_gradle_plugin(): Property<Boolean>.
```

 Do you know what I can do here to mark it as an input?

adam-enko: Could be caused by a weird Gradle issue where boolean properties aren't allowed to start with `is`?
https://docs.gradle.org/8.14/userguide/upgrading_version_8.html#groovy_boolean_properties

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
```diff
@@ -290,19 +304,93 @@ constructor(
             publicationEnabled.convention(true)
             onlyIf("publication must be enabled") { publicationEnabled.getOrElse(true) }
 
-            generator.dokkaSourceSets.addAllLater(
-                providers.provider {
-                    // exclude suppressed source sets as early as possible, to avoid unnecessary dependency resolution
-                    dokkaExtension.dokkaSourceSets.filterNot { it.suppress.get() }
+            generator.dokkaSourceSets.addAllLater(enabledSourceSets)
+
+            doFirst {
```
adam-enko: I think it'd make more sense to move this into a function inside `DokkaBaseTask`. `doFirst {}` blocks give me a 'hacky' feeling :) wdyt?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
```diff
@@ -290,19 +304,93 @@ constructor(
...
     }
 
+    private data class SourceRootIntersections(
+        val s1Name: String,
+        val s2Name: String,
+        val intersectedPaths: Set<File>,
+    )
+
+    // based on SourceRootIndependentChecker
+    private fun detectSourceRootsIntersections(sourceSets: List<DokkaSourceSetSpec>): List<SourceRootIntersections> {
```
adam-enko: Does this check, or the one inside `SourceRootIndependentChecker`, ignore empty directories? 

DGP could automatically ignore empty dirs, so if a user has flavours, but no specific code, Dokka could still work?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
whyoleg: Nice, thanks!
Changing it to `basedOnAndroidVariant` fixed all the problems!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: Yeah, agree!
For some reason, I thought that I would have no access to the `basedOnAndroidVariant` flag there...
But, I think, that I found an even better place for it: `DokkaParametersBuilder`.
I think it's better to do it here, since it feels like the right place for additional validation logic that already exists there.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: None of them will probably ignore empty directories. Also, I don't know how AA will work in this case, as we still provide those directories to them.

> DGP could automatically ignore empty dirs, so if a user has flavours, but no specific code, Dokka could still work?

But, it's not the case here? The problem will be If user has sources in `main` or `release` - and I think, that this is the most popular placement. So, if we filter empty dirs, we will really filter out only `paid`/`paidRelease` and so on.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/AndroidAdapter.kt
```diff
@@ -157,6 +163,10 @@ private interface AndroidExtensionWrapper {
         sourceSetName: String
     ): Provider<FileCollection>
 
+    fun isBasedOnAndroidVariant(
```
adam-enko, line 171: Could you add some brief, dev focused kdoc here please?

I think it'd be good to move the comment "Required for the improved..." from above to this kdoc.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/DokkaSourceSetSpec.kt
```diff
@@ -455,6 +455,10 @@ constructor(
             }
         }
 
+    @get:Input
+    @get:Optional
+    internal abstract val basedOnAndroidVariant: Property<Boolean>
```
adam-enko, line 470: Please add some brief kdoc.

dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/AndroidIntersectedSourceRootsErrorTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        "1.8" -> 8
+        else -> specVersion.toInt()
+    }
+    val hasJava17 = javaVersion >= 17
+
+    "when project has multiple flavors".config(enabled = hasJava17) {
+        val project = initAndroidProject(selectedVariantName = null)
+
+        project.runner
+            .withGradleVersion("9.1.0")
```
adam-enko: Could you extract this version out to a variable that explains why this specific version is required?

In the future when we update Dokka to use 9.1.0 we can just remove this explicit version.

I'm thinking of something like this:
https://github.com/Kotlin/dokka/blob/ef1fd1628b7a0480de9b810d4275f0468517e8ee/dokka-integration-tests/gradle/src/main/kotlin/org/jetbrains/dokka/it/gradle/gradleRunnerUtils.kt#L54-L68

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
```diff
@@ -151,4 +154,94 @@ internal class DokkaParametersBuilder(
             values = spec.jsonEncode(),
         )
     }
+
+    private fun verifySourceRootsIntersectionsBasedOnAndroidVariants(sourceSets: Set<DokkaSourceSetSpec>) {
```
adam-enko:
```suggestion
    /**
     * Throws an actionable error if AGP project has intersected source roots.
     */
    private fun verifySourceRootsIntersectionsBasedOnAndroidVariants(sourceSets: Set<DokkaSourceSetSpec>) {
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
```diff
@@ -151,4 +154,94 @@ internal class DokkaParametersBuilder(
...
+            |    }
+            |}
+            |Where VARIANT_NAME could be one of the following: $intersectedSourceSets
+            |For more information regarding the reasoning behind this restriction: https://github.com/Kotlin/dokka/issues/4472
+            |Common source roots:
+            |${
+                sourceRootsIntersections.joinToString("\n") { (s1Name, s2Name, intersectedPaths) ->
+                    "  '$s1Name' and '$s2Name' have common source roots:\n${
+                        intersectedPaths.joinToString("\n") {
+                            "    - ${it.absolutePath}"
```
adam-enko: nitpick: it might be nicer to log paths relative to the root project dir? It'd be easier to see the difference between the paths if the common parent paths were trimmed.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
adam-enko: > Also, I don't know how AA will work in this case, as we still provide those directories to them.

Yeah, I mean that DGP just removes empty dirs so AA never sees the intersecting src dirs.

> But, it's not the case here? The problem will be If user has sources in main or release - and I think, that this is the most popular placement. So, if we filter empty dirs, we will really filter out only paid/paidRelease and so on.

Agreed, it won't help in all cases. I just thought maybe it'd help a little bit.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
adam-enko: I was actually thinking it should be next to the other validation functions in `DokkaGenerateTask#checkModulePathsAreInsidePublicationDir`

https://github.com/Kotlin/dokka/blob/523252b84b30836fdee0ad877261b66e3ea5b601/dokka-runners/dokka-gradle-plugin/src/main/kotlin/tasks/DokkaGenerateTask.kt#L187

I'd really like to keep `DokkaParametersBuilder` as just 'DGP DSL to Dokka parameters' translator. 

wdyt?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
adam-enko: actually, it'd be really nice to this part of the message be tested, but that's tough at the moment because of the absolute paths the different Windows/Unix path separators.

It'd be nice to use both relativised & with invariant separators paths.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: I was thinking about it, but we need access to `basedOnAndroidVariant`, which is not present in `DokkaConfiguration` (and shouldn't be), only in `DokkaSourceSetSpec`.
My thinking is the opposite, to be honest: I do think we need to move the current checks into `DokkaParametersBuilder`, since it already contains similar checks, like in `buildModuleDescriptors`.
IMO, it's more correct to perform all validations for `DokkaConfiguration` during its construction in 'DGP DSL to Dokka parameters' rather than outside of it.
This way, all mapping and validation will be in `DokkaParametersBuilder`, and execution will be in `DokkaGenerateTask`.
I think it's just a matter of perspective?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: > Yeah, I mean that DGP just removes empty dirs so AA never sees the intersecting src dirs.

I think that we are going in circles here. :)
Just to remind: https://github.com/Kotlin/dokka/pull/4412/changes#r2681766608 - we removed filtering for non-existing dirs because it could easily hide issues (and it hid), the same will happen if we filter empty dirs.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
whyoleg: Nice idea! Do you know if there is a better way to provide a project directory than just passing it as an input to the task? (https://github.com/Kotlin/dokka/pull/4473/changes/56d2bf963ef8f5e2d4fc1fa5ed125c1bbfb29a44)
I know the `project` property is available on `Task`, but I'm afraid to touch it because of the CC. :)

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
adam-enko: Ahh okay, got it. Thanks for the reminder.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
adam-enko: I'd just create a property in the task. 

```
/**
 * Just used for pretty logging.
 */
private val projectDir: File = ...
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
adam-enko: > I was thinking about it, but we need access to basedOnAndroidVariant, which is not present in DokkaConfiguration (and shouldn't be), only in DokkaSourceSetSpec.

ah right, yeah, I see!

>  My thinking is the opposite, to be honest: I do think we need to move the current checks into DokkaParametersBuilder, since it already contains similar checks, like in buildModuleDescriptors.

The checks in `DokkaParametersBuilder` are more like 'last resort' checks than validations of user input. I agree it's a subtle difference though.

adam-enko: I really like keeping the validation check in the task, like this:

```kotlin
/*
 */
package org.jetbrains.dokka.gradle.tasks

import kotlinx.serialization.json.JsonElement
import org.gradle.api.file.ArchiveOperations
import org.gradle.api.file.ConfigurableFileCollection
import org.gradle.api.file.DirectoryProperty
import org.gradle.api.file.RegularFileProperty
import org.gradle.api.model.ObjectFactory
import org.gradle.api.provider.Property
import org.gradle.api.tasks.*
import org.gradle.kotlin.dsl.newInstance
import org.gradle.kotlin.dsl.submit
import org.gradle.workers.WorkerExecutor
import org.jetbrains.dokka.DokkaConfiguration
import org.jetbrains.dokka.DokkaConfiguration.DokkaModuleDescription
import org.jetbrains.dokka.DokkaConfigurationImpl
import org.jetbrains.dokka.gradle.DokkaBasePlugin.Companion.jsonMapper
import org.jetbrains.dokka.gradle.engine.parameters.DokkaGeneratorParametersSpec
import org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec
import org.jetbrains.dokka.gradle.engine.parameters.builders.DokkaParametersBuilder
import org.jetbrains.dokka.gradle.internal.DokkaPluginParametersContainer
import org.jetbrains.dokka.gradle.internal.InternalDokkaGradlePluginApi
import org.jetbrains.dokka.gradle.workers.ClassLoaderIsolation
import org.jetbrains.dokka.gradle.workers.DokkaGeneratorWorker
import org.jetbrains.dokka.gradle.workers.ProcessIsolation
import org.jetbrains.dokka.gradle.workers.WorkerIsolation
import org.jetbrains.dokka.toPrettyJsonString
import java.io.File
import javax.inject.Inject

/**
 * Base task for executing Dokka Generator, producing documentation.
 *
 * The Dokka Plugins added to the generator classpath determine the type of documentation generated.
 */
@CacheableTask
abstract class DokkaGenerateTask
@InternalDokkaGradlePluginApi
@Inject
constructor(
    objects: ObjectFactory,
    archives: ArchiveOperations,
    private val workers: WorkerExecutor,

    /**
     * Configurations for Dokka Generator Plugins. Must be provided from
     * [org.jetbrains.dokka.gradle.formats.DokkaPublication.pluginsConfiguration].
     */
    pluginsConfiguration: DokkaPluginParametersContainer,
) : DokkaBaseTask() {

    init {
        // Hide the 'generate module' and 'generate publication' tasks from the task list,
        // because they are lower level and potentially confusing
        // (for most users it's not obvious what the difference between a 'module' and 'publication' is).
        // For general generation usage the lifecycle tasks should be used instead.
        setGroup(null)
    }

    private val dokkaParametersBuilder = DokkaParametersBuilder(archives)

    /**
     * Directory containing the generation result. The content and structure depends on whether
     * the task generates a Dokka Module or a Dokka Publication.
     */
    @get:OutputDirectory
    abstract val outputDirectory: DirectoryProperty

    /**
     * Classpath required to run Dokka Generator.
     *
     * Contains the Dokka Generator, Dokka plugins, and any transitive dependencies.
     */
    @get:Classpath
    abstract val runtimeClasspath: ConfigurableFileCollection

    /** @see org.jetbrains.dokka.gradle.formats.DokkaPublication.cacheRoot */
    @get:LocalState
    abstract val cacheDirectory: DirectoryProperty

    /** @see org.jetbrains.dokka.gradle.formats.DokkaPublication.enabled */
    @get:Input
    abstract val publicationEnabled: Property<Boolean>

    @get:Nested
    val generator: DokkaGeneratorParametersSpec = objects.newInstance(pluginsConfiguration)

    /**
     * Control how Dokka Gradle Plugin launches Dokka Generator.
     *
     * Defaults to [org.jetbrains.dokka.gradle.DokkaExtension.dokkaGeneratorIsolation].
     *
     * @see org.jetbrains.dokka.gradle.DokkaExtension.dokkaGeneratorIsolation
     * @see org.jetbrains.dokka.gradle.workers.ProcessIsolation
     */
    @get:Nested
    abstract val workerIsolation: Property<WorkerIsolation>

    /**
     * All [org.jetbrains.dokka.DokkaGenerator] logs will be saved to this file.
     * This can be used for debugging purposes.
     */
    @get:Internal
    abstract val workerLogFile: RegularFileProperty

    /**
     * The [DokkaConfiguration] by Dokka Generator can be saved to a file for debugging purposes.
     * To disable this behaviour use [Property.unsetConvention] to clear the default value.
     */
    @InternalDokkaGradlePluginApi
    @get:Internal
    abstract val dokkaConfigurationJsonFile: RegularFileProperty

    /**
     * Completely override the default Dokka configuration with JSON encoded
     * Dokka Configuration.
     *
     * This should only be used for local debugging.
     */
    @get:Input
    @get:Optional
    @InternalDokkaGradlePluginApi
    abstract val overrideJsonConfig: Property<String>

    /** Only needed to be able to relativize paths in error messages. */
    private val projectDir: File = project.projectDir

    @InternalDokkaGradlePluginApi
    enum class GeneratorMode {
        Module,
        Publication,
    }

    @InternalDokkaGradlePluginApi
    protected fun generateDocumentation(
        generationType: GeneratorMode,
        outputDirectory: File,
    ) {
        verifySourceRootsIntersectionsBasedOnAndroidVariants()

        val dokkaConfiguration =
            if (overrideJsonConfig.isPresent) {
                logger.warn("w: [$path] Overriding DokkaConfiguration with overrideJsonConfig")
                DokkaConfigurationImpl(overrideJsonConfig.get())
            } else {
                createDokkaConfiguration(generationType, outputDirectory)
            }

        logger.info("dokkaConfiguration: $dokkaConfiguration")
        verifyDokkaConfiguration(dokkaConfiguration)
        dumpDokkaConfigurationJson(dokkaConfiguration)

        logger.info("DokkaGeneratorWorker runtimeClasspath: ${runtimeClasspath.asPath}")

        val isolation = workerIsolation.get()
        logger.info("[$path] running with workerIsolation $isolation")
        val workQueue = when (isolation) {
            is ClassLoaderIsolation ->
                workers.classLoaderIsolation()

            is ProcessIsolation ->
                workers.processIsolation {
                    forkOptions {
                        isolation.defaultCharacterEncoding.orNull?.let(this::setDefaultCharacterEncoding)
                        isolation.debug.orNull?.let(this::setDebug)
                        isolation.enableAssertions.orNull?.let(this::setEnableAssertions)
                        isolation.maxHeapSize.orNull?.let(this::setMaxHeapSize)
                        isolation.minHeapSize.orNull?.let(this::setMinHeapSize)
                        isolation.jvmArgs.orNull?.filter { it.isNotBlank() }?.let(this::setJvmArgs)
                        isolation.systemProperties.orNull?.let(this::systemProperties)
                    }
                }
        }

        workQueue.submit(DokkaGeneratorWorker::class) {
            this.dokkaClasspath.from(runtimeClasspath)
            this.dokkaParameters.set(dokkaConfiguration)
            this.logFile.set(workerLogFile)
            this.taskPath.set(this@DokkaGenerateTask.path)
        }
    }

    /**
     * Run some helper checks to log warnings if the [DokkaConfiguration] looks misconfigured.
     */
    private fun verifyDokkaConfiguration(dokkaConfiguration: DokkaConfiguration) {
        checkModulePathsAreInsidePublicationDir(dokkaConfiguration)
        checkModulePathsAreDistinct(dokkaConfiguration)
    }

    private fun checkModulePathsAreInsidePublicationDir(
        dokkaConfiguration: DokkaConfiguration,
    ) {
        fun DokkaModuleDescription.resolvedOutputDir(): File =
            dokkaConfiguration.outputDir.resolve(relativePathToOutputDirectory).normalize()


        val modulesWithOutputDirOutsidePublicationDir = dokkaConfiguration.modules
            .filter { module ->
                val moduleOutputDir = module.resolvedOutputDir()
                !moduleOutputDir.startsWith(dokkaConfiguration.outputDir.normalize())
            }
        check(modulesWithOutputDirOutsidePublicationDir.isEmpty()) {
            val modules = modulesWithOutputDirOutsidePublicationDir
                .map { "${it.name} (modulePath: '${it.relativePathToOutputDirectory.invariantSeparatorsPath}')" }
                .sorted()
                .joinToString("\n") { "  - $it" }
            """
            |[$path] Found ${modulesWithOutputDirOutsidePublicationDir.size} modules with directories that are outside the Dokka output directory.
            |  All module output directories must be a subdirectory inside the Dokka output directory.
            |  Update the `modulePath` in these modules:
            |$modules
            |""".trimMargin()
        }


        val modulesWithOutputDirSameAsPublicationDir = dokkaConfiguration.modules
            .filter { module ->
                val moduleOutputDir = module.resolvedOutputDir()
                moduleOutputDir == dokkaConfiguration.outputDir.normalize()
            }
        check(modulesWithOutputDirSameAsPublicationDir.isEmpty()) {
            val modules = modulesWithOutputDirSameAsPublicationDir
                .map { "${it.name} (modulePath: '${it.relativePathToOutputDirectory.invariantSeparatorsPath}')" }
                .sorted()
                .joinToString("\n") { "  - $it" }
            """
            |[$path] Found ${modulesWithOutputDirSameAsPublicationDir.size} modules with output directories that resolve to the same directory as the Dokka output directory.
            |  All module output directories must be a subdirectory inside the Dokka output directory.
            |  Specify `modulePath` in these modules:
            |$modules
            |""".trimMargin()
        }
    }

    private fun checkModulePathsAreDistinct(
        dokkaConfiguration: DokkaConfiguration,
    ) {
        val modulesWithDuplicatePaths = dokkaConfiguration.modules
            .groupBy { it.relativePathToOutputDirectory.toString() }
            .filterValues { it.size > 1 }

        if (modulesWithDuplicatePaths.isNotEmpty()) {
            val modulePaths = modulesWithDuplicatePaths.entries
                .map { (path, modules) ->
                    "${modules.joinToString { "'${it.name}'" }} have modulePath '$path'"
                }
                .sorted()
                .joinToString("\n") { "  - $it" }
            logger.warn(
                """
                |[$path] Duplicate `modulePath`s in Dokka Generator parameters.
                |  modulePaths must be distinct for each module, as they are used to determine the output
                |  directory. Duplicates mean Dokka Generator may overwrite one module with another.
                |$modulePaths
                |
                """.trimMargin()
            )
        }
    }

    /**
     * Throws an actionable error if the project with multiple Android variants has intersected source roots.
     *
     * See [org.jetbrains.dokka.gradle.engine.parameters.DokkaSourceSetSpec.basedOnAndroidVariant]
     * for more information on why we need it.
     */
    private fun verifySourceRootsIntersectionsBasedOnAndroidVariants() {
        val sourceSets = generator.dokkaSourceSets

        // no check in case there is only single source-set
        if (sourceSets.size <= 1) return

        // check only source sets based on Android variants;
        // everything else should be handled by the default checker in analysis
        val sourceRootsIntersections = detectSourceRootsIntersections(
            sourceSets.filter { it.basedOnAndroidVariant.getOrElse(false) }
        )

        if (sourceRootsIntersections.isEmpty()) return

        val intersectedSourceSets = mutableSetOf<String>().apply {
            sourceRootsIntersections.forEach { (s1Name, s2Name, _) ->
                add(s1Name)
                add(s2Name)
            }
        }

        error(
            """
            |Dokka cannot generate documentation for Android projects with multiple enabled variants that have common source roots.
            |Please suppress all variants except ONE via the following configuration:
            |dokka {
            |    dokkaSourceSets.configureEach {
            |        suppress.set(name != VARIANT_NAME)
            |    }
            |}
            |Where VARIANT_NAME could be one of the following: $intersectedSourceSets
            |For more information regarding the reasoning behind this restriction: https://github.com/Kotlin/dokka/issues/4472
            |Common source roots:
            |${
                sourceRootsIntersections.joinToString("\n") { (s1Name, s2Name, intersectedPaths) ->
                    "  '$s1Name' and '$s2Name' have common source roots (relative to project directory):\n${
                        intersectedPaths.joinToString("\n") {
                            "    - ${it.relativeTo(projectDir).invariantSeparatorsPath}"
                        }
                    }"
                }
            }""".trimMargin()
        )
    }

    /**
     * Dump the [DokkaConfiguration] JSON to a file ([dokkaConfigurationJsonFile]) for debugging
     * purposes.
     */
    private fun dumpDokkaConfigurationJson(
        dokkaConfiguration: DokkaConfiguration,
    ) {
        val destFile = dokkaConfigurationJsonFile.asFile.orNull ?: return
        destFile.parentFile.mkdirs()
        destFile.createNewFile()

        val compactJson = dokkaConfiguration.toPrettyJsonString()
        val json = jsonMapper.decodeFromString(JsonElement.serializer(), compactJson)
        val prettyJson = jsonMapper.encodeToString(JsonElement.serializer(), json)

        destFile.writeText(prettyJson)

        logger.info("[$path] Dokka Generator configuration JSON: ${destFile.toURI()}")
    }

    private fun createDokkaConfiguration(
        generationType: GeneratorMode,
        outputDirectory: File,
    ): DokkaConfiguration {

        val delayTemplateSubstitution = when (generationType) {
            GeneratorMode.Module -> true
            GeneratorMode.Publication -> false
        }

        val moduleOutputDirectories = generator.moduleOutputDirectories.toList()
        logger.info("[$path] got ${moduleOutputDirectories.size} moduleOutputDirectories: $moduleOutputDirectories")

        return dokkaParametersBuilder.build(
            spec = generator,
            delayTemplateSubstitution = delayTemplateSubstitution,
            outputDirectory = outputDirectory,
            moduleDescriptorDirs = moduleOutputDirectories,
            cacheDirectory = cacheDirectory.asFile.orNull,
        )
    }
}


private data class SourceRootIntersections(
    val s1Name: String,
    val s2Name: String,
    val intersectedPaths: Set<File>,
)

// based on SourceRootIndependentChecker
private fun detectSourceRootsIntersections(sourceSets: List<DokkaSourceSetSpec>): List<SourceRootIntersections> {
    fun Set<File>.normalize() = mapTo(mutableSetOf(), File::normalize)
    fun intersectOfNormalizedPaths(normalizedPaths: Set<File>, normalizedPaths2: Set<File>): Set<File> {
        val result = mutableSetOf<File>()
        for (p1 in normalizedPaths) {
            for (p2 in normalizedPaths2) {
                if (p1.startsWith(p2) || p2.startsWith(p1)) {
                    result.add(p1)
                    result.add(p2)
                }
            }
        }
        return result
    }

    fun intersect(paths: Set<File>, paths2: Set<File>): Set<File> {
        return intersectOfNormalizedPaths(paths.normalize(), paths2.normalize())
    }

    // no check in case there is only single source-set
    if (sourceSets.size <= 1) return emptyList()

    val intersections = mutableListOf<SourceRootIntersections>()
    for (i in sourceSets.indices) {
        for (j in i + 1 until sourceSets.size) {
            val intersection = intersect(sourceSets[i].sourceRoots.files, sourceSets[j].sourceRoots.files)
            if (intersection.isNotEmpty()) {
                intersections.add(
                    SourceRootIntersections(
                        s1Name = sourceSets[i].name,
                        s2Name = sourceSets[j].name,
                        intersectedPaths = intersection
                    )
                )
            }
        }
    }
    return intersections
}
```

dokka-runners/dokka-gradle-plugin/src/main/kotlin/engine/parameters/builders/DokkaParametersBuilder.kt
whyoleg: Nice, thanks!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: > The checks in DokkaParametersBuilder are more like 'last resort' checks than validations of user input. I agree it's a subtle difference though.

@adam-enko, yeah.. It's still really strange to me to have those three different validation patterns, which look very-very similar...
In any case, I've moved the verification to `DokkaGenerateTask`, as you suggested.

PR #4475, opened 2026-03-13
Move Enum templates into the analysis-kotlin-api project and update the `Enum.entries` KDoc

### LINKED ISSUE #4443: [K2] Synthetic documentation for Enum.entries/Enum.values/Enum.valueOf is missing
**Describe the bug**

[kotlinlang stdlib 2.1](https://kotlinlang.org/api/core/2.1/kotlin-stdlib/kotlin/-deprecation-level/) built by older Dokka with K1 - the declarations are documented (though, `name`/`ordinal` documentation is missing)

Image

[kotlinlang stdlib 2.2/2.3](https://kotlinlang.org/api/core/2.2/kotlin-stdlib/kotlin/-deprecation-level/) built by newer Dokka with K2 - the declarations **missing** documentation (but `name`/`ordinal` are documented now)

Image

I can't reproduce the issue locally with the kotlinx libraries and Dokka 2.2.0-Beta, so it looks like it's something kotlin stdlib-specific.

**Expected behaviour**

Enum.entries/Enum.values/Enum.valueOf should have synthetic documentation coming from [templates](https://github.com/Kotlin/dokka/tree/76417b9e167abbd7ba5ed24b3c0edc3a01451e89/dokka-subprojects/plugin-base/src/main/resources/dokka/docs/kdoc).

### LINKED ISSUE #4452: Refactor `EnumTemplatesTest` to use latest Kotlin release
> There's also `EnumTemplatesTest`, but it assess sources of the stdlib whose version is `2.0.21`. I'm not sure how useful that test is.

_Originally posted by @fzhinkin in https://github.com/Kotlin/dokka/issues/4450#issuecomment-3836527492_

---

Looks like we need to convert this test back to IT so we can run it with a custom Kotlin version, not the one used for building Dokka.

vmishenev:
Fixes #4443, #4452 (can still be actual: releasing Dokka is behind publishing the stdlib )

The stdlib uses DGPv1 that has an isolated classloader for `analysis-kotlin-*` artefacts, but the enum templates were located in the base plugin.  It seems they should be in `analysis-kotlin-*` artefacts.

Also, this PR
- enables a test for `Enum.entries`
- updates the `Enum.entries` KDoc
- downloads the latest stdlib sources #4452


dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/test/EnumTemplatesTest.kt
```diff
@@ -90,16 +88,31 @@ class EnumTemplatesTest {
...
+                { messagePrefix(message) + "Expected the char sequence to contain the substring.\nCharSequence <$charSequence>, substring <$other>, ignoreCase <$ignoreCase>." },
+                charSequence.contains(other, ignoreCase)
+            )
+        }
+
+        private fun loadResource(@Language("file-reference") path: String): String {
+            return EnumTemplatesTest::class.java.getResource(path)
+                ?.readText()
                 ?: error("Failed to load resource: $path")
-            return Paths.get(resource)
```
vmishenev, line 97: It no longer works because the resources are now bundled inside the `.jar`

dokka-subprojects/analysis-kotlin-api/src/test/kotlin/org/jetbrains/dokka/test/EnumTemplatesTest.kt
```diff
@@ -1,15 +1,15 @@
 /*
```
whyoleg, line 1: minor: the file has an empty package

whyoleg (approved):
Nice!

PR #4476, opened 2026-03-16
Add test for KDoc links to classpath

### LINKED ISSUE #4425: Enable new KDoc link resolution by default

vmishenev:


whyoleg (approved):
That's a nice way to test this!

whyoleg:
@vmishenev, should we merge this?

PR #4477, opened 2026-03-16
Print JavaDoc link location

### LINKED ISSUE #4465: Add detail to logging of unresolved javadoc links
The logging for [unresolved javadoc links](https://github.com/Kotlin/dokka/blob/e716676024d56cb6e1eed2855ce7ad7836ff99a9/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/PsiElementToHtmlConverter.kt#L176-L179) added in #4414 only includes the text of the unresolved tag, which makes it difficult to tell where the issue is coming from. It would be helpful if these messages included more detail like the containing element and ideally the filepath like what was added in #4430 for unresolved kdoc links.

vmishenev:
fixes #4465

### DIFF
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiUtil.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiUtil.kt
@@ -132,6 +132,7 @@ internal val PsiElement.kotlinFqNameProp: String?
     }
 
 /**
+ * Copy-paste from [org.jetbrains.dokka.analysis.kotlin.symbols.utils.getLocation]
  * @return a string in the format `file://file.kt:<line>:<column>`,
  * or `null` if [psiElement] is not from a source file
  */
--- a/dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
@@ -1149,8 +1149,7 @@ class SnippetTest : BaseAbstractTest() {
 
                 val warnMessage = logger.warnMessages.first()
 
-                val path = modules.first().sourceSets.first().sourceRoots.first().absolutePath
-                    .replace("\\", "/") // for Win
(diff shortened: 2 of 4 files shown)


dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiUtil.kt
```diff
@@ -130,3 +130,20 @@ internal val PsiElement.kotlinFqNameProp: String?
         is PsiQualifiedNamedElement -> element.qualifiedName
         else -> null
     }
+
+/**
+ * @return a string in the format `file://file.kt:<line>:<column>`,
+ * or `null` if [psiElement] is not from a source file
+ */
+internal fun getLocation(psiElement: PsiElement): String? {
```
whyoleg, line 139: It's the same function as in kotlin symbols, right? It would be nice to add cross-links between the two declarations, since they are copies.

dokka-subprojects/plugin-base/src/test/kotlin/parsers/javadoc/SnippetTest.kt
```diff
@@ -1149,10 +1149,13 @@ class SnippetTest : BaseAbstractTest() {
 
                 val warnMessage = logger.warnMessages.first()
 
+                val path = modules.first().sourceSets.first().sourceRoots.first().absolutePath
```
whyoleg: Minor: we could probably use [invariantSeparatorsPath](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io/invariant-separators-path.html) here to avoid manual `/` manipulations.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/util/PsiUtil.kt
vmishenev: I have added, but cross-links between the independent projects are unresolved.

whyoleg: Yeah, that's fine, at least we know that it's used somewhere :)

Issue #948, opened 2020-06-08
`tailrec` and `external` modifiers should not be documented


Kordyjan:
`tailrec` shouldn't be documented on any platform.
`external` should be documented only on JS.

sellmair:
This can be done/closed, right?

PR #4496, opened 2026-04-08
Add WasmWasi and WasmJs platforms

### LINKED ISSUE #4482: Set `WasmWasi` platform for Analysis API
Dokka does not differentiate between `wasmWasi` and `wasmJs`, [it just sets the platform to `unspecifiedWasmPlatform`](https://github.com/Kotlin/dokka/blob/master/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/KotlinAnalysis.kt#L33)
However, this affects analysis because the default imports are different:
https://github.com/JetBrains/kotlin/blob/master/analysis/analysis-api/testData/imports/defaultImportProvider/wasmJs.default.txt
https://github.com/JetBrains/kotlin/blob/master/analysis/analysis-api/testData/imports/defaultImportProvider/wasmWasi.default.txt

For example, it leads to 
```kt
// wasmWasi sourceset
@ExperimentalWasmInterop // is unresolved in Dokka
```

We can consider extending `org.jetbrains.dokka.Platform`.
~~Alternatively, we could treat wasmWasi as js. Does the DGP already do this?~~

vmishenev:
fixes #4482


dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/KmpSharedWasmTest.kt
```diff
@@ -34,6 +36,7 @@ class KmpSharedWasmTest : FunSpec({
                     "--stacktrace",
                 )
                 .forwardOutput()
+                .withDebug(true)
```
whyoleg: leftover?

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
```diff
@@ -439,16 +440,23 @@ private class KotlinCompilationDetailsBuilder(
...
 
+
+        val target = compilation.target
+
         return KotlinCompilationDetails(
-            target = compilation.target.name,
-            kotlinPlatform = KotlinPlatform.fromString(compilation.platformType.name),
+            target = target.name,
+            kotlinPlatform =
+                if (target is KotlinWasmTargetDsl && target.wasmTargetType != null) // Dokka needs to distinguish wasmWasi and wasmJs for analyzing
```
whyoleg, line 449: Minor: It feels a bit weird that we constructed the platform from a string even before those changes, but now it feels even more cumbersome to me.
@adam-enko, this looks like something is coming from Dokkatoo, so maybe you remember the reason for this `fromString` pattern rather than just switching based on `KotlinPlatformType` enum? Is it for compatibility in case the enum values change?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
```diff
@@ -43,7 +43,7 @@ public class KotlinSignatureProvider(
         ExtraModifiers.KotlinOnlyModifiers.External
     )
     private val platformSpecificModifiers: Map<ExtraModifiers, Set<Platform>> = mapOf(
-        ExtraModifiers.KotlinOnlyModifiers.External to setOf(Platform.js, Platform.wasm)
+        ExtraModifiers.KotlinOnlyModifiers.External to setOf(Platform.js, @Suppress("DEPRECATION") Platform.wasm, Platform.wasmWasi, Platform.wasmJs)
     )
```
whyoleg, line 47: While the changes in PR seem correct, I'm a bit confused about why we need this additional handling at all.
I mean, why don't we show `tailrec` and `external` by default, and still show `external` for JS/Wasm platforms but not on the JVM? Is it just because it doesn't affect the call sites in the JVM/Native but does affect JS/Wasm? Or what? WDYT?

Original issue: https://github.com/Kotlin/dokka/issues/948, sad it doesn't provide a reasoning...

dokka-subprojects/plugin-base/src/test/kotlin/content/annotations/SinceKotlinTest.kt
```diff
@@ -194,7 +194,7 @@ class SinceKotlinTest : AbstractRenderingTest() {
                 @Suppress("UNCHECKED_CAST") val funcs = module.children.single { it.name == "test" }
                     .children.filter { it.name == "ring" && it is DFunction } as List<DFunction>
                 with(funcs) {
-                    val sinceKotlin = mapOf(
+                    @Suppress("DEPRECATION") val sinceKotlin = mapOf(
```
whyoleg: Maybe instead of suppressing deprecation here, we could just update `configuration` to use `wasmJs`?

dokka-integration-tests/gradle/src/testExampleProjects/expectedData/exampleProjects/kotlin-multiplatform-example/html/dokka-multiplatform-example/org.kotlintestmpp.coroutines/async-with-delay.html
```diff
@@ -60,7 +60,7 @@
                         data-filter=":/macosMain">macos</button>
                 <button class="platform-tag platform-selector native-like" data-active="" aria-pressed="true"
                         data-filter=":/mingwMain">mingw</button>
-                <button class="platform-tag platform-selector wasm-like" data-active="" aria-pressed="true"
+                <button class="platform-tag platform-selector wasmJs-like" data-active="" aria-pressed="true"
```
whyoleg: Why did those files change?
Note that we don't have such CSS classes (`wasmJs-like`), only `wasm-like`
As I see, we handle those in `createPlatformTagBubbles`, so I would expect this to be `wasm-like`.

whyoleg (changes requested):
Nice! Looks good overall, but I'm concerned about why the `html` output changed.

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/signatures/KotlinSignatureProvider.kt
vmishenev: > why don't we show tailrec and external by default, and still show external for JS/Wasm platforms but not on the JVM? 

I do not know. To me, this handling seems redundant since modifiers are already source-set dependent. It feels like a bad hack. I could remove it, but I prefer not to touch it if it works.

whyoleg: No need to remove it, I was mostly curious about the reasons here.

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
vmishenev: I only kept the logic of Dokkatoo. It might affect compatibility with the KGP API, but I do not know exactly.
This question is for @adam-enko.

whyoleg (approved):
Good to go!

dokka-runners/dokka-gradle-plugin/src/main/kotlin/adapters/KotlinAdapter.kt
adam-enko: > maybe you remember the reason for this fromString pattern rather than just switching based on KotlinPlatformType enum? Is it for compatibility in case the enum values change?

Yes, I think that's why. I think initially Dokkatoo was compatible with Kotlin 1.6, and `wasm` was added in 1.6.20.
https://github.com/JetBrains/kotlin/commit/841f5a583e493c24a66825f54b187b99ebc50d25

Issue #4250, opened 2025-09-10
Define representation of synthetic Java properties


whyoleg:
Continuation of https://github.com/Kotlin/dokka/issues/3576

After several discussions, it's still not clear how to represent Java's getter/setter/field in cases of different visibility and inheritance (by both Kotlin and Java).
We should research and finally define the mental model for how to represent them consistently.
See linked issues for specific K1/K2 differences.

whyoleg:
Might be related: #2603

vmishenev:
When Java synthetic properties are formed by merging getters and setters, documentation from their accessors can be lost. We should also explore how to display such merged documentation.

Issue #3611, opened 2024-05-15
Do not show `open` for Java fields


vmishenev:
Dokka  renders the field from Java:
`int prop = 1;`
as 
![image](https://github.com/Kotlin/dokka/assets/9636937/b6b35e2b-ed64-4ef9-b8a6-810d48d55447)
It should not have the `open` modifier since fields can not be overridden in Kotlin.

PR #4502, opened 2026-05-06
Enhance Java symbols analysis 

### LINKED ISSUE #4463: Align Java unit tests with K2 analysis
There are tests that do not work under K2 analysis; they are marked with @OnlyJavaPsi.
They should be fixed to work with K2.

For example,

> There are ~60 unit tests left that are failing under new experimental flag. Most of them are related to problems with annotations, type conversion in AA (e.g. java's Object become kotlin's Any, int[] into IntArray, etc), modifiers.

Such tests can be duplicated and adapted for K2.

vmishenev:
Total of 5 unit tests are marked with `@OnlyJavaPsi`:
 - 3 test are caused by #4250 - having a non-private field with a getter
  ```java
public  int a
public  String getA()
  ```
 - 2 tests are due to mapping Java supertypes to Kotlin.
 
 
 TODO 
#4503


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -441,19 +460,20 @@ internal class DokkaSymbolVisitor(
         val javaFields = callables.filterIsInstance<KaJavaFieldSymbol>()
             .filterOutSyntheticJavaPropBackingField()
 
+
         fun List<KaNamedFunctionSymbol>.filterOutSyntheticJavaPropAccessors() = filterNot { fn ->
-            if ((fn.origin == KaSymbolOrigin.JAVA_SOURCE || fn.origin == KaSymbolOrigin.JAVA_LIBRARY) && fn.callableId != null)
-                syntheticJavaProperties.any { fn.callableId == it.javaGetterSymbol.callableId || fn.callableId == it.javaSetterSymbol?.callableId }
+            if ((fn.origin == KaSymbolOrigin.JAVA_SOURCE || fn.origin == KaSymbolOrigin.JAVA_LIBRARY) && fn.psi != null)
+                syntheticJavaProperties.any { fn.psi == it.javaGetterSymbol.psi || fn.psi == it.javaSetterSymbol?.psi }
```
vmishenev, line 467: `callableId` does not cover the case of overloads

vmishenev:
Also, I tested [gradle](https://github.com/gradle/gradle), and noticed some differences:
-  In  `valueOf`, the type `String` leads  to kotlinlang instead of Java references.
- (a lot) A getter and a setter collapse into a synthetic property without documentation:
image

image


Original: 
https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api/-ant-builder/get-lifecycle-log-level.html
and 
https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api/-ant-builder/set-lifecycle-log-level.html

whyoleg:
Q: What does `TODO https://github.com/Kotlin/dokka/issues/4503` mean in the description? Do we need to do something there right now?

whyoleg:
Regarding Gradle:
- `In valueOf, the type String leads to kotlinlang instead of Java references.` - sounds ok for me right now. Overall, I think we might really think about using Kotlin types in the future and just have a custom view/plugin in the stdlib for mapped types, to just say it there. AFAIK, for overload resolution, those types will be the same, so maybe we need to rethink what we discussed initially on DM regarding mapped types.
- `(a lot) A getter and a setter collapse into a synthetic property without documentation:` - that's really sad, and we need to fix this. What if there is only a `getter`? Do we also drop documentation in that case? Hm, but how could we have a documentation for the property from just getter/setter docs? Any thoughts on what the user will expect here?

vmishenev:
JIC Summary report from Claude:

image

dokka-subprojects/plugin-base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
```diff
@@ -388,15 +388,15 @@ class DefaultPsiToDocumentableTranslatorTest : BaseAbstractTest() {
...
+    @OnlyJavaPsi("synthetic: #4250 AA generate an extra synthetic property `a: String` with the getter String getA()")
     @Test
     fun `should preserve regular functions that are named like getters, but are not getters`() {
         testInline(
             """
             |/src/main/java/test/A.java
             |package test;
             |public class A {
-            |    private int a = 1;
+            |    public int a = 1;
```
whyoleg: What is the reason for this change?

dokka-subprojects/plugin-base/src/test/kotlin/transformers/ObviousAndInheritedFunctionsDocumentableFilterTest.kt
```diff
@@ -223,7 +222,7 @@ class ObviousAndInheritedFunctionsDocumentableFilterTest : BaseAbstractTest() {
         }
     }
 
-    @OnlyJavaPsi
+    @OnlyJavaPsi("type-mapping: AA's combinedMemberScope does not include all java.lang.Object methods (notify/notifyAll) - mapped types")
```
whyoleg, line 225: Should we create a single issue for mapped types?

dokka-subprojects/plugin-base/src/test/kotlin/superFields/PsiSuperFieldsTest.kt
```diff
@@ -130,14 +158,15 @@ class PsiSuperFieldsTest : BaseAbstractTest() {
         }
     }
 
-    @OnlyJavaPsi
+    @OnlyJavaPsi("synthetic: #4250 AA will have accessors, even if there is non in Kotlin")
```
whyoleg: Is it a Dokka or an AA issue? If AA - let's create an issue there? For me, this specific issue feels like a simple bug and is far less complex than other synthetic problems.

dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
```diff
@@ -750,7 +750,7 @@ class SignatureTest : BaseAbstractTest() {
         """.trimMargin()
     ) {
         renderedContent("root/example/-enum-class/index.html").firstSignature().matchIgnoringSpans(
-            "enum", A("EnumClass"),
+            "enum", A("EnumClass"), ":", A("Enum"), "<", A("EnumClass"), "?>" // TODO: `?` should it be here?
```
whyoleg: Let's drop the comment. Still, I'm curious if we can remove this `?` here.

vmishenev:
> Q: What does TODO https://github.com/Kotlin/dokka/issues/4503 mean in the description? Do we need to do something there right now?

Not really. I would say it is the next step.
Right now, it is still compatible with the current behavior. Otherwise, it would break the unit tests and make comparing the diffs more complicated.

>  What if there is only a getter? Do we also drop documentation in that case? 

Yes, it also drops documentation. 

> Hm, but how could we have a documentation for the property from just getter/setter docs? Any thoughts on what the user will expect here?

I think this is a part of https://github.com/Kotlin/dokka/issues/4250 
In a general case, I am not sure if we can seamless merge the documentation of a getter and a setter.

Maybe we should not filter out Java getters and setters, so we don’t break Gradle’s documentation at least.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/TypeReferenceFactory.kt
```diff
@@ -72,7 +72,7 @@ private fun KaSession.getTypeReferenceFromPossiblyRecursive(
         // For example, most Java types T in Kotlin are flexible (T..T?) or T!
         // see https://github.com/JetBrains/kotlin/blob/master/spec-docs/flexible-java-types.md (warn: it is not official spec)
         is KaFlexibleType -> getTypeReferenceFromPossiblyRecursive(
-            type.lowerBound,
+            type.upperBound,
```
whyoleg: Should we revert this, as we now parse Java types via PSI?

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/TypeTranslator.kt
```diff
@@ -133,13 +133,10 @@ internal class TypeTranslator(
...
-                toBoundFrom(type.upperBound, location),
-                toBoundFrom(type.lowerBound, location),
-                extra = PropertyContainer.withAll(
-                    getDokkaAnnotationsFrom(type)?.toSourceSetDependent()?.toAnnotations()
-                )
-            )
+            // Flexible types represent Java platform types (e.g., `String` in Java becomes `String..String?`).
+            // Ideally, we should use the upper bound (nullable) since Java types have no nullability guarantees,
+            // PSI module doesn't wrap Java types in Nullable !!!
+            is KaFlexibleType -> toBoundFrom(type.upperBound, location)
```
whyoleg: same here - should we revert this, or use `lowerBound`?

dokka-subprojects/plugin-base/src/test/kotlin/model/JavaTest.kt
```diff
@@ -51,146 +63,217 @@ class JavaTest : AbstractModelTest("/src/main/kotlin/java/Test.java", "java") {
...
+                    val generic =
+                        ((superclassType.typeConstructor.projections.single() as Invariance<*>).inner as Nullable).inner as GenericTypeConstructor
+                    assertEquals(generic.dri.classNames, "String")
+                }
             }
         }
     }
 
     @Test
+    //@OnlyJavaPsi("mapped-types: AA represents int[] as IntArray, not Array<PrimitiveJavaType>")
```
whyoleg: could be removed

dokka-subprojects/plugin-base/src/test/kotlin/model/JavaTest.kt
```diff
@@ -470,78 +609,87 @@ class JavaTest : AbstractModelTest("/src/main/kotlin/java/Test.java", "java") {
...
-                assertEquals(expectedParams.second, annotation?.params?.entries?.first()?.value)
+                    assertEquals(expectedDri, annotation?.dri)
+                    assertEquals(expectedParams.first, annotation?.params?.entries?.first()?.key)
+                    assertEquals(expectedParams.second, annotation?.params?.entries?.first()?.value)
+                }
             }
         }
     }
 
+    //@OnlyJavaSymbols("PSI doesn't mark types as nullable")
```
whyoleg: should be dropped

dokka-subprojects/plugin-base/src/test/kotlin/model/JavaTest.kt
```diff
@@ -51,146 +63,217 @@ class JavaTest : AbstractModelTest("/src/main/kotlin/java/Test.java", "java") {
...
+                    extra[ImplementedInterfaces]?.interfaces?.entries?.single()?.value?.map { it.dri.sureClassNames }
+                        ?.sorted() equals listOf("Highest", "Lower").sorted()
+                }
             }
         }
     }
 
-    @Test fun allImplementedInterfacesWithGenericsInJava() {
-        inlineModelTest(
+    @OnlyJavaSymbols("PSI doesn't mark types as nullable")
```
whyoleg: If I remember correctly, this was because we didn't use PSI for types; now we use it, so does this still hold?
(there is one more such case below in this file)

whyoleg (commented):
Nice! It looks like we really have minimal differences here!

whyoleg:
> In a general case, I am not sure if we can seamless merge the documentation of a getter and a setter.

Yeah, that's what I was thinking - I think it makes sense to mention this in #4250, as it could affect the design.

> Maybe we should not filter out Java getters and setters, so we don’t break Gradle’s documentation at least.

Yeah, I think that it will be more correct. Does this mean we will never need to show synthetic `var` properties to be correct?
Could you also check what `UserClassFilePermissions` will look like with symbols?

dokka-subprojects/plugin-base/src/test/kotlin/translators/DefaultPsiToDocumentableTranslatorTest.kt
vmishenev: leftover

dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
vmishenev: > Still, I'm curious if we can remove this ? here.

It comes from the AA/compiler. We could only have a hard-coded fix for the `Enum` supertype.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/ObviousAndInheritedFunctionsDocumentableFilterTest.kt
vmishenev: This is more of a limitation, but I filed [the issue](https://github.com/Kotlin/dokka/issues/4511).

dokka-subprojects/plugin-base/src/test/kotlin/model/JavaTest.kt
vmishenev: I have elaborated the reason. 
This is the same situation as with `Enum` above. We don't use PSI for analyzing a list of supertypes/interfaces. 
We could, but a scope would still be mapped to Kotlin, so the situation would be worse.

dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
vmishenev: JIC, `lowerBound` fixed it, but after #4503 it will be actual again.

dokka-subprojects/plugin-base/src/test/kotlin/model/JavaTest.kt
vmishenev: JIC, `lowerBound` fixed it, but after https://github.com/Kotlin/dokka/issues/4503 it will be actual again.

dokka-subprojects/plugin-base/src/test/kotlin/superFields/PsiSuperFieldsTest.kt
vmishenev: I'm not sure that this is on AA's side since it provides a Kotlin-view.
I fixed it in Dokka for the sake of the test.

dokka-subprojects/plugin-base/src/test/kotlin/transformers/ObviousAndInheritedFunctionsDocumentableFilterTest.kt
vmishenev: > This is more of a limitation

However, for example, I have an idea here to use `declaredMemberScope` (not including inherited members) instead of `combinedMemberScope`.... but collecting inherited symbols by Dokka is a bit complicated.

vmishenev:
> Maybe we should not filter out Java getters and setters, so we don’t break Gradle’s documentation at least.

I'm going to do it in a separate PR.

>  Does this mean we will never need to show synthetic var properties to be correct?

No. Dokka will show synthetic properties together with their declared accessors.
E.g. `var someProp`, `fun getSomeProp()` and `fun setSomeProp(..)`

dokka-subprojects/plugin-base/src/test/kotlin/transformers/ObviousAndInheritedFunctionsDocumentableFilterTest.kt
whyoleg: Let's not do it in this PR, feel free to share the idea behind those proposed changes in #4511 while it's still fresh in your mind. :)

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/expectedData/html/it-android/it.android/-integration-test-activity/index.html
```diff
@@ -166,7 +166,7 @@ <h2 class="tableheader">Properties</h2>
                 <div class="title">
                   <div class="platform-hinted  with-platform-tabs" data-platform-hinted="data-platform-hinted">
                     <div class="platform-bookmarks-row" data-toggle-list="data-toggle-list"><button class="platform-bookmark" data-filterable-current=":/androidMain" data-filterable-set=":/androidMain" data-active="" data-toggle=":/androidMain">android</button></div>
-<div class="content sourceset-dependent-content" data-active="" data-togglable=":/androidMain"><div class="symbol monospace"><span class="token keyword">val </span><a href="index.html#1391867424%2FProperties%2F1031138095">application</a><span class="token operator">: </span><a href="https://developer.android.com/reference/kotlin/android/app/Application.html">Application</a><span class="token operator">?</span></div></div>                  </div>
```
whyoleg, line 169: We lost nullability here (and below) because of `lowerBound` usage, right?

whyoleg (approved):
Nice! Great job!

Could you also please rebase and check the CI failure?
Should we now create follow-up issues (e.g., enable new analysis by default via Gradle property, Gradle docs related changes, and so on) and close #4463?

dokka-subprojects/plugin-base/src/test/kotlin/signatures/FunctionalTypeConstructorsSignatureTest.kt
```diff
@@ -283,12 +277,12 @@ class FunctionalTypeConstructorsSignatureTest : BaseAbstractTest() {
         }
     }
 
-    @OnlyJavaPsi
+    @OnlyJavaSymbols("#3611 - do not show open for Java fields")
```
whyoleg, line 280: BTW: Can we say that #3611 is fixed in Java Symbols analysis?

dokka-integration-tests/gradle/projects/it-android-kotlin-mp-builtin/expectedData/html/it-android/it.android/-integration-test-activity/index.html
vmishenev: Yes, but `lowerBound` should not affect it.
I have fixed it https://github.com/Kotlin/dokka/pull/4502/commits/30d65d8cc63105718c46bb9c7f550841418a3f56

Issue #1306, opened 2020-08-13
Inherited methods with generic type parameters have DRIs pointing to the inheritor, not parent


kamildoleglo:
Please see the `it-android-0` test for reference.

``` kotlin
class IntegrationTestActivity : AppCompatActivity() {}
```
Here the `IntegrationTestActivity` inherits methods like:
```kotlin
fun <T : View> requireViewById(p0: Int): T
```
The `T` parameter has the DRI pointing to the `IntegrationTestActivity`: 
```
it.android/IntegrationTestActivity/requireViewById/#kotlin.Int/PointingToGenericParameters(0)/
```
while it should be pointing to the `AppCompatActivity` instead

PR #4509, opened 2026-05-13
Removes workardound for #1306

whyoleg:
I've checked the test locally, and it doesn't fail anymore, so it looks like it's something very-very old.

I've also checked the HTML output here, and in this case, generic parameters just link to the same page...
It looks like #3054, but here it's about generic parameters.

After that, I've checked the codebase, and overall it looks like Dokka doesn't really support **rendering** links to any kind of **parameters** via DRI, or am I missing something?
We do check DRI targets when [creating sections](https://github.com/Kotlin/dokka/blob/c821d5bf6ca4ba36f7050987f640c81f1a30e533/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DescriptionSections.kt#L119-L124), but those are coming from `Param` tag only.
I don't say it's a bug, just an interesting observation.


vmishenev (approved):

vmishenev:
> After that, I've checked the codebase, and overall it looks like Dokka doesn't really support rendering links to any kind of parameters via DRI, or am I missing something?

The old Dokka (v0....) supported it. Now it is not supported.

PR #4512, opened 2026-05-15
Set `finalizeCoroutines` back to `true` and setup a test for metaspace leaks with DGPv2

whyoleg:
Looks like we have an issue with metaspace memory, again...
But now, it's somewhere in workers' usage.

Reproduced with both 2.2 and 2.1. Not reproduced in 2.0.
I would say the cause could be #4182. Investigating.

New test added reproduces the issue locally - OOM fails after ±6 project generation:

```
> Task :subproject_0:dokkaGenerateModuleHtml
> Task :subproject_1:dokkaGenerateModuleHtml
> Task :subproject_2:dokkaGenerateModuleHtml
> Task :subproject_3:dokkaGenerateModuleHtml
> Task :subproject_4:dokkaGenerateModuleHtml
> Task :subproject_5:dokkaGenerateModuleHtml
> Task :subproject_6:dokkaGenerateModuleHtml

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "DefaultDispatcher-worker-1"
> Task :subproject_7:dokkaGenerateModuleHtml
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "RMI TCP Connection(idle)"
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "RMI TCP Connection(idle)"
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "RMI TCP Connection(idle)"
```

After that, the project couldn't even recover, as the worker failed with OOM, and just stuck...

The reproducer is super simple: have N projects with a **single** worker reused all the time.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -170,7 +170,7 @@ constructor(
             enabled.convention(true)
             cacheRoot.convention(dokkaExtension.dokkaCacheDirectory)
             failOnWarning.convention(false)
-            finalizeCoroutines.convention(false)
+            finalizeCoroutines.convention(true)
             moduleName.convention(dokkaExtension.moduleName)
             moduleVersion.convention(dokkaExtension.moduleVersion)
             offlineMode.convention(false)
--- a/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MemoryStressTest.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MemoryStressTest.kt
@@ -64,6 +64,7 @@ private fun setupProject(
                 |  dokkaGeneratorIsolation.set(ProcessIsolation {
                 |      maxHeapSize.set("1g")
                 |      jvmArgs.add("-XX:MaxMetaspaceSize=$metaspaceMemory") // limit metaspace
+                |      jvmArgs.add("-XX:SoftRefLRUPolicyMSPerMB=10")
                 |  })
                 |}
                 |
--- a/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MemoryStressTest.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MemoryStressTest.kt
@@ -9,8 +9,8 @@ import org.jetbrains.dokka.gradle.utils.*
 import org.jetbrains.dokka.gradle.utils.projects.defaultKgpTestVersion
 
 class MemoryStressTest : FunSpec({
-    context("when subprojects have 10 modules and uses single worker") {
-        val project = setupProject(10)
+    context("when subprojects have 50 modules and uses single worker") {
+        val project = setupProject(50, "500m")
 
         test("generate HTML publication") {
             project.runner
@@ -31,7 +31,7 @@ class MemoryStressTest : FunSpec({
 
 private fun setupProject(
     numberOfProjects: Int,
-    metaspaceMemory: String = "500m"
+    metaspaceMemory: String
 ): GradleProjectTest = gradleKtsProjectTest("MemoryStressTest-$numberOfProjects") {
     val projectNames = List(numberOfProjects) { "subproject_$it" }
     settingsGradleKts += projectNames.joinToString("\n") { """include(":$it")""" }
--- a/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MemoryStressTest.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/testFunctional/kotlin/MemoryStressTest.kt
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package org.jetbrains.dokka.gradle
+
+import io.kotest.core.spec.style.FunSpec
+import org.jetbrains.dokka.gradle.internal.DokkaConstants
+import org.jetbrains.dokka.gradle.utils.*
+import org.jetbrains.dokka.gradle.utils.projects.defaultKgpTestVersion
+
+class MemoryStressTest : FunSpec({
+    context("when subprojects have 10 modules and uses single worker") {
+        val project = setupProject(10)
+
+        test("generate HTML publication") {
+            project.runner
+                .addArguments(
+                    ":dokkaGenerate",
+                    "--rerun-tasks",
+                    "--stacktrace",
+                    "--no-parallel",
+                    "--max-workers=1" // reuse single worker all the time to reproduce the issue
+                )
+                .forwardOutput()
+                .build {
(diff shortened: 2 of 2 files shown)


whyoleg:
## What has happened?

During the Dokka execution in projects with many modules, a Metaspace OOM occurs. It happens because, after #4182, we started creating custom class loaders (again) to execute Dokka, and those are never garbage-collected for various reasons.

One of the main contributors to memory leaks is the `finalizeCoroutines` configuration flag. This flag controls two things:

- The Dokka engine uses coroutines for analysis and HTML rendering under the hood; default dispatchers should be closed after the work is done. It happens [here](https://github.com/Kotlin/dokka/blob/9531f10f96c713686711178e210ec3e92acdcbd1/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/DokkaGenerator.kt#L54).
- At the same time, Dokka uses the Analysis API, which in turn uses the IntelliJ platform, which should be closed after the work is done. It happens [here](https://github.com/Kotlin/dokka/blob/9531f10f96c713686711178e210ec3e92acdcbd1/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/SymbolsAnalysisPlugin.kt#L59).

Previously, with DGPv1, it was always set to `true`, but with DGPv2, we migrated Dokka analysis to use Gradle workers and worker classpath, so it was set to `false` because workers with the same classpath could be reused, and when no work for workers is needed anymore, Gradle will terminate the worker process. No coroutines/resource management on our side!

**This was the default behavior as of Dokka 2.0.0 where DGPv2 was first introduced.**

At some point, because of the update of Kotlin Analysis API, which is built with the new version of Kotlin stdlib, we stumbled on a forward-incompatible change somewhere in Kotlin stdlib (#4178). Overall, it could have happened with any other version, as in reality, Gradle workers still provide some bootstrap classpath when using`WorkerExecutor.processIsolation` (https://github.com/gradle/gradle/issues/33987), and it provides their own Kotlin stdlib version (https://github.com/gradle/gradle/issues/34442), which shadows any other kotlin stdlib version we provide via a `classpath` property of `WorkerExecutor.processIsolation`.

So, Kotlin Analysis API started to depend on some declaration in Kotlin stdlib, which is not available in the version of Kotlin stdlib that the Gradle workers' bootstrap classpath is using.

To overcome this issue, we decided (in #4182) to fall back to the old way of executing Dokka via `DokkaBootstrap`, creating a new classloader for every execution, but now, inside of a worker.

Though we haven't set `finalizeCoroutines` back to `true`. And so, for every Dokka Gradle plugin task (which is, in most cases, one per module), we created a new classloader that was never garbage-collected. We didn't have a test for DGPv2 for this "classloader memory leak" because we thought Gradle workers would make it not an issue. :)

**The fix, as well as the test for this issue in DGPv2, is in the current PR.**

## Anything else?

During the investigation, I've found additional issues that could cause Dokka to leak classloaders.

### `SoftRefLRUPolicyMSPerMB` on JVM 8

On JVM 8: GC doesn't collect soft references fast enough. In DGPv1 tests, we do use `SoftRefLRUPolicyMSPerMB` system property to overcome this issue (by setting `-XX:SoftRefLRUPolicyMSPerMB=10`). The same property is used in the newly added test. Nice article about the property itself and how it affects everything: https://www.jasonpearson.dev/softreflrupolicymspermb-in-jvm-builds/.

Overall, this has been a known issue for a while, but the only place where this flag is needed is our tests, which run on different JVM versions. In reality, most projects use newer JVM versions to build, e.g., AGP requires JDK 11 or 17 (depending on the version), and Gradle 9 requires JDK 17.

**Note**: this is about the JVM version where Gradle runs, and not which version is used to compile the project (e.g., via toolchains).

**The fix: nothing to be done here, just use the property in our tests**

### `Thread.inheritedAccessControlContext` on JVM lower than 24

During several follow-up investigations, I have also found that as we create those classloaders, we also _could_ leak classloader via `Thread.inheritedAccessControlContext` - private field, which is needed to work with `SecurityManager`, now deprecated and removed in **JDK 24** ([JEP 486](https://openjdk.org/jeps/486)).

This is coming from some thread in `ForkJoinPool-commonPool`. I haven't been able to pinpoint when or why it happens, but it's probably because something inside the Gradle worker starts the thread on its own classpath first.
In my tests, I have seen only a single classloader leaked this way in multi-module projects, so I would not consider this an issue worth fixing at this point, especially since there is already JDK 25 released, which is a _kind of LTS version_ (see [Java 25 is ALSO no LTS Version](https://www.youtube.com/watch?v=x6-kyQCYhNo&) video for more details on the "kind of").

**The fix: use JDK24+. :)**

### com.intellij.util.concurrency.AppDelayQueue$TransferThread

In addition to the issues above, another problematic case was found: sometimes `com.intellij.util.concurrency.AppDelayQueue$TransferThread` could still be running even after we release resources in the IntelliJ platform, and so it holds the classloader and prevents it from being GC-ed.

It doesn't happen on every run, but I was able to reproduce it periodically in [cryptography-kotlin](https://github.com/whyoleg/cryptography-kotlin). So far, I think it might be related to https://youtrack.jetbrains.com/issue/IJPL-1619, which describes a race condition during closure.

The fix for it was done in https://github.com/JetBrains/intellij-community/commit/0f3694c0298faa4d06f1f562a54975957d529bbd and so available in`261.20362.25` version of the IntelliJ platform.

Dokka can't easily update the platform version, because it should be done first in the Kotlin compiler and Analysis API. Since the fix touches many areas, I wasn't able to create a patch for Dokka, as with other fixes.

**The issue could be revisited once Kotlin updates the IntelliJ platform version: https://youtrack.jetbrains.com/issue/KT-82657/Update-IntelliJ-SDK-dependency-to-261.9214.**

Dokka issue: https://github.com/Kotlin/dokka/issues/4533

## In the end.

The current PR fixes the biggest issue with metaspace leaks, which is reliable, reproducible, and affects everyone. Other issues are harder to fix, investigate, and reproduce. As well as they most likely were there all the time, as they don't depend on Dokka being executed by workers or not.

Still, there is one more potential improvement that could reduce Dokka's memory usage. Currently, we don't configure `MaxMetaspaceSize` for worker execution, which means it grows infinitely as we create new classloaders for each task. Configuring `MaxMetaspaceSize` to some value like 512 or 1024 MB by default will force the JVM to unload created classloaders when the limit is reached, which could reduce overall Dokka's memory consumption in multi-module projects.

Currently, it's already possible to do so by configuring `jvmArgs` for the Dokka worker, as shown in the test:

```kotlin
dokka {
    dokkaGeneratorIsolation.set(ProcessIsolation {
        jvmArgs.add("-XX:MaxMetaspaceSize=512m")
    })
}
```

But the problem here is that when we create a new instance of `ProcessIsolation`, we will clean up all the arguments set there by default in Dokka, including `maxHeapSize`.

So the proposal here is to set up a default during the `ProcessIsolation` function call, and set a reasonable `MaxMetaspaceSize`value in a separate PR (https://github.com/Kotlin/dokka/pull/4534)

vmishenev (approved):

PR #4514, opened 2026-05-25
Add support for KEEP-0449: the companion block and the companion modifier on extensions

### LINKED ISSUE #2344: When Java code is included in a module alongside Kotlin, static methods on interfaces aren't rendered differently
Create a Kotlin module that also includes Java code. Create a Java interface with a static method. Render it to Dokka HTML format. Look at the Java interface. Nothing calls out that the method is static, probably because Kotlin doesn't have such a concept. However, when rendering code written in Java, this concept does need to be represented.

### LINKED ISSUE #4526: Implement representation of companion extensions and blocks
[KEEP-0449: Companion extensions and blocks](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0449-companions-block-extension.md)

> Results of internal discussion:
> 
> **Companion blocks:**
> - Represent companion blocks, Java static declarations, and Enum synthetic declarations (valueOf/values/entries) in the same way.
> - Two new sections, "Companion functions" and "Companion properties", need to be added in addition to the existing ones: Constructors, Types, Functions, Properties.
>   - The new sections should be added after the "Types" section.
> - Support those in DRI to distinguish between instance and companion block declarations.
> 
> **Companion extensions:**
> - Support rendering of `companion` modifier in the signature
> - Support those in DRI to distinguish between: companion extension, companion object extension, and top-level function
> - Render those extensions on the `Members & Extensions` tab, in the same way other extensions are rendered there.
> 
 _Originally posted by @whyoleg in [#4438](https://github.com/Kotlin/dokka/issues/4438#issuecomment-4406215498)_

vmishenev:
**What changes**

• Model. New experimental CompanionBlockMember extra (and a `Callable.isCompanion` flag on the DRI) representing companion-block scope. Used to unify Java statics, enum synthetic `values/valueOf/entries`, Kotlin companion-block members, and companion extensions under one concept. 
• Analysis. The K2 symbol translator and the Java PSI parser tag matching declarations with `CompanionBlockMember`. 
• Rendering. Two new sections — `Companion functions` and `Companion properties` — appear on classlike pages after `Types` and before `Functions/Properties`. The `companion` keyword is emitted in Kotlin signatures only for companion extensions.

----

Most unit tests are generated by AI.

### DIFF
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
@@ -466,7 +466,7 @@ internal class DokkaPsiParser(
                     ObviousMember.takeIf { psi.isObvious(inheritedFrom) },
                     helper.getCheckedExceptionDRIs(psi).takeIf { it.isNotEmpty() }
                         ?.let { CheckedExceptions(it.toSourceSetDependent()) },
-                    @OptIn(ExperimentalDokkaApi::class) CompanionBlockMember.takeIf { isCompanionBlock },
+                    @OptIn(ExperimentalDokkaApi::class) IsCompanion.takeIf { isCompanionBlock },
                 )
             }
         )
@@ -642,7 +642,7 @@ internal class DokkaPsiParser(
                     annotations.toSourceSetDependent().toAnnotations(),
                     helper.getConstantExpression(psi)?.let { DefaultValue(it.toSourceSetDependent()) },
                     takeIf { isVar }?.let { IsVar },
-                    @OptIn(ExperimentalDokkaApi::class) CompanionBlockMember.takeIf { psi.hasModifier(JvmModifier.STATIC) },
+                    @OptIn(ExperimentalDokkaApi::class) IsCompanion.takeIf { psi.hasModifier(JvmModifier.STATIC) },
                 )
             }
         )
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
@@ -421,10 +421,9 @@ internal class DokkaPsiParser(
         parentDRI: DRI? = null,
     ): DFunction {
         val isCompanionBlock = !isConstructor && psi.hasModifier(JvmModifier.STATIC)
-        val baseDri = parentDRI?.let { dri ->
+        val dri = parentDRI?.let { dri ->
             DRI.from(psi).copy(packageName = dri.packageName, classNames = dri.classNames)
         } ?: DRI.from(psi)
-        val dri = baseDri.markStaticIfNeeded(isCompanionBlock)
         val docs = psi.getDocumentation()
         return DFunction(
             dri = dri,
@@ -472,15 +471,6 @@ internal class DokkaPsiParser(
             }
         )
     }
-
-    /**
-     * Marks the callable part of a [DRI] as static so static Java members (and any other
-     * companion-block scope declarations) can be distinguished from instance ones with the
-     * same name on the same classlike.
-     */
-    private fun DRI.markStaticIfNeeded(isCompanionBlock: Boolean): DRI =
-        if (!isCompanionBlock) this
-        else copy(callable = callable?.copy(isCompanion = true))
 
     private fun PsiNamedElement.parseSources(): SourceSetDependent<DocumentableSource> {
         return when {
@@ -617,8 +607,7 @@ internal class DokkaPsiParser(
         setter: DFunction?,
         inheritedFrom: DRI? = null
     ): DProperty {
-        val isCompanionBlock = psi.hasModifier(JvmModifier.STATIC)
-        val dri = DRI.from(psi).markStaticIfNeeded(isCompanionBlock)
+        val dri = DRI.from(psi)
 
         // non-final java field without accessors should be a var
         // setter should be not null when inheriting kotlin vars
@@ -653,7 +642,7 @@ internal class DokkaPsiParser(
                     annotations.toSourceSetDependent().toAnnotations(),
                     helper.getConstantExpression(psi)?.let { DefaultValue(it.toSourceSetDependent()) },
                     takeIf { isVar }?.let { IsVar },
-                    @OptIn(ExperimentalDokkaApi::class) CompanionBlockMember.takeIf { isCompanionBlock },
+                    @OptIn(ExperimentalDokkaApi::class) CompanionBlockMember.takeIf { psi.hasModifier(JvmModifier.STATIC) },
                 )
             }
         )
--- a/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
+++ b/dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
@@ -11,6 +11,7 @@ import com.intellij.psi.*
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import org.jetbrains.dokka.DokkaConfiguration
+import org.jetbrains.dokka.ExperimentalDokkaApi
 import org.jetbrains.dokka.analysis.java.BreakingAbstractionKotlinLightMethodChecker
 import org.jetbrains.dokka.analysis.java.SyntheticElementDocumentationProvider
 import org.jetbrains.dokka.analysis.java.getVisibility
@@ -419,9 +420,11 @@ internal class DokkaPsiParser(
         inheritedFrom: DRI? = null,
         parentDRI: DRI? = null,
     ): DFunction {
-        val dri = parentDRI?.let { dri ->
+        val isCompanionBlock = !isConstructor && psi.hasModifier(JvmModifier.STATIC)
+        val baseDri = parentDRI?.let { dri ->
             DRI.from(psi).copy(packageName = dri.packageName, classNames = dri.classNames)
         } ?: DRI.from(psi)
+        val dri = baseDri.markStaticIfNeeded(isCompanionBlock)
         val docs = psi.getDocumentation()
         return DFunction(
             dri = dri,
@@ -463,11 +466,21 @@ internal class DokkaPsiParser(
                         .toAnnotations(),
                     ObviousMember.takeIf { psi.isObvious(inheritedFrom) },
                     helper.getCheckedExceptionDRIs(psi).takeIf { it.isNotEmpty() }
-                        ?.let { CheckedExceptions(it.toSourceSetDependent()) }
+                        ?.let { CheckedExceptions(it.toSourceSetDependent()) },
+                    @OptIn(ExperimentalDokkaApi::class) CompanionBlockMember.takeIf { isCompanionBlock },
                 )
             }
         )
     }
+
+    /**
+     * Marks the callable part of a [DRI] as static so static Java members (and any other
+     * companion-block scope declarations) can be distinguished from instance ones with the
+     * same name on the same classlike.
+     */
+    private fun DRI.markStaticIfNeeded(isCompanionBlock: Boolean): DRI =
+        if (!isCompanionBlock) this
+        else copy(callable = callable?.copy(isCompanion = true))
 
     private fun PsiNamedElement.parseSources(): SourceSetDependent<DocumentableSource> {
         return when {
@@ -604,7 +617,8 @@ internal class DokkaPsiParser(
         setter: DFunction?,
         inheritedFrom: DRI? = null
     ): DProperty {
-        val dri = DRI.from(psi)
+        val isCompanionBlock = psi.hasModifier(JvmModifier.STATIC)
+        val dri = DRI.from(psi).markStaticIfNeeded(isCompanionBlock)
 
         // non-final java field without accessors should be a var
         // setter should be not null when inheriting kotlin vars
@@ -638,7 +652,8 @@ internal class DokkaPsiParser(
                     it.toSourceSetDependent().toAdditionalModifiers(),
                     annotations.toSourceSetDependent().toAnnotations(),
                     helper.getConstantExpression(psi)?.let { DefaultValue(it.toSourceSetDependent()) },
-                    takeIf { isVar }?.let { IsVar }
+                    takeIf { isVar }?.let { IsVar },
+                    @OptIn(ExperimentalDokkaApi::class) CompanionBlockMember.takeIf { isCompanionBlock },
                 )
             }
         )
--- a/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
+++ b/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
@@ -89,7 +89,19 @@ public data class Callable(
     val params: List<TypeReference>,
     @property:ExperimentalDokkaApi
     val contextParameters: List<TypeReference> = emptyList(),
-    val isProperty: Boolean = false
+    val isProperty: Boolean = false,
+    /**
+     * Distinguishes callables that do not have an instance receiver from instance ones.
+     *
+     * `true` for:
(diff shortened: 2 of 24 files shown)


dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
```diff
@@ -463,12 +466,22 @@ internal class DokkaPsiParser(
...
             }
         )
     }
 
+    /**
+     * Marks the callable part of a [DRI] as static so static Java members (and any other
+     * companion-block scope declarations) can be distinguished from instance ones with the
+     * same name on the same classlike.
+     */
+    private fun DRI.markStaticIfNeeded(isCompanionBlock: Boolean): DRI =
```
whyoleg: Curious: Is there a reason why it's not performed inside `DRI.from(psi)`? At least, I would expect it to be there.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -128,17 +140,39 @@ public data class Callable(
...
+         * The `static` marker is appended only when [isCompanion] is `true` to preserve the existing
+         * signature of regular instance callables.
          */
         val contextParameters = @OptIn(ExperimentalDokkaApi::class) contextParameters
-        return if (contextParameters.isNotEmpty())
+        val base = if (contextParameters.isNotEmpty())
             "${contextParameters.joinToString("#")}#${receiver?.toString().orEmpty()}#${params.joinToString("#")}"
         else
             "${receiver?.toString().orEmpty()}#${params.joinToString("#")}"
+        return if (@OptIn(ExperimentalDokkaApi::class) isCompanion) "$base/static" else base
```
whyoleg: Why `static` and not `companion`?
Also, why `/` and not `#` like with parameters?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableProperties.kt
```diff
@@ -43,6 +44,26 @@ public object ObviousMember : ExtraProperty<Documentable>, ExtraProperty.Key<Doc
...
+ * meaning it has no instance receiver from the enclosing class.
+ *
+ *
+ * This covers, in a unified way:
+ * - Java `static` methods and fields
+ * - Enum synthetic declarations (`values`, `valueOf`, `entries`)
+ * - Kotlin companion-block members
+ *   (see [KEEP-0449](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0449-companions-block-extension.md))
+ *
+ * The same condition is reflected in the [org.jetbrains.dokka.links.Callable.isCompanion]
```
whyoleg, line 60: Now comes the question: why do we need this `CompanionBlockMember` extra if we can get the same information from the DRI?

Also, the name feels wrong, as for extensions those are called `companion extesnsions`, so maybe we need to rename it to `IsCompanion`, like with `IsVar`?

For me, it seems, that it might be enough to not have it at all, as we have this information in DRI, but I'm not sure, WDYT?

dokka-subprojects/plugin-base/src/test/kotlin/transformers/CompanionSectionsTest.kt
```diff
@@ -0,0 +1,443 @@
...
+        ) {
+            pagesTransformationStage = { root ->
+                val enumPage = root.dfs { it.name == "E" } as? ClasslikePageNode
+                    ?: error("E page not found")
+                val companionFns = enumPage.sectionTexts("Companion functions")
+                    ?: error("Companion functions section must exist on enum page")
+
+                assertEquals(listOf("valueOf", "values"), companionFns, "expected synthetic 'valueOf' and 'values' under Companion functions, got: $companionFns")
+
+                val regularFns = enumPage.sectionTexts("Functions").orEmpty()
```
whyoleg, line 142: shouldn't we have just an empty list here and so assert against it?

dokka-subprojects/plugin-base/src/test/kotlin/signatures/SignatureTest.kt
```diff
@@ -1765,6 +1765,155 @@ class SignatureTest : BaseAbstractTest() {
...
+        """
+            |/src/main/kotlin/example/Util.kt
+            |package example
+            |class Vector(val x: Double, val y: Double)
+            |
+            |fun Vector.length(): Double = 0.0
+        """.trimMargin()
+    ) {
+        val sig = renderedContent("root/example/length.html").firstSignature().toString()
+        assertTrue("companion" !in sig, "expected no 'companion' keyword for a plain extension, got: $sig")
```
whyoleg: here and below, I think that instead of `assertTrue("companion" !in sig` it would be better and clearer to assert for real signature, as in all other places.

dokka-subprojects/plugin-base/src/test/kotlin/enums/KotlinEnumsTest.kt
```diff
@@ -361,7 +362,8 @@ class KotlinEnumsTest : BaseAbstractTest() {
                 }
 
                 root.contentPage<ClasslikePageNode>("TestEnum") {
-                    assertHasFunctions("toBeImplemented", "valueOf", "values")
+                    assertHasFunctions("toBeImplemented")
+                    assertHasCompanionFunctions("valueOf", "values")
```
whyoleg, line 366: It might make sense to add a check, that we also handle `entries` property introduced in Kotlin 1.9 (if I remember correctly).
Maybe in separate PR? I don't see tests for `entries` properties at all in this file right now...

dokka-subprojects/plugin-base/src/test/kotlin/model/JavaTest.kt
```diff
@@ -476,6 +477,128 @@ class JavaTest : BaseAbstractTest() {
...
+                        @OptIn(ExperimentalDokkaApi::class) valueOf.dri.callable?.isCompanion,
+                        "DRI for 'valueOf' should be marked isStatic=true"
+                    )
+                }
+            }
+        }
+    }
+
+    @Test
+    fun `Callable signature distinguishes static from instance for same-name callables`() {
```
whyoleg: It feels weird, that this test is in `JavaTest` - don't we have a better place for DRI tests?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -89,7 +89,19 @@ public data class Callable(
     val params: List<TypeReference>,
     @property:ExperimentalDokkaApi
     val contextParameters: List<TypeReference> = emptyList(),
-    val isProperty: Boolean = false
+    val isProperty: Boolean = false,
+    /**
+     * Distinguishes callables that do not have an instance receiver from instance ones.
```
whyoleg, line 94: So, just to make sure I understand correctly.
- it's a `companion block` callable, if `isCompanion=true` and `receiver=null`
- it's a `companion extension` callable, `isCompanion=true` and `receiver=XXX`
- it's a regular member callable if `isCompanion=false` and `receiver=null`
- it's a regular extension callable if `isCompanion=false` and `receiver=XXX`

Right?

What will be in place of `receiver` in case of `companion extension`? just `TypeConstructor` instance?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
```diff
@@ -89,7 +89,19 @@ public data class Callable(
...
+     *
+     * `true` for:
+     * - Java static methods/fields
+     * - Enum synthetic declarations (`values`, `valueOf`, `entries`)
+     * - Kotlin companion-block members (see [KEEP-0449](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0449-companions-block-extension.md))
```
whyoleg, line 99: companion extensions too?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -416,32 +417,50 @@ public open class DefaultPageCreator(
         extensions: List<Documentable>,
     ) = contentBuilder.contentFor(dri, sourceSets) {
         typesBlock(types)
+
         val (extensionProperties, extensionFunctions) = extensions.splitPropsAndFuns()
+
+        // Companion-block scope members (Java statics, enum synthetic, KEEP-0449 companion blocks)
+        // are rendered right after Types and before instance Functions/Properties.
+        val (companionExtensionFunctions, extensionInstanceFunctions) = extensionFunctions.partition { it.isCompanionBlock() }
```
whyoleg: Curious: is it really true, that we don't need to adapt `ExtensionExtractorTransformer` for `companion extensions`?
If true, that's nice! Just wanted to double check it.

dokka-subprojects/plugin-base/src/test/kotlin/model/CompanionBlockTest.kt
```diff
@@ -0,0 +1,268 @@
...
+            |
+            |companion fun Vector.unit(): Vector = Vector(1.0, 1.0)
+            """.trimIndent()
+        ) {
+            val pkg = packages.single()
+            val ext = pkg.functions.firstOrNull { it.name == "unit" }
+                .assertNotNull("top-level companion extension 'unit'")
+            ext.extra[CompanionBlockMember].assertNotNull("CompanionBlockMember on companion extension 'unit'")
+            // sanity-check it is in fact an extension: receiver is non-null and resolves to Vector
+            assertTrue(ext.receiver != null, "companion extension must keep its receiver in the documentable")
```
whyoleg, line 214: It would be nice to assert what specifically is in `receiver` (here and below)

whyoleg (changes requested):
Nice!

My main concern is that we have both `DRI.isCompanion` and `CompanionBlockMember`, where the latter seems redundant, has a strange name, and could cause confusion. E.g., when should I use one or the other?

Could you also recheck all comments in tests and code to make them more readable and mention both companion blocks and extensions where needed? Currently, it feels like all of them are only talking about "companion blocks," but they describe behavior for both companion blocks and **companion extensions**.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: Yes
like a regular extension

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
vmishenev: TBH, I have no idea or preference for how to present it.
Would `/companion` be ok?

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableProperties.kt
vmishenev: > Also, the name feels wrong, as for extensions those are called companion extesnsions, so maybe we need to rename it to IsCompanion, like with IsVar?

Good idea. Renamed.

> why do we need this CompanionBlockMember extra if we can get the same information from the DRI?

In my view, Documentable and DRI are independent entities: DRI is for identification, while Documentable is for description. Getting such information (or e.g. `name`) directly from Documentable is more convenient and obvious than from DRI.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/links/DRI.kt
whyoleg: Let's go with `/companion` to be aligned with the overall feature naming.
As KEEP says, there is no `static` in the Kotlin language; it exists only in underlying platforms (backends).

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableProperties.kt
whyoleg: Thanks! Agree, let's have it like this.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
```diff
@@ -419,6 +420,7 @@ internal class DokkaPsiParser(
         inheritedFrom: DRI? = null,
         parentDRI: DRI? = null,
     ): DFunction {
+        val isCompanionBlock = !isConstructor && psi.hasModifier(JvmModifier.STATIC)
```
whyoleg, line 423: Mostly curious: why do we need to check for `isConstructor` here? It's not possible to have a static constructor in Java, right?

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -416,32 +417,50 @@ public open class DefaultPageCreator(
...
             functionsBlock("Inherited functions", inheritedFunctions, inheritedExtensionFunctions)
         } else {
-            propertiesBlock("Properties", properties, extensionProperties)
-            functionsBlock("Functions", functions, extensionFunctions)
+            propertiesBlock("Properties", instanceProperties, extensionInstanceProperties)
+            functionsBlock("Functions", instanceFunctions, extensionInstanceFunctions)
         }
     }
 
+    private fun <T> T.isCompanionBlock(): Boolean where T : Documentable, T : WithExtraProperties<T> {
```
whyoleg: The name is a bit misleading now - it's actually check if the declaration is a companion block member **or** companion extension.
Let's rename it to `isCompanion` or even just inline, as there are just two usages, and `isCompanion` feels really ambiguous, especially since we can call this function on classes. :)

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
```diff
@@ -678,6 +697,32 @@ public open class DefaultPageCreator(
...
     }
 
+    private fun DocumentableContentBuilder.companionFunctionsBlock(declarations: List<DFunction>, extensions: List<DFunction>) {
+        functionsOrPropertiesBlock(
+            name = "Companion functions",
+            contentKind = ContentKind.Functions,
+            contentType = when {
+                declarations.isEmpty() -> BasicTabbedContentType.EXTENSION_FUNCTION
+                else -> BasicTabbedContentType.FUNCTION
+            },
```
whyoleg, line 707: BTW: I haven't spotted it before - is it fine that we re-use the same `contentKind`/`contentType` for both companion and regular functions/properties?
It probably should be fine?

whyoleg (approved):
Nice! Good to go after fixing DRI signature rendering!

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
vmishenev: >  is it fine that we re-use the same contentKind/contentType for both companion and regular functions/properties?

It is fine till we display them under the `Members` tab. For now, it makes no sense to create extra `contentKind`/`contentType`.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/DokkaPsiParser.kt
vmishenev: Just an extra check for no reason, even though there are no static constructors in Java

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/translators/documentables/DefaultPageCreator.kt
vmishenev: >  especially since we can call this function on classes. 

I have added the constraint  `T : org.jetbrains.dokka.model.Callable`

PR #4528, opened 2026-06-03
Update Analysis API to 2.4.20-dev-5364

vmishenev:
Contains: 
- https://youtrack.jetbrains.com/issue/KT-85112/AA-does-not-see-packages-from-unpacked-klibs
- https://youtrack.jetbrains.com/issue/KT-86370/KDoc-links-to-extensions-members-of-a-companion-block-should-be-resolved

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
@@ -15,10 +15,10 @@ import org.jetbrains.kotlin.analysis.api.analyze
 import org.jetbrains.kotlin.analysis.api.projectStructure.KaSourceModule
 import org.jetbrains.kotlin.analysis.api.projectStructure.contextModule
 import org.jetbrains.kotlin.analysis.api.symbols.*
-import org.jetbrains.kotlin.idea.references.mainReference
 import org.jetbrains.kotlin.kdoc.psi.api.KDoc
 import org.jetbrains.kotlin.kdoc.psi.impl.KDocLink
 import org.jetbrains.kotlin.kdoc.psi.impl.KDocName
+import org.jetbrains.kotlin.psi.KtExperimentalApi
 import org.jetbrains.kotlin.psi.KtPsiFactory
 
 /**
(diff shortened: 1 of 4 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/ResolveKDocLink.kt
```diff
@@ -162,7 +162,8 @@ private fun resolveKDocLinkToDRI(kDocLink: KDocLink): DRI? {
 
 private fun KaSession.resolveToSymbol(kDocLink: KDocLink): KaSymbol? {
     val lastNameSegment = kDocLink.children.filterIsInstance<KDocName>().lastOrNull()
-    return lastNameSegment?.mainReference?.resolveToSymbols()?.sortedWith(linkCandidatesComparator)?.firstOrNull()
+    @OptIn(KaExperimentalApi::class, KtExperimentalApi::class)
+    return lastNameSegment?.resolveSymbols()?.sortedWith(linkCandidatesComparator)?.firstOrNull()
```
vmishenev, line 166: https://youtrack.jetbrains.com/issue/KT-66039

dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
```diff
@@ -1332,7 +1332,7 @@ class LinkTest : BaseAbstractTest() {
                             "Storage",
                             Callable("setValue", null, listOf(JavaClassReference("java.lang.String")))
                         ),
-                        "Storage.prop" to DRI("example", "Storage", Callable("getProp", null, emptyList()))
+                        "Storage.prop" to DRI("example", "Storage", Callable("prop", null, emptyList(), isProperty = true))
```
whyoleg, line 1336: Could you add a YT issue related to this change to the comment here?

Overall, it seems correct.

whyoleg (approved):
Nice!

dokka-subprojects/plugin-base/src/test/kotlin/markdown/LinkTest.kt
vmishenev: BTW, we now have one more reason not to ignore synthetic properties.

whyoleg: true...
It started to feel that for synthetic properties, we need to copy both getter/setter docs and render them with `get`/`set` prefixes in the main docs.

PR #4534, opened 2026-06-16
Move `ProcessIsolation` convention values to `ProcessIsolation` factory function

whyoleg:
Move `ProcessIsolation` convention values to `ProcessIsolation` factory function, so that if final users override just maxHeapSize, everything else will use preconfigured values.
If users need to modify `jvmArgs`, they can use `jvmArgs.set` to replace or `jvmArgs.add` to append to' jvmArgs' when creating a `ProcessIsolation` object. Which is most likely never needed.

In addition, update the `ProcessIsolation` convention to limit metaspace memory.

### DIFF
--- a/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
+++ b/dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
@@ -111,30 +111,7 @@ constructor(
             dokkaEngineVersion.convention(DokkaConstants.DOKKA_VERSION)
         }
 
-        dokkaExtension.dokkaGeneratorIsolation.convention(
-            dokkaExtension.ProcessIsolation {
-                maxHeapSize.convention("2g")
-                debug.convention(false)
-                jvmArgs.convention(
-                    listOf(
-                        //"-XX:MaxMetaspaceSize=512m",
-                        "-XX:+HeapDumpOnOutOfMemoryError",
-                        "-XX:+AlwaysPreTouch", // https://github.com/gradle/gradle/issues/3093#issuecomment-387259298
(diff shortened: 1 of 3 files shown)


dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
```diff
@@ -111,30 +111,7 @@ constructor(
...
 
-        dokkaExtension.dokkaGeneratorIsolation.convention(
-            dokkaExtension.ProcessIsolation {
-                maxHeapSize.convention("2g")
-                debug.convention(false)
-                jvmArgs.convention(
-                    listOf(
-                        //"-XX:MaxMetaspaceSize=512m",
-                        "-XX:+HeapDumpOnOutOfMemoryError",
-                        "-XX:+AlwaysPreTouch", // https://github.com/gradle/gradle/issues/3093#issuecomment-387259298
```
vmishenev, line 122: Why has this argument been removed?

vmishenev (approved):

dokka-runners/dokka-gradle-plugin/src/main/kotlin/DokkaBasePlugin.kt
whyoleg: Because it doesn't really make any difference and might potentially harm. :)
Previous [internal discussion](https://jetbrains.slack.com/archives/C02H0GSH474/p1731418903394129) with @adam-enko, which boils down to "let's leave it there". The only single mention of this flag in relation to Gradle workers is in https://github.com/gradle/gradle/issues/3093#issuecomment-387259298. I haven't found any other Gradle plugin that could use this flag.

Also, before this PR, overriding isolation properties would drop this flag anyway, and a couple of projects have done so.

So I don't really see a reason for this flag to be there at all. Though I forgot to mention this in the PR description, sorry. :(

PR #4535, opened 2026-06-17
Update kotlinx-collections-immutable to 0.5.0

DmitryNekrasov:
0.5.0 is the first stable release of kotlinx.collections.immutable after 0.4.0. Dokka consumes it only as a `runtimeOnly` dependency, so the bump needs no source changes.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,7 +12,7 @@ gradlePlugin-kotlin = "2.3.21"
 gradlePlugin-android = "7.3.1"
 
 kotlinx-coroutines = "1.7.3"
-kotlinx-collections-immutable = "0.3.6"
+kotlinx-collections-immutable = "0.5.0"
 kotlinx-serialization = "1.6.0"
 kotlinx-bcv = "0.17.0"
 


whyoleg:
@DmitryNekrasov, thanks for the PR, but as far as I remember, we don't use this library at all. We only need it because the Kotlin Analysis API uses it, so to avoid any incompatibilities, it's better to use the same version used in the Kotlin Analysis API.
cc @vmishenev, just in case I'm missing something.

So I think we don't need to update it. Soon, we will just use a single dependency Analysis API artifact, and there won't be any problems with dependencies.

vmishenev:
As far as I remember, it was needed for the Analysis API/compiler somehow.
Although now `kotlinx-collections-immutable` comes from `platform:core:251.27812.49` with the newer version `0.3.8`.
image

I think we could just get rid of `kotlinx-collections-immutable`.

whyoleg:
Thanks!
Let's change nothing for now, then, since the fact that the IntelliJ platform provides it doesn't mean we don't need to explicitly state that it's a dependency of AA, similar to caffeine.

PR #4538, opened 2026-06-17
Exit from CLI runner when an exception is thrown

### LINKED ISSUE #4537: [Bug] Dokka cli runner failing to exit after thrown exception

If there is an exception thrown during docs generation when running dokka with the CLI runner, the process doesn't exit correctly and instead hangs indefinitely.

### Reproducer

To reproduce:

* In `SingleModuleGeneration.kt`, add the line `throw IllegalStateException("exit")` at the end of the `try` block in the `generate` function.
* Run `./gradlew publishToMavenLocal`
* Set up the following project with any kotlin file in `src/main/kotlin`, and run `gradlew run`

`settings.gradle.kts`:
```
pluginManagement {
    repositories {
        mavenLocal()
        mavenCentral()
        gradlePluginPortal()
    }
}

    repositories {
        mavenLocal()
        mavenCentral()
    }
}
```

`build.gradle.kts`:
```
plugins {
    kotlin("jvm") version "2.3.21"
}

@CacheableTask
abstract class DokkaRunner : DefaultTask() {
    @get:Inject abstract val execOperations: ExecOperations
    @get:Classpath abstract val dokkaClasspath: ConfigurableFileCollection
    @get:Classpath abstract val pluginsClasspath: ConfigurableFileCollection
    @get:[InputFiles PathSensitive(PathSensitivity.NONE)] abstract val sourceDirectory: DirectoryProperty
    @get:OutputDirectory abstract val outputDirectory: DirectoryProperty

    @TaskAction
    fun exec() {
        execOperations.javaexec {
            classpath = dokkaClasspath
            args = listOf(
                "-pluginsClasspath",
                pluginsClasspath.files.joinToString(";"),
                "-sourceSet",
                "-src ${sourceDirectory.get().asFile.absolutePath}",
                "-outputDir",
                outputDirectory.get().asFile.absolutePath,
            )
        }
    }
}

val runTask = tasks.register<DokkaRunner>("run") {
    dokkaClasspath.from(configurations.create("runnerJar") {
        dependencies.add(
            project.dependencies.create("org.jetbrains.dokka:dokka-cli:2.3.0-SNAPSHOT")
        )
    })
    pluginsClasspath.from(
        configurations.create("pluginsJars") {
            dependencies.add(
                project.dependencies.create("org.jetbrains.dokka:dokka-base:2.3.0-SNAPSHOT")
            )
            dependencies.add(
                project.dependencies.create("org.jetbrains.dokka:analysis-kotlin-symbols:2.3.0-SNAPSHOT")
            )
        }
    )
    sourceDirectory.set(project.layout.projectDirectory.dir("src/main/kotlin"))
    outputDirectory.set(project.layout.buildDirectory.dir("docs"))
}
```

### Dokka version

2.3.0-SNAPSHOT

_No response_

_No response_

### Build tool

CLI

### Project type

- [ ] Android
- [ ] JVM
- [ ] Other

### Anything else? (optional)

I did also try this with the `finalizeCoroutines` option in a configuration JSON file, and still found it was hanging.

juliamcclellan:
Currently, if an exception is thrown when dokka is invoked through the CLI runner, the process does not exit correctly. This addresses the issue 

To test:
* In `SingleModuleGeneration.kt`, add the line `throw IllegalStateException("exit")` at the end of the `try` block in the `generate` function.
* Run `./gradlew publishToMavenLocal` from the checkout root directory
* From `examples/cli/dokka-cli-example`, run `../../../gradlew run`

Without this change, the process hangs after the error is thrown. With this change, it prints the error stack trace and then exits.

Fixes: #4537

### DIFF
--- a/dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
+++ b/dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
@@ -133,7 +133,7 @@ class CliIntegrationTest : AbstractCliIntegrationTest() {
         val result = process.awaitProcessResult()
         assertEquals(1, result.exitCode, "Expected exitCode 1 (Fail)")
 
-        assertTrue(result.output.contains("Exception in thread \"main\" org.jetbrains.dokka.DokkaException: Failed with warningCount"))
+        assertTrue(result.output.contains("org.jetbrains.dokka.DokkaException: Failed with warningCount"))
     }
 
     @Test
--- a/dokka-runners/runner-cli/src/main/kotlin/org/jetbrains/dokka/main.kt
+++ b/dokka-runners/runner-cli/src/main/kotlin/org/jetbrains/dokka/main.kt
@@ -7,11 +7,20 @@ package org.jetbrains.dokka
 import org.jetbrains.dokka.DokkaConfiguration.ExternalDocumentationLink
 import org.jetbrains.dokka.utilities.*
 import java.nio.file.Paths
+import kotlin.system.exitProcess
 
 public fun main(args: Array<String>) {
     val globalArguments = GlobalArguments(args)
     val configuration = initializeConfiguration(globalArguments)
-    DokkaGenerator(configuration, globalArguments.logger).generate()
+    var exitCode = 0
+    try {
+        DokkaGenerator(configuration, globalArguments.logger).generate()
+    } catch (e: Throwable) {
+        exitCode = 1
+        e.printStackTrace()
+    } finally {
+        exitProcess(exitCode)
+    }
 }
 
 public fun initializeConfiguration(globalArguments: GlobalArguments): DokkaConfiguration {
--- a/examples/cli/dokka-cli-example/README.md
+++ b/examples/cli/dokka-cli-example/README.md
@@ -1,0 +1,7 @@
+# Dokka CLI example
+
+This directory contains a small project which is set up to run Dokka through the CLI runner.
+It does this by setting up a Gradle task that invokes the CLI runner.
+
+To run it, first run `./gradlew publishToMavenLocal` from the checkout root.
+Then run `../../../gradlew run` from this directory.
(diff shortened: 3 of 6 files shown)


whyoleg:
@juliamcclellan, we do already have integration tests for CLI, and even for the fail case:
https://github.com/Kotlin/dokka/blob/d11819cd3b85590691c4eea96bac8177fb2ee114/dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt#L111

And it looks like it currently works fine with errors.

juliamcclellan:
Sorry, I didn't find that test. However, I am consistently finding that when there is an exception while using the CLI runner through a `JavaExec` gradle task, the process does not exit. I think this is because there are additional threads which are not cleaned up without the explicit `exit`, and using `finalizeCoroutines` didn't fix this.

vmishenev:
>  I think this is because there are additional threads which are not cleaned up without the explicit exit,

Alright, it seems https://github.com/Kotlin/dokka/blob/e704740b8d4799a8d3b177a09cac777c74934d2d/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/plugin/SymbolsAnalysisPlugin.kt#L59 is never called when an exception occurs.

Apparently, in the test https://github.com/Kotlin/dokka/blob/d11819cd3b85590691c4eea96bac8177fb2ee114/dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt#L111
the AA is not initialized yet.

juliamcclellan:
@vmishenev thanks for looking into this! That dispose step never running does explain things, if I move `runPostActions()` into the `finally` block in `SingleModuleGeneration` that also fixes the issue. Do you think it would make sense to do this? I don't know whether all post actions make sense to run if docs generation has failed. Another option could be having a separate extension point for clean up steps that must run even if there's a failure?

vmishenev:
>  Do you think it would make sense to do this? 

I think so. For example, we could mark such a `PostAction` as a disposer via a marker interface and run it in the `finally` block.

> Another option could be having a separate extension point for clean up steps that must run even if there's a failure?

Your option works for me as well.

juliamcclellan:
Created #4545 instead.

PR #4545, opened 2026-06-23
Add clean up extension point which runs in finally block

### LINKED ISSUE #4537: [Bug] Dokka cli runner failing to exit after thrown exception

juliamcclellan:
Currently, if an exception is thrown during generation but before the post actions are run, the dokka process doesn't terminate because `disposeKotlinAnalysisPostAction` from `SymbolsAnalysisPlugin` is never run. I've added a test in `CliIntegrationTest` that would demonstrate this issue by forcing an error when trying to write to the output directory.

To address this, this PR adds a new `CleanUpAction` extension point, which unlike `PostAction` will run in the `finally` block of generation so that it is guaranteed to run even if generation fails. This switches `disposeKotlinAnalysisPostAction` from `SymbolsAnalysisPlugin` to a `CleanUpAction`.

Fixes #4537

### DIFF
--- a/dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
+++ b/dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
@@ -5,6 +5,7 @@ package org.jetbrains.dokka.it.cli
 
 import org.jetbrains.dokka.it.awaitProcessResult
 import java.io.File
+import java.util.concurrent.TimeUnit
 import kotlin.test.*
 
 class CliIntegrationTest : AbstractCliIntegrationTest() {
@@ -117,6 +118,7 @@ class CliIntegrationTest : AbstractCliIntegrationTest() {
             "-outputDir", dokkaOutputDir.path,
             "-pluginsClasspath", dokkaPluginsClasspath,
             "-moduleName", "Basic Project",
+            // The failure here happens after all generation is complete.
             "-failOnWarning",
             "-sourceSet",
             buildString {
@@ -134,6 +136,41 @@ class CliIntegrationTest : AbstractCliIntegrationTest() {
         assertEquals(1, result.exitCode, "Expected exitCode 1 (Fail)")
 
         assertTrue(result.output.contains("Exception in thread \"main\" org.jetbrains.dokka.DokkaException: Failed with warningCount"))
+    }
+
+    @Test
+    fun failCliDuringGeneration() {
+        val dokkaOutputDir = File(projectDir, "output")
+        assertTrue(dokkaOutputDir.mkdirs())
+        // Making the output unwritable causes a failure before all steps of generation are complete. This verifies that
+        // clean up actions run successfully (if they didn't, the test would time out).
+        assertTrue(dokkaOutputDir.setWritable(false))
+
+        val process = ProcessBuilder(
+            "java",
+            "-jar", dokkaCliJarPath,
+            "-outputDir", dokkaOutputDir.path,
+            "-pluginsClasspath", dokkaPluginsClasspath,
+            "-moduleName", "Basic Project",
+            "-sourceSet",
+            buildString {
+                append(" -sourceSetName cliMain")
+                append(" -src ${File(projectDir, "src").path}")
+                append(" -jdkVersion 8")
+                append(" -analysisPlatform jvm")
+            }
+        )
+            .redirectErrorStream(true)
+            .start()
+
+        // Forces the test to time out if the process does not terminate within 30 seconds.
+        val complete = process.waitFor(30, TimeUnit.SECONDS)
+        assertTrue(complete)
+
+        val result = process.awaitProcessResult()
+        assertEquals(1, result.exitCode, "Expected exitCode 1 (Fail)")
+
+        assertTrue(result.output.contains("Failed to write"))
     }
 
     @Test
--- a/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/CoreExtensions.kt
+++ b/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/CoreExtensions.kt
@@ -4,6 +4,7 @@
 
 package org.jetbrains.dokka
 
+import org.jetbrains.dokka.generation.CleanUpAction
 import org.jetbrains.dokka.generation.Generation
 import org.jetbrains.dokka.plugability.ExtensionPoint
 import org.jetbrains.dokka.renderers.PostAction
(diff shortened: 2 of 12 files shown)


dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/generation/SingleModuleGeneration.kt
whyoleg, line 1: Could you please also update [AllModulesPageGeneration](https://github.com/Kotlin/dokka/blob/87901eb6a1cad7e7a4d50afe39bfbae676be7a80/dokka-subprojects/plugin-all-modules-page/src/main/kotlin/org/jetbrains/dokka/allModulesPage/AllModulesPageGeneration.kt) with the same logic?

whyoleg: Also, [TestProjectAnalyzer](https://github.com/Kotlin/dokka/blob/407f0531104e6ff1dbf1af2cdbc03eca43b01027/dokka-subprojects/analysis-kotlin-api/src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/analysis/TestProjectAnalyzer.kt#L253) should now use `cleanUp`.

dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/CoreExtensions.kt
```diff
@@ -36,6 +37,8 @@ public object CoreExtensions {
 
     public val postActions: ExtensionPoint<PostAction> by coreExtensionPoint<PostAction>()
 
+    public val cleanUp: ExtensionPoint<CleanUpAction> by coreExtensionPoint<CleanUpAction>()
```
whyoleg:
```suggestion
    public val cleanUpActions: ExtensionPoint<CleanUpAction> by coreExtensionPoint<CleanUpAction>()
```

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/generation/SingleModuleGeneration.kt
```diff
@@ -55,10 +55,11 @@ public class SingleModuleGeneration(private val context: DokkaContext) : Generat
 
         report("Running post-actions")
         runPostActions()
+    } finally {
+        report("Cleaning up")
+        cleanUp()
 
         reportAfterRendering()
```
whyoleg: I'm not totally sure that we need to move this into `finally` (same for the changes in `BaseDokkaTestGenerator`).
1. Because `reportAfterRendering` can throw, and so could theoretically hide the exception, which is more important.
2. As far as I see, `reportAfterRendering` is a last step in "non-exceptional" Dokka execution, and so should be handled as part of the pipeline, and so `cleanUp` should be called afterward.

whyoleg (changes requested):
Thanks! I do like the direction of the PR, though there are some things we need to fix first so that the change to the new `cleanUp` actions will work more robustly and across all places.

dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
```diff
@@ -136,6 +138,41 @@ class CliIntegrationTest : AbstractCliIntegrationTest() {
...
+                append(" -src ${File(projectDir, "src").path}")
+                append(" -jdkVersion 8")
+                append(" -analysisPlatform jvm")
+            }
+        )
+            .redirectErrorStream(true)
+            .start()
+
+        // Forces the test to time out if the process does not terminate within 30 seconds.
+        val complete = process.waitFor(30, TimeUnit.SECONDS)
```
vmishenev, line 167: Thank you for submitting this PR

It seems `awaitProcessResult` below also has a 60-second timeout

dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/generation/SingleModuleGeneration.kt
juliamcclellan: Moved it back to the `try`.

dokka-integration-tests/cli/src/cliIntegrationTest/kotlin/org/jetbrains/dokka/it/cli/CliIntegrationTest.kt
juliamcclellan: It seems like it should time out, but when I run the test without this `waitFor` it doesn't time out (I let it run for 3 minutes).

whyoleg (approved):
Nice! Thanks for the contribution!

vmishenev (approved):

PR #4550, opened 2026-07-03
Update Analysis API to 2.4.20-dev-8332

vmishenev:
Contains  https://youtrack.jetbrains.com/issue/KT-87143/findKDoc-returns-parent-property-KDoc-for-accessors-of-an-overriding-property

### DIFF
--- a/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
+++ b/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
@@ -22,6 +22,7 @@ import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtFile
 
 /**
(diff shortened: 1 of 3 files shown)


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
```diff
@@ -99,7 +100,8 @@ internal class AnnotationTranslator(private val logger: DokkaLogger) {
                 )
             },
             mustBeDocumented = mustBeDocumented(annotation),
-            scope = annotation.useSiteTarget?.toDokkaAnnotationScope() ?: Annotations.AnnotationScope.DIRECT
+            scope = (annotation.psi as? KtAnnotationEntry)?.useSiteTarget?.getAnnotationUseSiteTarget()
```
whyoleg, line 103: Hm, why are we going into PSI here instead of the already available AA API?
Is there an issue/thread/etc about why this was changed?

whyoleg (commented):
Looks good, and it's nice that we have one less version to depend on!

Though I'm not sure, what is the reason for the changes in how annotations are handled...

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/AnnotationTranslator.kt
vmishenev: https://youtrack.jetbrains.com/issue/KT-73796

vmishenev: It says "Clients should work with PSI if needed since useSiteTarget is syntax, not semantic"

whyoleg: Thanks! Now I get the idea.

whyoleg (approved):

Issue #4513, opened 2026-05-19
Enable Java Symbol Analysis by default


vmishenev:
Checklist:
- [x] Recheck the  gradle project
- [ ] (**POSTPONED**) Do not filter out accessors of synthetic Java properties  https://github.com/Kotlin/dokka/pull/4515 
- [ ] Define a final Documentable repsentation

Will fix:

#3611

juliamcclellan:
Java annotation changes that are probably WAI from a Kotlin API perspective: Java-source annotations now have `DConstructor`s, and Java-source annotation values are now `DProperty`s instead of `DFunction`s.

PR #4559, opened 2026-07-21
Update Annotations merge strategy to combine annotations for each source set

### LINKED ISSUE #4558: [Bug] Package-level annotation dropped due to inconsistent `Annotations` merge

If a package with both java and kotlin source files has package-level annotations in the `package-info.java` files, I expect the annotations to be present on the `DPackage` after the java and kotlin `DModule`s have been merged. This was the previous dokka behavior, but between versions and 2.3.0-dev-423 and 2.3.0-dev-424 the behavior changed and the annotations are no longer present on the merged `DPackage`.

From debugging, this is because the behavior of `Annotations.mergeStrategyFor` is inconsistent depending on which `Annotations` is `left` and which is `right`:

```kotlin
override fun mergeStrategyFor(left: Annotations, right: Annotations): MergeStrategy<AnnotationTarget> =
            MergeStrategy.Replace(Annotations(left.myContent + right.myContent))
```

Because `myContent` is `SourceSetDependent<List<Annotation>>`, typealias for `Map<DokkaSourceSet, List<Annotation>>`. When adding two `Map`s together, if the same key is present in both `Map`s, the corresponding entries aren't merged in any way, the second entry is just used. So when merging, if the java package `Annotations` are `left` while they empty `Annotations` from the kotlin `DPackage` are `right`, the resultant `DPackage` has empty `Annotations`.

### Reproducer

`/src/com/sample/package-info.java`
```java
@Deprecated
package com.sample;
```
`/src/com/sample/Foo.kt`
```kotlin
package com.sample
class Foo
```
`/src/com/sample/Bar.java`
```java
package com.sample;
public class Bar {}
````

### Dokka version

2.3.0-dev-424

_No response_

_No response_

### Build tool

Gradle

### Project type

- [ ] Android
- [ ] JVM
- [ ] Other

### Anything else? (optional)

_No response_

juliamcclellan:
Currently, `Annotations.mergeStrategyFor` uses `+` on two `Map`s. This means that if the same key is present in both `Map`s, the value for that key in the resultant `Map` is the value from the second `Map`, with the value from the first `Map` getting ignored.

This behavior means that if a mixed java and kotlin package has annotations in the `package-info.java`, they might get dropped during the merge of the java and kotlin `DPackage`s.

This PR updates the merge behavior to avoid dropping annotations. I'm not sure if this is always the best way to handle annotation merging (for instance, if there were the same annotations in both `left` and `right`, should the annotations be duplicated?). Let me know if a different approach would make more sense.

Fixes #4558

### DIFF
--- a/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/additionalExtras.kt
+++ b/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/additionalExtras.kt
@@ -32,8 +32,18 @@ public data class Annotations(
     private val myContent: SourceSetDependent<List<Annotation>>
 ) : ExtraProperty<AnnotationTarget> {
     public companion object : ExtraProperty.Key<AnnotationTarget, Annotations> {
-        override fun mergeStrategyFor(left: Annotations, right: Annotations): MergeStrategy<AnnotationTarget> =
-            MergeStrategy.Replace(Annotations(left.myContent + right.myContent))
+        override fun mergeStrategyFor(left: Annotations, right: Annotations): MergeStrategy<AnnotationTarget> {
+            val sourceSets = left.myContent.keys + right.myContent.keys
+            // For each source set, merge the annotations from the `left` and `right`, if they exist.
+            return MergeStrategy.Replace(
+                Annotations(
+                    sourceSets.associateWith { sourceSet ->
+                        left.myContent.getOrDefault(sourceSet, emptyList()) +
+                                right.myContent.getOrDefault(sourceSet, emptyList())
+                    }
+                )
+            )
+        }
     }
 
     override val key: ExtraProperty.Key<AnnotationTarget, *> = Annotations
--- a/dokka-subprojects/plugin-base/src/test/kotlin/model/annotations/JavaAnnotationsTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/model/annotations/JavaAnnotationsTest.kt
@@ -9,6 +9,7 @@ import org.jetbrains.dokka.base.testApi.testRunner.BaseAbstractTest
 import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.model.*
 import translators.findClasslike
+import translators.findPackage
 import kotlin.test.*
 
 class JavaAnnotationsTest : BaseAbstractTest() {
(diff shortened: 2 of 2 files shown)


vmishenev (commented):
Thank you for your contribution! 👍 This is a really interesting case. I hope [the Java Symbols analysis](https://github.com/Kotlin/dokka/issues/4513) will also fix it.

whyoleg:
Thanks for the fix!

> This PR updates the merge behavior to avoid dropping annotations. I'm not sure if this is always the best way to handle annotation merging (for instance, if there were the same annotations in both left and right, should the annotations be duplicated?). Let me know if a different approach would make more sense.

Overall, I share your concern, but in reality, I think it's not an issue that could happen. Dokka analyses every source file once in the context of one source set. This situation, where we need to merge anything (like annotations for a package) for a declaration present twice in one source set, could happen only in mixed Kotlin/Java projects, as we do have two analyses there: one for Kotlin and one for Java files. And, as Vadim mentioned above, if we start to analyze everything via a single translator (e.g., there will no longer be `DefaultPsiToDocumentableTranslator`, just `DefaultSymbolToDocumentableTranslator`), this issue could not happen at all.

At first, I was thinking that we might need to adapt other properties, which could be merged like `AdditionalModifiers` in the file above the changed `Annotations` property (or any other property really, like [ImplementedInterfaces](https://github.com/Kotlin/dokka/blob/41d657abee2467ab717f082868d805ec780c7c49/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableProperties.kt#L25) or [CheckedExceptions](https://github.com/Kotlin/dokka/blob/41d657abee2467ab717f082868d805ec780c7c49/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/documentableProperties.kt#L88)).
But realistically, we need to only adapt the case with **annotations on packages**, and it's just safe to do so as presented in the PR.
- In a KMP project,  the `DokkaSourceSet` in `SourceSetDependent` will always be different for each source-set, so no problem at all to begin with.
- In a Kotlin/Java project, only a package with the same name is some entity that could be present in Kotlin and Java files, and so each analysis could produce its own `DPackage` with the same `DokkaSourceSet`, but we don't have annotations on Kotlin packages, so packages produced from Java will be the only ones with annotations.

In all other cases, like classes, properties, functions, any declaration with the same signature will produce an error during compilation, and so there will be no other entities apart from packages, which could be merged in the same source set. 

At least, this is how I understand the current situation :).

whyoleg (approved):

vmishenev (commented):
> In a Kotlin/Java project, only a package with the same name is some entity that could be present in Kotlin and Java files, and so each analysis could produce its own DPackage with the same DokkaSourceSet, but we don't have annotations on Kotlin packages, so packages produced from Java will be the only ones with annotations.

Oh, I didn't notice that in the first review. `DefaultSymbolToDocumentableTranslator` can return `null` instead of an empty list  [here](https://github.com/Kotlin/dokka/blob/71f648117699b0231b01aaa8886d9f0747d98da7/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt#L183) for package annotations. It usually returns `null` instead of an empty list for annotations, and so on.

In this case, the merger should work correctly. I think this could be an additional fix on top of the existing one.

juliamcclellan:
> At least, this is how I understand the current situation :).

That makes sense to me, and from testing this on the AndroidX codebase I didn't see any other changes besides for package annotations.

> I hope https://github.com/Kotlin/dokka/issues/4513 will also fix it.

> DefaultSymbolToDocumentableTranslator can return null instead of an empty list [here](https://github.com/Kotlin/dokka/blob/71f648117699b0231b01aaa8886d9f0747d98da7/dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt#L183) for package annotations.

Good points, I think this will fix the issue for now but might become unnecessary in the future.

juliamcclellan:
I don't think I have permission to merge, but is it alright to do so?

PR #4563, opened 2026-07-24
Update Analysis API to 2.5.0-dev-2030 and Intellij platform to 261

vmishenev:
Updates the Analysis API to `2.5.0-dev-2030` and, with it, the IntelliJ platform to
`261.24374.151`.
The Kotlin Gradle plugin used to build Dokka goes `2.3.21` → `2.4.0`.

## Analysis API changes (https://youtrack.jetbrains.com/issue/KT-86591)

* `org.jetbrains.kotlin.analysis.api.analyze` moved to `...analysis.api.session.analyze`.
* `analyze { }` no longer puts `KaSession` in scope as a receiver — the session is obtained via
  `contextOf<KaSession>()`. 
  
  
## IntelliJ platform 261 changes

**1. Core PSI moved out of `java-psi`.** `com.intellij.psi.PsiElement` and friends are no longer on
the `java-psi(-impl)` classpath; they live in the platform `core` module. 

**2. `PsiKeyword` → `com.intellij.java.syntax.parser.JavaKeywords`** in
`DefaultPsiToDocumentableTranslator.getVisibility()`.

**3. Dropped the vendored `com.intellij.util.lang.JavaVersion`.** It was a workaround for
[KT-73967](https://youtrack.jetbrains.com/issue/KT-73967) and became actively harmful at 261.

**4. `JavaDocParser` crashes on JDK < 21.** 261 ships a new `java-syntax` artifact in which the
Java-doc parser was rewritten in Kotlin. The class targets Java 8 (major version 52), so it loads
everywhere, but it compiles `closingStatusList.removeLast()` down to
`java.util.List.removeLast()` — a `SequencedCollection` method
([JEP 431](https://openjdk.org/jeps/431)) that only exists at runtime on JDK 21+:
```
Caused by: java.lang.NoSuchMethodError: java.util.List.removeLast()Ljava/lang/Object;
at com.intellij.java.syntax.parser.JavaDocParser.parseDataItem(JavaDocParser.kt:117)
at com.intellij.java.syntax.parser.JavaDocParser.parseDocCommentText(JavaDocParser.kt:66)
at com.intellij.psi.impl.source.tree.JavaDocElementType$DocCommentElementType.lambda$new$0(JavaDocElem@KotlinDocCommentParser.kt
```

Worked around by keeping a copy of `JavaDocParser` in `analysis-java-psi`, which shadows
the artifact's version on the classpath (same trick already used for
`BinaryFileTypeDecompilers`), with the single call rewritten as `removeAt(lastIndex)` →
`java.util.List.remove(int)`. **

[ IDEA-392386](https://youtrack.jetbrains.com/issue/IDEA-392386/Dokka-fails-on-JDK-8-due-to-removeLast-usage.) has been created to address this issue in the future.

**5. `PsiInlineDocTag.dataElements` no longer contains whitespaces.** After IDEA-379008
`WHITE_SPACE` was removed from `PsiInlineDocTagImpl.VALUE_BIT_SET`.

**6. `PsiMethod.findSuperMethods()` silently returns an empty array.**    `com.intellij.superMethodsSearch` has no executor.   Fixed by vendoring `MethodSuperSearcher` and registering it into the EP once the analysis session (and thus
the application environment) exists, via `registerJavaPsiSearchExecutors(disposable)`.

[KT-88362](https://youtrack.jetbrains.com/projects/KT/issues/KT-88362/PsiMethod.findSuperMethods-always-returns-an-empty-array)  might resolve it. _(I will keep an eye)_

_Its only executor,
`MethodSuperSearcher`, was [moved to `java-indexing-impl` (https://github.com/JetBrains/kotlin/commit/87c7798742a43f93249bbb87ec790fc6d448546d),
which Dokka doesn't depend on._


gradle/libs.versions.toml
```diff
@@ -20,13 +20,13 @@ ktor = "2.3.11"
...
 ## Analysis
-kotlin-compiler-k2 = "2.4.20-dev-8332"
+kotlin-compiler-k2 = "2.5.0-dev-2030"
 
 # MUST match the version of the intellij platform used in the kotlin compiler,
 # otherwise this will lead to different versions of psi API and implementations
 # on the classpath and will fail with hard to debug problems in runtime.
-# See: https://github.com/JetBrains/kotlin/blob/79b4def109dc73f882ec8a291a69d0197ad55c58/gradle/versions.properties#L1
-intellij-platform = "251.27812.49"
+# See: https://github.com/JetBrains/kotlin/blob/build-2.5.0-dev-2030/gradle/versions.properties#L1
```
whyoleg, line 28: Oh, I like that we can use the tag here aligned with the AA version!

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/kdoc/moduledocs/ModuleAndPackageDocumentationParsingContext.kt
```diff
@@ -45,7 +46,7 @@ internal fun ModuleAndPackageDocumentationParsingContext(
         MarkdownParser(
             externalDri = { link ->
                 analyze(sourceModule) {
-                    resolveKDocTextLink(link, contextPackageFQN, locationInformation, logger, sourceSet)
+                    contextOf<KaSession>().resolveKDocTextLink(link, contextPackageFQN, locationInformation, logger, sourceSet)
```
whyoleg: Minor: We might as well update all of our functions to use context parameters instead of the receiver for sessions.
In this or in a follow-up PR.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolExternalDocumentablesProvider.kt
```diff
@@ -48,9 +49,10 @@ internal class SymbolExternalDocumentablesProvider(val context: DokkaContext) :
                 lightMethodChecker = lightMethodChecker
             )
 
-            val parentDRI = (symbol.containingSymbol as? KaDeclarationSymbol)?.let { getDRIFromSymbol(it) } ?: /* top level */ DRI(dri.packageName)
+
+            val parentDRI = with(contextOf<KaSession>()) { (symbol.containingSymbol as? KaDeclarationSymbol) }?.let { contextOf<KaSession>().getDRIFromSymbol(it) } ?: /* top level */ DRI(dri.packageName)
```
whyoleg: minor: `with(contextOf<KaSession>())` seems redundant, as we anyway call `getDRIFromSymbol` on explicit receiver, no?

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/PsiElementToHtmlConverter.kt
```diff
@@ -161,7 +161,7 @@ internal class PsiElementToHtmlConverter(
         }
 
         private fun dataElementsAsText(tag: PsiInlineDocTag): String {
-            return tag.dataElements.joinToString("") {
+            return tag.dataElementsWithWhitespaces().joinToString("") {
```
whyoleg, line 164: TBH, I'm not sure I understand this change, and the introduced functions feel like a hack.

Can't we just update the `joinToString` call to accept a space as the separator, or is that not possible because of newlines?

whyoleg (approved):
Nice work, especially with all of the investigations and fixes! Mostly minor comments.

dokka-subprojects/analysis-java-psi/src/main/kotlin/org/jetbrains/dokka/analysis/java/parsers/doctag/PsiElementToHtmlConverter.kt
vmishenev: It does not cover cases with more than one consecutive whitespace
For, example
```java
* @{code
* a
*
* b
* }
```

whyoleg (approved):

PR #4567, opened 2026-07-29
Fix getting DRI to context parameter for properties

### LINKED ISSUE #4566: [Bug] Documentation of a context parameter via `@param` throws with "Error in translating of symbol (null)"

After adding `@param theContextParameterName ...` line in docstring of a property, Dokka worker throws an exception. But it seems it shouldn't.

### Reproducer

```kotlin
/**
 * Foo!
 */
interface Foo {
    fun work()
}

/**
 * work!
 * @param foo Foo?
 */
context(foo: Foo)
val work: Unit get() = foo.work()
```

The line ` * @param foo Foo?` is crucial.

### Dokka version

2.2.0

2.4.20-Beta2

_No response_

### Build tool

Gradle

### Project type

- [x] Kotlin Multiplatform (KMP)
- [ ] Android
- [ ] JVM
- [ ] Other

### Anything else? (optional)

_No response_

whyoleg:
Fix #4566


vmishenev (approved):
Can you add some tests for the Documentable model as well?

vmishenev (approved):

PR #4568, opened 2026-07-29
Exclude companion extensions from subclass and implementor pages

### LINKED ISSUE #4562: Parent companion extensions look redundant on child's page
```kotlin
open class ParentClassCompanionExtContainer
companion val ParentClassCompanionExtContainer.parentClassCompanionExtProperty = ""

class ChildClassCompanionExtContainer: ParentClassCompanionExtContainer()
```
the `ChildClassCompanionExtContainer` page will look like
Image

But in general nothing connects `parentClassCompanionExtProperty` with `ChildClassCompanionExtContainer`

vmishenev:
fixes #4562
Implements KEEP-0449 rules in `ExtensionExtractorTransformer` to ensure companion extensions of supertypes are not inherited.

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/transformers/CompanionSectionsTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/transformers/CompanionSectionsTest.kt
@@ -414,6 +414,77 @@ class CompanionSectionsTest : BaseAbstractTest() {
         }
     }
 
+    /**
+     * The companion scope of a supertype is not accessible through the name of a subtype
+     * (see KEEP-0449 §2.5), so companion extensions must not be listed on subclass pages.
+     *
+     * See https://github.com/Kotlin/dokka/issues/4562
+     */
+    @Test
+    fun `companion extension of a parent does not appear on a child page`() {
+        testInline(
+            """
+            |/src/main/kotlin/example/Util.kt
+            |package example
+            |open class Parent
+            |class Child : Parent()
+            |
+            |companion val Parent.companionExtProperty: String get() = ""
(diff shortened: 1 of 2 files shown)


dokka-subprojects/plugin-base/src/test/kotlin/transformers/CompanionSectionsTest.kt
```diff
@@ -414,6 +414,77 @@ class CompanionSectionsTest : BaseAbstractTest() {
...
+                assertNull(childPage.sectionTexts("Companion functions"))
+                // while plain extensions of the parent are applicable to the child and must be kept
+                assertEquals(listOf("instanceExtProperty"), childPage.sectionTexts("Properties"))
+                assertEquals(listOf("instanceExtFunction"), childPage.sectionTexts("Functions"))
+            }
+        }
+    }
+
+    @Test
+    fun `companion extension of an interface does not appear on an implementor page`() {
```
whyoleg, line 461: TBH: I can't spot the difference in what this test tests in addition to the one above it.

whyoleg (approved):
Nice!

dokka-subprojects/plugin-base/src/test/kotlin/transformers/CompanionSectionsTest.kt
vmishenev: The former tests an open class and its subclass; the latter tests an interface and its implementation.

whyoleg: Ah, now I see, `interface` vs `class`. Thanks!

PR #4569, opened 2026-07-30
Display unmapped Java types only if `enableExperimentalSymbolsJavaAnalysis` is enabled

### LINKED ISSUE #4564: Changed behavior for inherited primitive wrappers Java in 2.3.0
```kotlin
// java
public class Parent{
    public String propertyString = null;
    public int propertyInt = 0;
    public Integer propertyInteger = null;
}

// kotlin
class Child: Parent()
```

In 2.2.0 Dokka `Integer` and `String` are rendered as nullable types with reference to Kotlin stdlib
Image

In 2.3.0-dev they are rendered and `Integer` and `String` with references to JDK (that is not obvious till the click/hover)

Image

For the `Parent` page nothing changed – both linked to JDK and non nullable

vmishenev:
fixes #4564
Restores previous behavior


vmishenev:
In the future, #4513 could break it.

whyoleg (approved):

PR #4576, opened 2026-08-11
Update AA to 2.5.0-dev-3229

vmishenev:


dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/services/SymbolFullClassHierarchyBuilder.kt
```diff
@@ -50,7 +51,7 @@ internal class SymbolFullClassHierarchyBuilder(context: DokkaContext) : FullClas
     ) {
         val (dri, kotlinType) = driWithKType
         if (supersMap[dri] == null) {
-            val supertypes = kotlinType.directSupertypes(shouldApproximate = true).filterNot { it.isAnyType }.toList()
+            val supertypes = kotlinType.directSupertypes(shouldApproximate = true).filterNot { (it as? KaClassType)?.classId == StandardClassIds.Any }.toList()
```
whyoleg: minor: I think that it makes sense to extract this into a function, like `Sequence<KaType>.filterOutAny` (name is just an example) and use it in all places, as the filter condition now reads not so easily.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -646,7 +648,8 @@ internal class DokkaSymbolVisitor(
     ): DFunction = withExceptionCatcher(propertyAccessorSymbol) {
         val isGetter = propertyAccessorSymbol is KaPropertyGetterSymbol
         // it also covers @JvmName annotation
-        @OptIn(KaExperimentalApi::class) // due to javaGetterName/javaSetterName
+        @OptIn(KaExperimentalApi::class)
+        @Suppress("DEPRECATION")
```
whyoleg: Why suppress, and not replace with an alternative (`propertyAccessorSymbol.javaMethodName`)? 
E.g., I see that on [CI](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Dokka_ScheduledDokkaTestsK2/6449140), the message suggests using `javaMethodName` for the getter/setter.
Is there some problem with it?

whyoleg (changes requested):

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
vmishenev: My bad, I missed it. Junie generated it.

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
```diff
@@ -646,8 +648,11 @@ internal class DokkaSymbolVisitor(
     ): DFunction = withExceptionCatcher(propertyAccessorSymbol) {
         val isGetter = propertyAccessorSymbol is KaPropertyGetterSymbol
         // it also covers @JvmName annotation
-        @OptIn(KaExperimentalApi::class) // due to javaGetterName/javaSetterName
-        val name = (if (isGetter) propertySymbol.javaGetterName else propertySymbol.javaSetterName)?.asString() ?: ""
+        @OptIn(KaExperimentalApi::class)
+        // `javaMethodName` is null when there is no addressable JVM method (e.g. accessors of `const` or `@JvmField` properties),
+        val name = propertyAccessorSymbol.javaMethodName
+            ?: propertySymbol.name.asString()
+                .let { if (isGetter) JvmAbi.getterName(it) else JvmAbi.setterName(it) }
```
vmishenev, line 655: Just to have some names for the DRIs

whyoleg: Hm, maybe then we don't need to provide either a `getter` or a `setter` in this case at all?
While it might be a bit strange that we will rely on the logic of `javaMethodName` for this, it feels more correct, since with `const val`, there really is no getter.

Though I don't insist, the current approach is also fine, and it's compatible.

JIC, note that `JvmAbi` might become inaccessible at any moment, as it's probably an internal compiler API.

whyoleg (approved):

dokka-subprojects/analysis-kotlin-symbols/src/main/kotlin/org/jetbrains/dokka/analysis/kotlin/symbols/translators/DefaultSymbolToDocumentableTranslator.kt
vmishenev: > Hm, maybe then we don't need to provide either a getter or a setter in this case at all?

Maybe, but it rather represents only the Java view, not KMP. I'm just trying to keep compatibility.

We can't rely on `javaMethodName `. It returns `null` in some cases if it isn't `isJavaIdentifier`:
``val `foo bar` `` or  mangled `getX-impl / getX-5dsf87`

However, we can add the condition `if(property.isConst) getter=null`, but it looks like a separated change


> JIC, note that JvmAbi might become inaccessible at any moment, as it's probably an internal compiler API

It is bad(
Maybe we could provide special names  `<getter>` or `<setter>` for such cases. WDYT?

whyoleg: But we will still need to use the property name in the accessor, so would it be `<getter>-NAME`?
I would just keep compatibility with the old way for now, and let's decide everything else later. It looks like `JvmAbi` does this.
Let's leave it like it is now, and we will fix it later after the merge in Kotlin - it might not be the only usage of compiler APIs that might become unavailable when we start depending only on AA.

PR #4579, opened 2026-08-18
Update Jsoup to the latest version

whyoleg:
Update Jsoup to the latest version (1.23.2, [changelog](https://github.com/jhy/jsoup/blob/master/CHANGES.md)) to avoid vulnerabilities.

By simply updating the version, several things were broken:

1. escaping of `'` and `"` - fixed in a7b108bc66c7dbf6d824d6bd9aa4403aa68197f9
2. `inheritDoc` injection - fixed in 16b585dc58b99bc800757822f421ce8718b6d258
3. accessors in javadoc format tests couldn't be found - fixed in 9cb675340522b66612ab3337be6f79de4d505b2b
4. expectedData used `viewbox` attribute for some reason, while [original](https://github.com/Kotlin/dokka/blob/d4d457d357627fd8100e4f0d47868020485996ec/dokka-subprojects/plugin-versioning/src/main/resources/dokka/not-found-version.html#L136-L144) page always contained `viewBox` - fixed in 5f1eacb1e5b3451a750519ce84ab6a11b40d64fa

### DIFF
--- a/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/ParseUtils.kt
+++ b/dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/ParseUtils.kt
@@ -10,7 +10,7 @@ import org.jetbrains.dokka.InternalDokkaApi
 import org.jetbrains.dokka.model.doc.DocTag
 import org.jetbrains.dokka.model.doc.Text
 import org.jsoup.internal.StringUtil
-import org.jsoup.nodes.Entities
+import org.jsoup.nodes.TextNode
 
 @InternalDokkaApi
 public fun String.parseHtmlEncodedWithNormalisedSpaces(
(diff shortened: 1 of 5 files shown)


dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/ParseUtils.kt
```diff
@@ -21,12 +21,16 @@ public fun String.parseHtmlEncodedWithNormalisedSpaces(
...
+            if (renderWhiteCharactersAsSpaces) {
+                if (!lastWasWhite) {
+                    accum.append(' ')
+                    lastWasWhite = true
+                }
+            } else {
+                accum.appendCodePoint(c)
             }
-        } else if (Compat.codePointToString(c).let { it != Entities.escape(it) }) {
+        } else if (Compat.codePointToString(c).let { it != TextNode(it).toString() }) {
```
vmishenev: What about escaping other characters? Is it unnecessary?

whyoleg: Not sure I understand the question: could you expand a bit?

vmishenev: Before the change, there was some escaping using Entities.escape(it). After the change, it became `TextNode(it).toString()`. Is this also a kind of escaping, but without escaping `'` and `"` (non-attribute value)? 
Is this reliable way? 
 _I have not read the  `toString` documentation._

whyoleg: Yes, `TextNode.toString` returns HTML based on the passed text, so it will escape it (and will use a function that `Entities.escape` calls under the hood, but with a slightly different argument). I will add a comment about this.

Also, note that we only need to check whether the symbol needs escaping here, so it could be significantly simplified, though I decided to keep using the same jsoup to avoid breaking anything.

vmishenev: Thank you. I see.

Would replacing it with `TextNode.outerHtml()` be more straightforward than using `TextNode.toString()`? `outerHtml` makes it more explicit

vmishenev (approved):

dokka-subprojects/analysis-markdown-jb/src/main/kotlin/org/jetbrains/dokka/analysis/markdown/jb/ParseUtils.kt
whyoleg: Hm, yes! That sounds like a good idea! Thanks! Replaced and added a comment there.

PR #4577, opened 2026-08-13
Prepare the Dokka modules for the move to the Kotlin repo (KT-88283)

vmishenev:
The first part of KT-88283:
- Remove modules that don't make sense for KDM: current runners (Gradle/Maven plugins, CLI) and Dokka plugins. Effectively, only core, base and analysis modules are needed.
- Drop the content model and related rendering machinery.

After merging this PR, remained Dokka's modules will be merged into a single module.

PR #4580, opened 2026-08-18
[KDM] Merge modules into single module (KT-88283)

vmishenev:
The previous PR is #4577

^KT-88283


whyoleg (changes requested):
Overall, the changes look logical and simple, but because we dropped a lot of files, it's really difficult to review them as a whole in both IDEA and on GitHub.
So, maybe it makes sense to split PR into two, then?
1. Drop modules that don't make sense anymore (tests, runners, dev-docs, etc).
2. Merging modules and cleaning up around it.
Just for the review and potential future simplicity in git blame.

vmishenev:
> So, maybe it makes sense to split PR into two, then?

It's already divided into separate commits, so you can review them one commit at a time.

The problem is that the PR removes many files at once. 
Additionally, I split the commit that deletes the files into three separate commits.

> Just for the review and potential future simplicity in git blame.
 
I'm not going to squash the PR. Otherwise, such a commit would be a nightmare.

src/test/kotlin/org/jetbrains/dokka/analysis/test/SourceRootChecker.kt
```diff
@@ -31,13 +31,13 @@ class SourceRootChecker {
             DokkaSourceSetImpl(
                 analysisPlatform = Platform.jvm,
                 displayName = "KotlinJvmSourceSet2",
-                sourceSetID = KotlinJvmTestProject.DEFAULT_SOURCE_SET_ID,
+                sourceSetID = KotlinJvmTestProject.DEFAULT_SOURCE_SET_ID.copy(sourceSetName = "kotlin2"),
```
whyoleg, line 34: Why are those changes needed?

src/testFixtures/kotlin/org/jetbrains/dokka/analysis/test/api/kotlin/KotlinTestData.kt
```diff
@@ -8,9 +8,11 @@ import org.jetbrains.dokka.analysis.test.api.TestData
 import org.jetbrains.dokka.analysis.test.api.TestDataFile
 
 /**
+ *
+ * [org.jetbrains.dokka.analysis.test.api.kotlin.KotlinTestData] iss
```
whyoleg: Accidental changes?

gradle/libs.versions.toml
```diff
@@ -118,6 +111,11 @@ kotlin-symbol-light-classes = { module = "org.jetbrains.kotlin:symbol-light-clas
 #### Java analysis ####
 intellij-java-psi-api = { module = "com.jetbrains.intellij.java:java-psi", version.ref = "intellij-platform" }
 intellij-java-psi-impl = { module = "com.jetbrains.intellij.java:java-psi-impl", version.ref = "intellij-platform" }
+# The `*-syntax` artifacts come in transitively via `java-psi(-impl)`, but they must also be declared
+# explicitly to win over the stripped copies bundled in `kotlin-compiler`, see plugin-base
```
whyoleg: there is no `plugin-base` anymore

build.gradle.kts
```diff
@@ -2,96 +2,128 @@
...
+    implementation(libs.jsoup)
+    implementation(libs.jetbrains.markdown)
+
+//Java
+    // This must be explicit so the full `java-psi` API takes precedence over
+    // stripped copies that may be present in compiler-related artifacts.
+
+    // this is a `hack` to include classes `intellij-java-psi-api` in shadowJar
+    // which are not present in `kotlin-compiler`
+    // should be `api` since we already have it in :analysis-java-psi
```
whyoleg: there is no `:analysis-java-psi` anymore

build.gradle.kts
```diff
@@ -2,96 +2,115 @@
...
 
-fun addDependencyOnSameTasksOfIncludedBuilds(vararg taskNames: String) {
-    taskNames.forEach { taskName ->
-        tasks.named(taskName) {
-            dependsOn(includedBuildTasks(taskName))
+    // Other
+    implementation(libs.kotlin.reflect)
+    implementation(libs.kotlinx.coroutines.core)
+    implementation(libs.jackson.kotlin)
+    constraints {
```
whyoleg: minor: this constraint is redundant now, as we use single version for both kotlin and databind modules.

whyoleg (approved):
Nice, left just minor comments!

> Additionally, I split the commit that deletes the files into three separate commits.

Thanks! That's helped get "changed files" under the 3000 limit on GH, so it can show them all.

> I'm not going to squash the PR. Otherwise, such a commit would be a nightmare.

Yeah, that makes sense, thanks!
Splitting the PR was just more in line with how we've done this before. :)

src/test/kotlin/org/jetbrains/dokka/analysis/test/SourceRootChecker.kt
vmishenev: After merging the plugins, a check for unique `sourceSetId`s was added. There were two source sets with the same name.

whyoleg: Ah, from `base` plugin, got it!

PR #4581, opened 2026-08-19
Update Jackson to the latest version

whyoleg:
Update Jackson to the latest version (2.22.2) to avoid vulnerabilities, and update Gradle versions in IT to use Gradle 7.6.6, which includes a fix that allows multi-release JARs.
In theory, it means we drop support for Gradle 7.6.3 and below, though I think that those who use Gradle 7.6 could easily update the patch version if needed.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -39,7 +39,7 @@ jetbrains-markdown = "0.7.3"
 
 ## JSON
 # jackson 2.15.3 is the latest version, which works correctly with Kotlin API/Language 1.4 and Gradle 7.6.3+
-jackson = "2.15.3"
+jackson = "2.22.2"
 
 ## Maven
 apacheMaven-core = "3.5.0"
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -39,7 +39,7 @@ jetbrains-markdown = "0.7.3"
 
 ## JSON
 # jackson 2.15.3 is the latest version, which works correctly with Kotlin API/Language 1.4 and Gradle 7.6.3+
```
MGaetan89: Shouldn't the comment be reviewed?

vmishenev (approved):
It seems like the comment above should be updated.

gradle/libs.versions.toml
whyoleg: Thanks, yes! I was in a bit of a rush to create a PR, so I totally forgot about it...

PR #4592, opened 2026-09-02
Add tests to reproduce #4588

### LINKED ISSUE #4588: Nested interface with clashed properties from parents has a new page for the property
The scenario:
```kt
interface Schema
interface NarrowedSchema : Schema

interface Base {
    /**
     * KDOC TO DISAPPEAR
     */
    val schema: Schema
}

interface NarrowedBase : Base {
    override val schema: NarrowedSchema
}

interface Container {
    /**
     * This nested interface has schema that miss kdoc
     */
    interface Nested : Base, NarrowedBase
}

/**
 * The top level interface that has kdoc
 */
interface TopLevel : Base, NarrowedBase
```
The problem is for `Container.Nested`. `schema` property there miss the kdoc and link leads to a new page under `Nested`.
As an example there is `TopLevel` where the `schema` leads to the page under Base class (but the return value from `NarrowBase` 🤔)

In Dokka 2.2.0 `Container.Nested` and `TopLevel` behaved in the same way (as `TopLevel` in 2.3.0-test).

Dokka 2.2.0

Image

Dokka 2.3.0-test

Image

The scenario from aws-sdk-kotlin. Class: `dynamodb-mapper/aws.sdk.kotlin.hll.dynamodbmapper.model/-table/-partition-key/index.html` and `schema` property.

vmishenev:
#4588 has been fixed by updating AA to 2.5.0-dev-3229 (#4576)

### DIFF
--- a/dokka-subprojects/plugin-base/src/test/kotlin/model/PropertyTest.kt
+++ b/dokka-subprojects/plugin-base/src/test/kotlin/model/PropertyTest.kt
@@ -12,6 +12,8 @@ import utils.name
 import utils.text
 import kotlin.test.Test
 import org.jetbrains.dokka.ExperimentalDokkaApi
+import org.jetbrains.dokka.base.signatures.KotlinSignatureUtils.driOrNull
+import org.jetbrains.dokka.links.DRI
 import org.jetbrains.dokka.links.Nullable
 import org.jetbrains.dokka.links.TypeConstructor
 import org.jetbrains.dokka.links.TypeParam
@@ -279,6 +281,7 @@ class PropertyTest : AbstractModelTest("/src/main/kotlin/property/Test.kt", "pro
                 fun DProperty.assertInheritedFromBase() {
                     dri.classNames equals "Base"
                     dri.callable?.name equals "schema"
+                    type.driOrNull equals DRI("property", "NarrowedSchema")
 
                     val inheritedFrom = extra[InheritedMember]?.inheritedFrom?.values?.single()
                         .assertNotNull("inheritedFrom")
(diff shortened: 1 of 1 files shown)


vmishenev:
I couldn't find a related issue in YT.

dokka-subprojects/plugin-base/src/test/kotlin/model/PropertyTest.kt
```diff
@@ -245,6 +246,104 @@ class PropertyTest : AbstractModelTest("/src/main/kotlin/property/Test.kt", "pro
...
+            |
+            |interface Base {
+            |    /**
+            |     * base property docs
+            |     */
+            |    val schema: Schema
+            |}
+            |
+            |interface NarrowedBase : Base {
+            |    override val schema: NarrowedSchema
```
whyoleg, line 270: Could you also assert in the test the return type of the property?
In the original issue, it has `NarrowedSchema`, so I'm curious what will it be here.

dokka-subprojects/plugin-base/src/test/kotlin/model/PropertyTest.kt
whyoleg, line 1: (top-level comment to create a thread)

> has been fixed by updating AA to 2.5.0-dev-3229 (https://github.com/Kotlin/dokka/pull/4576)

So, added test, fails on commit before that update, right? Specifically on the 2.3.0 branch, right?

whyoleg (commented):
It's nice that it's already working as expected with the new AA. But, this means we will need to cherry-pick the AA update into the 2.3.0 branch, so we will need to retest everything one more time, right? cc @atyrin

dokka-subprojects/plugin-base/src/test/kotlin/model/PropertyTest.kt
vmishenev: Yes, I have checked

whyoleg (approved):

PR #4593, opened 2026-09-02
Fix table header parsing: use Th nodes for Markdown tables

### LINKED ISSUE #4586: Broken markdown table rendering: header replaces with cells
```
/**
 * Minimal case: one column, one data row, no alignment, no bold in the header.
 *
 * | header |
 * |--------|
 * | value  |
 */
public class SingleColumnTable
```

Dokka 2.2.0

Image

Dokka 2.3.0-test

Image

or with

```
/**
 * The same table with no alignment markers at all.
 *
 * | **id** | **type** | **manufacturer** |
 * |--------|----------|------------------|
 * | 1      | car      | Ford             |
 * | 2      | bike     | Schwinn          |
 */
public class PlainTable
```

Dokka 2.2.0

Image

Dokka 2.3.0-test

Image

vmishenev:
fixes #4586


whyoleg (approved):
That was really unexpected for me, that our `Table` children have a totally different structure when coming from HTML vs. MD...

